PR #4, opened 2024-02-10
Initial OpenTelemetry Tracer to stdout

shehabgamin:
Initial tracer setup for OpenTelemetry.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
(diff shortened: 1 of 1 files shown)


crates/framework-telemetry/src/telemetry.rs
```diff
@@ -0,0 +1,57 @@
...
+                os_resource
+                    .merge(&process_resource)
+                    .merge(&sdk_resource)
+                    .merge(&env_resource)
+                    .merge(&telemetry_resource),
+            ),
+        )
+        .build();
+
+    provider.tracer("Test") // TODO: should be service.name
```
shehabgamin, line 53: Spent too long trying to figure this out, don't think it's worth our time. When we setup the collector, this line won't even exist and `service.name` will be set properly.

PR #10, opened 2024-04-01
Support explode function

linhr:


crates/spark-connect-server/examples/pyspark-sql/src/app/__main__.py
```diff
@@ -39,6 +42,30 @@ def add_one(x):
     # df = spark.read.json("/tmp/df.json/")
     # print(df.toPandas())
 
+    print(
```
linhr, line 45: Our implementation can support nested `explode` functions and multiple `explode` expressions in the same `SELECT` clause. This expands the semantics of Spark SQL.

PR #14, opened 2024-04-06
Basic Python UDF Implementation

shehabgamin:
Basic Python UDF Implementation for CommonInlineUserDefinedFunction. More work is needed but I think that this is a good foundation to merge in so that we can unblock other work.

### DIFF
--- a/crates/framework-python/src/error.rs
+++ b/crates/framework-python/src/error.rs
@@ -1,6 +1,7 @@
 use datafusion_common::DataFusionError;
 use pyo3::prelude::PyErr;
 
+// TODO: Support implicit conversion by implementing the From trait
 pub fn convert_pyerr_to_dferror(py_err: PyErr) -> DataFusionError {
     DataFusionError::Execution(format!("Python error: {:?}", py_err))
 }
--- a/crates/framework-python/src/utils.rs
+++ b/crates/framework-python/src/utils.rs
@@ -213,6 +213,7 @@ pub fn execute_python_function(
     output_type: &DataType,
 ) -> Result<ArrayRef, DataFusionError> {
     Python::with_gil(|py| {
+        // TODO: Avoid loading the function on every batch.
         let python_function = load_python_function(py, &function_bytes)?;
 
         let processed_array = match &output_type {


crates/framework-python/src/error.rs
```diff
@@ -0,0 +1,6 @@
+use datafusion_common::DataFusionError;
+use pyo3::prelude::PyErr;
+
+pub fn convert_pyerr_to_dferror(py_err: PyErr) -> DataFusionError {
```
linhr, line 5: OK for now, but we could support implicit conversion by implementing the `From` trait for an error type defined in our crates. (We cannot implement `From<PyErr>` for `DataFusionError` though, since neither of them is defined in our crates.)

Here is an example: <https://github.com/lakehq/framework/blob/16457698c15f247994616079f2a55916dbfbf6e2/crates/spark-connect-server/src/error.rs#L68-L72>

crates/framework-python/src/udf.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        let (array_ref, is_scalar) = match &args[0] {
+            ColumnarValue::Array(arr) => {
+                (arr.clone(), false)
+            }
+            ColumnarValue::Scalar(scalar) => {
+                let arr = scalar
+                    .to_array().
+                    map_err(|e| {
```
linhr: Let's try `cargo fmt` and see if it fixes the style of the chained function call here.

You could also enable "format on save" for RustRover.

docker/lakesail.Dockerfile
```diff
@@ -9,9 +9,19 @@ RUN apt-get update && \
...
+    python3-pip \
+    python3-venv && \
     rm -rf /var/lib/apt/lists/*
 
+# Install PySpark
+ARG PYSPARK_VERSION=3.5.1
+ENV PYSPARK_VERSION $PYSPARK_VERSION
+RUN python3 -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+RUN python3 -m pip install "pyspark[connect,sql]==${PYSPARK_VERSION}" # TODO: Look into only capturing pyspark.cloudpickle
```
linhr: Regarding the TODO comment, I feel we might not be able to only install `pyspark.cloudpickle` since the UDF refers to Spark data types which are part of the PySpark library.

crates/spark-connect-server/src/expression.rs
```diff
@@ -74,9 +77,6 @@ pub(crate) fn from_spark_expression(
             if func.is_distinct {
                 return Err(SparkError::unsupported("distinct function"));
             }
-            if func.is_user_defined_function {
-                return Err(SparkError::unsupported("user defined function"));
-            }
```
linhr: Let's add this back. I don't think we support registering UDF to the catalog yet.

crates/spark-connect-server/src/expression.rs
```diff
@@ -240,8 +240,78 @@ pub(crate) fn from_spark_expression(
...
+        ExprType::CommonInlineUserDefinedFunction(udf) => {
+            use sc::common_inline_user_defined_function::Function::PythonUdf;
+            use sc::PythonUdf as PythonUDFStruct;
+            use pyo3::prelude::Python;
+
+            let function_name: &str = &udf
+                .function_name;
+
+            let deterministic: bool = udf
+                .deterministic;
```
linhr: I guess you wanted style consistency with the chained expressions below, but I feel it's a bit more readable if we have such simple statements in a single line.

crates/spark-connect-server/src/expression.rs
```diff
@@ -240,8 +240,78 @@ pub(crate) fn from_spark_expression(
...
+                .command;
+
+            let python_ver: &str = &function
+                .python_ver;
+
+            let pyo3_python_version: String = Python::with_gil(|py| {
+                py.version().to_string()
+            });
+
+            if !pyo3_python_version.starts_with(python_ver) {
```
linhr, line 288: Nice!

crates/framework-python/src/utils.rs
```diff
@@ -0,0 +1,478 @@
...
+    py: Python<'py>,
+    python_function: &Py<PyAny>,
+) -> Result<ArrayRef, DataFusionError>
+    where
+        TOutput: ArrowPrimitiveType,
+        TOutput::Native: FromPyObject<'py>,
+{
+    match &array_ref.data_type() {
+        DataType::Null => {
+            unimplemented!()
```
linhr: We should avoid panic since we have a long-running server. Let's return `DataFusionError::NotImplemented` with an error message instead.

crates/framework-python/src/udf.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    fn invoke(&self, args: &[ColumnarValue]) -> Result<ColumnarValue> {
+        if args.len() != 1 {
+            return Err(DataFusionError::Internal(format!(
+                "{:?} should only be called with a single argument",
+                self.name()
+            )));
+        }
+
+        let (array_ref, is_scalar) = match &args[0] {
```
linhr, line 70: A simpler approach is to use `ColumnarValue::values_to_arrays` to convert the input, and returns `ColumnarValue::Array` after the UDF call regardless of whether the input is a scalar or not. In this way we don't need the `array_ref_to_columnar_value` function.

The documentation says `ColumnarValue::values_to_arrays` is not efficient for scalar, but I think it's fine for now. (I believe it's inefficient because it converts scalar to a single-element array, which is similar to what we are doing in `array_ref_to_columnar_value`.)

crates/framework-python/src/utils.rs
```diff
@@ -0,0 +1,478 @@
...
+
+pub fn execute_python_function(
+    array_ref: &ArrayRef,
+    function_bytes: &[u8],
+    output_type: &DataType,
+) -> Result<ArrayRef, DataFusionError> {
+    Python::with_gil(|py| {
+        let python_function = load_python_function(py, &function_bytes)?;
+
+        let processed_array = match &output_type {
```
linhr, line 219: OK for now, but we might be able to dispatch on data type in one place (instead of dispatching twice in two functions). We'll probably get more idea when we support complex data types.

crates/framework-python/src/utils.rs
```diff
@@ -0,0 +1,478 @@
...
+    }
+}
+
+pub fn execute_python_function(
+    array_ref: &ArrayRef,
+    function_bytes: &[u8],
+    output_type: &DataType,
+) -> Result<ArrayRef, DataFusionError> {
+    Python::with_gil(|py| {
+        let python_function = load_python_function(py, &function_bytes)?;
```
linhr, line 217: Let's add a TODO statement to avoid loading the function on every batch.

crates/framework-python/src/utils.rs
shehabgamin: @linhr I tried doing that, but couldn't figure out a way to do it elegantly.

crates/framework-python/src/udf.rs
shehabgamin: @linhr I think we should keep it as is until we better understand the warning in values_to_arrays. Also values_to_arrays assumes that all args have the same length. Will there ever be a scenario where the arguments have mixed length?

crates/framework-python/src/error.rs
shehabgamin: Added a TODO!

crates/framework-python/src/utils.rs
shehabgamin: Added a TODO!

crates/framework-python/src/udf.rs
linhr: I haven't looked into the assumptions for `values_to_arrays` either. Yeah, let's revisit this later.

PR #15, opened 2024-04-25
DataFusion 37.1.0 - Fixing Various Breaking Changes 

shehabgamin:


crates/spark-connect-server/src/expression.rs
```diff
@@ -444,10 +456,10 @@ pub(crate) fn get_scalar_function(
         }
         // TODO: contains
         "startswith" => {
-            return Ok(expr::Expr::ScalarFunction(expr::ScalarFunction {
-                func_def: ScalarFunctionDefinition::BuiltIn(BuiltinScalarFunction::StartsWith),
-                args,
-            }));
+            return Ok(functions::expr_fn::starts_with(
+                args[0].clone(),
+                args[1].clone(),
```
linhr, line 464: Let's return an error if `args` is not of size 2. Otherwise, the code may panic for out-of-bound indexing.

crates/spark-connect-server/src/expression.rs
```diff
@@ -517,6 +531,9 @@ pub(crate) fn get_scalar_function(
                 negated: false,
             }));
         }
+        "abs" => {
+            return Ok(functions::expr_fn::abs(args[0].clone()));
```
linhr, line 541: Same here. We need to make sure `args` has exactly one element.

PR #16, opened 2024-04-26
More Pickle Stuff

shehabgamin:


crates/framework-python/src/py_object_pyspark.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde::de::{self, value::BorrowedBytesDeserializer, Deserialize, Deserializer, Visitor};
+use serde_bytes::Bytes;
+
+#[derive(Debug, Clone)]
+pub struct PythonObjectWrapper {
+    pub function: PyObject,
+    pub return_type: PyObject,
+}
+
+struct PyObjectVisitor;
```
linhr: We could probably name these `PyFunctionWrapper` and `PyFunctionVisitor` for clarity and consistency.

crates/framework-python/src/lib.rs
```diff
@@ -1,3 +1,4 @@
 mod error;
+pub mod py_object_pyspark;
```
linhr: The module name may be a bit too specific. OK for now though, since we will get better ideas around module structure and naming when we add more functionality.

crates/framework-python/src/py_object_pyspark.rs
shehabgamin: Done!

PR #17, opened 2024-04-27
Spark Connect Tests Setup

shehabgamin:


.gitignore
```diff
@@ -40,3 +40,6 @@ Chart.lock
 *.tar
 *.gz
 *.tgz
+
+# Spark
+spark/
```
linhr, line 45: I'm wondering if we can remove this, and in README we can say that the Spark project should be cloned outside the framework project.

scripts/spark-connect-server/run-framework-server.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -euo 'pipefail'
+
+cd "${FRAMEWORK_PROJECT_PATH}" || exit
```
linhr:
```suggestion
cd "${FRAMEWORK_PROJECT_PATH}"
```

`exit` is no longer needed since we now have `set -e`.

scripts/spark-connect-server/setup-spark-env.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euo 'pipefail'
+
+cd "${SPARK_PROJECT_PATH}" || exit
```
linhr:
```suggestion
cd "${SPARK_PROJECT_PATH}"
```

.gitignore
shehabgamin: I personally prefer having it inside the framework project so that I don't need to have two JetBrains windows open.

PR #18, opened 2024-04-29
Fix Python SIGINT issue

linhr:


linhr:
I made some small code changes to get rid of all compiler warnings as well.

PR #21, opened 2024-05-06
Set up GitHub Actions

linhr:
1. Add a workflow to build and test Rust code.
2. Add a workflow to run Spark tests in pull requests.


.github/actions/setup-rust/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        set -x
+        protoc --version
+        python -V
+        rustc -vV
+        set +x
+        rust_release="$(rustc -vV | sed -ne 's/^release: *//p')"
+        echo "rust_release=${rust_release}" >> "$GITHUB_OUTPUT"
+      shell: bash
+
+    - uses: actions/cache@v4
```
linhr, line 29: There are third-party actions such as [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) which supports restoring Rust build cache by partial cache key match, and has the logic to clean up unused artifacts from the cache.

Here I use the official cache action instead, and use exact cache key match. Our Rust build only takes a few minutes even without cache, so we can discard cache entirely when `Cargo.toml` changes. I feel this makes it easier to reason about the cache behavior for our use case.

.github/workflows/run-spark-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+    uses: ./.github/workflows/run-spark-tests-internal.yml
+    with:
+      ref: ${{ github.ref }}
+      name: head
+
+  test-base:
+    uses: ./.github/workflows/run-spark-tests-internal.yml
+    with:
+      # FIXME: use `${{ github.base_ref }}`
+      ref: ${{ github.ref }}
```
linhr, line 19: This needs to be updated in a separate PR.

README.md
```diff
@@ -26,32 +26,33 @@ one for the Spark project and the other for the framework Python examples.
 Run the following command to clone the Spark project.
 
 ```bash
-git clone git@github.com:apache/spark.git
+git clone git@github.com:apache/spark.git opt/spark
```
linhr, line 29: `opt/spark` is now the required location for the Spark project in the repository.

scripts/spark-tests/run-server.sh
```diff
@@ -2,14 +2,17 @@
 
 set -euo 'pipefail'
 
-cd "${FRAMEWORK_PROJECT_PATH}"
+project_path="$(dirname "$0")/../.."
 
-source "${FRAMEWORK_PROJECT_PATH}"/examples/python/.venv/bin/activate
+source "${project_path}"/examples/python/.venv/bin/activate
+
+python_version=$(python -c 'import sys; print("%s.%s" % (sys.version_info.major, sys.version_info.minor))')
```
shehabgamin, line 9: nice!

.github/actions/setup-rust/action.yml
shehabgamin: Great work

PR #22, opened 2024-05-08
Use PyArrow For UDF

shehabgamin:


crates/spark-connect-server/src/extension/function/struct_function.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, arg_types: &[DataType]) -> Result<DataType> {
+        let return_fields = arg_types
+            .iter()
+            .enumerate()
+            .map(|(pos, dt)| Field::new(self.field_names[pos].clone(), dt.clone(), true))
+            .collect::<Vec<Field>>();
```
linhr: Let's handle look up error here. In `.map()` you can return a `Result` and in `.collect()` you can convert element error to an error over the entire collection like this:

```rust
.collect::<Result<Vec<Field>>>()?;
```

crates/spark-connect-server/src/extension/function/struct_function.rs
```diff
@@ -0,0 +1,83 @@
...
+    Ok(Arc::new(StructArray::from(vec)))
+}
+
+fn to_struct_expr(args: &[ColumnarValue], field_names: &[String]) -> Result<ColumnarValue> {
+    let arrays = ColumnarValue::values_to_arrays(args)?;
+    Ok(ColumnarValue::Array(to_array_struct(
+        arrays.as_slice(),
+        field_names,
+    )?))
+}
```
linhr, line 42: Add an empty line here. (I'm not sure why `rustfmt` didn't capture this.)

crates/spark-connect-server/src/extension/function/struct_function.rs
```diff
@@ -0,0 +1,83 @@
...
+            Ok((
+                Arc::new(Field::new(
+                    field_names[i].as_str(),
+                    arg.data_type().clone(),
+                    true,
+                )),
+                arg.clone(),
+            ))
+        })
+        .collect::<Result<Vec<_>>>()?;
```
linhr: Let's use `.zip()` instead of array indexing to avoid panic on different array length.

crates/spark-connect-server/src/expression.rs
```diff
@@ -622,13 +623,35 @@ pub(crate) fn get_scalar_function(
...
             return Ok(functions::expr_fn::to_unixtime(args));
         }
+        "struct" => {
+            let field_names: Vec<String> = args.iter().map(|x| {
+                match x {
+                    expr::Expr::Column(column) => Ok(column.name.to_string()),
+                    _ => {
+                        Err(SparkError::invalid("get_scalar_function: struct function should have expr::Expr::Column as arguments"))
+                    }
+                }.unwrap()
```
linhr: Replace `.unwrap()` with `?`.

crates/framework-python/src/lib.rs
```diff
@@ -18,3 +19,28 @@ signal.signal(signal.SIGINT, signal.SIG_DFL)
 pub fn init_python() -> PyResult<()> {
     Python::with_gil(|py| py.run_bound(PYTHON_INIT_SCRIPT, None, None))
 }
+
+#[macro_export]
+macro_rules! impl_py_state_serialization {
```
linhr, line 24: How is this used?

shehabgamin: Ahh it's not anymore. But we may need it later so i just kept it.

PR #27, opened 2024-05-09
Support user-defined types

linhr:
This PR adds support for the Spark user-defined types (UDT). Spark uses schema field metadata to store UDT information, so we need to keep track of field metadata in the Arrow schema.

### DIFF
--- a/scripts/spark-tests/spark-3.5.1.patch
+++ b/scripts/spark-tests/spark-3.5.1.patch
@@ -154,6 +154,19 @@ index e1ec97928f7..76dd13a6867 100644
  
      def _test_apply_in_pandas_with_state_basic(self, func, check_results):
          input_path = tempfile.mkdtemp()
+diff --git a/python/pyspark/testing/sqlutils.py b/python/pyspark/testing/sqlutils.py
+index 077d854b1dd..513a8f5d516 100644
+--- a/python/pyspark/testing/sqlutils.py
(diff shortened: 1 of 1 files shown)


scripts/spark-tests/spark-3.5.1.patch
```diff
@@ -154,6 +154,19 @@ index e1ec97928f7..76dd13a6867 100644
...
+diff --git a/python/pyspark/testing/sqlutils.py b/python/pyspark/testing/sqlutils.py
+index 077d854b1dd..513a8f5d516 100644
+--- a/python/pyspark/testing/sqlutils.py
++++ b/python/pyspark/testing/sqlutils.py
+@@ -84,7 +84,7 @@ class ExamplePointUDT(UserDefinedType):
+ 
+     @classmethod
+     def module(cls):
+-        return "pyspark.sql.tests"
++        return "pyspark.testing.sqlutils"
```
linhr, line 166: I'm patching this because the value seems incorrect. The original test somehow works even with the incorrect value. I guess the reason is that the UDT defines its Scala counterpart as `org.apache.spark.sql.test.ExamplePointUDT`, and the Python UDT class is correctly defined in the Scala class:

```scala
  override def pyUDT: String = "pyspark.testing.sqlutils.ExamplePointUDT"
```

So the Python UDT class happens to be corrected after the round-trip to the JVM. For us, there is no Scala UDT involved, so the incorrect Python class does not work.

PR #28, opened 2024-05-10
Support join and window

linhr:
1. Support join in logical plans.
2. Support window expressions.

### DIFF
--- a/scripts/spark-tests/run-server.sh
+++ b/scripts/spark-tests/run-server.sh
@@ -10,6 +10,7 @@ python_version=$(python -c 'import sys; print("%s.%s" % (sys.version_info.major,
 
 export RUST_LOG="${RUST_LOG:-spark_connect_server=debug}"
 export RUST_BACKTRACE="${RUST_BACKTRACE:-full}"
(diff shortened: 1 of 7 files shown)


scripts/spark-tests/run-server.sh
```diff
@@ -10,6 +10,7 @@ python_version=$(python -c 'import sys; print("%s.%s" % (sys.version_info.major,
 
 export RUST_LOG="${RUST_LOG:-spark_connect_server=debug}"
 export RUST_BACKTRACE="${RUST_BACKTRACE:-full}"
+export RUST_MIN_STACK="${RUST_MIN_STACK:-8388608}"
```
linhr, line 13: The default stack size (2 MB) is too small for deeply nested expressions/plans.

linhr:
There are a few known issues but we will address these in follow-up PRs.

1. The join condition does not work if the two tables have the same join column name. We need to resolve the column reference.
2. Some window functions return a `int64` data type (row number, rank, etc.) while Spark expects `int32`.

PR #29, opened 2024-05-15
Catalogz

shehabgamin:


crates/spark-connect-server/src/plan.rs
```diff
@@ -643,8 +682,297 @@ pub(crate) async fn from_spark_relation(
...
+                CatType::CurrentDatabase(_current_database) => {
+                    let results: DataFrame = ctx
+                        .sql("SELECT CAST($1 AS STRING)")
+                        .await?
+                        .with_param_values(vec![(
+                            "1",
+                            ScalarValue::Utf8(Some(
+                                state.config().options().catalog.default_schema.to_string(),
+                            )),
+                        )])?;
```
linhr: It seems we use positional parameters here, so I wonder if we can pass the value vector instead a map from `"1"` to the value. (It seems a bit surprising to me that `"1"` is accepted as a valid parameter name.)

Same for other places in this function.

crates/spark-connect-server/src/schema.rs
```diff
@@ -358,3 +359,161 @@ pub(crate) fn cast_record_batch(
...
+    pub fn to_simple_string(&self) -> SparkResult<String> {
+        match &self.kind {
+            Some(kind) => kind.to_simple_string(),
+            None => Ok("none".to_string()),
+        }
+    }
+}
+
+impl sdt::Kind {
+    pub fn to_simple_string(&self) -> SparkResult<String> {
```
linhr, line 385: This looks quite solid!

crates/spark-connect-server/src/sql/utils.rs
```diff
@@ -0,0 +1,59 @@
...
+                // Ignore pattern syntax errors
+            }
+        }
+    }
+
+    func_names
+}
+
+pub(crate) fn parse_identifiers(s: &str) -> SparkResult<Vec<Ident>> {
+    let mut parser = Parser::new(&GenericDialect {}).try_with_sql(s)?;
```
linhr: Let's use `SparkDialect`, which extends `GenericDialect`.

crates/spark-connect-server/src/sql/utils.rs
```diff
@@ -0,0 +1,59 @@
...
+            Ok(regex) => {
+                for &name in &names {
+                    let name = name.to_string();
+                    if regex.is_match(&name) && !func_names.contains(&name) {
+                        func_names.push(name);
+                    }
+                }
+            }
+            Err(_) => {
+                // Ignore pattern syntax errors
```
linhr: Let's handle the error here. We should return `SparkResult<_>` for this function since the client may specify an invalid regex.

crates/spark-connect-server/src/plan.rs
```diff
@@ -643,8 +682,297 @@ pub(crate) async fn from_spark_relation(
...
+                            let data_type = from_spark_built_in_data_type(schema)?;
+                            match data_type {
+                                DataType::Struct(fields) => fields,
+                                _ => return Err(SparkError::invalid("external table schema")),
+                            }
+                        }
+                        None => Fields::empty(),
+                    };
+                    let options: CaseInsensitiveHashMap<String> =
+                        CaseInsensitiveHashMap::from(create_table.options.clone());
```
linhr: This seems unused.

Let's review all the warnings as well and fix those if possible. You can add `#[allow(dead_code)]` for structs/enums not in use yet.

crates/spark-connect-server/src/plan.rs
shehabgamin: Oh yeah good point lol!

crates/spark-connect-server/src/sql/utils.rs
shehabgamin: @linhr The equivalent function in Spark ignores pattern syntax errors. So I think we should match that functionality.
```
  /**
   * This utility can be used for filtering pattern in the "Like" of "Show Tables / Functions" DDL
   * @param names the names list to be filtered
   * @param pattern the filter pattern, only '*' and '|' are allowed as wildcards, others will
   *                follow regular expression convention, case insensitive match and white spaces
   *                on both ends will be ignored
   * @return the filtered names list in order
   */
  def filterPattern(names: Seq[String], pattern: String): Seq[String] = {
    val funcNames = scala.collection.mutable.SortedSet.empty[String]
    pattern.trim().split("\\|").foreach { subPattern =>
      try {
        val regex = ("(?i)" + subPattern.replaceAll("\\*", ".*")).r
        funcNames ++= names.filter{ name => regex.pattern.matcher(name).matches() }
      } catch {
        case _: PatternSyntaxException =>
      }
    }
    funcNames.toSeq
  }
  ```

linhr: I see. Could you add a note in the code?

PR #39, opened 2024-05-24
Support plan abstraction

linhr:
1. Support converting Spark Connect data structures to "unresolved" plans.
2. Use the "unresolved" plan to generate DataFusion logical plans. The code is cleaner, and should also be more efficient, since we consume the input plan instead of having a reference to the input (which requires various `.clone()` operations downstream).


linhr:
There are a few more Spark tests failing since we no longer uses `SqlToRel` for SQL to plan conversion. Such tests will be fixed in follow-up PRs when we add more functionalities to `from_ast_statement`.

PR #40, opened 2024-05-24
Upgrade DataFusion to 38.0

linhr:


crates/spark-connect-server/src/expression.rs
```diff
@@ -629,20 +611,15 @@ pub(crate) fn get_scalar_function(
...
                 func_def: ScalarFunctionDefinition::UDF(Arc::new(ScalarUDF::from(
                     StructFunction::new(field_names),
                 ))),
-                args: args,
-            }));
+                args,
+            }))
         }
-        _ => {}
+        _ => Err(SparkError::invalid(format!("unknown function: {}", name))),
```
linhr, line 621: More test cases in the function gold set are failing since we haven't added all the available functions here. This will be done in follow up PRs.

crates/spark-connect-server/src/plan.rs
```diff
@@ -701,9 +702,15 @@ pub(crate) async fn from_spark_relation(
...
-            // ctx.state().config().options_mut().catalog.default_schema = db_name;
+        PlanNode::SetCurrentDatabase { database_name } => {
+            ctx.state_weak_ref()
+                .upgrade()
+                .ok_or_else(|| SparkError::internal("invalid session context"))?
+                .write()
+                .config_mut()
+                .options_mut()
+                .catalog
+                .default_schema = database_name;
```
linhr, line 713: `ctx.state()` returns a copy of the state, so we grab a write lock instead and modify the state via a mutable reference.

PR #41, opened 2024-05-27
Refactor catalog commands

linhr:
1. Define a logical plan extension node for all catalog commands. In this way, `from_spark_relation` will not have the side-effect of actually executing DDL. Such side-effect is dangerous since a DDL can be nested inside an `EXPLAIN` SQL statement, and the DDL should not be executed in such cases.
2. Define a `SessionCatalogContext` struct and move all catalog functions to it. Refactor the implementation of these functions.
3. Use the [serde_arrow](https://docs.rs/serde_arrow/latest/serde_arrow/) crate to convert Rust struct to Arrow `RecordBatch` and derive schema automatically.
4. Update the `metadata` field in `spec::Field` to have type `HashMap<String, String>` instead of `Option<HashMap<String, String>>`.

### DIFF
--- a/crates/spark-connect-server/src/extension/logical/command.rs
+++ b/crates/spark-connect-server/src/extension/logical/command.rs
@@ -1,4 +1,3 @@
-use arrow::datatypes::{FieldRef, Schema, SchemaRef};
 use std::fmt::Formatter;
 use std::sync::Arc;
 
@@ -8,6 +7,7 @@ use crate::sql::session_catalog::database::DatabaseMetadata;
 use crate::sql::session_catalog::function::FunctionMetadata;
 use crate::sql::session_catalog::table::TableMetadata;
 use crate::sql::session_catalog::{EmptyMetadata, SessionCatalogContext, SingleValueMetadata};
+use arrow::datatypes::{FieldRef, Schema, SchemaRef};
 use datafusion::common::{DFSchemaRef, Result};
 use datafusion::datasource::{provider_as_source, MemTable};
 use datafusion::logical_expr::{Expr, LogicalPlan, UserDefinedLogicalNodeCore};
@@ -188,12 +188,11 @@ impl CatalogCommand {
                 to_record_batch(fields, &rows)
             }
             CatalogCommand::DatabaseExists { database_name } => {
-                let value = ctx
-                    .list_databases(
-                        Some(ctx.default_catalog()?.as_str()),
-                        Some(database_name.as_str()),
-                    )?
-                    .is_empty();
+                let databases = ctx.list_databases(
+                    Some(ctx.default_catalog()?.as_str()),
+                    Some(database_name.as_str()),
+                )?;
+                let value = !databases.is_empty();
                 let rows = vec![SingleValueMetadata { value }];
                 to_record_batch(fields, &rows)
             }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -61,6 +61,7 @@ arrow-cast = { version = "51.0.0" }
 sqlparser = { version = "0.46.0", features = ["serde", "visitor"] }
 polars = { version = "0.39.2", features = ["serde", "polars-io", "dtype-struct"] }
 polars-arrow = { version = "0.39.2", features = ["serde", "arrow_rs"] }
+serde_arrow = { version = "0.11", features = ["arrow-51"] }
```
linhr, line 64: This crate is not widely used, but I found it useful for us. If it ever becomes unmaintained in the future, we can switch to our own manual way to build Arrow `RecordBatch`.

crates/spark-connect-server/src/extension/logical/command.rs
```diff
@@ -0,0 +1,329 @@
...
+        Ok(Self {
+            name: format!("CatalogCommand: {}", command.name()),
+            schema: DFSchemaRef::new(DFSchema::try_from(schema)?),
+            command,
+        })
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub(crate) enum CatalogCommand {
```
linhr, line 39: The `CatalogCommand` enum differs from `spec::PlanNode` in that all the schemas and plans (e.g. the logical plan for the `CreateTable` command) are **resolved** and are represented using DataFusion data structures.

crates/spark-connect-server/src/sql/session_catalog/table.rs
```diff
@@ -1,279 +1,133 @@
...
-                        if filter != table_type {
-                            continue;
-                        }
-                    }
-                    None => {}
+    pub(crate) async fn list_tables(
+        &self,
+        catalog_pattern: Option<&str>,
+        database_pattern: Option<&str>,
+        table_pattern: Option<&str>,
```
linhr, line 92: For such listing functions, we now takes optional catalog/database patterns but do not replace the pattern with the default catalog/database when it is `None`. Such logic is moved to the logical plan command node. When the pattern is `None`, we visit all the catalogs/databases. This makes `SessionCatalogContext` much simpler.

crates/spark-connect-server/src/sql/utils.rs
```diff
@@ -37,23 +36,13 @@ pub(crate) fn filter_pattern(names: Vec<&str>, pattern: Option<&str>) -> Vec<Str
...
+
+#[allow(dead_code)]
 pub(crate) fn parse_identifiers(s: &str) -> SparkResult<Vec<Ident>> {
     let mut parser = Parser::new(&SparkDialect {}).try_with_sql(s)?;
     let idents = parser.parse_multipart_identifier()?;
     Ok(idents)
 }
-
-pub(crate) fn build_schema_reference(schema_name: &str) -> SparkResult<SchemaReference> {
-    let mut idents = parse_identifiers(schema_name)?;
```
linhr, line 47: In general, we would not have any SQL parsing in `spec::Plan`. All multi-part identifiers should have been tokenized before building the plan.

(There are still a few `unparsed` fields in `spec::Plan`, and I'll turn them into "parsed" data in follow-up PRs.)

crates/spark-connect-server/src/utils.rs
```diff
@@ -1,22 +0,0 @@
-use std::collections::HashMap;
-
-#[derive(Debug, Clone)]
-pub(crate) struct CaseInsensitiveHashMap<T> {
```
linhr, line 4: I removed this because I realized that the better way to do this is to define `struct LowerCase(String)` and implement required traits so that it can be used as map keys. In this way, we do not need to wrap `HashMap` and we will get all the functionalities of `HashMap` for free.

crates/framework-common/src/utils.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_export]
+macro_rules! unwrap_or {
```
shehabgamin, line 2: nice!

PR #50, opened 2024-06-05
Implement show string

linhr:
1. Implement `ShowString`.
2. Implement `HtmlString`.
3. Change data type for a few integer fields in the plan specification.

### DIFF
--- a/crates/framework-plan/src/extension/logical/range.rs
+++ b/crates/framework-plan/src/extension/logical/range.rs
@@ -90,13 +90,14 @@ impl RangeNode {
             return plan_err!("the number of partitions must be greater than 0");
         }
         let fields = vec![Field::new("id", DataType::Int64, false)];
+        let schema = Arc::new(DFSchema::from_unqualifed_fields(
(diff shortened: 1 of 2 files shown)


crates/framework-plan/src/extension/logical/show_string.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    fn show_footer(&self, num_rows: usize, has_more: bool) -> String {
+        match (has_more, num_rows) {
+            (true, 1) => "only showing top 1 row\n".to_string(),
+            (true, n) => format!("only showing top {} rows\n", n),
+            _ => "".to_string(),
+        }
+    }
+
+    fn show_string(&self, batch: &RecordBatch, has_more: bool) -> Result<String> {
```
linhr, line 92: We cannot use functions such as `pretty_format_batches` from Arrow, since there are minor differences in the resulted table string (padding, alignment, etc.). But the function here is inspired by the implementation in Arrow.

PR #52, opened 2024-06-07
Support set operations and sort within partitions

linhr:
1. Support set operations.
2. Support sort within partitions.
3. Fix an issue with the `ShowString` physical plan.


crates/framework-plan/src/extension/physical/show_string.rs
```diff
@@ -61,6 +61,14 @@ impl ExecutionPlan for ShowStringExec {
         vec![Distribution::SinglePartition]
     }
 
+    fn maintains_input_order(&self) -> Vec<bool> {
```
linhr, line 64: This is needed, otherwise the physical plan optimizer may remove the child `SortExec`.

PR #53, opened 2024-06-09
PySpark UDF Cont.

shehabgamin:


scripts/spark-tests/run-server.sh
```diff
@@ -2,18 +2,5 @@
...
-
-export RUST_LOG="${RUST_LOG:-framework_spark_connect=debug}"
-export RUST_BACKTRACE="${RUST_BACKTRACE:-full}"
-export RUST_MIN_STACK="${RUST_MIN_STACK:-8388608}"
-# We have to set `PYTHONPATH` even if we are using the virtual environment.
-# This is because the Python executable is the Rust program itself, and there is
-# no `pyvenv.cfg` at its required location (one directory above the executable).
-export PYTHONPATH="${project_path}/python/.venv/lib/python${python_version}/site-packages"
-
+source "$(dirname "$0")/prepare-server.sh"
```
linhr, line 5: We can probably revert the changes here since I found that we can set up debugger without using a script: #55.

shehabgamin: @linhr Nice find! I think we should still keep the changes though. It's useful to be able to prepare the server without actually starting the server.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
```diff
@@ -23,21 +25,21 @@ pub const PY_SPARK_SQL_GROUPED_MAP_PANDAS_UDF_WITH_STATE: i32 = 208;
 pub const PY_SPARK_SQL_TABLE_UDF: i32 = 300;
 pub const PY_SPARK_SQL_ARROW_TABLE_UDF: i32 = 301;
 
-pub fn is_pyspark_arrow_udf(udf_type: i32) -> bool {
-    udf_type == PY_SPARK_SQL_ARROW_BATCHED_UDF
-        || udf_type == PY_SPARK_SQL_MAP_ARROW_ITER_UDF
-        || udf_type == PY_SPARK_SQL_ARROW_TABLE_UDF
+pub fn is_pyspark_arrow_udf(udf_type: &i32) -> bool {
```
linhr, line 28: Is there any reason for these changes?

I feel comparing values are more intuitive here, even though comparing by reference is also correct. There is probably no performance difference since the value is an integer.

You can write `is_pyspark_arrow_udf(*x)` if `x` is `&i32`.

crates/framework-common/src/config.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SparkUdfConfig {
+    pub timezone: ConfigKeyValue,
+    pub pandas_window_bound_types: ConfigKeyValue,
+    pub pandas_grouped_map_assign_columns_by_name: ConfigKeyValue,
+    pub pandas_convert_to_arrow_array_safely: ConfigKeyValue,
+    pub arrow_max_records_per_batch: ConfigKeyValue,
+}
```
linhr, line 17: 1. I'm not sure if we want `SparkUdfConfig` in `framework-common` either since it is specific to the plan resolver. I feel it's better to assume that `SparkUdfConfig` is always part of `PlanConfig`, so that we don't need to duplicate general information such as the session timezone.
2. Let's only have the values in `SparkUdfConfig`. I don't think we should store the key since they are the same for all the sessions.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
```diff
@@ -132,15 +131,63 @@ impl PartialEq for PartialPySparkUDF {
...
+    let mut data: Vec<u8> = Vec::new();
+    data.extend(&eval_type.to_be_bytes()); // Add eval_type for extraction in visit_bytes
+    if is_pyspark_arrow_udf(eval_type) || is_pyspark_pandas_udf(eval_type) {
+        let configs = [
+            &spark_udf_config.timezone,
+            &spark_udf_config.pandas_window_bound_types,
+            &spark_udf_config.pandas_grouped_map_assign_columns_by_name,
+            &spark_udf_config.pandas_convert_to_arrow_array_safely,
+            &spark_udf_config.arrow_max_records_per_batch,
+        ];
```
linhr, line 169: As suggested in the other comment, let's construct the key-value pairs here. I'd prefer to store only minimum information in `PlanConfig`, and `PlanConfig` should not address the concern how the information is used (e.g. as key-value pairs to be passed to Python workers).

crates/framework-python/src/udf/mod.rs
```diff
@@ -1,3 +1,115 @@
...
+    }
+}
+
+impl PythonFunction for PartialPySparkUDF {
+    fn get_inner<'py>(&self, py: Python<'py>) -> Bound<'py, PyAny> {
+        self.0.clone_ref(py).into_bound(py)
+    }
+}
+
+pub trait CommonPythonUDF {
```
linhr, line 32: Let's add a TODO to return `DataFusionError::External` for `PyErr`, since we're now able to propagate errors after #59.

crates/framework-spark-connect/src/session.rs
```diff
@@ -66,15 +81,52 @@ impl Session {
...
+            .config
+            .get(SPARK_SQL_SESSION_TIME_ZONE)?
+            .map(|x| x.to_string())
+            .unwrap_or_else(|| "UTC".into());
+        let spark_udf_config = SparkUdfConfig {
+            timezone: ConfigKeyValue {
+                key: "spark.sql.session.timeZone".to_string(),
+                value: Some(time_zone.clone()),
+            },
+            pandas_window_bound_types: ConfigKeyValue {
```
linhr, line 95: Let's add a FIXME here. It seems to me that `pandas_window_bound_types` is not a proper Spark configuration.

crates/framework-common/src/config.rs
```diff
@@ -0,0 +1,23 @@
+use std::fmt::Debug;
+use std::hash::Hash;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum TimestampType {
```
linhr: This is only meaningful in the context of the plan resolver right now. So I'd prefer to move this back to `framework-plan`.

In general, we can consider adding things to `framework-common` when there are at least two use cases in different crates.

crates/framework-plan/src/function/common.rs
```diff
@@ -38,6 +38,13 @@ pub(crate) fn get_three_arguments(
 pub(crate) struct FunctionBuilder;
 
 impl FunctionBuilder {
+    pub fn nullary<F>(f: F) -> Function
+    where
+        F: Fn() -> expr::Expr + Send + Sync + 'static,
+    {
+        Arc::new(move |_args| Ok(f()))
```
linhr: Let's return `Err` if `args` is not empty.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
```diff
@@ -132,15 +131,63 @@ impl PartialEq for PartialPySparkUDF {
 impl Eq for PartialPySparkUDF {}
 
 pub fn deserialize_partial_pyspark_udf(
+    python_version: &str,
     command: &[u8],
     eval_type: &i32,
     num_args: &i32,
```
linhr, line 138: I think having `i32` as the argument type is more readable.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
```diff
@@ -77,25 +76,16 @@ impl<'de> Visitor<'de> for PartialPySparkUDFVisitor {
...
-        data.extend(&self.num_args.to_be_bytes()); // num_args
-        for index in 0..self.num_args {
-            data.extend(&index.to_be_bytes()); // arg_offsets
-        }
-        data.extend(&1i32.to_be_bytes()); // num functions
-        data.extend(&(v.len() as i32).to_be_bytes()); // len of the function
-        data.extend_from_slice(v);
-        let data: &[u8] = data.as_slice();
-
+        let (eval_type_bytes, v) = v.split_at(std::mem::size_of::<i32>());
```
linhr, line 80: `split_at()` will panic if the slice is too short.

crates/framework-python/src/udf/mod.rs
```diff
@@ -1,3 +1,115 @@
...
+    fn get_python_function<'py>(&self, py: Python<'py>) -> Result<Bound<'py, PyAny>> {
+        let python_function: Bound<PyAny> = self
+            .python_function()
+            .get_inner(py)
+            .get_item(0)
+            .map_err(|err| DataFusionError::Internal(format!("python_function {}", err)))?;
+        Ok(python_function)
+    }
+
+    fn get_python_builtins<'py>(&self, py: Python<'py>) -> Result<Bound<'py, PyModule>> {
```
linhr: It's unclear to me why we define `get_python_xxx()` as trait methods. The implementation is not meant to be customized, so it might be better to simply define top-level utility functions.

crates/framework-common/src/config.rs
shehabgamin: @linhr 
1. `framework-python` and `framework-plan` both need `SparkUdfConfig` which is why i can't put it in framework-plan.
2. Spark's `read_udfs` in `worker.py` expects both the keys and values.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
shehabgamin: @linhr Do you want the callers of `deserialize_partial_pyspark_udf` to pass along the entire `PlanConfig`?

linhr: We can construct the key-value pairs on the fly in `framework-plan` and pass that as an argument to `deserialize_partial_pyspark_udf`. `framework-python` only needs the config as `Vec<(String, String)>` since it only needs to pass the entire config to the worker without accessing individual keys in the config. It's cleaner to keep both `PlanConfig` and the inner `SparkUdfConfig` as data structure specific to `framework-plan`.

I think `framework-python` is similar to `worker.py` in PySpark, while `framework-plan` is like the Spark Scala code which prepares the config and serializes it using some wire protocol.

crates/framework-python/src/udf/mod.rs
shehabgamin: Good point haha

crates/framework-python/src/cereal/partial_pyspark_udf.rs
shehabgamin: @linhr Is that possible considering that when we do this `eval_type` is guaranteed to always be an i32?
```
pub fn build_pyspark_udf_payload(
    command: &[u8],
    eval_type: &i32,
    num_args: &i32,
    spark_udf_config: &SparkUdfConfig,
) -> Vec<u8> {
    let mut data: Vec<u8> = Vec::new();
    data.extend(&eval_type.to_be_bytes()); // Add eval_type for extraction in visit_bytes
```

crates/framework-python/src/cereal/partial_pyspark_udf.rs
linhr: Actually, I just realized that even `framework-python` does not have access to the Spark config keys defined in `framework-spark-connect`, so we cannot construct key-value pairs on the fly. In this case, we should probably only store `HashMap<String, String>` as `spark_udf_config` in `PlanConfig`, instead of leaking the Spark-specific implementation details all the way to `framework-{plan,python,common}`. None of these crates needs to know what's inside in the config.

Sorry for the confusion. I didn't realize this when we discussed the design the other day.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
shehabgamin: Imo the readibily is the same. Happy to change it if you feel strongly about it.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
shehabgamin: Upstream (everywhere) we get the value as a reference and I thought it made more sense to make those changes instead of de-referencing with `*`

crates/framework-python/src/cereal/partial_pyspark_udf.rs
shehabgamin: @linhr Do you want me to add a `TODO` and pick this up in my next pr, since i'll be touching `pandas_window_bound_types` and our requirements may change?

linhr: That works for me!

crates/framework-python/src/cereal/partial_pyspark_udf.rs
linhr: I see. It might be a good idea to add a note here, since the assumption across functions is not obvious to the reader. Also the function may be used in other use cases in the future, so it's good to be defensive in the input.

In fact, I'm not particular concerned with this particular case. A more interesting question is how we prevent panic as our codebase grows, especially in cases not captured by the linter.

crates/framework-python/src/cereal/partial_pyspark_udf.rs
linhr: I see. This does not affect correctness, though people coming from another programming language may think comparing by reference is wrong. (I was confused the first time I looked at the code.) I'm fine to keep this for now.

PR #55, opened 2024-06-11
Update README

linhr:


README.md
```diff
@@ -104,13 +125,52 @@ jq -r -f scripts/spark-tests/show-passed-tests.jq \
...
+### Running the Rust Debugger in RustRover
+
+Since we use PyO3 to support Python binding in Rust, we need some additional setup to run the Rust debugger in RustRover.
+In **Run** > **Edit Configurations**, add a new **Cargo** configuration with the following settings:
+
+1. Name: **Run Spark Connect server** (You can use any name you like.)
+2. Command: `run -p framework-spark-connect`
+3. Environment Variables:
+    - (required) `PYTHONPATH`: `python/.venv/lib/python<version>/site-packages` (Please replace `<version>` with the actual Python version, e.g. `3.11`.)
+    - (required) `PYO3_PYTHON`: `<project>/python/.venv/bin/python` (Please replace `<project>` with the actual project path. **This must be an absolute path.**)
```
linhr, line 160: The debugger seems to work if we correctly set the `PYO3_PYTHON` environment variable. In this way we do not need to run a separate script as a "Before launch" step. In fact, I'm not sure if running the script works, since I notice that RustRover runs the script as a separate process.

PR #57, opened 2024-06-12
Implement plan configuration

linhr:
This PR defines the `PlanConfig` struct that is used by the plan resolver and catalog operations.

`PlanConfig` provides an abstraction over Spark configuration. For example, with `PlanConfig` we are now able to get the session time zone when converting data types and literals.

### DIFF
--- a/crates/framework-plan/src/config.rs
+++ b/crates/framework-plan/src/config.rs
@@ -15,6 +15,8 @@ pub trait DataTypeFormatter: DynObject + Debug + Send + Sync {
     fn to_simple_string(&self, data_type: spec::DataType) -> PlanResult<String>;
 }
 
+impl_dyn_object_traits!(DataTypeFormatter);
+
 #[derive(Debug, PartialEq, Eq, Hash)]
 struct DefaultDataTypeFormatter;
 
@@ -23,8 +25,6 @@ impl DataTypeFormatter for DefaultDataTypeFormatter {
         Err(PlanError::unsupported("default data type formatter"))
     }
 }
-
-impl_dyn_object_traits!(DataTypeFormatter);
 
 // The generic type parameter is used to work around the issue deriving `PartialEq` for `dyn` trait.
 // See also: https://github.com/rust-lang/rust/issues/78808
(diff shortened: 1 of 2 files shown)


crates/framework-plan/src/resolver/data_type.rs
```diff
@@ -0,0 +1,300 @@
...
+            }
+            DataType::Struct { fields } => Ok(adt::DataType::Struct(self.resolve_fields(fields)?)),
+            DataType::Decimal { scale, precision } => {
+                Ok(adt::DataType::Decimal128(precision, scale))
+            }
+            DataType::Char { .. } => Ok(adt::DataType::Utf8),
+            DataType::VarChar { .. } => Ok(adt::DataType::Utf8),
+            DataType::Date => Ok(adt::DataType::Date32),
+            DataType::Timestamp => {
+                let timezone: Arc<str> = Arc::from(self.config.time_zone.as_str());
```
linhr, line 46: The data type conversion function is moved from the `framework-common` crate. The only difference is that we are now able to resolve the time zone.

crates/framework-plan/src/resolver/literal.rs
```diff
@@ -0,0 +1,95 @@
...
+                    value,
+                    precision,
+                    scale,
+                } = decimal;
+                Ok(ScalarValue::Decimal128(Some(value), precision, scale))
+            }
+            Literal::String(x) => Ok(ScalarValue::Utf8(Some(x))),
+            Literal::Date { days } => Ok(ScalarValue::Date32(Some(days))),
+            Literal::Timestamp { microseconds } => {
+                let timezone: Arc<str> = Arc::from(self.config.time_zone.as_str());
```
linhr, line 34: Same here. We are now able to resolve the time zone when converting literals to scalar values.

crates/framework-common/src/object.rs
```diff
@@ -0,0 +1,42 @@
+use std::any::Any;
+use std::hash::{Hash, Hasher};
+
+/// A trait that facilitates deriving `PartialEq`, `Eq`, and `Hash` for `dyn` trait objects.
```
shehabgamin, line 4: Nice!

PR #59, opened 2024-06-13
Map errors in gRPC status

linhr:
This PR continues our previous attempt to return better error details in gRPC invocations. The error mapping can be further refined in the future.

### DIFF
--- a/crates/framework-spark-connect/src/error.rs
+++ b/crates/framework-spark-connect/src/error.rs
@@ -253,7 +253,9 @@ impl From<SparkError> for Status {
             }
             SparkError::ArrowError(e)
             | SparkError::DataFusionError(DataFusionError::ArrowError(e, _)) => match e {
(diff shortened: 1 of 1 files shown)


crates/framework-spark-connect/src/error.rs
```diff
@@ -146,44 +148,159 @@ where
...
 impl From<SparkError> for Status {
     fn from(error: SparkError) -> Self {
         match error {
+            SparkError::ArrowError(ArrowError::ExternalError(e))
+            | SparkError::DataFusionError(DataFusionError::ArrowError(
+                ArrowError::ExternalError(e),
+                _,
+            ))
+            | SparkError::DataFusionError(DataFusionError::External(e)) => {
+                if let Some(e) = e.downcast_ref::<PyErr>() {
```
linhr, line 247: We can now return Python errors to the client by wrapping `PyErr` using `DataFusionError::External`.

PR #60, opened 2024-06-13
Update data type parsing functions

linhr:


crates/framework-spark-connect/src/proto/data_type.rs
```diff
@@ -18,11 +18,15 @@ pub(crate) const SPARK_DECIMAL_SYSTEM_DEFAULT_PRECISION: u8 = 38;
...
+/// Parse a Spark data type string of various forms.
+/// Reference: org.apache.spark.sql.connect.planner.SparkConnectPlanner#parseDatatypeString
+pub(crate) fn parse_spark_data_type(schema: &str) -> SparkResult<spec::DataType> {
+    if let Ok(dt) = parse_data_type(schema) {
         Ok(dt)
-    } else if let Ok(dt) = parse_spark_data_type(format!("struct<{schema}>").as_str()) {
+    } else if let Ok(dt) = parse_data_type(format!("struct<{schema}>").as_str()) {
+        // The SQL parser supports both `struct<name: type, ...>` and `struct<name type, ...>` syntax.
+        // Therefore, by wrapping the input with `struct<...>`, we do not need separate logic
+        // to parse table schema input (`name type, ...`).
```
linhr, line 29: It seems that we already support table schema parsing, so this function has the same behavior as the `parseDatatypeString` Scala function. So I just renamed the function to make things cleaner.

PR #61, opened 2024-06-17
Assign better names for projection expressions

linhr:
1. Assign better names for literal and function expressions, so that we can have column names that match those in Spark SQL.
2. Define `PlanResolverState`. This is not used in this PR yet.

### DIFF
--- a/crates/framework-plan/src/function/common.rs
+++ b/crates/framework-plan/src/function/common.rs
@@ -16,9 +16,7 @@ impl FunctionBuilder {
         F: Fn() -> expr::Expr + Send + Sync + 'static,
     {
         Arc::new(move |args| {
-            if args.len() != 0 {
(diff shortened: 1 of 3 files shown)


crates/framework-spark-connect/src/schema.rs
```diff
@@ -1,190 +1,11 @@
...
-    pub fn to_simple_string(&self) -> SparkResult<String> {
-        match &self.kind {
-            Some(kind) => kind.to_simple_string(),
-            None => Ok("none".to_string()),
-        }
-    }
-}
-
-impl sdt::Kind {
-    pub fn to_simple_string(&self) -> SparkResult<String> {
```
linhr, line 44: The logic is moved to `crates/framework-plan/src/formatter.rs`.

PR #69, opened 2024-06-27
Generate field names in logical plan schema

linhr:
1. Use generated field names for resolved logical plans, so that we can support duplicated field names in the schema.
2. Implement "named expression", which is a cleaner way to derive names for complex expressions. Previously we add alias at each level when resolving expressions.
3. Add the logic to handle plan ID, so that we can support PySpark logical plans such as `df.select(df["a"])`.

### DIFF
--- a/crates/framework-python/src/udf/pyspark_udf.rs
+++ b/crates/framework-python/src/udf/pyspark_udf.rs
@@ -13,7 +13,8 @@ use crate::cereal::partial_pyspark_udf::PartialPySparkUDF;
 use crate::cereal::{is_pyspark_arrow_udf, is_pyspark_pandas_udf, PY_SPARK_SQL_BATCHED_UDF};
 use crate::udf::{
     build_pyarrow_array_kwargs, get_pyarrow_array_function, get_pyarrow_output_data_type,
-    get_python_builtins_list_function, get_python_builtins_str_function, PythonFunctionType,
+    get_python_builtins_list_function, get_python_builtins_str_function, get_udf_name,
+    PythonFunctionType,
 };
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


crates/framework-python/src/udf/pyspark_udtf.rs
```diff
@@ -34,11 +35,15 @@ pub struct PySparkUserDefinedTable {
 }
 
 impl PySparkUserDefinedTable {
-    pub fn new(return_schema: SchemaRef, batches: Vec<RecordBatch>) -> Self {
-        Self {
+    pub fn try_new(return_schema: SchemaRef, batches: Vec<RecordBatch>) -> Result<Self> {
+        let batches = batches
+            .into_iter()
+            .map(|batch| cast_record_batch(batch, return_schema.clone()))
```
linhr, line 41: The casting is needed since the Arrow data returned by Python functions may have a different schema. The problem did not arise previously, since previously the response was streamed directly from `MemoryExec`. Now we have a projection plan on top of `MemoryExec` and the projected `RecordBatch` cannot be created when there is a schema mismatch.

crates/framework-common/src/utils.rs
shehabgamin, line 1: Nice!

crates/framework-python/src/udf/pyspark_udf.rs
```diff
@@ -35,6 +35,16 @@ impl PySparkUDF {
         python_function: PartialPySparkUDF,
         output_type: DataType,
     ) -> Self {
+        // We generate a unique function name with the memory address of the Python function.
+        // Otherwise, lambda functions with the name "<lambda>" will be treated as the same function
+        // by logical plan optimization rules (e.g. common sub-expression elimination), resulting in
+        // incorrect logical plans.
+        // TODO: Apply this to all types of Python UDFs.
```
shehabgamin: @linhr This `TODO` seems simple enough to finish in this PR

python/src/app/__main__.py
```diff
@@ -92,6 +93,8 @@ def add(x, y):
     print(df.select(add(F.col("a"), F.col("a"))).toPandas())
 
     # FIXME: not working
+    # print(spark.sql("SELECT 1 AS text WHERE $1 > 'a'", ["b"]).toPandas())
```
shehabgamin, line 96: @linhr Why are these now failing?

crates/framework-python/src/udf/pyspark_udf.rs
linhr: Done!

python/src/app/__main__.py
linhr: This is not related to this PR. This has been failing for a while after we have `PlanNode::WithParameters`. I just moved it under `FIXME`. To fix this we probably need to store parameter type information in resolver state so that the placeholder can be processed by DataFusion.

PR #71, opened 2024-06-28
Update To Patched SqlParser 47

shehabgamin:


crates/framework-spark-connect/tests/gold_data/plan/plan_select.json
```diff
@@ -2064,39 +2064,7 @@
...
-                    "metadata": null
-                  }
-                }
-              ]
-            },
-            "planId": null,
-            "sourceInfo": null
-          }
-        }
+        "failure": "invalid argument: sql parser error: Expected an expression, found: from at Line: 1, Column: 13"
```
shehabgamin, line 2067: Seems to be failing at `let statement = parser.parse_statement()?;` in `parse_sql_statement`

linhr: This is the Scala test code in Spark:

```scala
assertEqual("select from tbl", OneRowRelation().select($"from".as("tbl")))
```

This is such a weird syntax. Maybe it's fine to not support this.

shehabgamin: Yeah it's very weird syntax. I believe it's the equivalent of SELECT * though.

PR #82, opened 2024-07-06
AST To Plan - DDL Pt. 2

shehabgamin:


crates/framework-sql/src/statement.rs
```diff
@@ -85,6 +92,126 @@ fn parse_explain_statement(parser: &mut Parser) -> SqlResult<Statement> {
...
+    if parser.parse_keyword(Keyword::LIKE) {
+        return Err(SqlError::todo("CREATE TABLE LIKE"));
+    }
+    let (columns, mut constraints): (Vec<ast::ColumnDef>, Vec<ast::TableConstraint>) =
+        parser.parse_columns()?;
+    let mut file_format: Option<String> = None;
+    if parser.parse_keyword(Keyword::USING) {
+        file_format = Some(parse_option_value(parser)?);
+    }
+    // FIXME: Function to loop and do parse_one_of_keywords for all table clauses + hive format
```
shehabgamin: Will be addressed in this PR.

crates/framework-common/src/spec/plan.rs
```diff
@@ -734,6 +734,7 @@ pub struct StorageLevel {
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
linhr, line 737: Nice catch!

crates/framework-plan/src/catalog/table.rs
```diff
@@ -120,6 +123,50 @@ impl<'a> CatalogManager<'a> {
...
+            file_sort_order,
+            if_not_exists,
+            or_replace: _, // TODO: support or_replace
+            unbounded,
+            options,
+            definition,
+        } = create_table
+        else {
+            unreachable!()
+        };
```
linhr: You could do this via `if let ... else ...` so that `matches!` is not needed.

We should probably start having separate structs for each `CatalogCommand` variant since the number of fields is getting large. But this can be done in follow-up PRs.

crates/framework-plan/src/config.rs
```diff
@@ -56,6 +62,9 @@ impl Default for PlanConfig {
                     value: None,
                 },
             },
+            default_bounded_table_file_format: "PARQUET".to_string(),
+            default_unbounded_table_file_format: "ARROW".to_string(),
+            default_warehouse_directory: "/sail-warehouse".to_string(),
```
linhr:
```suggestion
            default_warehouse_directory: "sail-warehouse".to_string(),
```

I'd prefer using the current directory instead of the root directory by default, since the user may not have write permission to the root directory.

crates/framework-plan/src/extension/logical/catalog.rs
```diff
@@ -283,22 +291,57 @@ impl CatalogCommand {
...
                     .is_ok();
                 let rows = vec![SingleValueMetadata { value }];
-                build_record_batch(schema, &rows)?
+                build_record_batch(self_schema, &rows)?
             }
-            CatalogCommand::CreateTable { .. } => {
-                return not_impl_err!("create table");
+            CatalogCommand::CreateTable {
+                table,
+                schema,
```
linhr, line 298: You can do `schema: table_schema,` so that you don't need to rename the outer `schema` variable.

If you plan to keep the rename, I feel `command_schema` is a more descriptive name than `self_schema`.

crates/framework-spark-connect/src/session.rs
```diff
@@ -122,12 +122,26 @@ impl Session {
...
         Ok(Arc::new(PlanConfig {
             time_zone,
             // TODO: get the default timestamp type from configuration
             timestamp_type: TimestampType::TimestampLtz,
             plan_formatter: Arc::new(DefaultPlanFormatter),
             spark_udf_config,
+            default_bounded_table_file_format,
+            default_unbounded_table_file_format: PlanConfig::default()
+                .default_unbounded_table_file_format,
+            default_warehouse_directory,
```
linhr:
```suggestion
            default_bounded_table_file_format,
            default_warehouse_directory,
            ..PlanConfig::default(),
```

crates/framework-sql/src/parse.rs
```diff
@@ -0,0 +1,104 @@
...
+    let next_token = parser.next_token();
+    let value = match next_token.token {
+        // e.g. things like "snappy" or "gzip" that may be keywords
+        Token::Word(word) => Ok(ast::Value::SingleQuotedString(word.value)),
+        Token::SingleQuotedString(s) => Ok(ast::Value::SingleQuotedString(s)),
+        Token::DoubleQuotedString(s) => Ok(ast::Value::DoubleQuotedString(s)),
+        Token::EscapedStringLiteral(s) => Ok(ast::Value::EscapedStringLiteral(s)),
+        Token::Number(ref n, l) => match n.parse() {
+            Ok(n) => Ok(ast::Value::Number(n, l)),
+            // The tokenizer should have ensured `n` is an integer
```
linhr:
```suggestion
            // The tokenizer should have ensured `n` is a number
```

I guess floating point numbers are also valid here?

PR #96, opened 2024-07-17
Fix SQL query issues for TPC-H benchmark

linhr:


python/framework/examples/spark/tpch.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
```
linhr, line 1: This example script can be used to send queries or explore the data when we troubleshoot the TPC-H benchmark. But we may remove this once we have a better benchmark setup.

crates/framework-plan/src/function/scalar/conditional.rs
```diff
@@ -1,5 +1,22 @@
...
+use crate::error::PlanResult;
 use crate::function::common::Function;
 
+fn case(args: Vec<expr::Expr>) -> PlanResult<expr::Expr> {
+    let mut when_then_expr = Vec::new();
+    let mut iter = args.into_iter();
+    while let (Some(condition), Some(result)) = (iter.next(), iter.next()) {
+        when_then_expr.push((Box::new(condition), Box::new(result)));
+    }
+    let else_expr = iter.next().map(Box::new); // The last element, if any, is the else expression
```
linhr, line 12: This is quite elegant!

PR #101, opened 2024-07-23
Update executor logic

linhr:
1. Implement synchronization for the executor state, so that we do not need to close the response stream for the `ExecutePlan` request when `ReleaseExecute` requests arrive concurrently. (The client keeps sending `ReleaseExecute` requests in the background to free the buffer on the server side.) This avoids frequent `ReattachExecute` requests made by the client to resume plan execution.
2. Add a workaround to terminate the server process when tasks in the execution stream are still running.

### DIFF
--- a/crates/framework-spark-connect/src/executor.rs
+++ b/crates/framework-spark-connect/src/executor.rs
@@ -63,7 +63,7 @@ impl ExecutorBuffer {
     }
 
     fn add(&mut self, output: ExecutorOutput) {
-        if self.inner.len() == self.inner.capacity() {
+        if self.inner.len() >= self.inner.capacity() {
             self.inner.pop_front();
         }
         self.inner.push_back(output);


crates/framework-spark-connect/src/executor.rs
```diff
@@ -169,64 +196,81 @@ impl Executor {
...
-            if let ExecutorState::Pending(context) = &mut self.state {
-                context.remove_output_until(&response_id);
-            }
+            buffer.lock()?.remove_until(&response_id);
         }
         Ok(())
     }
 }
 
-impl Drop for Executor {
```
linhr, line 222: Explicit `Drop` implementation is not needed since the resources are dropped when they go out of scope.

crates/framework-spark-connect/src/executor.rs
```diff
@@ -47,6 +49,38 @@ impl ExecutorOutput {
...
+    fn new() -> Self {
+        // TODO: use "spark.connect.execute.reattachable.observerRetryBufferSize"
+        // TODO: limit the size based on serialized message size instead of element count
+        Self {
+            inner: VecDeque::with_capacity(128),
+        }
+    }
+
+    fn add(&mut self, output: ExecutorOutput) {
+        if self.inner.len() == self.inner.capacity() {
```
shehabgamin: Maybe this should be >= incase there are race conditions.

linhr: Actually `self.inner.len()` won’t be greater than `self.inner.capacity()`. (Also there will be no race condition since only one thread can have `&mut` at a time.)

But let me use `>=` here to be more consistent with the style of defensive coding.

PR #104, opened 2024-07-25
Expr InList

shehabgamin:


crates/framework-plan/src/resolver/expression.rs
```diff
@@ -1000,6 +1008,22 @@ impl PlanResolver<'_> {
...
+        expr: spec::Expr,
+        list: Vec<spec::Expr>,
+        negated: bool,
+        schema: &DFSchema,
+        state: &mut PlanResolverState,
+    ) -> PlanResult<NamedExpr> {
+        let expr = Box::new(self.resolve_expression(expr, schema, state).await?);
+        let list = self.resolve_expressions(list, schema, state).await?;
+        Ok(NamedExpr::new(
+            vec!["in_list".to_string()],
```
linhr, line 1024: Let's add a TODO to construct better names for the expression (e.g. `a IN (b, c)`).

PR #110, opened 2024-07-29
Support PySpark shell

linhr:


python/framework/spark/__main__.py
```diff
@@ -0,0 +1,65 @@
...
+
+    spark = SparkSession.builder.remote(f"sc://localhost:{port}").getOrCreate()
+    namespace = {"spark": spark}
+    readline.parse_and_bind("tab: complete")
+    readline.set_completer(Completer(namespace).complete)
+
+    python_version = platform.python_version()
+    (build_number, build_date) = platform.python_build()
+    # Simulate the messages similar to `shell.py` in PySpark.
+    banner = rf"""Welcome to
```
shehabgamin, line 21: Should we have this say `Welcome to LakeSail (ASCII Art Font) PySpark version` instead? People already know it's a Spark shell if they're running that command.

python/framework/spark/__main__.py
```diff
@@ -0,0 +1,65 @@
...
+      ____              __
+     / __/__  ___ _____/ /__
+    _\ \/ _ \/ _ `/ __/  '_/
+   /__ / .__/\_,_/_/ /_/\_\   version {pyspark.__version__}
+      /_/
+
+Using Python version {python_version} ({build_number}, {build_date})
+Client connected to the Spark Connect server at localhost:{port}
```
shehabgamin:
```suggestion
Client connected to the Sail Spark Connect server at localhost:{port}
```

PR #116, opened 2024-08-01
Add release workflow

linhr:
I generated the workflow using `maturin generate-ci github` and then manually edited the file.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
-name = "PySail"
-version = "0.0.1"
+name = "pysail"
+version = "0.1.0.dev0"
```
linhr, line 3: <https://packaging.python.org/en/latest/specifications/version-specifiers/>

PR #122, opened 2024-08-07
Expanding Function and DDL Support

shehabgamin:


crates/sail-plan/src/catalog/table.rs
```diff
@@ -158,6 +159,12 @@ impl<'a> CatalogManager<'a> {
             }));
             // TODO: process the output
             _ = self.ctx.execute_logical_plan(ddl).await?;
+            if let Some(copy_to_plan) = copy_to_plan {
+                _ = self
+                    .ctx
+                    .execute_logical_plan((*copy_to_plan).clone())
+                    .await?;
```
linhr, line 168: This does not actually execute the copy. `execute_logical_plan` only execute DDL statements and other statements remain unchanged. We need to execute the physical plan instead ([reference](https://github.com/lakehq/sail/blob/0a5084694c04b719dcdd102d779569baa36cc423/crates/sail-spark-connect/src/executor.rs#L281)).

I'm fine to simply add a TODO comment. It might be easier to work with this once we have better unit tests for this.

crates/sail-plan/src/extension/function/raise_error.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Null)
+    }
+
+    fn invoke(&self, args: &[ColumnarValue]) -> Result<ColumnarValue> {
+        if args.len() != 1 {
+            return Err(DataFusionError::Internal(format!(
+                "raise_error should only be called with two arguments, got {}",
```
linhr:
```suggestion
                "raise_error should only be called with one argument, got {}",
```

crates/sail-plan/src/function/scalar/misc.rs
```diff
@@ -1,12 +1,48 @@
...
+            args.len()
+        )));
+    }
+
+    let (col, list) = args.at_least_one()?;
+    let err_msg = if list.len() == 1 {
+        list[0].clone()
+    } else {
+        lit(ScalarValue::Utf8(Some(format!("'{}' is not true!", col))))
+    };
```
linhr: You can structure the code in this way to avoid `list[0].clone()`.

```rust
let (col, err_msg) = if args.len() == 1 {
    let col = args.one()?;
    // create the error message
} else if args.len() == 2 {
    args.two()?
} else {
    return Err(PlanError::invalid(format!(
        "assert_true expects one or two arguments, got {}",
        args.len()
    )));
}
```

crates/sail-plan/src/function/scalar/misc.rs
```diff
@@ -1,12 +1,48 @@
...
+
+    Ok(expr::Expr::Case(expr::Case {
+        expr: None,
+        when_then_expr: vec![(
+            Box::new(expr::Expr::Not(Box::new(col.clone()))),
+            Box::new(expr::Expr::ScalarFunction(expr::ScalarFunction {
+                func: Arc::new(ScalarUDF::from(RaiseError::new())),
+                args: vec![err_msg.clone()],
```
linhr: I guess `.clone()` are not needed here.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -1543,16 +1546,19 @@ impl PlanResolver<'_> {
...
-        // TODO: handle query
-        //  1. Resolve query to get schema
-        //  2. (optional) if columns are specified in the definition, validate the schema
-        //  3. create external table
-        //  4. fill external table from query table (copy to)
-        let fields = self.resolve_fields(schema.fields)?;
-        let schema = Arc::new(DFSchema::from_unqualified_fields(fields, HashMap::new())?);
+
+        let (schema, query_logical_plan) = if let Some(query) = query {
+            let logical_plan = self.resolve_query_plan(*query, state).await?;
```
linhr, line 1555: Let's add a FIXME here. The query plan has all columns renamed to `#1`, `#2`, etc. so I think there is more work here. We can revisit this when we have unit tests that cover this.

crates/sail-plan/src/function/scalar/misc.rs
```diff
@@ -1,12 +1,48 @@
...
+    }
+
+    let (col, list) = args.at_least_one()?;
+    let err_msg = if list.len() == 1 {
+        list[0].clone()
+    } else {
+        lit(ScalarValue::Utf8(Some(format!("'{}' is not true!", col))))
+    };
+
+    Ok(expr::Expr::Case(expr::Case {
```
linhr, line 31: This seems a bit fragile to me. Even if the input is all `true`, `RaiseError` may still be executed as part of the child expression evaluation, if the physical plan executor is not optimized enough.

Let's add a FIXME to revisit this in the future.

crates/sail-plan/src/function/scalar/array.rs
```diff
@@ -12,13 +15,59 @@ fn array_repeat(element: expr::Expr, count: expr::Expr) -> expr::Expr {
...
+        data_type: DataType::Int64,
+    });
+    let end = expr::Expr::BinaryExpr(BinaryExpr {
+        left: Box::new(start.clone()),
+        op: Operator::Plus,
+        right: Box::new(expr::Expr::BinaryExpr(BinaryExpr {
+            left: Box::new(length),
+            op: Operator::Minus,
+            right: Box::new(lit(ScalarValue::Int64(Some(1)))),
+        })),
```
linhr, line 38: DataFusion mentions that "the to index is exclusive" for `array_slice`. (I'm not sure if the doc is correct though.) Could you take another look? Maybe you can construct some test cases locally and see if they work as expected.

crates/sail-plan/src/function/scalar/array.rs
```diff
@@ -12,13 +15,59 @@ fn array_repeat(element: expr::Expr, count: expr::Expr) -> expr::Expr {
...
+}
+
+fn sort_array(args: Vec<expr::Expr>) -> PlanResult<expr::Expr> {
+    let (array, asc) = args.two()?;
+    let (sort, nulls) = if asc == lit(ScalarValue::Boolean(Some(true))) {
+        (
+            lit(ScalarValue::Utf8(Some("ASC".to_string()))),
+            lit(ScalarValue::Utf8(Some("NULLS FIRST".to_string()))),
+        )
+    } else if asc == lit(ScalarValue::Boolean(Some(false))) {
```
linhr: I'd prefer `match` on `Expr::Literal(ScalarValue::Boolean(Some(true)))` etc. rather than `==`. In this way, we don't need to worry about how `Eq` or `PartialEq` behave for these types.

crates/sail-plan/src/function/scalar/misc.rs
shehabgamin: Ahh good point!

crates/sail-plan/src/function/scalar/misc.rs
shehabgamin: nice catch!

crates/sail-plan/src/function/scalar/misc.rs
shehabgamin: @linhr I'm not sure I follow.

crates/sail-plan/src/function/scalar/array.rs
shehabgamin: Nice catch!

crates/sail-plan/src/function/scalar/array.rs
shehabgamin: @linhr Yeah I actually read the source code for this because we needed to make sure it was 1-indexed as well. My understanding is that it is exclusive, but i will add a test as well to make sure!

crates/sail-plan/src/function/scalar/misc.rs
linhr: When evaluating an expression tree, my intuition is that the child expressions (`RaiseError` and `col`) is evaluated first, before evaluating the parent expression (`When`). So I'm afraid that the error may be raised before evaluating the `col` values. Some optimizer rules may avoid this, but there is no guarantee.

Let's add a TODO here to add some PySpark tests once we have the pytest setup for the library.

crates/sail-plan/src/function/scalar/array.rs
```diff
@@ -38,7 +87,57 @@ pub(super) fn list_built_in_array_functions() -> Vec<(&'static str, Function)> {
...
         ("sequence", F::ternary(expr_fn::gen_series)),
         ("shuffle", F::unknown("shuffle")),
-        ("slice", F::unknown("slice")),
-        ("sort_array", F::unknown("sort_array")),
+        ("slice", F::ternary(slice)),
+        ("sort_array", F::custom(sort_array)),
     ]
 }
+
+#[cfg(test)]
```
linhr, line 95: Nice!

linhr (approved):
Great work!

PR #131, opened 2024-08-15
DataFusion 41 + With Clause + More Catalog & Function Support

shehabgamin:


crates/sail-sql/src/query.rs
```diff
@@ -33,7 +34,14 @@ pub(crate) fn from_ast_query(query: ast::Query) -> SqlResult<spec::QueryPlan> {
     if for_clause.is_some() {
         return Err(SqlError::unsupported("FOR clause"));
     }
-    let plan = from_ast_set_expr(*body)?;
+
+    let mut ctes = if let Some(with) = with {
```
shehabgamin: Not sure if `ctes` needs to be an arg of `from_ast_query` which would mean that `from_ast_with` would insert into the existing `HashMap` instead of creating a new one.

crates/sail-spark-connect/src/service/plan_executor.rs
```diff
@@ -314,11 +315,14 @@ pub(crate) async fn handle_execute_create_dataframe_view(
...
-    let _ = view.is_global;
+    let _ = view.is_global; // TODO: global view
     if view.replace {
         ctx.deregister_table(table_ref.clone())?;
     }
-    ctx.register_table(table_ref, df.into_view())?;
+    ctx.register_table(
+        table_ref,
+        Arc::new(TemporaryTableProvider::new(df.logical_plan().clone())),
+    )?;
```
linhr: We no longer need `df`. Just pass `plan` into the table provider without cloning.

crates/sail-sql/src/query.rs
```diff
@@ -559,9 +568,22 @@ fn with_ast_table_alias(
...
 }
+
+pub(crate) fn from_ast_with(
+    with: ast::With,
+) -> SqlResult<Vec<(spec::Identifier, spec::QueryPlan)>> {
+    let mut ctes: Vec<(spec::Identifier, spec::QueryPlan)> = Vec::new();
+    for cte in with.cte_tables {
+        let cte_name = spec::Identifier::from(normalize_ident(cte.alias.name.clone()));
+        let plan = from_ast_query(*cte.query)?;
+        let plan = with_ast_table_alias(plan, Some(cte.alias))?;
```
linhr: We no longer need the table alias since the CTE name is the hash map key.

linhr (approved):
Great work!! 🚀

crates/sail-spark-connect/src/service/plan_executor.rs
shehabgamin: oh good point lol

crates/sail-sql/src/query.rs
shehabgamin: Nice catch!

linhr (approved):

PR #132, opened 2024-08-16
Refactor GitHub Actions workflows

linhr:


scripts/spark-tests/generate-test-report.sh
```diff
@@ -3,12 +3,12 @@
 set -euo 'pipefail'
 
 if [ "$#" -ne 2 ]; then
-    echo "Usage: $0 <head-test-logs> <base-test-logs>"
+    echo "Usage: $0 <test-logs-after> <test-logs-before>"
```
linhr, line 6: The terms "base" and "head" are not accurate. For pull requests, we are essentially evaluating the commit from the merge branch (maintained by GitHub automatically) rather than the head branch. So we now refer to the two commits as "before" and "after".

PR #140, opened 2024-08-20
More SQL Support + More Function Support + Show String Fixes

shehabgamin:


linhr (approved):
Great work! All the enhancements and bug fixes look really nice to me!

PR #142, opened 2024-08-21
Set up Python unit tests

linhr:


pyproject.toml
```diff
@@ -30,13 +35,16 @@ path = ".venvs/default"
 python = "3.11"
 installer = "pip"
 dependencies = [
-    "pyspark[connect,sql]==3.5.1",
+    "pyspark[connect]==3.5.1",
```
shehabgamin, line 38: Why remove sql?

linhr: The extra dependencies for `sql` is a subset of `connect`, so only `connect` is enough.

PR #146, opened 2024-08-26
Support S3 data via object store

linhr:


crates/sail-plan/src/object_store/config.rs
```diff
@@ -0,0 +1,31 @@
+use aws_config::{BehaviorVersion, SdkConfig};
+
+#[derive(Debug, Clone)]
+pub struct ObjectStoreConfig {
```
linhr, line 4: We have to define the object store configuration and pass it to the registry. The AWS configuration initialization is async, so we cannot do this inside the registry trait, which has only sync method to get object stores.

PR #148, opened 2024-08-27
Refactor Spark Connect operations

linhr:
1. Move the logic of a few Spark Connect operations to the plan resolver.
2. Fix the `datafusion-functions-nested` dependency.
3. Update the type of `metadata` in the plan specification.

### DIFF
--- a/crates/sail-common/src/spec/data_type.rs
+++ b/crates/sail-common/src/spec/data_type.rs
@@ -1,10 +1,8 @@
-use std::collections::HashMap;
-
 use serde::{Deserialize, Serialize};
 
 use crate::error::{CommonError, CommonResult};
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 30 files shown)


crates/sail-common/src/spec/data_type.rs
```diff
@@ -67,23 +65,23 @@ impl DataType {
...
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct Field {
     pub name: String,
     pub data_type: DataType,
     pub nullable: bool,
-    pub metadata: HashMap<String, String>,
+    pub metadata: Vec<(String, String)>,
```
linhr, line 81: Here are the reasons for this change:
1. Having `metadata` as `Vec<(String, String)>` ensures that the spec is deterministic.
2. I'd like to derive `Hash` for `DataType`, which is not possible if it contains `HashMap`.

PR #149, opened 2024-08-27
Functions and DF

shehabgamin:


crates/sail-plan/src/resolver/plan.rs
```diff
@@ -2026,6 +2039,146 @@ impl PlanResolver<'_> {
...
+                                Operator::Plus,
+                                Box::new(predicate?),
+                            ))) as PlanResult<Expr>
+                        })?;
+                    non_null_count.gt_eq(lit(min_non_nulls))
+                } else {
+                    conjunction(not_null_exprs).unwrap()
+                }
+            }
+            None => conjunction(not_null_exprs).unwrap(),
```
linhr: A schema may have zero fields, so we should return an error instead of calling `unwrap()`.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -2026,6 +2039,146 @@ impl PlanResolver<'_> {
...
+        let columns: Vec<String> = columns.into_iter().map(|x| x.into()).collect();
+
+        let fill_na_exprs = schema
+            .columns()
+            .into_iter()
+            .map(|c| {
+                let column_expr = col(c.clone());
+                let column_data_type = column_expr.get_type(schema)?;
+                let column_name = state.get_field_name(c.name())?;
+                let expr = if columns.is_empty() || columns.iter().any(|col| col == column_name) {
```
linhr, line 2068: I guess there may be a way to avoid checking `col == column_name` twice. Let's add a TODO here.

crates/sail-plan/src/function/scalar/array.rs
```diff
@@ -2,6 +2,7 @@ use arrow::datatypes::DataType;
 use datafusion::functions_nested::expr_fn;
 use datafusion_common::ScalarValue;
 use datafusion_expr::{expr, lit, BinaryExpr, Operator};
+use datafusion_functions_nested::position::array_position_udf;
```
linhr: We have `use datafusion_functions_nested` and `use datafusion::functions_nested`. I'm not sure if there is a way to prefer one over the other in the linter.

linhr (approved):
Great work!

crates/sail-plan/src/resolver/plan.rs
shehabgamin: Agreed! The logic is a little messy, and took a while to figure out.

crates/sail-plan/src/function/scalar/array.rs
shehabgamin: Nice catch. I changed it to `datafusion::functions_nested` since we use that a few lines above.

linhr (approved):

PR #163, opened 2024-09-04
Decimal128 and Decimal256

shehabgamin:


crates/sail-spark-connect/tests/gold_data/expression/numeric.json
```diff
@@ -134,17 +134,51 @@
     },
     {
       "input": "1.20E-38BD",
-      "exception": "\n[DECIMAL_PRECISION_EXCEEDS_MAX_PRECISION] Decimal precision 40 exceeds max precision 38..(line 1, pos 0)\n\n== SQL ==\n1.20E-38BD\n^^^\n",
```
shehabgamin, line 137: removed this exception and created a new test for this exception below

crates/sail-spark-connect/tests/gold_data/expression/numeric.json
```diff
@@ -134,17 +134,51 @@
...
+              ],
+              "precision": 40,
+              "scale": 40
+            }
+          }
+        }
+      }
+    },
+    {
+      "input": "1200000000000000000000000000000000000.00000",
```
shehabgamin: new test

crates/sail-spark-connect/tests/gold_data/expression/numeric.json
```diff
@@ -134,17 +134,51 @@
...
+              ],
+              "precision": 42,
+              "scale": 5
+            }
+          }
+        }
+      }
+    },
+    {
+      "input": "1.20E-76BD",
```
shehabgamin: new test

crates/sail-common/src/spec/literal.rs
```diff
@@ -54,13 +56,20 @@ pub enum Literal {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Decimal {
+pub enum DecimalType {
```
linhr: Let's remove `pub` and name it `DecimalLiteral` (It's not a "type" actually.) Also let's keep only `#[derive(Debug)]` and remove `#[serde(...)]` so that it remains an internal data structure that cannot be used outside `spec`.

shehabgamin: Nice catch!

linhr: Also `decimal_type` local variables in various places need to be updated accordingly.

crates/sail-spark-connect/tests/gold_data/expression/numeric.json
linhr: As discussed offline, manually added/edited tests have to be removed until we figure out a way to manage gold data test cases manually. Right now all gold data files are generated and cannot be edited.

crates/sail-common/src/spec/literal.rs
shehabgamin: We can't remove pub actually because `crates/sail-spark-connect/src/proto/literal.rs` matches on `DecimalLiteral`

linhr: Can we use `TryFrom` in `proto/literal.rs` as well? This is similar to how you handle that in `sail-sql`.

linhr: Anyway I just realized this is more work. We can keep `pub` and add a TODO for refactoring.

crates/sail-spark-connect/tests/gold_data/plan/ddl_drop_view.json
```diff
@@ -10,6 +10,7 @@
                 "db",
                 "view"
               ],
+              "kind": null,
```
shehabgamin, line 13: idk where this came from

linhr: It seems to be related to #91. But I'm not sure why it shows up here.

crates/sail-common/src/spec/literal.rs
shehabgamin: As discussed privately if we use `TryFrom`, `parse_decimal_string` may be called twice. So we decided to optimize for performance

linhr (approved):

PR #166, opened 2024-09-07
Integrate Fastrace + Enable GZIP and ZSTD in Tonic

shehabgamin:
#169

### DIFF
--- a/crates/sail-telemetry/src/trace_layer.rs
+++ b/crates/sail-telemetry/src/trace_layer.rs
@@ -10,6 +10,7 @@ use tower::{Layer, Service};
 pub struct TraceLayer {
     name: &'static str, // TODO: Arc needed?
 }
+
 impl TraceLayer {
     pub fn new(name: &'static str) -> Self {
         Self { name }
@@ -26,11 +27,13 @@ impl<S> Layer<S> for TraceLayer {
         }
     }
 }
+
 #[derive(Clone)]
 pub struct TraceService<S> {
     inner: S,
     name: &'static str, // TODO: Arc needed?
 }
+
 impl<S, R> Service<Request<R>> for TraceService<S>
 where
     S: Service<Request<R>> + Send + 'static,


crates/sail-telemetry/src/telemetry.rs
```diff
@@ -1,86 +1,91 @@
...
+        let reporter = OpenTelemetryReporter::new(
+            opentelemetry_otlp::new_exporter()
+                .tonic()
+                .with_endpoint(url)
+                .with_protocol(opentelemetry_otlp::Protocol::Grpc)
+                .with_timeout(Duration::from_secs(
+                    opentelemetry_otlp::OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT,
+                ))
+                .build_span_exporter()
+                .expect("initialize oltp exporter"),
```
linhr: Let's return error instead.

crates/sail-telemetry/src/telemetry.rs
```diff
@@ -1,86 +1,91 @@
...
 }
 
-// TODO: init_metrics
-pub fn init_metrics() {}
+// We use DummyReporter instead of ConsoleReporter to avoid duplicate span output to the console.
+// Although using ConsoleReporter and suppressing span output in the logger is possible, it would
+// result in a different format.
+// Disabling fastrace when logging only to the console is another option, but retaining trace and
+// span IDs in the logs is useful.
+pub struct DummyReporter;
```
linhr, line 87: Nice!

crates/sail-telemetry/src/telemetry.rs
```diff
@@ -1,86 +1,91 @@
...
+impl Reporter for DummyReporter {
```
linhr, line 89: Let's add one empty line before the `impl` block.

crates/sail-telemetry/src/trace_layer.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use fastrace::prelude::*;
+use tonic::codegen::http::Request;
+use tower::{Layer, Service};
+
+#[derive(Clone)]
+pub struct TraceLayer {
+    name: &'static str, // TODO: Arc needed?
+}
+impl TraceLayer {
```
linhr, line 14: Let's add one empty line to separate top-level blocks in this file.

Cargo.toml
```diff
@@ -69,6 +59,14 @@ aws-config = "1.5.5"
...
 clap = { version = "4.5.17", features = ["derive"] }
 num_enum = "0.7.3"
+log = "0.4.22"
+env_logger = "0.11.5"
+fastrace = { version = "0.7.2", features = ["enable"] }
+fastrace-opentelemetry = "0.7.2"
+opentelemetry = "0.24.0"
+opentelemetry_sdk = "0.24.1"
+opentelemetry-otlp = { version = "0.17.0", features = ["tls", "tls-roots"] }
+anyhow = "1.0.87"
```
linhr: Is this used anywhere?

linhr (approved):
This is great!

crates/sail-telemetry/src/telemetry.rs
```diff
@@ -1,86 +1,91 @@
...
+pub fn init_logger(use_collector: bool) -> TelemetryResult<()> {
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
+        .format(move |buf, record| {
+            if use_collector {
+                Event::add_to_local_parent(record.level().as_str(), || {
+                    [("message".into(), record.args().to_string().into())]
+                });
+            }
+            let level = record.level();
+            let target = record.module_path().unwrap();
```
linhr: Let's return error here as well.

Cargo.toml
shehabgamin: nice catch, not anymore!

crates/sail-telemetry/src/telemetry.rs
shehabgamin: Nice catch! The `module_path` isn't guaranteed to be set so we should instead do `.unwrap_or("SOME DEFAULT");`

shehabgamin: Nvm ignore this, we should be using `.target()` instead.

linhr (approved):

PR #167, opened 2024-09-07
Refactor UDF functions

linhr:
1. Remove the `PythonFunctionType` enum and define the `PythonFunction` trait.
2. Remove the `Serialize` trait for UDF functions.
4. Remove the `Deserialize` trait for UDF functions. The deserialization logic is now part of the `PythonFunction` trait so that we can return the original `PyErr` instead of wrapping it in a serde error.
5. Simplify `Hash` and `PartialEq` implementations for UDF functions.
6. Rename "partial" UDF to UDF "object" since such UDF objects will be used for UDAF and UDWF as well.
7. Remove "Python UDF" since they are not used anywhere and we do not have a way to test it.


crates/sail-python-udf/src/cereal/mod.rs
```diff
@@ -1,7 +1,14 @@
-pub mod partial_pyspark_udf;
-pub mod partial_python_udf;
+use datafusion_common::Result;
+use pyo3::{Bound, PyAny, Python};
+
+pub mod pyspark_udf;
 pub mod pyspark_udtf;
 
+pub trait PythonFunction: Sized {
```
shehabgamin, line 7: nice!

PR #181, opened 2024-09-10
Fix `INSERT INTO` statement

linhr:
After this PR, I'm able to write to table in my manual testing. But I cannot add unit tests for this since `DROP TABLE` does not fully work yet. (The data is not deleted when the table is dropped.)

More work is needed in future PRs. I've added TODO comments in various places.

### DIFF
--- a/crates/sail-plan/src/config.rs
+++ b/crates/sail-plan/src/config.rs
@@ -66,7 +66,7 @@ impl Default for PlanConfig {
             },
             default_bounded_table_file_format: "PARQUET".to_string(),
             default_unbounded_table_file_format: "ARROW".to_string(),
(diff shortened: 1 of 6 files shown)


crates/sail-plan/src/config.rs
```diff
@@ -66,7 +66,7 @@ impl Default for PlanConfig {
             },
             default_bounded_table_file_format: "PARQUET".to_string(),
             default_unbounded_table_file_format: "ARROW".to_string(),
-            default_warehouse_directory: "sail-warehouse".to_string(),
+            default_warehouse_directory: "spark-warehouse".to_string(),
```
linhr, line 69: I made this change so that we still follow the Spark convention for warehouse directory.

PR #183, opened 2024-09-11
Expand Parsing Support for Create & Replace Table

shehabgamin:
1. Added support for parsing `REPLACE TABLE`, aligned with the already-supported `CREATE OR REPLACE`.
2. Fixed `PARTITIONED BY` parsing, now fully supported except for expressions (e.g., `bucket(16, some_col)`).
3. Support parsing of table names that begin with numbers (e.g., `1m.2g`).
4. Nearly all parsing cases for `CREATE/REPLACE TABLE` are now supported!!

Follow-Up Work:
1. False Positives - `CREATE_TABLE_COLUMN_DESCRIPTOR_DUPLICATE`: Parsing should **fail** when encountering duplicate column descriptors.
2. False Negatives - Expressions in `PARTITIONED BY` (e.g., `bucket(16, some_col)`): These expressions should be parsed correctly.
3. False Negatives - `CLUSTERED BY`: Parsing support for `CLUSTERED BY` needs to be implemented.
4. Parse comments within column definitions (there are no Gold Data tests for this).

**Note:** Any validation within the `sql` crate must occur there, as the Gold Data tests specifically test for parsing errors.

### DIFF
--- a/crates/sail-sql/src/statement/common.rs
+++ b/crates/sail-sql/src/statement/common.rs
@@ -19,6 +19,7 @@ pub const VALID_FILE_FORMATS_FOR_ROW_FORMAT_SERDE: [&str; 3] =
     ["TEXTFILE", "SEQUENCEFILE", "RCFILE"];
 pub const VALID_FILE_FORMATS_FOR_ROW_FORMAT_DELIMITED: [&str; 1] = ["TEXTFILE"];
 
+#[derive(Debug, PartialEq)]
 pub(crate) enum Statement {
     Standard(ast::Statement),
     Explain {
--- a/crates/sail-sql/src/statement/create.rs
+++ b/crates/sail-sql/src/statement/create.rs
@@ -328,3 +328,91 @@ pub(crate) fn calc_inline_constraints_from_columns(
     }
     constraints
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::parser::SparkDialect;
+
+    #[test]
+    fn test_parse_create_statement() -> SqlResult<()> {
+        // Gold tests thoroughly test the parser. Adding a test for coverage of number identifiers
+        let parse = |sql: &str| -> SqlResult<Statement> {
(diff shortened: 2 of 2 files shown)


crates/sail-spark-connect/tests/gold_data/plan/ddl_create_table.json
```diff
@@ -501,7 +957,53 @@
...
+              "table": [
+                "my_tab"
+              ],
+              "schema": {
+                "fields": [
+                  {
+                    "name": "a",
+                    "dataType": "integer",
+                    "nullable": true,
+                    "metadata": []
```
linhr, line 972: It seems the column comment is not parsed. Maybe we can add a TODO somewhere in the code.

crates/sail-sql/src/parse.rs
```diff
@@ -95,8 +97,99 @@ pub fn parse_file_format(parser: &mut Parser) -> SqlResult<String> {
...
+            .into_iter()
+            .flat_map(|ident| ident.split('.').map(String::from).collect::<Vec<String>>())
+            .collect()
+    }
+    Ok(spec::ObjectName::from(idents))
+}
+
+pub fn parse_normalized_identifier(parser: &mut Parser) -> SqlResult<spec::Identifier> {
+    let next_token = parser.next_token();
+    let ast_ident = match next_token.token {
```
linhr, line 127: Should we support identifiers such as `foo.1m`?

A few unit tests may be useful for this function.

linhr (approved):
Great work! 🚀

crates/sail-spark-connect/tests/gold_data/plan/ddl_create_table.json
shehabgamin: Nice catch!

crates/sail-sql/src/parse.rs
shehabgamin: Nice catch let me add this in!

PR #185, opened 2024-09-12
Define error type for Python UDF

linhr:
This removes the nearly 100 lines of `.map_err(|e| DataFusionError::External(e.into()))` in the Python UDF crate.


crates/sail-python-udf/src/udf/pyspark_udf.rs
```diff
@@ -79,7 +80,7 @@ impl ScalarUDFImpl for PySparkUDF {
         let args: Vec<ArrayRef> = ColumnarValue::values_to_arrays(args)?;
 
         if self.eval_type.is_arrow_udf() {
-            let array_data: Result<ArrayData, DataFusionError> = Python::with_gil(|py| {
+            let array_data = Python::with_gil(|py| -> PyUdfResult<_> {
```
shehabgamin, line 83: nice!

PR #200, opened 2024-09-17
more err tolerant build-pyspark.sh

skewballfox:
the changes to build-script.sh from the other branch. Use lockfiles to avoid repeating steps in the event of an early exit

### DIFF
--- a/scripts/spark-tests/build-pyspark.sh
+++ b/scripts/spark-tests/build-pyspark.sh
@@ -3,20 +3,61 @@
 set -euo 'pipefail'
 
 project_path="$(git rev-parse --show-toplevel)"
-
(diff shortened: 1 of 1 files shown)


skewballfox:
btw, I think I can add a script to set up pre-commit hooks for the lints / formatting from the docs. the creation of the pre-commit hook needs to be scripted because client side hooks can't be setup automatically

linhr:
> btw, I think I can add a script to set up pre-commit hooks for the lints / formatting from the docs. the creation of the pre-commit hook needs to be scripted because client side hooks can't be setup automatically

Thanks for the suggestion! I've created an issue to track this: #233.

PR #213, opened 2024-09-26
Expanding Spark SQL Coverage 

shehabgamin:


crates/sail-plan/src/function/scalar/conversion.rs
```diff
@@ -1,7 +1,26 @@
...
+use crate::error::{PlanError, PlanResult};
 use crate::function::common::Function;
+use crate::utils::ItemTaker;
+
+fn cast(args: Vec<expr::Expr>, _config: Arc<PlanConfig>) -> PlanResult<expr::Expr> {
+    let (source_expr, target_type) = args.two()?;
+    match target_type {
+        expr::Expr::Literal(literal) => Ok(expr::Expr::Cast(expr::Cast {
+            expr: Box::new(source_expr),
+            data_type: literal.data_type(),
```
linhr: What's this literal? Is this the data type string? I'm not sure how `literal.data_type()` works here.

crates/sail-plan/src/function/scalar/string.rs
```diff
@@ -83,6 +83,53 @@ fn unbase64(expr: expr::Expr) -> expr::Expr {
...
+        // TODO: optional third argument
+        return Err(PlanError::todo(
+            "position with 3 arguments is not supported yet",
+        ));
+    }
+    Err(PlanError::invalid("position requires 2 arguments"))
+}
+
+fn space(n: expr::Expr) -> expr::Expr {
+    expr_fn::repeat(lit(""), n)
```
linhr: Should this be `lit(" ")`?

crates/sail-plan/src/utils.rs
```diff
@@ -73,3 +87,97 @@ impl<T: Debug> ItemTaker for Vec<T> {
...
+        ("zzzz", "%Z"),
+        ("zzz", "%Z"),
+        ("zz", "%Z"),
+        ("z", "%Z"),
+        ("OOOO", "%Z"),
+        ("OO", "%Z"),
+        ("VV", "%Z"),
+    ];
+
+    let mut result = format.to_string();
```
linhr, line 171: Let's add a TODO to create regex using `lazy_static`.

linhr (approved):
Great work!! 🚀

crates/sail-plan/src/function/scalar/string.rs
shehabgamin: Nice catch!

crates/sail-plan/src/function/scalar/conversion.rs
shehabgamin: Nice catch. This logic is incorrect, it will always return a string datatype lol. What we want to do instead is map the string value to the proper datatype. I'll undo this code as it's out of scope for this pr.

PR #214, opened 2024-09-27
Add basic support for UDAF

linhr:


crates/sail-python-udf/src/udf/pyspark_udf.rs
```diff
@@ -79,175 +253,39 @@ impl ScalarUDFImpl for PySparkUDF {
...
-                    Ok(result)
-                })
-                .collect::<Result<Vec<_>, _>>()?;
-
-            let results = pyarrow_module_array.call((results,), Some(&pyarrow_array_kwargs))?;
-
-            Ok(ArrayData::from_pyarrow_bound(&results)?)
-        })?;
-
+        let array_data = if self.eval_type.is_arrow_udf() {
```
linhr, line 256: After this small refactoring, the duplicated code becomes more obvious. We can clean up them in future PRs.

crates/sail-plan/src/resolver/expression.rs
```diff
@@ -871,36 +875,100 @@ impl PlanResolver<'_> {
...
-        let (function_name, argument_names, arguments, is_distinct) = match window_function {
-            spec::Expr::UnresolvedFunction {
-                function_name,
-                arguments,
-                is_user_defined_function,
-                is_distinct,
-            } => {
-                if is_user_defined_function {
-                    return Err(PlanError::unsupported("user defined window function"));
+        let (function, function_name, argument_names, arguments, is_distinct) =
```
linhr, line 878: This is a bit messy. I'll clean up the duplicated code in follow-up PRs.

PR #235, opened 2024-10-05
Prepare v0.1.4

shehabgamin:


shehabgamin:
It seems `maturin-action` fails since it tries to install Python package in system Python [1][2][3]. The workaround is to add a `setup-python` step so that `pip install` works.

* [1] <https://github.com/PyO3/maturin-action/issues/291>
* [2] <https://github.com/PyO3/maturin/issues/2244>
* [3] <https://github.com/PyO3/maturin/issues/2245>

PR #244, opened 2024-10-15
Set up distributed processing

linhr:
This PR adds the setup for distributed processing.

1. The driver serializes the physical plan and sends it to the worker. After the worker notifies the driver that the task has started running, the driver fetches the output stream from the worker via Arrow Flight RPC and forwards the stream to the Spark Connect client.
2. The driver and worker runs as separate threads in the same process. This is similar to the "pseudo cluster" in Spark local mode. The driver and worker communicates via gRPC.
3. The entire job is represented as a single physical plan. The job is not divided into shuffle stages yet.
4. A new worker is created for each job. Worker reuse and release is not implemented yet.

This PR does not alter the behavior of the existing Spark Connect server. In a future PR, some configuration can be added to enable distributed processing.


crates/sail-server/src/builder.rs
```diff
@@ -0,0 +1,113 @@
...
+        // The TCP listener from the standard library does not work with graceful shutdown.
+        // See also: https://github.com/hyperium/tonic/issues/1424
+        listener: TcpListener,
+        signal: F,
+    ) -> Result<(), Box<dyn std::error::Error>>
+    where
+        F: Future<Output = ()>,
+    {
+        let reflection_server = self.reflection_server_builder.build_v1()?;
+        let router = self.router.add_service(reflection_server);
```
shehabgamin, line 102: Turn off in production

crates/sail-spark-connect/src/entrypoint.rs
```diff
@@ -15,68 +13,23 @@ use crate::spark::connect::spark_connect_service_server::SparkConnectServiceServ
...
-        Some(signal) => {
-            server
-                .serve_with_incoming_shutdown(incoming, signal)
-                .await?;
-        }
-    }
-
-    Ok(())
+    let service = SparkConnectServiceServer::new(server)
+        .max_encoding_message_size(GRPC_MAX_MESSAGE_LENGTH_DEFAULT)
```
shehabgamin: Comment out encoding for now, I'm sure about the `GRPC_MAX_MESSAGE_LENGTH_DEFAULT` being relevant to `max_decoding_message_size`.

crates/sail-server/src/builder.rs
linhr: I added a TODO comment for this.

crates/sail-spark-connect/src/entrypoint.rs
linhr: Done!

PR #254, opened 2024-10-20
Fix Resolve Expression Attribute Candidates With Qualifier

shehabgamin:


crates/sail-plan/src/resolver/expression.rs
```diff
@@ -673,6 +673,13 @@ impl PlanResolver<'_> {
     ) -> PlanResult<Vec<(String, DataType, Column)>> {
         // TODO: handle qualifier and nested fields
         let name: Vec<String> = name.clone().into();
+        let name = if name.iter().any(|ident| ident.contains('.')) {
```
linhr: The identifier is considered "raw". If the nested identifier is not parsed correctly, it should be fixed upstream in the protobuf or SQL processor.

shehabgamin: This is coming from proto. Fixing it upstream would require doing it everywhere in proto? Or is there somewhere common to place this?

linhr: Let’s see if fixing this FIXME would fix the issue.

<https://github.com/lakehq/sail/blob/fca61be759ea02071b9f290b3969e6f2a9ccb3a6/crates/sail-spark-connect/src/proto/expression.rs#L38-L39>

shehabgamin: worked!

linhr (approved):

PR #256, opened 2024-10-22
Set up unit tests for TPC-H and TPC-DS

linhr:
This PR adds unit tests for queries from the TPC-H and TPC-DS benchmark. The data generation is done via DuckDB.

I copied the queries from [apache/datafusion-benchmarks](https://github.com/apache/datafusion-benchmarks). Then I used the following command to format the SQL files.

```bash
find python/pysail/data -name '*.sql' -exec pnpm exec sql-formatter --fix {} \;
```

1. These queries were generated from templates so they are "static". In the future, we may want to support generating "dynamic" queries (with randomness in parameters) using the original TPC-H and TPC-DS tools.
2. These queries are different from the ones in the DuckDB repository. There are minor syntactic differences due to the SQL dialect used. Also the parameters are different.

The tests are currently skipped. The developer can remove the `@pytest.mark.skip` marker locally to run these tests.

To run the tests against a separate Spark Connect server without building the Python package in Maturin, the following command can be used. (This instruction is in the documentation already.)

```bash
env SPARK_REMOTE="sc://localhost:50051" hatch run pytest
```


python/pysail/tests/spark/test_tpch.py
```diff
@@ -0,0 +1,41 @@
...
+        text = f.read()
+    for sql in text.split(";"):
+        sql = sql.strip()  # noqa: PLW2901
+        if not sql:
+            continue
+        actual = spark.sql(sql.replace("create view", "create temp view"))
+        expected = duck.sql(sql)
+        if "create view" in sql or "drop view" in sql:
+            continue
+        assert_frame_equal(actual.toPandas(), expected.df())
```
shehabgamin: We shouldn't be comparing the `duckdb` response to our response. We should be comparing the `spark` response to our response.

python/pysail/tests/spark/test_tpcds.py
```diff
@@ -0,0 +1,43 @@
...
+        sql = sql.strip()  # noqa: PLW2901
+        if not sql:
+            continue
+        # Spark and DuckDB has different syntax for quoted identifiers.
+        #   Spark: SELECT 1 AS `v`
+        #   DuckDB: SELECT 1 AS "v"
+        # In TPC-DS, quoted identifiers are used when column names contain spaces.
+        actual = spark.sql(sql)
+        expected = duck.sql(sql.replace("`", '"'))
+        assert_frame_equal(actual.toPandas(), expected.df())
```
shehabgamin: We shouldn't be comparing the `duckdb` response to our response. We should be comparing the `spark` response to our response.

PR #261, opened 2024-10-24
FIX Query Plan Sorts + More Spark DF Coverage

shehabgamin:
Resolve sorting issues using recursion, addressing all remaining resolution errors in derived TPC-DS. This PR brings the total support to 74 out of the 99 derived TPC-DS queries.

### DIFF
--- a/crates/sail-plan/src/resolver/plan.rs
+++ b/crates/sail-plan/src/resolver/plan.rs
@@ -2731,13 +2731,10 @@ impl PlanResolver<'_> {
         columns: Vec<spec::Identifier>,
         state: &mut PlanResolverState,
     ) -> PlanResult<LogicalPlan> {
-        use datafusion::functions_aggregate::{
-            average::avg_udaf,
-            count::count_udaf,
-            min_max::{max_udaf, min_udaf},
-            stddev::stddev_udaf,
-        };
-        // use datafusion_expr::expr::Cast;
+        use datafusion::functions_aggregate::average::avg_udaf;
+        use datafusion::functions_aggregate::count::count_udaf;
+        use datafusion::functions_aggregate::min_max::{max_udaf, min_udaf};
+        use datafusion::functions_aggregate::stddev::stddev_udaf;
 
         let input = self.resolve_query_plan(input, state).await?;
         let columns: Vec<Column> = if columns.is_empty() {
@@ -2760,7 +2757,7 @@ impl PlanResolver<'_> {
                 order_by: None,
                 null_treatment: None,
             })
-            .alias(state.register_field(&format!("count_{}", column.name())));
+            .alias(state.register_field(format!("count_{}", column.name())));
             all_aggregates.push(count);
 
             if let Ok(field) = input.schema().field_from_column(column) {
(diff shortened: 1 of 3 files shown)


crates/sail-plan/src/resolver/plan.rs
```diff
@@ -1276,16 +1287,31 @@ impl PlanResolver<'_> {
         for col in columns {
```
shehabgamin, line 1333: The strange logic in this for loop is intentional in order to satisfy all of the test conditions in
`pyspark/sql/tests/connect/test_parity_dataframe.py::DataFrameParityTests::test_drop_join`,
`pyspark/sql/tests/connect/test_parity_dataframe.py::DataFrameParityTests::test_drop`,
and `pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.drop`

python/pysail/tests/spark/test_tpcds.py
```diff
@@ -13,28 +13,66 @@ def duck():
...
 
 @pytest.fixture(scope="module", autouse=True)
-def data(sail, spark, duck):
+def data(sail, spark, duck):  # noqa: ARG001
     tables = list(duck.sql("SHOW TABLES").df()["name"])
     for table in tables:
         df = duck.sql(f"SELECT * FROM {table}").df()  # noqa: S608
         sail.createDataFrame(df).createOrReplaceTempView(table)
-        spark.createDataFrame(df).createOrReplaceTempView(table)
+        # spark.createDataFrame(df).createOrReplaceTempView(table)
```
shehabgamin, line 21: This breaks for me locally so I commented it out

python/pysail/data/tpcds/queries/q90.sql
```diff
@@ -2,7 +2,7 @@
 -- TPC-DS queries are Copyright 2021 Transaction Processing Performance Council.
 -- This query was generated at scale factor 1.
 select
-  cast(amc as decimal(15, 4)) / cast(pmc as decimal(15, 4)) am_pm_ratio
```
shehabgamin, line 5: `DuckDB` SQL allows for division by 0, so `DuckDB` actually returns `NaN` here. We adjust the SQL query to use `CASE` because we are getting the same result as `DuckDB`.

linhr: Could we avoid modifying the query? The queries are actually generated from some templates. If we support query templates in the future, the change will be overridden.

I copied the queries from here: <https://github.com/apache/datafusion-benchmarks/tree/main/tpcds/queries-spark>

I'm not sure why DuckDB is involved here. I thought in our previous PR we have set up the unit tests to compare our results with Spark results. DuckDB is only used for data generation.

shehabgamin: Other versions of TPC-DS have the case statement here. The queries are suppose to be generated based off of the data. But in this case we have the data from DuckDB and the query from DataFusion comet.

We have no other choice but to modify the query. The point I was making is that Sail successfully processes this query but it will always fail because of the "/ 0" if we don't add the case statement.
DuckDB allows for "/ 0" which is why it doesn't fail for DuckDB. And this is relevant because DuckDB generated the data knowing that it would work for its workload.
We get the same exact result, we just don't allow for "/0" which is why we have to modify the query. Spark will fail here too, because Spark doesn't allow for "/0" either.

PR #262, opened 2024-10-24
Improve actor implementation

linhr:
1. Implement `ActorContext` to keep track of all Tokio tasks spawned by the actor.
2. Refactor `ServerMonitor` to be an `enum` so that it can manage its state better.

### DIFF
--- a/crates/sail-server/src/actor.rs
+++ b/crates/sail-server/src/actor.rs
@@ -1,18 +1,29 @@
+use log::error;
 use tokio::sync::{mpsc, watch};
+use tokio::task::{AbortHandle, JoinSet};
 
 const ACTOR_CHANNEL_SIZE: usize = 8;
 
 pub trait Actor: Sized + Send + 'static {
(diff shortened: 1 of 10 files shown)


crates/sail-server/src/actor.rs
```diff
@@ -1,18 +1,29 @@
...
+    fn start(&mut self, ctx: &mut ActorContext<Self>) -> Result<(), Self::Error>;
+    /// Process one message and return the next action.
+    /// This method should only return errors when they are not recoverable.
+    /// In such a situation, the actor will be stopped.
+    /// If the actor can recover from the error, it should handle it inside the method
+    /// and return [Ok].
+    /// This method should not invoke any blocking functions, otherwise the actor event loop
+    /// would be blocked since all messages are processed sequentially in a single thread.
+    /// If the actor needs to perform async operations, it should spawn tasks via
+    /// [ActorContext::spawn].
```
shehabgamin, line 22: Nice comment!!

PR #265, opened 2024-10-25
Implement distributed job stages

linhr:
1. Implement the physical plans for shuffle read and shuffle write.
2. Update the codec for physical plans.
3. Implement pipelined shuffle. Define the data structure that can later be used to support blocking shuffle via local files or remote storage.
4. Implement the logic to start and stop workers for each job. Right now each job owns a fixed number of workers, and the workers are stopped after the job finishes. Dynamic worker allocation and worker reuse will be future work.

The shuffle physical plans and the corresponding planner are inspired by the DataFusion Ballista and DataFusion Ray projects.

* <https://github.com/apache/datafusion-ballista>
* <https://github.com/apache/datafusion-ray>


crates/sail-execution/proto/sail/plan/physical.proto
```diff
@@ -13,6 +13,10 @@ message ExtendedPhysicalPlanNode {
   oneof NodeKind {
     RangeExecNode range = 1;
     ShowStringExecNode show_string = 2;
+    ShuffleReadExecNode shuffle_read = 3;
+    ShuffleWriteExecNode shuffle_write = 4;
+    // TODO: This can be removed once it is supported by DataFusion.
+    MemoryExecNode memory = 101;
```
shehabgamin, line 19: Needed to serialize memory data as part of the plan. Data itself is part of the plan when dealing with Spark `LocalRelation` or instances where data is in memory (meaning that the `LogicalPlan` has the data itself).

crates/sail-execution/src/driver/planner.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+fn build_job_graph(
+    plan: Arc<dyn ExecutionPlan>,
+    graph: &mut JobGraph,
+) -> ExecutionResult<Arc<dyn ExecutionPlan>> {
+    let children = plan
+        .children()
+        .into_iter()
+        .map(|x| build_job_graph(x.clone(), graph))
```
shehabgamin, line 50: In a DAG, one stage can be reused by multiple downstream stages. But, that is not the case here because of the recursive nature.

crates/sail-execution/src/driver/planner.rs
```diff
@@ -0,0 +1,115 @@
...
+        return Err(ExecutionError::DataFusionError(plan_datafusion_err!(
+            "expected one child but got {}: {:?}",
+            plan.children().len(),
+            plan
+        )));
+    }
+    Ok(plan.children()[0].clone())
+}
+
+fn create_shuffle(
```
shehabgamin, line 95: 1 shuffle pair (write, read) between each boundary. Read shuffle comes from previous stage, write shuffle comes from current stage.

crates/sail-execution/src/worker_manager/local.rs
```diff
@@ -50,6 +50,7 @@ impl WorkerManager for LocalWorkerManager {
             worker_listen_port: 0,
             worker_external_host: "127.0.0.1".to_string(),
             worker_external_port: None,
+            memory_stream_buffer: 16,
```
shehabgamin, line 53: Number of batches per stream. Used for pipeline shuffle

crates/sail-server/src/actor.rs
```diff
@@ -1,15 +1,18 @@
+use std::fmt::Debug;
+
 use log::error;
 use tokio::sync::mpsc;
 use tokio::task::{AbortHandle, JoinSet};
 
 const ACTOR_CHANNEL_SIZE: usize = 8;
```
shehabgamin, line 7: Max buffer of 8 events per Actor. If queue is full, then it is blocking event send from threads spawned by Actor. Events are messages, e.g. task started or task completed.

crates/sail-execution/src/job/runner.rs
```diff
@@ -51,6 +51,8 @@ impl ClusterJobRunner {
             driver_listen_port: 0,
             driver_external_host: "127.0.0.1".to_string(),
             driver_external_port: None,
+            worker_count_per_job: 4,
+            job_output_buffer: 16,
```
shehabgamin, line 55: Stream of output from worker to driver. Max buffer size of 16 batches before blocking execution of tasks.

PR #275, opened 2024-11-03
Clean Up Dependencies & Small Python UDF Refactor

shehabgamin:


crates/sail-python-udf/src/udf/pyspark_udf.rs
```diff
@@ -70,16 +71,20 @@ fn call_arrow_udf(
...
 
     let py_args = args
         .iter()
         .map(|arg| {
             let arg = arg
                 .into_data()
                 .to_pyarrow(py)?
-                // FIXME: Should be to_pandas here for performance (Zero-Copy),
-                //  but behavior of results is inconsistent with PySpark's expectations.
-                .call_method0(py, intern!(py, "to_pylist"))?
```
shehabgamin, line 82: When I put this `FIXME` in, I recall that many more tests used to fail when we used `to_pandas()`. Now, only 3 additional tests fail when we use `to_pandas()`, and the failures are due to differences in expected output, which I've pasted below:
```
self = <pyspark.sql.tests.connect.test_parity_arrow_python_udf.ArrowPythonUDFParityTests testMethod=test_register>

    def test_register(self):
        df = self.spark.range(1).selectExpr(
            "array(1, 2, 3) as array",
        )
        str_repr_func = self.spark.udf.register("str_repr", udf(lambda x: str(x), useArrow=True))

        # To verify that Arrow optimization is on
>       self.assertEquals(
            df.selectExpr("str_repr(array) AS str_id").first()[0],
            "[1, 2, 3]",  # The input is a NumPy array when the Arrow optimization is on
        )
E       AssertionError: '[1 2 3]' != '[1, 2, 3]'
E       - [1 2 3]
E       + [1, 2, 3]
E       ?   +  +
```
```
self = <pyspark.sql.tests.connect.test_parity_arrow_python_udf.ArrowPythonUDFParityTests testMethod=test_nested_array>

    def test_nested_array(self):
        df = self.spark.range(1).selectExpr("array(array(1, 2), array(3, 4)) as nested_array")
        # Input
        row = df.select(udf(lambda x: str(x))("nested_array")).first()
>       self.assertEquals(row[0], "[[1, 2], [3, 4]]")
E       AssertionError: '[array([1, 2], dtype=int32) array([3, 4], dtype=int32)]' != '[[1, 2], [3, 4]]'
E       - [array([1, 2], dtype=int32) array([3, 4], dtype=int32)]
E       + [[1, 2], [3, 4]]
```
```
self = <pyspark.sql.tests.connect.test_parity_arrow_python_udf.ArrowPythonUDFParityTests testMethod=test_nested_array_input>

    def test_nested_array_input(self):
        df = self.spark.range(1).selectExpr("array(array(1, 2), array(3, 4)) as nested_array")
>       self.assertEquals(
            df.select(
                udf(lambda x: str(x), returnType="string", useArrow=True)("nested_array")
            ).first()[0],
            "[[1, 2], [3, 4]]",
        )
E       AssertionError: '[array([1, 2], dtype=int32) array([3, 4], dtype=int32)]' != '[[1, 2], [3, 4]]'
E       - [array([1, 2], dtype=int32) array([3, 4], dtype=int32)]
E       + [[1, 2], [3, 4]]
```

As you can see, these aren't concerning differences. They're all related to how these strings are formatted, but the underlying values are actually the same.

crates/sail-python-udf/Cargo.toml
```diff
@@ -10,13 +10,8 @@ thiserror = { workspace = true }
...
-serde_bytes = { workspace = true }
 pyo3 = { workspace = true }
-arrow = { workspace = true }
-arrow-cast = { workspace = true }
-arrow-select = { workspace = true }
-sqlparser = { workspace = true }
 async-trait = { workspace = true }
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
shehabgamin, line 17: According to PyO3 docs, `cdylib` is required:
https://pyo3.rs/v0.22.5/#using-rust-from-python
https://pyo3.rs/v0.22.5/getting-started#cargotoml

linhr (approved):

PR #278, opened 2024-11-05
This Should Have Been Split Into Multiple PRs

shehabgamin:


crates/sail-execution/src/plan/shuffle_write.rs
```diff
@@ -152,8 +152,11 @@ impl ExecutionPlan for ShuffleWriteExec {
         }
         let stream = self.plan.execute(partition, context)?;
         // TODO: support metrics in batch partitioner
-        let partitioner =
-            BatchPartitioner::try_new(self.shuffle_partitioning.clone(), Default::default())?;
+        let shuffle_partitioning = match &self.shuffle_partitioning {
+            Partitioning::UnknownPartitioning(size) => Partitioning::RoundRobinBatch(*size),
```
linhr, line 156: I think this is a reasonable approach for now. 👍 

I'm curious how "unknown partitioning" appears in physical plans. Let's add a TODO to revisit this in the future.

crates/sail-plan/src/extension/function/explode.rs
```diff
@@ -5,46 +5,64 @@ use datafusion::common::Result;
 use datafusion::logical_expr::{ColumnarValue, ScalarUDFImpl, Signature, Volatility};
 use datafusion_common::plan_err;
 
+pub fn explode_name_to_kind(name: &str) -> ExplodeKind {
+    match name {
+        "explode" => ExplodeKind::Explode,
+        "explode_outer" => ExplodeKind::ExplodeOuter,
+        "posexplode" => ExplodeKind::PosExplode,
+        "posexplode_outer" => ExplodeKind::PosExplodeOuter,
+        _ => unreachable!(),
```
linhr: We should return `Err` instead of panicking here.

crates/sail-plan/src/extension/function/explode.rs
```diff
@@ -5,46 +5,64 @@ use datafusion::common::Result;
...
-    fn name(&self) -> &str {
+    pub fn kind_to_name(&self) -> &str {
         match self.kind {
             ExplodeKind::Explode => "explode",
             ExplodeKind::ExplodeOuter => "explode_outer",
             ExplodeKind::PosExplode => "posexplode",
             ExplodeKind::PosExplodeOuter => "posexplode_outer",
         }
     }
+}
```
linhr: I'm a bit confused about the changes in these methods.

`name_to_kind()` is not needed since we have `kind()` already. Also we should remove `kind_to_name` and move the implementation to `name()` in `impl ScalarUDFImpl` (which is how it works before these changes).

crates/sail-plan/src/resolver/expression.rs
```diff
@@ -939,7 +937,7 @@ impl PlanResolver<'_> {
                     };
                     let output_type = self.resolve_data_type(output_type)?;
 
-                    let python_function: PySparkUdfObject = deserialize_partial_pyspark_udf(
+                    let python_bytes: Vec<u8> = build_pyspark_udf_payload(
```
linhr, line 940: This now makes sense to me. By deferring deserialization to the physical execution (via `OnceLock`), we avoid the serde overhead in the driver in distributed processing. Well done!

crates/sail-plan/src/resolver/statistic.rs
```diff
@@ -261,4 +258,98 @@ impl PlanResolver<'_> {
...
+            .alias_qualified(
+                right_column.relation.clone(),
+                right_column.name.clone(),
+            )])?
+            .build()?;
+        let distinct_values = LogicalPlanBuilder::from(projected_plan)
+            .project(vec![Expr::Column(right_column.clone())])?
+            .distinct()?
+            .build()?;
+        let distinct_values_batches = self
```
linhr, line 291: Let's add a TODO here to call out that this can be expensive for large input datasets.

crates/sail-python-udf/src/udf/mod.rs
```diff
@@ -11,12 +13,15 @@ use pyo3::types::PyDict;
...
 /// Without this, lambda functions with the name `<lambda>` will be treated as the same function
 /// by logical plan optimization rules (e.g. common sub-expression elimination), resulting in
 /// incorrect logical plans.
-pub fn get_udf_name(function_name: &str, function: &PyObject) -> String {
-    format!("{}@0x{:x}", function_name, function.as_ptr() as usize)
+pub fn get_udf_name(function_name: &str, function_bytes: &[u8]) -> String {
+    let mut hasher = DefaultHasher::new();
+    function_bytes.hash(&mut hasher);
+    let hash = hasher.finish();
+    format!("{function_name}@0x{hash:x}")
```
linhr, line 25: Let's add a FIXME here for hash collision.

linhr (approved):
Great work!!! 🚀

crates/sail-plan/src/extension/function/explode.rs
linhr: Maybe this function is not needed at all, based on the other comment I had on this file.

PR #279, opened 2024-11-06
Set up application configuration

linhr:
1. Set up configuration at the application level using [Figment](https://github.com/sergiobenitez/figment).
2. Remove `ConfigKeyValueList` in favor of using `Vec<ConfigKeyValue>` directly.
3. Move `SparkUdfConfig` to the `sail-python-udf` crate.
4. Refactor `DynamicObjectStoreRegistry` to use global object store configuration.
5. Refactor Spark session. We now have `SparkSession` as an extension in the DataFusion `SessionContext`. Previously, we have the `Session` struct that wraps DataFusion `SessionContext`.

### DIFF
--- a/crates/sail-common/src/config/application.rs
+++ b/crates/sail-common/src/config/application.rs
@@ -1,0 +1,69 @@
+use figment::providers::{Env, Format, Toml};
+use figment::Figment;
+use serde::{Deserialize, Serialize};
+
+use crate::error::{CommonError, CommonResult};
+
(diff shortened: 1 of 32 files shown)


crates/sail-common/src/config/application.rs
```diff
@@ -0,0 +1,69 @@
+use figment::providers::{Env, Format, Toml};
```
shehabgamin, line 1: Let's add comments here or somewhere else about what some of these configs are. We did PR comments for some of the ambiguous ones last time, so we could just pull those in.

PR #280, opened 2024-11-08
Set up Kubernetes

linhr:
1. Implement Kubernetes worker manager. Each worker is launched as a pod.
2. Support dynamic worker allocation and worker reuse.
3. Support idle worker removal.
4. Support detecting worker launch timeout and task scheduling timeout.


crates/sail-execution/src/driver/actor/handler.rs
```diff
@@ -227,77 +286,260 @@ impl DriverActor {
...
+        let slots_per_worker = self.options().worker_task_slots;
+        let active_workers = self.state.count_active_workers();
+        let used_slots = self.state.count_active_tasks();
+        let pending_slots = self.state.count_pending_tasks();
+        let available_slots = active_workers * slots_per_worker;
+        if available_slots >= used_slots + pending_slots {
+            return;
+        }
+        let missing_slots = used_slots + pending_slots - available_slots;
+        let missing_workers = (missing_slots + slots_per_worker - 1) / slots_per_worker;
```
shehabgamin, line 304: Has to do with integer division rounding.
`9 / 3 = (9 + 3 - 1) / 3`. But if you have 8 missing slots you’d need `(8 + 3 - 1) / 3 = 3` instead of `8 / 3 = 2`.

PR #288, opened 2024-11-15
Add Kubernetes docs and update config

linhr:


docs/guide/deployment/kubernetes.md
```diff
@@ -0,0 +1,91 @@
+---
+title: Kubernetes
+rank: 1
+---
+
+# Kubernetes
+
+Sail supports distributed data processing on Kubernetes clusters.
+This guide demonstrates how to deploy Sail on a Kubernetes cluster and connect to it via Spark Connect.
```
shehabgamin, line 9: ❤️

docs/guide/deployment/kubernetes.md
```diff
@@ -0,0 +1,91 @@
...
+```bash
+docker build -t sail:latest -f docker/dev/Dockerfile .
+```
+
+:::
+
+You will then need to make the Docker image available to your Kubernetes cluster.
+The exact steps depend on your Kubernetes environment.
+For example, you may want to push the Docker image to your private Docker registry accessible from the Kubernetes cluster.
+If you are using a local Kubernetes cluster, you may want to load the Docker image into the cluster. The command varies depending on the Kubernetes distribution you are using.
```
shehabgamin, line 41: Let's add an example command for loading the docker image into a local Kubernetes cluster.

linhr: Yeah this is a good point. Let me set up kind/minikube/k3s/k3d and show the right command in a future PR.

PR #289, opened 2024-11-16
DataFusion 43 Patch + Doc Updates + Dockerfile Updates

shehabgamin:
Update DataFusion patch to include fixes for WindowUDF


linhr (approved):
Great work!!

PR #297, opened 2024-11-20
Use explicit pip version in docs

linhr:


README.md
```diff
@@ -16,14 +16,14 @@ Please refer to the [Kubernetes Deployment Guide](https://docs.lakesail.com/sail
...
-pip install pysail
+pip install "pysail==0.2.0.dev0"
 ```
 
 Alternatively, you can install Sail from source for better performance for your hardware architecture.
 You would need [rustup](https://rustup.rs/) and [protoc](https://protobuf.dev/) in your environment for this.
 
 ```bash
-env RUSTFLAGS="-C target-cpu=native" pip install pysail -v --no-binary pysail
+env RUSTFLAGS="-C target-cpu=native" pip install "pysail==0.2.0.dev0" -v --no-binary pysail
```
linhr, line 26: This is a temporary change. We do not want to update the version manually in README, so we'll revert this change once we stop publishing pre-release versions.

PR #302, opened 2024-11-20
Update Kubernetes Docker setup

linhr:
1. Update Kubernetes Docker setup.
2. Update Kubernetes deployment documentation.


docker/Dockerfile
```diff
@@ -1,86 +0,0 @@
...
-ARG RUST_PROFILE
-ARG RUSTFLAGS
-ARG RELEASE_TAG
-ARG PYSPARK_VERSION
-
-ENV RUST_VERSION=${RUST_VERSION}
-ENV RUST_PROFILE=${RUST_PROFILE}
-ENV RUSTFLAGS=${RUSTFLAGS}
-ENV RELEASE_TAG=${RELEASE_TAG}
-ENV PYSPARK_VERSION=${PYSPARK_VERSION}
```
linhr, line 80: I've removed these `ENV` instructions since I plan to storing build information in the Sail binary. This is tracked in #303.

docker/dev/build-dev.sh
```diff
@@ -1,20 +0,0 @@
-#!/bin/bash
```
linhr, line 1: After we split the Dockerfile based on different use cases, we no longer need these scripts to manage build context.

docs/guide/deployment/kubernetes.md
```diff
@@ -1,13 +1,14 @@
 ---
```
shehabgamin, line 1: Will we need to update the URL of the blog post?

linhr:
I updated VitePress and the docs build time went down from ~50 s to ~5 s, thanks to optimizations in its recent versions.

docker/release/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/rustup-init && \
+    chmod a+x /tmp/rustup-init && \
+    /tmp/rustup-init -y --no-modify-path --profile minimal --default-toolchain ${RUST_VERSION}
+
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+WORKDIR /app
+
+RUN git clone --depth 1 --branch "${RELEASE_TAG}" https://github.com/lakehq/sail.git .
```
shehabgamin, line 34: This won't work for commit hashes, which won't make it possible to test release builds with this Dockerfile.

docs/guide/deployment/kubernetes.md
linhr: No, I decided to not change the URL so that there is least confusion for the user.

docker/release/Dockerfile
linhr: We can specify a branch name for `RELEASE_TAG`. This does not work for commit hashes though, but it should cover the majority use cases in our development.

PR #310, opened 2024-11-22
Support map partitions

linhr:
1. Implement map partitions for Pandas and Arrow UDF. The key complexity is the adapter between Rust `SendableRecordBatchStream` (async) and Python iterator (blocking).
2. Implement codec for map partitions plan. The known issues are tracked in #312.
3. Move Python and PyArrow helper functions to `sail_python_udf::utils` and make them more modular.
4. Change `PartialOrd` implementation.


crates/sail-plan/src/extension/logical/catalog.rs
```diff
@@ -182,103 +193,293 @@ pub(crate) enum CatalogCommand {
...
+            } => CatalogCommandOrd::CreateDatabase {
+                database,
+                if_not_exists: *if_not_exists,
+                comment,
+                location,
+                properties,
             },
-            // For all other cases, use the default derived comparison
-            _ => {
-                CatalogCommandHelper(self.clone()).partial_cmp(&CatalogCommandHelper(other.clone()))
```
linhr, line 276: This implementation of `PartialOrd` does not work since this would cause infinite recursion. Here is the code to reproduce:
```rust
let c = CatalogCommand::CurrentCatalog;
println!("{:?}", c.partial_cmp(&c));
```

Here is the output:
```text
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
```

shehabgamin: Nice catch!

scripts/spark-tests/plugins/spark.py
```diff
@@ -201,6 +201,10 @@ class TestMarker:
         keywords=["SparkSessionTestCase", "test_session.py"],
         reason="Client keeps retrying setting session config for invalid remote endpoints",
     ),
+    TestMarker(
+        keywords=["JobCancellationTests", "test_session.py"],
+        reason="Slow test not fully working yet",
+    ),
```
linhr, line 207: This slow test now becomes noticeable since it uses `.mapInPandas()`, which this PR supports.

PR #313, opened 2024-11-27
Support group map and co-group map

linhr:
1. Support Pandas group map.
2. Support Pandas co-group map.
3. Clean up Spark UDF configuration.
4. Clean up UDF codec.


docker/dev/Dockerfile
```diff
@@ -38,7 +38,7 @@ RUN --mount=type=cache,target=/root/.cargo/registry/ \
     --mount=type=cache,target=/app/target/ \
     RUST_TARGET_SUBDIR=$(case "${RUST_PROFILE}" in \
         dev|test) echo "debug" ;; \
-        release|prod|bench) echo "release" ;; \
+        release|bench) echo "release" ;; \
```
linhr, line 41: `prod` is not a built-in Cargo profile.

<https://doc.rust-lang.org/cargo/reference/profiles.html>

crates/sail-python-udf/src/cereal/pyspark_udf.rs
```diff
@@ -1,79 +1,75 @@
...
-use crate::cereal::{check_python_udf_version, PythonFunction};
+use crate::cereal::check_python_udf_version;
 use crate::config::SparkUdfConfig;
 use crate::error::{PyUdfError, PyUdfResult};
 
-#[derive(Debug)]
-pub struct PySparkUdfObject(pub PyObject);
+pub struct PySparkUdfObject {
+    data: Vec<u8>,
+    cell: GILOnceCell<PyObject>,
```
linhr, line 16: We now use `GILOnceCell ` instead of `OnceLock` for lazy initialization of the Python UDF object. In this way, we can use `GILOnceCell::get_or_try_init` instead of `OnceLock::get_or_init` for better error propagation. `OnceLock::get_or_try_init` is not available on stable Rust yet.

shehabgamin: Nice find!

crates/sail-python-udf/src/udf/mod.rs
```diff
@@ -1,19 +1,26 @@
...
 /// Without this, lambda functions with the name `<lambda>` will be treated as the same function
 /// by logical plan optimization rules (e.g. common sub-expression elimination), resulting in
 /// incorrect logical plans.
-pub fn get_udf_name(function_name: &str, function_bytes: &[u8]) -> String {
-    // FIXME: Hash collision is possible
-    let mut hasher = DefaultHasher::new();
-    function_bytes.hash(&mut hasher);
-    let hash = hasher.finish();
-    format!("{function_name}@0x{hash:x}")
+pub fn get_udf_name(name: &str, payload: &[u8]) -> String {
```
shehabgamin, line 15: Nice!

PR #316, opened 2024-12-01
Support memory stream replay and improve error handling

linhr:
1. Support stream replay for shuffle plans under `CoalescePartitionsExec`.
2. Implement worker heartbeat.
3. Implement retry logic for RPC requests.

### DIFF
--- a/crates/sail-execution/src/rpc.rs
+++ b/crates/sail-execution/src/rpc.rs
@@ -2,7 +2,7 @@ use std::future::Future;
 use std::sync::Arc;
 
 use arrow_flight::flight_service_client::FlightServiceClient;
-use tokio::sync::{oneshot, Mutex, MutexGuard, OnceCell};
+use tokio::sync::{oneshot, OnceCell};
 use tokio::task::JoinHandle;
 use tonic::transport::Channel;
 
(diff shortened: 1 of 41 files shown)


crates/sail-execution/src/rpc.rs
```diff
@@ -119,35 +119,41 @@ impl_client_builder!(DriverServiceClient<Channel>);
...
 pub struct ClientHandle<T> {
+    /// The client options.
     options: Arc<ClientOptions>,
-    inner: OnceCell<Arc<Mutex<T>>>,
+    /// The shared gRPC client which is lazily initialized.
+    /// Note that this must be `Arc<OnceCell<T>>` instead of `OnceCell<Arc<T>>`.
+    /// If we use the latter, when the client is not initialized, an empty `OnceCell` would be
+    /// cloned and later initialized independently, resulting in multiple connections.
+    /// This could then easily overwhelm the server, and the client would see the
+    /// "connection refused" Tonic transport error.
```
linhr, line 133: This was the subtle bug that caused "transport error" when cluster load is high.

shehabgamin: nice find!

PR #317, opened 2024-12-01
Sail Native Types

shehabgamin:
1. Refactor Sail `DataType` and `Literal` to match Arrow types
2. Clean up crate dependencies by removing unnecessary arrow crates.

### DIFF
--- a/crates/sail-spark-connect/src/proto/data_type.rs
+++ b/crates/sail-spark-connect/src/proto/data_type.rs
@@ -351,7 +351,8 @@ impl TryFrom<spec::DataType> for DataType {
             spec::DataType::Timestamp { time_unit: spec::TimeUnit::Microsecond, timezone_info: spec::TimeZoneInfo::NoTimeZone } => {
                 Ok(Kind::TimestampNtz(sdt::TimestampNtz::default()))
             }
-            spec::DataType::Timestamp { time_unit: spec::TimeUnit::Microsecond, timezone_info: spec::TimeZoneInfo::LocalTimeZone }
+            spec::DataType::Timestamp { time_unit: spec::TimeUnit::Microsecond, timezone_info: spec::TimeZoneInfo::Configured }
+            | spec::DataType::Timestamp { time_unit: spec::TimeUnit::Microsecond, timezone_info: spec::TimeZoneInfo::LocalTimeZone }
             | spec::DataType::Timestamp { time_unit: spec::TimeUnit::Microsecond, timezone_info: spec::TimeZoneInfo::TimeZone { timezone: _ } }=> {
                 Ok(Kind::Timestamp(sdt::Timestamp::default()))
             }
@@ -361,12 +362,6 @@ impl TryFrom<spec::DataType> for DataType {
                 // This error theoretically should never be reached.
                 Err(SparkError::unsupported(
                     "TryFrom spec::DataType::Timestamp { time_unit: Second | Millisecond | Nanosecond } to Spark Kind",
-                ))
-            }
-            spec::DataType::Timestamp { time_unit: _, timezone_info: spec::TimeZoneInfo::Configured } => {
-                // This error theoretically should never be reached.
-                Err(SparkError::invalid(
-                    "TryFrom spec::DataType::Timestamp { time_unit: _, timezone_info: Configured } to Spark Kind. Timezone must be NoTimeZone, LocalTimeZone, or TimeZone",
                 ))
             }
             spec::DataType::Interval { interval_unit: spec::IntervalUnit::MonthDayNano, start_field: _, end_field: _ } => {


crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,319 @@
...
-    Byte,
-    Short,
-    Integer,
-    Long,
-    Float,
-    Double,
-    Decimal128 {
-        precision: u8,
-        scale: i8,
-    },
```
linhr: I prefer this way over `Decimal128(u8, i8)`. This makes our JSON representation more readable. I'd suggest follow our previous pattern instead of copying all Arrow types as is.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,319 @@
...
+use crate::error::{CommonError, CommonResult};
 
+pub const LOCAL_TIME_ZONE_IDENTIFIER: &str = "ltz";
+pub const NO_TIME_ZONE_IDENTIFIER: &str = "ntz";
+
+/// Native Sail data types that convert to Arrow types.
+/// These types usually directly match to [arrow_schema::DataType] variants when there is a corresponding type.
+/// Exceptions to this are: [`Interval`].
+/// Additionally, custom data types are supported for cases not covered by Arrow (e.g. [`UserDefined`]).
+/// [Credit]: Comments within the enum are copied from [`arrow_schema::DataType`].
```
linhr: Do we have a way to keep our comment in sync if the upstream ever changes? Maybe we should remove the long description and only keep a one-line summary for each data type. You can use links such as `[arrow_schema::DataType::Int8]` (no need for backtick) in the doc to refer to the Arrow type and the long description is easily accessible in the IDE by following the link.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -88,28 +333,172 @@ pub struct Field {
     pub metadata: Vec<(String, String)>,
 }
 
+/// [Credit]: The implementations for [`Field`], [`FieldRef`], [`Fields`], and [`UnionFields`] are
+/// copied from [`arrow_schema::Field`], [`arrow_schema::FieldRef`], [`arrow_schema::Fields`], and [`arrow_schema::UnionFields`].
+///
+/// A reference counted [`Field`]
+pub type FieldRef = Arc<Field>;
```
linhr, line 240: I wonder if we could keep our previous definition of `Field`, `Fields`, etc. The spec does not need to be shared across threads, so the overhead of `Arc` is probably unnecessary.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,319 @@
...
+    /// The keys and values are each respectively contiguous.
+    /// The key and value types are not constrained, but keys should be
+    /// hashable and unique.
+    /// Whether the keys are sorted can be set in the `bool` after the `Field`.
+    ///
+    /// In a field with Map type, the field has a child Struct field, which then
+    /// has two children: key type and the second the value type. The names of the
+    /// child fields may be respectively "entries", "key", and "value", but this is
+    /// not enforced.
+    Map(FieldRef, bool),
```
linhr: Defining `Map` like this is very hard to work with. I'd suggest using our existing definition. Same for other container types such as `List`.

crates/sail-common/src/spec/data_type.rs
shehabgamin: Discussed offline, will keep `FieldRef`

Cargo.toml
```diff
@@ -91,10 +91,6 @@ datafusion-functions-json = "0.43.0"
 # auto-initialize: Changes [`Python::with_gil`] to automatically initialize the Python interpreter if needed.
 pyo3 = { version = "0.22.0", features = ["auto-initialize", "serde"] }
 # Should be the equivalent of enabling the pyarrow feature in datafusion since we already have pyo3 in the workspace
-arrow = { version = "53.1.0", features = ["ffi", "pyarrow"] }
-arrow-cast = { version = "53.1.0" }
-arrow-schema = { version = "53.1.0" }
-arrow-select = { version = "53.1.0" }
```
linhr, line 97: Nice!

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,219 @@
...
+/// The style of expressing the type may not always be exactly the same as Arrow's.
+/// The spec is designed to have an easy-to-read JSON representation,
+/// making language interoperability easier in the future.
+/// This is achieved by eliminating the use of tuple structs and using named fields instead.
+/// [Credit]: Comments within the enum are copied from [arrow_schema::DataType].
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase", rename_all_fields = "camelCase")]
 pub enum DataType {
+    /// Null type.
+    /// Corresponds to [arrow_schema::DataType::Null].
```
linhr, line 22: I was wrong when I talked about the doc syntax last time. It turns out that we can use the following (bracket with backtick) to have the link text rendered as code. (When we use bracket only, we get the link but the text is rendered as normal text.)

```text
[`arrow_schema::DataType::Null`]
```

No need to change it for now though, but this is something to keep in mind in the future. We may be able to use some linter to make the doc style more consistent later.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,219 @@
...
-use crate::error::CommonError;
+use crate::error::{CommonError, CommonResult};
 
+/// Native Sail data types that convert to Arrow types.
+/// Types directly match to [arrow_schema::DataType] variants when there is a corresponding type.
+/// Additionally, custom data types are supported for cases not covered by Arrow.
+/// The style of expressing the type may not always be exactly the same as Arrow's.
+/// The spec is designed to have an easy-to-read JSON representation,
+/// making language interoperability easier in the future.
+/// This is achieved by eliminating the use of tuple structs and using named fields instead.
```
linhr, line 16: Nice comment!

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,219 @@
...
+    LargeUtf8,
+    /// A variable-length string in Unicode with UTF-8 encoding
+    /// Logically the same as [DataType::Utf8].
+    /// Corresponds to [arrow_schema::DataType::Utf8View].
+    Utf8View,
+    /// A list of some logical data type with variable length.
+    /// A single List array can store up to [`i32::MAX`] elements in total.
+    /// Corresponds to [arrow_schema::DataType::List].
+    List {
+        field: FieldRef,
```
linhr, line 132: This should be `value_type` and `nullable`, so that we don't have the ambiguity how the field is named ("entry", "item", or some other names). The same applies to `FixedSizeList` and `LargeList`.

(This follows the same design we have for the `Map` type.)

You can make the change in follow-up PRs.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,219 @@
...
+    },
+    /// A Map is a logical-nested type that is represented as
+    /// `List<entries: Struct<key: K, value: V>>`
+    /// Corresponds to [arrow_schema::DataType::Map].
     Map {
         key_type: Box<DataType>,
         value_type: Box<DataType>,
-        value_contains_null: bool,
+        value_type_nullable: bool,
+        keys_are_sorted: bool,
```
linhr, line 183: This should be named as `keys_sorted`, so that the naming follows the same style as `value_type_nullable`.

crates/sail-plan/src/formatter.rs
```diff
@@ -31,51 +31,108 @@ impl_dyn_object_traits!(PlanFormatter);
...
-            DataType::Double => Ok("double".to_string()),
+            DataType::Int8 => Ok("tinyint".to_string()),
+            DataType::Int16 => Ok("smallint".to_string()),
+            DataType::Int32 => Ok("int".to_string()),
+            DataType::Int64 => Ok("bigint".to_string()),
+            DataType::UInt8 => Ok("unsigned tinyint".to_string()),
+            DataType::UInt16 => Ok("unsigned smallint".to_string()),
+            DataType::UInt32 => Ok("unsigned int".to_string()),
+            DataType::UInt64 => Ok("unsigned bigint".to_string()),
+            DataType::Float16 => Ok("half-float".to_string()),
```
linhr, line 74:
```suggestion
            DataType::Float16 => Ok("half_float".to_string()),
```

If we support float16 in SQL, `half-float` would not be a valid identifier.

crates/sail-common/src/spec/data_type.rs
```diff
@@ -1,74 +1,219 @@
...
+    /// Resolves to either [DataType::Utf8] or [DataType::LargeUtf8],
+    /// based on `config.arrow_use_large_var_types`.
+    /// Optional length parameter is currently unused but retained for potential future use.
+    /// Optional `ConfiguredUtf8Type` specifies if the type is a `varchar` or `char`.
+    // TODO: The [`sail-spark-connect`] crate uses `TryFrom` for mapping Spark types to Sail types,
+    //  whereas [`PlanResolver`] resolves `DataType` without `TryFrom`.
+    //  This makes [`PlanResolver::resolve_data_type`] more suitable here,
+    //  as it has access to `config` and can determine the type accordingly.
+    //  Refactor data type resolution in [`sail-spark-connect`] to avoid using `TryFrom`,
+    //  which would allow for removal of this variant.
```
linhr, line 203: I realized that we probably have to keep this since the SQL transformer does not have access to `PlanConfig`.

Maybe we could have a consolidated `String` type and use an enum (similar to `TimeZoneInfo`) to control how it maps to the Arrow type. This would be future work though.

linhr (approved):
Great work!! 🚀

Cargo.toml
```diff
@@ -91,10 +91,6 @@ datafusion-functions-json = "0.43.0"
 # auto-initialize: Changes [`Python::with_gil`] to automatically initialize the Python interpreter if needed.
 pyo3 = { version = "0.22.0", features = ["auto-initialize", "serde"] }
 # Should be the equivalent of enabling the pyarrow feature in datafusion since we already have pyo3 in the workspace
```
linhr, line 93: This comment line "Should be the equivalent..." can be removed now.

shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

crates/sail-common/src/spec/data_type.rs
shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

crates/sail-plan/src/formatter.rs
shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

crates/sail-common/src/spec/data_type.rs
shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

crates/sail-common/src/spec/data_type.rs
shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

crates/sail-common/src/spec/data_type.rs
shehabgamin: Addressing here:
https://github.com/lakehq/sail/pull/328

PR #319, opened 2024-12-03
Support removing idle Spark sessions

linhr:
1. Support removing idle Spark sessions.
2. Switch to a more robust approach for timeout event handling in all actors.

### DIFF
--- a/crates/sail-execution/src/driver/actor/handler.rs
+++ b/crates/sail-execution/src/driver/actor/handler.rs
@@ -170,7 +170,7 @@ impl DriverActor {
         if let WorkerState::Running { heartbeat_at, .. } = &worker.state {
             if *heartbeat_at <= instant {
                 warn!("worker {worker_id} heartbeat timeout");
-                let message = "worker heartbeat timeout ".to_string();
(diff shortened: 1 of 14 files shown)


crates/sail-execution/src/driver/actor/handler.rs
```diff
@@ -122,18 +122,12 @@ impl DriverActor {
             warn!("worker {worker_id} not found");
             return ActorAction::Continue;
         };
-        let timeout = self.options().worker_launch_timeout;
-        if let WorkerState::Pending { pending_at } = &worker.state {
-            if pending_at.elapsed() > timeout {
```
linhr, line 127: `x.elapsed() > timeout` is fragile since the outcome can be wrong if the probe event is handled slightly before the timeout is reached. In this PR, we take a different approach by storing the time instant inside the event.

linhr: Some probe events do not need the time instant inside the event since they will not be updated (e.g. `pending_at`). But other probe events do need the time instance since they can be updated (e.g. `updated_at`).

PR #323, opened 2024-12-05
Refactor PySpark UDF

linhr:
1. Refactor PySpark UDF implementation.
    1. Move input and output data conversion logic to Python.
    2. Create separate structs for different types of UDFs. 
2. Remove alias for function arguments.


crates/sail-python-udf/src/utils/spark.py
```diff
@@ -0,0 +1,435 @@
...
+except AttributeError:
+    _PYARROW_STRING_TYPE_INSTANCES = (pa.string(), pa.large_string())
+
+try:
+    _PYARROW_BINARY_TYPE_INSTANCES = (pa.binary(), pa.large_binary(), pa.binary_view())
+except AttributeError:
+    _PYARROW_BINARY_TYPE_INSTANCES = (pa.binary(), pa.large_binary())
+
+
+class Converter:
```
linhr, line 39: We now have non-trivial data conversion logic that is easier to write in Python than using PyO3. This Python module is loaded once and compiled to Python byte code, so the performance is similar to calling the Python API via PyO3 in Rust.

If we ever need to further improve the performance, we could rewrite part of the code using PyO3. For now the focus is to ensure the correctness for PySpark UDF, especially in edge cases.

shehabgamin: This makes sense to me!

PR #330, opened 2024-12-17
Sail Literal

shehabgamin:


crates/sail-spark-connect/src/proto/literal.rs
```diff
@@ -12,58 +12,234 @@ impl TryFrom<Literal> for spec::Literal {
...
-            LiteralType::Boolean(x) => spec::Literal::Boolean(x),
-            LiteralType::Byte(x) => spec::Literal::Byte(x as i8),
-            LiteralType::Short(x) => spec::Literal::Short(x as i16),
-            LiteralType::Integer(x) => spec::Literal::Integer(x),
-            LiteralType::Long(x) => spec::Literal::Long(x),
-            LiteralType::Float(x) => spec::Literal::Float(x),
-            LiteralType::Double(x) => spec::Literal::Double(x),
+            LiteralType::Null(data_type) => {
+                if data_type.kind.is_some() {
+                    let data_type: spec::DataType = data_type.try_into()?;
```
linhr, line 17: Ideally you should simply match against the Spark data type here. This would avoid various `SparkError::todo` below and various match arms which would never happen.

Or if we want to keep this, maybe we can move the logic to `impl Literal` in the spec. This can be a general `pub fn null(data_type: spec::DataType) -> Literal` constructor.

crates/sail-plan/src/resolver/state.rs
```diff
@@ -124,6 +126,22 @@ impl PlanResolverState {
...
+    //      https://github.com/search?q=repo%3Aapache%2Fspark%20%22useLargeVarTypes%22&type=code
+    //  2. We are likely overly liberal in setting this flag to `true`.
+    //     Evaluate if we are unnecessarily setting this flag to `true` anywhere.
+    pub fn register_apply_arrow_use_large_var_types_config(&mut self, apply: bool) {
+        self.apply_arrow_use_large_var_types_config = apply;
+    }
+
+    pub fn get_apply_arrow_use_large_var_types_config(&self) -> bool {
+        self.apply_arrow_use_large_var_types_config
+    }
```
linhr, line 144: This doesn't seem right to me since I don't see `register_apply_arrow_use_large_var_types_config` reverted after resolving UDF. I guess you need something like `enter_udf_scope()` instead.

crates/sail-sql/src/literal.rs
```diff
@@ -908,7 +913,29 @@ fn parse_multi_unit_interval(
...
+    if remaining_micros % 1000 == 0 {
+        spec::Literal::IntervalDayTime {
+            days: Some(total_days as i32),
+            milliseconds: Some((remaining_micros / 1000) as i32),
+        }
+    } else if microseconds % 1000 == 0 {
+        spec::Literal::IntervalDayTime {
+            days: Some(0),
+            milliseconds: Some((microseconds / 1000) as i32),
+        }
```
linhr, line 933: Is this `else if` possible?

crates/sail-common/src/spec/literal.rs
```diff
@@ -1,108 +1,177 @@
...
+    },
+    DurationNanosecond {
+        nanoseconds: Option<i64>,
+    },
+    IntervalYearMonth {
+        months: Option<i32>,
+    },
+    IntervalDayTime {
+        days: Option<i32>,
+        milliseconds: Option<i32>,
```
linhr, line 99: When we have multiple fields, having `Option` on each field is less ideal. We should have a struct for the inner data and then use `Option` to wrap the struct. This is because we do not want to allow the case when only some of the fields are `None`.

This can be future work since this PR is already large.

linhr (approved):
Great work!! 🚀

crates/sail-sql/src/literal.rs
shehabgamin: No

crates/sail-common/src/spec/literal.rs
shehabgamin: Yeah good point i meant to do this. Nice catch.

crates/sail-sql/src/literal.rs
shehabgamin: Updating here:
https://github.com/lakehq/sail/pull/339

crates/sail-common/src/spec/literal.rs
shehabgamin: Updating here:
https://github.com/lakehq/sail/pull/339

crates/sail-plan/src/resolver/state.rs
shehabgamin: Updating here:
https://github.com/lakehq/sail/pull/339

crates/sail-spark-connect/src/proto/literal.rs
shehabgamin: Updating here:
https://github.com/lakehq/sail/pull/339

PR #339, opened 2024-12-31
Literal Cleanup

shehabgamin:


crates/sail-plan/src/resolver/expression.rs
```diff
@@ -732,7 +732,9 @@ impl PlanResolver<'_> {
     ) -> PlanResult<NamedExpr> {
         if let Ok(udf) = self.ctx.udf(function_name.as_str()) {
             if let Some(_f) = udf.inner().as_any().downcast_ref::<PySparkUnresolvedUDF>() {
-                state.register_apply_arrow_use_large_var_types_config(true);
+                let mut scope = state.enter_config_scope();
+                let state = scope.state();
+                state.register_config_apply_arrow_use_large_var_types(true);
```
linhr: `scope` will go out of scope at the end of `if`, so setting the configuration in this way does not have an effect.

crates/sail-sql/src/literal.rs
```diff
@@ -28,7 +30,7 @@ lazy_static! {
         regex::Regex::new(r"^\s*(?P<year>\d{4})(-(?P<month>\d{1,2})(-(?P<day>\d{1,2})T?)?)?\s*$")
             .unwrap();
     static ref TIMESTAMP_REGEX: regex::Regex =
-        regex::Regex::new(r"^\s*(?P<year>\d{4})(-(?P<month>\d{1,2})(-(?P<day>\d{1,2})((\s+|T)(?P<hour>\d{1,2})(:((?P<minute>\d{1,2})(:((?P<second>\d{1,2})([.]((?P<fraction>\d{1,6})?(?P<tz>.*))?)?)?)?)?)?)?)?)?\s*$")
+        regex::Regex::new(r"^\s*(?P<year>\d{4})(-(?P<month>\d{1,2})(-(?P<day>\d{1,2})((\s+|T)(?P<hour>\d{1,2})(:(?P<minute>\d{1,2})(:(?P<second>\d{1,2})(\.(?P<fraction>\d{1,9}))?(?P<tz>.*)?)?)?)?)?)?\s*$")
```
linhr, line 33: Nice fix!

scripts/spark-tests/prepare-server.sh
```diff
@@ -13,7 +13,7 @@ else
 fi
 
 export PYO3_PYTHON="${python_env}/bin/python"
-export RUST_LOG="${RUST_LOG:-sail_spark_connect=debug}"
+export RUST_LOG="${RUST_LOG:-sail=debug}"
```
linhr, line 16: Nice! I wasn't aware that we can do prefix match on crate names.

crates/sail-plan/src/resolver/expression.rs
shehabgamin: Nice catch!

PR #340, opened 2025-01-03
Improve attribute resolver

linhr:
1. Improve the interface for opaque field ID (previously called "resolved field name") in `PlanResolverState`. Also each field can now map to multiple plan IDs (when `Expr::Column` is used to project the field as is).
2. Support nested field reference and nested wildcard.
3. Fix the issue in field extraction for array data type.
4. Fix the issue in join plan to avoid duplicated join key columns.
5. Fix the issue in aggregate expression rebase (TPC-DS q23).
6. Clean up the resolver logic for a few query plan nodes.

### DIFF
--- a/crates/sail-plan/src/resolver/plan.rs
+++ b/crates/sail-plan/src/resolver/plan.rs
@@ -2899,9 +2899,12 @@ impl PlanResolver<'_> {
         let input = self.resolve_query_plan(input, state).await?;
         let table_reference = self.resolve_table_reference(&table)?;
         let table_provider = self.ctx.table_provider(table_reference.clone()).await?;
-        let schema = self
-            .resolve_schema_projection(table_provider.schema(), &columns)
-            .await?;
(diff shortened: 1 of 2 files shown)


crates/sail-plan/src/function/aggregate.rs
```diff
@@ -258,9 +258,8 @@ fn list_built_in_aggregate_functions() -> Vec<(&'static str, AggFunction)> {
 }
 
 pub(crate) fn get_built_in_aggregate_function(name: &str) -> PlanResult<AggFunction> {
-    let name = name.to_lowercase();
```
linhr, line 261: The plan resolver is now responsible for ensuring the function name is converted to lower case. This behavior is consistent with the `udf()` and `table_function()` methods in DataFusion `SessionContext`, where case-insensitive match is not supported.

This change would also make it easier to support the `spark.sql.caseSensitive` configuration in the future.

PR #343, opened 2025-01-07
Fix more PySpark UDF issues

linhr:
1. Use the serializer in PySpark for all Arrow and Pandas UDF types. This avoids the custom logic for input/output conversion and fixes many failed tests.
2. Clean up PySpark UDF implementation.


crates/sail-plan/src/resolver/state.rs
```diff
@@ -13,7 +13,10 @@ pub(super) type FieldName = String;
 
 pub(super) type ResolvedFieldName = String;
 
-pub(super) const APPLY_ARROW_USE_LARGE_VAR_TYPES: &str = "APPLY_ARROW_USE_LARGE_VAR_TYPES";
+#[derive(Debug, Clone, Default)]
+pub(super) struct PlanResolverStateConfig {
```
shehabgamin, line 17: Nice!

crates/sail-python-udf/src/config.rs
```diff
@@ -1,32 +1,24 @@
-use sail_common::datetime::get_system_timezone;
-
-use crate::error::{PyUdfError, PyUdfResult};
+use pyo3::pyclass;
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd)]
-pub struct SparkUdfConfig {
+#[pyclass(frozen)]
```
shehabgamin, line 4: Nice find with `frozen`!

crates/sail-python-udf/src/config.rs
```diff
@@ -1,32 +1,24 @@
...
+pub struct PySparkUdfConfig {
```
shehabgamin, line 5: There is an inconsistency in how `PySparkUdfConfig` is used. Most files wrap an `Arc` around it, but at least one file (`crates/sail-python-udf/src/python/spark.rs`) does not.

Should we apply the same pattern used with `Field` and `FieldRef` in `crates/sail-common/src/spec/data_type.rs`?
```
#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd)]
#[pyclass(frozen)]
pub struct PySparkUdfConfig {
    #[pyo3(get)]
    pub session_timezone: String,
    #[pyo3(get, name = "window_bound_types")]
    pub pandas_window_bound_types: Option<String>,
    #[pyo3(get, name = "assign_columns_by_name")]
    pub pandas_grouped_map_assign_columns_by_name: bool,
    #[pyo3(get, name = "arrow_convert_safely")]
    pub pandas_convert_to_arrow_array_safely: bool,
    #[pyo3(get)]
    pub arrow_max_records_per_batch: usize,
}

pub type PySparkUdfConfigRef = Arc<PySparkUdfConfig>;
```

linhr: Nice observation on the inconsistency! I wasn't aware of it. But it seems the inconsistency may make sense, after I think about it now. The reason I did not use `&Arc<PySparkUdfConfig>` in `spark.rs` is that I cannot pass the `Arc` to Python, so I have to clone the struct. Therefore the functions in this file do not care about whether the config is wrapped in an `Arc` or not. All they need is a reference (`&`) to the config that can be cloned.

shehabgamin: I see that makes sense!

PR #354, opened 2025-01-17
Spark v3.5.4 Patch

shehabgamin:
`pyspark.sql.functions.current_database` is skipped as a test in Spark 3.5.4 which is why it shows up as a failure in the `Passed Tests Diff`


linhr (commented):
Could you update the proto files using the Spark `v3.5.4` tag? Otherwise it looks good!

linhr (approved):

PR #365, opened 2025-02-02
Test DataFusion 45.0.0 with Sail

shehabgamin:
### Test 1: Commit `68e372f` (DataFusion `main`)
**1.** Derived [TPC-DS Query 66](https://github.com/lakehq/sail/blob/main/python/pysail/data/tpcds/queries/q66.sql) fails on Sail, when it previously did not.
 ```
Exception: Error executing query q66 with error: Internal error:
Physical input schema should be the same as the one converted from logical input schema. 
Differences:
  - field nullability at index 7 [#98]: (physical) false vs (logical) true.
```

**2.** Another regression is that implementing the `is_nullable` function in the `ScalarUDFImpl` trait no longer works. For example:
```
impl ScalarUDFImpl for SparkArray {
    ...
    fn is_nullable(&self, _args: &[Expr], _schema: &dyn ExprSchema) -> bool {
        false
    }
    ...
}
```

### Test 2: Commit `26058ac` (DataFusion [`45.0.0-rc1`](https://github.com/apache/datafusion/tree/45.0.0-rc1))
**1.** Derived TPC-DS Query 66 still failing, with same error.

**2.** Same issue still. I ended up removing `is_nullable` and implementing `return_type_from_args`. This seems to be a breaking change rather than `is_nullable` simply being deprecated.

### DIFF
--- a/crates/sail-python-udf/src/cereal/pyspark_udf.rs
+++ b/crates/sail-python-udf/src/cereal/pyspark_udf.rs
@@ -31,7 +31,7 @@ impl PySparkUdfPayload {
         tuple
             .get_item(0)?
             .into_pyobject(py)
-            .map_err(|e| PyUdfError::InternalError(e.to_string()))
+            .map_err(|e| PyUdfError::PythonError(e.into()))
     }
 
     pub fn build(
--- a/crates/sail-python-udf/src/cereal/pyspark_udtf.rs
+++ b/crates/sail-python-udf/src/cereal/pyspark_udtf.rs
@@ -33,7 +33,7 @@ impl PySparkUdtfPayload {
         tuple
             .get_item(0)?
             .into_pyobject(py)
-            .map_err(|e| PyUdfError::InternalError(e.to_string()))
+            .map_err(|e| PyUdfError::PythonError(e.into()))
     }
 
     pub fn build(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1109,9 +1109,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.12"
+version = "1.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: crates/sail-python-udf/src/cereal/pyspark_udf.rs [hunk 1]
```diff
@@ -11,24 +11,27 @@ use crate::error::{PyUdfError, PyUdfResult};
...
             .call0()?;
-        let tuple = PyModule::import_bound(py, intern!(py, "pyspark.worker"))?
+        let tuple = PyModule::import(py, intern!(py, "pyspark.worker"))?
             .getattr(intern!(py, "read_udfs"))?
             .call1((serializer, infile, eval_type))?;
-        Ok(tuple.get_item(0)?.to_object(py))
+        tuple
+            .get_item(0)?
+            .into_pyobject(py)
+            .map_err(|e| PyUdfError::InternalError(e.to_string()))
```
linhr: Can this be `PyUdfError::PythonError(e.into())`?

crates/sail-python-udf/src/cereal/pyspark_udtf.rs
```diff
@@ -13,24 +13,27 @@ use crate::error::{PyUdfError, PyUdfResult};
...
             .call0()?;
-        let tuple = PyModule::import_bound(py, intern!(py, "pyspark.worker"))?
+        let tuple = PyModule::import(py, intern!(py, "pyspark.worker"))?
             .getattr(intern!(py, "read_udtf"))?
             .call1((serializer, infile, eval_type))?;
-        Ok(tuple.get_item(0)?.to_object(py))
+        tuple
+            .get_item(0)?
+            .into_pyobject(py)
+            .map_err(|e| PyUdfError::InternalError(e.to_string()))
```
linhr: Can this be `PyUdfError::PythonError(e.into())`?

linhr (approved):
🚀

linhr (approved):

PR #373, opened 2025-02-16
Rename `sail-sql` crate

linhr:
1. Rename the `sail-sql` crate to `sail-sql-analyzer`.
2. Remove the `sqlparser` dependency.

I ended up keeping the `SqlError` name instead of renaming it to `SqlAnalyzerError` accordingly, since this error type is used for both representing analysis errors and propagating parsing errors from the SQL parser.


python/pysail/tests/spark/test_group_by.py
```diff
@@ -147,10 +147,6 @@ def test_grouping_sets(sail):
     assert_frame_equal(actual, expected)
 
 
-# TODO: update sqlparser to support 'GROUP BY WITH' syntax
-#   https://github.com/sqlparser-rs/sqlparser-rs/pull/1323
```
linhr, line 151: This TODO is already done.

PR #376, opened 2025-02-18
Refactor SQL parser and speed up compilation

linhr:
1. Use chumsky's built-in support for token spans, instead of storing lexical spans inside the token. (This makes compilation slower due to the added type complexity. But thanks to the optimization mentioned below, the overall compilation is actually much faster after this PR.)
2. Make the parser generic over input.
3. Make the parsing API more modular by separating `parse_*` and `from_ast_*` functions.
4. Make the lifetime of `&ParserOptions` the same as the input in `TreeParser`.
5. Reduce compilation times via boxed parsers along with custom parsers for keywords, operators, literals, and identifiers. (More information is in the notes below.)

### Notes

It turns out that `cargo llvm-lines` ([link](https://github.com/dtolnay/cargo-llvm-lines/)) is helpful for troubleshooting long compilation times. However, after some experiments of tweaking the code, it turns out that, in our case, slow compilation is less correlated with the lines of LLVM IR emitted, but more correlated with the large types involved in the LLVM IR. Here are some statistics for the `sail-sql-analyzer` crate.
| | LLVM Lines | Largest Type Name Length |
| -- | -- | -- |
| Before this PR | 945178 | 1000016 |
| After this PR | 735008 | 173538 |

We can see that the largest type name goes from over 1 million characters to less than 200k characters. The compilation takes ~30 seconds now on my laptop with Apple M1 chip, compared to > 2 minutes previously. Usually GitHub Actions compiles code more slowly, so the time difference is amplified in CI, suggesting a noticeably better developer experience.

Here are some useful commands.
```bash
# Install the LLVM tool globally.
cargo install cargo-llvm-lines

cargo llvm-lines -p sail-sql-analyzer > tmp/llvm.txt
head -n 3 tmp/llvm.txt
cat tmp/llvm.txt | cut -c46- | tail -n +4 | awk '{ print(length($0)) }' | sort -nr | uniq -c | head
```


crates/sail-spark-connect/src/proto/data_type.rs
```diff
@@ -25,9 +26,11 @@ pub(crate) const SPARK_DECIMAL_SYSTEM_DEFAULT_SCALE: i8 = 18;
 /// Parse a Spark data type string of various forms.
 /// Reference: org.apache.spark.sql.connect.planner.SparkConnectPlanner#parseDatatypeString
 pub(crate) fn parse_spark_data_type(schema: &str) -> SparkResult<spec::DataType> {
-    if let Ok(dt) = parse_data_type(schema) {
+    if let Ok(dt) = parse_data_type(schema).and_then(from_ast_data_type) {
```
shehabgamin, line 29: Didn't know you could do this `and_then` pattern. Pretty useful!

shehabgamin: Ahh wait I've done this before lol. First time seeing it used with a shorthand function call though.

crates/sail-sql-parser/src/ast/keywords.rs
```diff
@@ -1,41 +1,52 @@
...
 
+// The keyword parser is implemented as a custom parser so that its type remains simple.
+// A simple parser type for keywords is crucial to reduce the overall type complexity of the
+// SQL parser, since keywords are used heavily in the grammar. The same applies to operators,
+// literals, and identifiers.
+// We could have implemented the keyword parser using parser combinators and wrapped it in a
+// boxed parser to reduce type complexity, but this would result in allocation when creating
+// the parser at runtime. The custom parser avoids such a runtime cost. The custom parser
+// implementation for these "elementary" AST nodes is also quite readable, so it does not add
+// maintenance overhead.
```
shehabgamin, line 48: Nice!!

PR #380, opened 2025-02-21
Refactor function builder and add more SQL syntax

linhr:
1. Consolidate the function builder "arguments" and "context" into a single "input".
2. Support more SQL syntax for functions (e.g. `IGNORE NULLS`, `WITHIN GROUP`, and `FILTER`).

### DIFF
--- a/crates/sail-common/src/spec/literal.rs
+++ b/crates/sail-common/src/spec/literal.rs
@@ -1,5 +1,8 @@
+use std::str::FromStr;
+
 pub use arrow_buffer::i256;
 use half::f16;
+use serde::de::Error;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::error::{CommonError, CommonResult};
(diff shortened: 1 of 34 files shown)


crates/sail-common/src/spec/literal.rs
```diff
@@ -153,14 +156,18 @@ pub enum Literal {
     Decimal128 {
         precision: u8,
         scale: i8,
+        #[serde(
+            serialize_with = "serialize_optional",
+            deserialize_with = "deserialize_optional"
+        )]
         value: Option<i128>,
```
linhr, line 163: I started getting serde errors such as "i128 is not supported". I'm not sure why it was not the case before. It seems the recommendation is to serialize 128-bit or 256-bit integers as strings, since other languages/libraries may not understand JSON numbers larger than 64-bit. So I changed the serde logic here.

crates/sail-plan/src/function/common.rs
```diff
@@ -2,50 +2,25 @@ use std::sync::Arc;
```
shehabgamin, line 2: I like the refactor of this file!

PR #385, opened 2025-02-25
Support more SQL syntax

linhr:
1. Support identifiers starting with a digit. The solution is to remove `Token::Number` from the lexer. Whether an unquoted word is parsed as a number literal or an identifier is now decided in the parser, depending on the context. When we used `sqlparser-rs`, this was only partially supported in `CREATE TABLE` statements since we did not have full control over tokenization.
2. Support concatenating multiple string literals into one literal expression.

### DIFF
--- a/crates/sail-spark-connect/tests/gold_data/plan/plan_select.json
+++ b/crates/sail-spark-connect/tests/gold_data/plan/plan_select.json
@@ -2203,7 +2203,7 @@
     {
       "input": "select * from t limit cast(9 / 4 as int)",
       "output": {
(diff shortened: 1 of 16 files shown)


crates/sail-spark-connect/tests/gold_data/plan/plan_select.json
```diff
@@ -2203,7 +2203,7 @@
     {
       "input": "select * from t limit cast(9 / 4 as int)",
```
shehabgamin, line 2204: This would be cool to have in a future PR if it's easy to knock out

PR #386, opened 2025-02-25
Fix window boundary issue in SQL

linhr:


linhr:
> Well done! Can we add some basic tests?

The changes are covered by Ibis tests. I looked at the test results and many tests involving window expressions are fixed after this PR.

PR #388, opened 2025-02-26
Test DataFusion 46.0.0 with Sail

shehabgamin:
Note To Self: Check everywhere we use `datafusion::functions_aggregate::count`

### Test 1: Commit `3193014` (DataFusion `main`)
**1.** The following query now fails when it previously did not:
```
SELECT count(*) FROM VALUES (NULL), (5), (5), (20) AS tab(col);
```
Error:
```
error in DataFusion: This feature is not implemented:
Physical plan does not support logical expression Wildcard { qualifier: None, options: WildcardOptions { ilike: None, exclude: None, except: None, replace: None, rename: None } }
```

The same error also appears 24 times in various `TPC-DS` AND `TPC-H` tests when it previously did not.

### Test 2: Commit `5502f73` (DataFusion `main`)

**1.** Still not resolved:
```
Physical plan does not support logical expression Wildcard { qualifier: None, options: WildcardOptions { ilike: None, exclude: None, except: None, replace: None, rename: None } }
```

**2.** `pyspark.sql.column.Column.substr` and: `pyspark.sql.functions.substring`:
```
UNEXPECTED EXCEPTION: ArrowInvalid('Unrecognized type: 24')
```
Looks like there is some regression in the `expr_fn::substr(string, position)` UDF.

**3.** `SparkConnectFunctionTests::test_count_star` is the same error as **1.**:
```
Physical plan does not support logical expression Wildcard { qualifier: None, options: WildcardOptions { ilike: None, exclude: None, except: None, replace: None, rename: None } }
```

### DIFF
--- a/crates/sail-plan/src/function/scalar/string.rs
+++ b/crates/sail-plan/src/function/scalar/string.rs
@@ -6,7 +6,7 @@ use datafusion::functions::expr_fn;
 use datafusion::functions::regex::regexpcount::RegexpCountFunc;
 use datafusion::functions::string::contains::ContainsFunc;
 use datafusion_common::ScalarValue;
-use datafusion_expr::{expr, lit, ScalarUDF};
+use datafusion_expr::{expr, lit, ExprSchemable, ScalarUDF};
 
 use crate::error::{PlanError, PlanResult};
 use crate::extension::function::string::levenshtein::Levenshtein;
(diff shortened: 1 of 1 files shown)


jayzhan211:
https://github.com/lakehq/sail/blob/c2a84ca4ac399b77110f5da3273de9ab5778113e/crates/sail-plan/src/resolver/plan.rs#L3413-L3434

I think you need to rewrite `Expr::Wildcard` with `count` to `count_all()` somewhere before `LogicalPlanBuilder::aggregate`

crates/sail-plan/src/function/window.rs
```diff
@@ -24,6 +24,8 @@ pub(crate) fn get_built_in_window_function(
         )),
         "sum" => Ok(expr::WindowFunctionDefinition::AggregateUDF(sum::sum_udaf())),
         "count" => Ok(expr::WindowFunctionDefinition::AggregateUDF(
+            // FIXME: CHECK HERE
+            //  DO NOT MERGE IN UNTIL COUNT WILDCARD IS FIXED LIKE WE DID IN `crates/sail-plan/src/function/aggregate.rs`
```
shehabgamin: Need to create `WindowFunctionInput`

linhr (approved):
Great work!

Issue #251, opened 2024-10-17
Outer JOIN column ambiguity


linhr:
Spark handles projection differently for outer join outputs, based on how the join column is specified in `.select()`. Here is an example for the behavior that we should support.

```text
>>> df = spark.createDataFrame([(2, "Alice"), (5, "Bob")]).toDF("age", "name")
>>> df2 = spark.createDataFrame([Row(height=80, name="Tom"), Row(height=85, name="Bob")])

>>> df.join(df2, 'name', 'outer').sort(desc("name")).show()
+-----+----+------+
| name| age|height|
+-----+----+------+
|  Tom|NULL|    80|
|  Bob|   5|    85|
|Alice|   2|  NULL|
+-----+----+------+

>>> df.join(df2, 'name', 'outer').select('name', 'height').sort(desc("name")).show()
+-----+------+
| name|height|
+-----+------+
|  Tom|    80|
|  Bob|    85|
|Alice|  NULL|
+-----+------+

>>> df.join(df2, 'name', 'outer').select(df.name, 'height').sort(desc("name")).show()
+-----+------+
| name|height|
+-----+------+
|  Bob|    85|
|Alice|  NULL|
| NULL|    80|
+-----+------+

>>> df.join(df2, 'name', 'outer').select(df2.name, 'height').sort(desc("name")).show()
+----+------+
|name|height|
+----+------+
| Tom|    80|
| Bob|    85|
|NULL|  NULL|
+----+------+

```

The example is extended from `DataFrame.join()` doctest in Spark.

linhr:
Here are more examples.

```text
>>> df.join(df2, 'name', 'outer').sort(desc(df.name)).show()
+-----+----+------+                                                             
| name| age|height|
+-----+----+------+
|  Bob|   5|    85|
|Alice|   2|  NULL|
|  Tom|NULL|    80|
+-----+----+------+

>>> df.join(df2, 'name', 'outer').sort(desc(df2.name)).show()
+-----+----+------+
| name| age|height|
+-----+----+------+
|  Tom|NULL|    80|
|  Bob|   5|    85|
|Alice|   2|  NULL|
+-----+----+------+

```

doki23:
The schemas are also different with each other:
sail:
```python
>>> df.join(df2, 'name', 'outer').sort(desc("name")).describe()
DataFrame[summary: string, age: string, height: string]
```

pyspark:
```python
>>> df.join(df2, 'name', 'outer').sort(desc("name")).describe()
[...]

PR #390, opened 2025-02-27
Support more SQL syntax

linhr:
1. Add spec for named function arguments.
2. Add spec for natural join.
3. Improve helper functions for `Identifier` and `ObjectName` in the spec.

### DIFF
--- a/crates/sail-plan/src/resolver/expression.rs
+++ b/crates/sail-plan/src/resolver/expression.rs
@@ -615,8 +615,7 @@ impl PlanResolver<'_> {
         schema: &DFSchemaRef,
         state: &mut PlanResolverState,
     ) -> PlanResult<Option<(String, expr::Expr)>> {
-        let name: Vec<&str> = name.into();
-        let candidates = Self::generate_qualified_nested_field_candidates(&name);
+        let candidates = Self::generate_qualified_nested_field_candidates(name.parts());
         let mut candidates = schema
             .iter()
             .flat_map(|(qualifier, field)| {
@@ -626,13 +625,15 @@ impl PlanResolver<'_> {
                 candidates
                     .iter()
                     .filter_map(|(q, name, inner)| {
-                        if qualifier_matches(q.as_ref(), qualifier) && info.matches(name, plan_id) {
+                        if qualifier_matches(q.as_ref(), qualifier)
+                            && info.matches(name.as_ref(), plan_id)
+                        {
                             let expr = Self::resolve_nested_field(
                                 col((qualifier, field)),
                                 field.data_type(),
                                 inner,
                             )?;
-                            let name = inner.last().unwrap_or(name).to_string();
+                            let name = inner.last().unwrap_or(name).as_ref().to_string();
                             Some((name, expr))
                         } else {
                             None
@@ -655,8 +656,7 @@ impl PlanResolver<'_> {
         schema: &DFSchemaRef,
         state: &mut PlanResolverState,
     ) -> PlanResult<Option<(String, expr::Expr)>> {
-        let name: Vec<&str> = name.into();
-        let candidates = Self::generate_qualified_field_candidates(&name);
+        let candidates = Self::generate_qualified_field_candidates(name.parts());
         let mut candidates = schema
             .iter()
             .flat_map(|(qualifier, field)| {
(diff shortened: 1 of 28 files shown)


crates/sail-plan/src/resolver/expression.rs
```diff
@@ -615,8 +615,7 @@ impl PlanResolver<'_> {
         schema: &DFSchemaRef,
         state: &mut PlanResolverState,
     ) -> PlanResult<Option<(String, expr::Expr)>> {
-        let name: Vec<&str> = name.into();
```
linhr, line 618: We no longer need to create `Vec<&str>` like this. `spec::Identifier` now implements `AsRef<str>` and I've changed many functions to accept `&[T]` with bound `T: AsRef<str>`.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -903,105 +901,96 @@ impl PlanResolver<'_> {
         let spec::Join {
```
shehabgamin, line 901: I realized that we may not have proper test coverage for join. The DataFrame DocTest for join fails because of different output due to the "Outer JOIN column ambiguity" issue (https://github.com/lakehq/sail/issues/251).

Let's manually add test coverage for the join DataFrame DocTest. Every test below should pass except for any test that has a difference in output due to the "Outer JOIN column ambiguity"  issue:
``` 
    def join(
        self,
        other: "DataFrame",
        on: Optional[Union[str, List[str], Column, List[Column]]] = None,
        how: Optional[str] = None,
    ) -> "DataFrame":
        """Joins with another :class:`DataFrame`, using the given join expression.

        .. versionadded:: 1.3.0

        .. versionchanged:: 3.4.0
            Supports Spark Connect.

        Parameters
        ----------
        other : :class:`DataFrame`
            Right side of the join
        on : str, list or :class:`Column`, optional
            a string for the join column name, a list of column names,
            a join expression (Column), or a list of Columns.
            If `on` is a string or a list of strings indicating the name of the join column(s),
            the column(s) must exist on both sides, and this performs an equi-join.
        how : str, optional
            default ``inner``. Must be one of: ``inner``, ``cross``, ``outer``,
            ``full``, ``fullouter``, ``full_outer``, ``left``, ``leftouter``, ``left_outer``,
            ``right``, ``rightouter``, ``right_outer``, ``semi``, ``leftsemi``, ``left_semi``,
            ``anti``, ``leftanti`` and ``left_anti``.

        Returns
        -------
        :class:`DataFrame`
            Joined DataFrame.

        Examples
        --------
        The following performs a full outer join between ``df1`` and ``df2``.

        >>> from pyspark.sql import Row
        >>> from pyspark.sql.functions import desc
        >>> df = spark.createDataFrame([(2, "Alice"), (5, "Bob")]).toDF("age", "name")
        >>> df2 = spark.createDataFrame([Row(height=80, name="Tom"), Row(height=85, name="Bob")])
        >>> df3 = spark.createDataFrame([Row(age=2, name="Alice"), Row(age=5, name="Bob")])
        >>> df4 = spark.createDataFrame([
        ...     Row(age=10, height=80, name="Alice"),
        ...     Row(age=5, height=None, name="Bob"),
        ...     Row(age=None, height=None, name="Tom"),
        ...     Row(age=None, height=None, name=None),
        ... ])

        Inner join on columns (default)

        >>> df.join(df2, 'name').select(df.name, df2.height).show()
        +----+------+
        |name|height|
        +----+------+
        | Bob|    85|
        +----+------+
        >>> df.join(df4, ['name', 'age']).select(df.name, df.age).show()
        +----+---+
        |name|age|
        +----+---+
        | Bob|  5|
        +----+---+

        Outer join for both DataFrames on the 'name' column.

        >>> df.join(df2, df.name == df2.name, 'outer').select(
        ...     df.name, df2.height).sort(desc("name")).show()
        +-----+------+
        | name|height|
        +-----+------+
        |  Bob|    85|
        |Alice|  NULL|
        | NULL|    80|
        +-----+------+
        >>> df.join(df2, 'name', 'outer').select('name', 'height').sort(desc("name")).show()
        +-----+------+
        | name|height|
        +-----+------+
        |  Tom|    80|
        |  Bob|    85|
        |Alice|  NULL|
        +-----+------+

        Outer join for both DataFrams with multiple columns.

        >>> df.join(
        ...     df3,
        ...     [df.name == df3.name, df.age == df3.age],
        ...     'outer'
        ... ).select(df.name, df3.age).show()
        +-----+---+
        | name|age|
        +-----+---+
        |Alice|  2|
        |  Bob|  5|
        +-----+---+
        """
```

linhr: As discussed offline, I'll create a separate PR to add the join tests.

PR #391, opened 2025-02-28
Spark Functions Coverage

shehabgamin:


crates/sail-plan/src/extension/function/error_utils.rs
```diff
@@ -0,0 +1,51 @@
+use datafusion::arrow::datatypes::DataType;
+// TODO: https://github.com/apache/spark/tree/master/common/utils/src/main/resources/error
+use datafusion_common::{exec_datafusion_err, internal_datafusion_err, DataFusionError};
+
+pub fn invalid_arg_count_exec_err(
```
linhr, line 5: This file looks nice!

crates/sail-plan/src/extension/function/math/mod.rs
```diff
@@ -1,6 +1,51 @@
...
+pub fn round_decimal_base(
+    precision: i32,
+    scale: i32,
+    target_scale: i32,
+    decimal_128: bool,
+) -> (u8, i8) {
+    let integral_least_num_digits = precision - scale + 1;
+    if target_scale < 0 {
+        let new_precision = max(integral_least_num_digits, -target_scale + 1) as u8;
+        (min(new_precision, DECIMAL128_MAX_PRECISION), 0)
```
linhr: Do we need to account for `DECIMAL256_MAX_PRECISION` here?

python/pysail/tests/spark/math/test_ceil.txt
```diff
@@ -0,0 +1,167 @@
+>>> spark.sql("SELECT ceil(-0.1), typeof(ceil(-0.1)), typeof(-0.1);").show()
```
linhr, line 1: Nice tests!

crates/sail-plan/src/extension/function/math/spark_pmod.rs
```diff
@@ -0,0 +1,369 @@
...
+                &[dividend.data_type(), divisor.data_type()],
+            )),
+        }
+    }
+
+    fn coerce_types(&self, arg_types: &[DataType]) -> Result<Vec<DataType>> {
+        if arg_types.len() != 2 {
+            return Err(invalid_arg_count_exec_err("pmod", (2, 2), arg_types.len()));
+        }
+        let (dividend, divisor) = (&arg_types[0], &arg_types[1]);
```
linhr, line 327: I found this trick would avoid explicit size checking and slice indexing:

```rust
let [dividend, divisor] = arg_types else {
    return Err(invalid_arg_count_exec_err("pmod", (2, 2), arg_types.len()));
}
```

No need to change anything for this PR though. I just wanted to share this so that it can be useful in the future.

linhr (approved):
Great work! 🚀

crates/sail-plan/src/extension/function/math/mod.rs
shehabgamin: Yes, nice catch!

linhr: Fixed.

PR #392, opened 2025-03-01
Fix join issues and add tests

linhr:
1. Fix column resolution issues with semi-join and anti-join.
2. Add doctests for join.


python/pysail/tests/spark/test_join.txt
```diff
@@ -0,0 +1,162 @@
...
+... ])
+
+>>> df1.join(df2, "name").show()
++----+---+------+
+|name|age|height|
++----+---+------+
+| Bob|  5|    85|
++----+---+------+
+
+>>> df1.join(df2, "name").select(df1.name, df2.height).show()  # doctest: +SKIP
```
shehabgamin: Why are we skipping this? This was the inner join that was failing:
```
        Inner join on columns (default)

        >>> df.join(df2, 'name').select(df.name, df2.height).show()
        +----+------+
        |name|height|
        +----+------+
        | Bob|    85|
        +----+------+
 ```
 
I looked through the PR from the Spark code base that you linked at the top of this file, there doesn't seem to be a reason why this specific test was removed either.

File: python/pysail/tests/spark/test_join.txt [hunk 2]
```diff
@@ -0,0 +1,162 @@
...
++----+------+
+
+>>> df1.join(df3, ["name", "age"]).show()
++----+---+------+
+|name|age|height|
++----+---+------+
+| Bob|  5|  NULL|
++----+---+------+
+
+>>> df1.join(df3, ["name", "age"]).select(df1.name, df1.age).show()  # doctest: +SKIP
```
shehabgamin: This is an inner join so https://github.com/lakehq/sail/issues/251 wouldn't apply here.

python/pysail/tests/spark/test_join.txt
linhr: The `name` column after join is a "coalesced" column from both relations, so referring it by `df.name` does not make much sense. I assume this is not a recommended way so it got removed from doctests. Supporting this requires tracking columns across plans, which I guess is somewhat related to the column ambiguity issue in #251.

python/pysail/tests/spark/test_join.txt
linhr: See my comment above.

python/pysail/tests/spark/test_join.txt
shehabgamin: I'm not sure I understand. The values in the name column are not coalesced because this is an inner join on name. So, selecting df.name or df2.name would yield the same result.

We use to support this behavior by just selecting either column if there was ambiguity. Can we add that behavior back in?

shehabgamin: Btw this means we have no support for selecting a specific column after an inner join which is really problematic.

linhr: I'm not sure if we had the logic to select either column. And I feel this is just an undocumented strange behavior in Spark. Even if the values are the same, conceptually there is no longer `df.name` in the joined relation since the join column is, by definition, joined. So this column does not have proper lineage with the original relation. The user should either write `df.join(df2, df.name == df2.name).select(df.name)` (since both columns are preserved in condition-based join) or just `df.join(df2, "name").select("name")` (i.e. projection by column name without referring to the original column).

A proper way to handle this strange behavior is to pass original columns as "hidden" ones so that they can be referred to in downstream projection or sort. This is way the skipped inner test is related to the outer join column ambiguity issue.

linhr: I've implemented the logic for "hidden" columns so that both this issue and #251 are resolved.

PR #393, opened 2025-03-03
Fix TPC-DS parity issues

### LINKED ISSUE #347: TPC-DS result parity
The following TPC-DS queries have different results between Sail and Spark. We need to investigate the correctness issue.

* `q33`
* `q34`
* `q56`
* `q60`

For the remaining queries (as well as all the TPC-H queries), the parity issue seems related to different column types, different column names, and floating point number precision.

(We do not enable the parity tests in pytest yet. I manually ran the skipped tests and looked at the results.)

linhr:
This PR fixes result parity issues for TPC-DS queries `q33`, `q59`, and `q60`.

It seems the parity issue for TPC-DS query `q34` has already been fixed before this PR, though it's unclear what's the root cause for the failure before.

Closes #347.

### DIFF
--- a/crates/sail-plan/src/formatter.rs
+++ b/crates/sail-plan/src/formatter.rs
@@ -860,7 +860,7 @@ impl PlanFormatter for DefaultPlanFormatter {
                 let arguments = arguments.join(", ");
                 Ok(format!("{name}({arguments})"))
             }
-            // FIXME: This is incorrect if the column name is named `*`:
+            // FIXME: This is incorrect if the column name is `*`:
             //   ```
             //   SELECT count(`*`) FROM VALUES 1 AS t(`*`)
             //   ```


crates/sail-plan/src/resolver/plan.rs
```diff
@@ -1085,20 +1085,12 @@ impl PlanResolver<'_> {
                 } else {
                     (left, right)
                 };
-                let (left, right) = (Arc::new(left), Arc::new(right));
-                let union_schema = left.schema().clone();
                 if is_all {
-                    Ok(LogicalPlan::Union(plan::Union {
-                        inputs: vec![left, right],
-                        schema: union_schema,
```
linhr, line 1093: It seems the issue is due to incorrect `functional_dependencies` in the union schema.

Although the issue is fixed, it shows that an invalid `LogicalPlan` can remain undetected and generate invalid results at runtime. There is room for improvement in DataFusion and/or Sail to ensure the validity of `LogicalPlan`s.

PR #397, opened 2025-03-05
Add missing data source codec

linhr:


crates/sail-execution/src/codec.rs
```diff
@@ -585,6 +587,53 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
                 input,
                 common_prefix_length,
             })
+        } else if let Some(data_source) = node.as_any().downcast_ref::<DataSourceExec>() {
```
linhr, line 590: This is mostly copy-paste from the logic somewhere above this line. The code duplication is fine since we'll remove `MemoryExec` etc. later.

Note that we only need to change the encoding logic. The decoding logic uses the deprecated `MemoryExec` etc. instead of `DataSourceExec` but it's fine for now.

crates/sail-execution/src/codec.rs
```diff
@@ -585,6 +587,53 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
...
+                    NodeKind::NdJson(gen::NdJsonExecNode {
+                        base_config,
+                        file_compression_type,
+                    })
+                } else if file_source.as_any().is::<ArrowSource>() {
+                    let base_config =
+                        self.try_encode_message(serialize_file_scan_config(file_scan, self)?)?;
+                    NodeKind::Arrow(gen::ArrowExecNode { base_config })
+                } else {
+                    return plan_err!("unsupported data source node: {data_source:?}");
```
linhr, line 608: Other file formats such as Parquet are handled inside DataFusion so it will not reach here.

shehabgamin: What about `CsvSource`?

linhr: CSV is also handled by DataFusion internally.

PR #404, opened 2025-03-06
Fix issue with task error reporting

linhr:
In distributed query execution, the task should return the error in the control flow and let the driver fail the job, but before this happens, in the data flow, the task stream may have been closed so the driver thinks the execution is successful with empty output.

This PR fixes this issue. In the longer term, we may want to consider whether the data flow should have the capability of returning errors from tasks, so that we can avoid such a race condition.

### DIFF
--- a/crates/sail-execution/src/driver/actor/output.rs
+++ b/crates/sail-execution/src/driver/actor/output.rs
@@ -12,10 +12,17 @@ use crate::id::JobId;
 
 pub(super) enum JobOutput {
     Pending {
+        /// The sender for the signal of job completion,
+        /// or [`None`] if the signal has already been sent.
+        tx: Option<oneshot::Sender<Option<String>>>,
+        /// The receiver for the signal of job completion.
+        rx: oneshot::Receiver<Option<String>>,
         result: oneshot::Sender<ExecutionResult<SendableRecordBatchStream>>,
     },
     Running {
-        signal: oneshot::Sender<String>,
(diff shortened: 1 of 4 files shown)


crates/sail-execution/src/driver/actor/output.rs
```diff
@@ -25,24 +32,23 @@ impl JobOutput {
...
-        signal: oneshot::Receiver<String>,
+        signal: oneshot::Receiver<Option<String>>,
         sender: mpsc::Sender<Result<RecordBatch>>,
     ) {
-        tokio::select! {
-            _ = Self::read(stream, sender.clone()) => {},
-            _ = Self::stop(signal, sender) => {},
-        }
+        Self::read(stream, sender.clone()).await;
+        Self::wait(signal, sender).await;
```
linhr, line 51: This does not fix the fundamental issue of the race condition between the control flow and data flow. Consider the following sequence of events:
1. The task fails and reports the status to the driver.
2. The driver marks the job as failed and writes the failure reason to the one-shot channel (whose receiver is `signal`).
3. The driver asks the worker to discard all task streams for the failed job.
4. The job output attempts to fetch the task stream in `Self::read()`.
5. The job output gets the `task stream not found` internal error and sends the error via `sender` inside `Self::read()`.
6. The internal error is returned to the user, before the job output calls `Self::wait()` to get the actual failure reason from `signal` and send the error via `sender`.

linhr:
This is superseded by #409.

PR #406, opened 2025-03-10
Rewrite Spark gold data generator logic in Rust

linhr:
This PR implements the Spark gold data generator logic in Rust to replace the original implementation based on jq scripts. This is to prepare for upcoming changes to collect more Spark tests. The jq scripts are not very flexible or maintainable for such tasks.

Interestingly, this PR also shows a bug in the jq scripts, where the first test case in each suite is dropped incorrectly.


scripts/spark-gold-data/process_data_type.jq
```diff
@@ -1,7 +0,0 @@
-{
-    "tests": [
-        inputs
```
linhr, line 3: The usage of `inputs` in the jq scripts is incorrect. `inputs` contains remaining JSON objects after the first one. So the first test case in each suite is dropped incorrectly.

PR #409, opened 2025-03-11
Improve task error reporting

linhr:
1. Update the task stream interface to support error reporting from task streams (via Tonic HTTP headers).
2. Ensure the control plane can return the precise error (via the driver `ReportTaskStatus` gRPC endpoint). This is useful when the task "running" and "failed" events are received out of order. In this case, the task stream is never created, so we have to rely on the control plane to return the error to the user.
3. Add `CommonErrorCause` that can recursively find the best error cause from an `Error`. This is used in both the `sail-execution` and `sail-spark-connect` crates.

I've run the PySpark tests locally in `local-cluster` mode. The number of passed tests is now the same as the number for `local` mode.


crates/sail-execution/src/driver/actor/handler.rs
```diff
@@ -838,13 +874,9 @@ impl DriverActor {
...
     }
 
-    fn cancel_job(&mut self, ctx: &mut ActorContext<Self>, job_id: JobId, reason: String) {
+    fn cancel_job(&mut self, ctx: &mut ActorContext<Self>, job_id: JobId, cause: CommonErrorCause) {
         if let Some(output) = self.job_outputs.remove(&job_id) {
-            output.fail(reason);
-        }
-        for worker_id in self.state.detach_job_from_workers(job_id) {
-            self.remove_worker_streams(ctx, worker_id, job_id);
-            self.schedule_idle_worker_probe(ctx, worker_id);
```
linhr, line 847: We should not prematurely remove streams from the worker here. Otherwise the job output may return a "task stream not found" error, which hides the actual error. Removing streams now happens exclusively in the `CleanUpJob` event handler.

PR #411, opened 2025-03-13
Fix rustup clippy installation issue

linhr:
We started getting the following error in GitHub Actions for Rust setup. This PR tries to fix the issue.

```text
error: invalid value 'clippy' for '[TOOLCHAIN]...': invalid toolchain name: 'clippy'
```


linhr:
The issue is fixed by writing the `rustup` commands in a different way. But I'm not sure why the issue started happening recently. The original commands are still working when I tried them locally.

PR #86, opened 2024-07-11
Fix Spark Connect library issue

linhr:
1. Hopefully fix the occasional Spark tests timeout, possibly due to deadlock in Spark Connect Python client.
2. Add fault handler in pytest, so that we can gather more information when timeout happens in tests in the future.
3. Move the PySpark packaging fix to the patch for `setup.py`.

The developer needs to rebuild the PySpark package locally after this change.

PR #413, opened 2025-03-18
Fix temporary view issues

linhr:
1. Fix temporary view issues.
2. Add catalog doctests.


python/pysail/tests/spark/conftest.py
```diff
@@ -52,9 +54,22 @@ def configure_spark_session(session: SparkSession):
     session.conf.set("spark.sql.execution.arrow.pyspark.enabled", "true")
 
 
-@pytest.fixture(scope="session", autouse=True)
+def patch_spark_connect_session(session: pyspark.sql.connect.session.SparkSession):
```
linhr, line 57: This patch is needed for the same reason as #86.

The deadlock becomes apparent in this PR since we now have one session per module in pytest.

PR #415, opened 2025-03-19
Support URL glob patterns

linhr:
DataFusion `ListingTableUrl` does not support glob patterns in URLs that are not file paths. Also, the glob syntax in the `glob` crate is different from the Spark glob syntax.

This PR implements parsing logic for URLs with glob patterns.


Cargo.toml
```diff
@@ -109,6 +110,37 @@ object_store = { version = "0.11.0", features = ["aws", "gcp", "azure", "http"]
 [patch.crates-io]
 # Override dependencies to use our forked versions.
 # You can use `path = "..."` to temporarily point to your local copy of the crates to speed up local development.
+# FIXME: use a commit from the main branch
```
linhr: This PR should not be merged before this is updated.

PR #416, opened 2025-03-20
Update Deps

shehabgamin:


Cargo.toml
```diff
@@ -5,82 +5,90 @@ members = [
 resolver = "2"
 
 [workspace.package]
+# Define LakeSail version
```
linhr: Let's change "LakeSail" to "Sail" here.

Or maybe we can simply remove this comment.

Cargo.toml
```diff
@@ -94,10 +102,10 @@ datafusion-proto = { version = "46.0.0" }
 datafusion-functions-nested = { version = "46.0.0" }
 datafusion-functions-json = { git = "https://github.com/lakehq/datafusion-functions-json.git", rev = "da30208" }
```
linhr: I noticed the DataFusion version has changed to 46.0.1 in `Cargo.lock`. Could we also change it here?

We may need to update our fork for `datafusion-functions-json` as well.

Pending confirmation: <https://github.com/apache/datafusion/issues/15328>

linhr (approved):

PR #421, opened 2025-03-24
Improve MCP server logging

linhr:
1. Emit Python log records to Rust loggers to make the logging style consistent. This is inspired by [dylanbstorey/pyo3-pylogger](https://github.com/dylanbstorey/pyo3-pylogger).
2. Support connecting the MCP server to a separate Spark remote endpoint.


crates/sail-cli/src/main.rs
```diff
@@ -30,3 +32,28 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
     Ok(())
 }
+
+fn run_python_interpreter() -> ! {
```
linhr, line 36: This is an existing function moved from `crates/sail-cli/src/python.rs`.

PR #432, opened 2025-04-03
Run object store in a separate Tokio runtime

linhr:
This PR adds a wrapper to `ObjectStore` so that the async IO tasks are run in a separate Tokio runtime. This ensures that IO tasks can be scheduled when there are compute-intensive tasks.

### References

* <https://github.com/apache/arrow-rs-object-store/issues/272>
* <https://github.com/apache/datafusion/issues/12393>
* <https://github.com/apache/datafusion/issues/13692>
* <https://github.com/apache/datafusion/pull/13424>
* <https://github.com/apache/datafusion/pull/13634>
* <https://github.com/apache/datafusion/pull/13690>
* <https://github.com/apache/datafusion/pull/14286>
* <https://github.com/datafusion-contrib/datafusion-dft/pull/248>
* <https://github.com/delta-io/delta-rs/pull/2789>


linhr:
After this change, the `error decoding response body` error has disappeared when running benchmarks involving large datasets (`q18` and sometimes `q21` for derived TPC-H benchmark at SF=1000).

But it's unclear to me whether there are performance issues after this change. The benchmark execution time has a high variance, so more analysis may be needed to ensure there is no performance degradation in other parts of the system after this change.

Cargo.toml
```diff
@@ -161,4 +161,4 @@ overflow-checks = false
 lto = true
 panic = 'unwind'
 incremental = false
-codegen-units = 1
+codegen-units = 16
```
shehabgamin: This should be 1

linhr: Yeah this is a temporary change to speed up compilation during testing. It didn't help much though. So I'll revert the change.

crates/sail-execution/src/driver/actor/handler.rs
```diff
@@ -1023,6 +1025,7 @@ impl TaskSlotAssigner {
     }
 
     pub fn next(&mut self) -> Option<WorkerId> {
+        self.slots.shuffle(&mut rng());
```
linhr: I tried randomizing task assignment and it didn't seem to help. I guess future work is needed for more sophisticated task assignment logic.

linhr:
I decided to put the feature behind configuration so that we can enable the separate Tokio runtime explicitly and continue experimenting with it.

This PR now also has the following changes.
1. Consolidate Tokio runtime creation logic for various entrypoints (Python library, PySpark shell, CLI, etc.).
2. Avoid creating a `SessionContext` for every task in the worker.

PR #438, opened 2025-04-15
Fix default time zone issue

linhr:
This PR fixes the bug of hard-coded default time zone for Spark sessions. After this change, various time zone adjustment workarounds (relying on system time zone) are no longer needed. Such workarounds only partially worked and have known issues, and were based on a few assumptions which turned out to be incorrect now.

The simplified timestamp logic relies on the following facts.
1. SQL timestamp literal (e.g. `TIMESTAMP '2025-01-01 00:00:00'`) assumes session time zone.
2. PySpark timestamp literals (e.g. `lit(datetime(2025, 1, 1))`) assumes client system time zone. But such literals are converted to or from a UTC timestamp before sent to or after received from the server. The server does not need to be aware of how the literal is represented on the client side. Therefore, we do not need to assume that the server and the client share the same system time zone.
3. For the Arrow `Timestamp` type, if the time zone is `None`, we treat the timestamp as `TIMESTAMP_NTZ`. If the time zone is `Some`, we treat the timestamp as `TIMESTAMP_LTZ`.
4. For the Arrow `Timestamp` type, changing the time zone from `Some` to a different `Some` is a metadata-only change. No adjustment should be made to the data since the value is always UTC.

This PR also adds PySpark unit tests with a focus on daylight saving time zones and the NTZ timestamp type. Many tests are marked as skipped due to formatting or type coercion issues. The issues will be fixed in future PRs since this PR is already large. But I've manually verified that the response is "logically" correct.


crates/sail-common-datafusion/src/datetime/date.rs
```diff
@@ -0,0 +1,26 @@
+use chrono::format::{parse_and_remainder, Item, Numeric, Pad, Parsed};
+use chrono::{NaiveDate, ParseError};
+use datafusion::arrow::datatypes::Date32Type;
+use datafusion_common::{exec_datafusion_err, Result};
+
+pub fn parse_date(s: &str) -> Result<NaiveDate> {
```
linhr, line 5: Right now I use the low-level `chrono` API for date and timestamp parsing. I'm considering rewriting the parsing logic in chumsky since Spark supports partially specified date and timestamp literals (e.g. `DATE '1970-01'`).

crates/sail-common-datafusion/src/datetime/timestamp.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn localize(&self, datetime: &NaiveDateTime) -> MappedLocalTime<DateTime<Utc>> {
+        match self {
+            TimeZoneValue::Fixed(offset) => {
+                offset.from_local_datetime(datetime).map(|x| x.to_utc())
+            }
+            TimeZoneValue::Tz(tz) => tz.from_local_datetime(datetime).map(|x| x.to_utc()),
+        }
+    }
+
+    pub fn localize_with_fallback(&self, datetime: &NaiveDateTime) -> Result<DateTime<Utc>> {
```
linhr, line 45: Spark has this fallback logic due to how datetime is handled in JDK. This is why we cannot use the timestamp parsing logic in Arrow, which returns error for ambiguous/invalid naive datetime.

crates/sail-spark-connect/tests/gold_data/expression/date.json
```diff
@@ -40,17 +34,19 @@
       "input": "DAtE 'mar 11 2016'",
       "exception": "\n[INVALID_TYPED_LITERAL] The value of the typed literal \"DATE\" is invalid: 'mar 11 2016'.(line 1, pos 0)\n\n== SQL ==\nDAtE 'mar 11 2016'\n^^^\n",
       "output": {
-        "failure": "invalid argument: date: mar 11 2016"
+        "success": {
+          "unresolvedDate": {
+            "value": "mar 11 2016"
+          }
+        }
```
linhr, line 41: This becomes false positive now since the error is detected at a later stage in the plan resolver. The SQL parser no longer parses date and timestamp literals. It is not a regression.

python/pysail/tests/spark/literal/test_timestamp.txt
```diff
@@ -0,0 +1,85 @@
+>>> spark.conf.set("spark.sql.session.timeZone", "Asia/Shanghai")
+>>> spark.sql("SELECT CAST(TIMESTAMP '1970-01-01 00:00:00' AS LONG)").show()  # doctest: +SKIP
++-----------------------------------------------+
+|CAST(TIMESTAMP '1970-01-01 00:00:00' AS BIGINT)|
++-----------------------------------------------+
+|                                         -28800|
++-----------------------------------------------+
```
linhr, line 7: When casting timestamps to integers, Spark returns the number of whole seconds while the Arrow result depends on the time unit. This needs to be fixed in a future PR.

crates/sail-spark-connect/src/proto/literal.rs
```diff
@@ -99,11 +99,11 @@ impl TryFrom<Literal> for spec::Literal {
             LiteralType::Date(x) => spec::Literal::Date32 { days: Some(x) },
             LiteralType::Timestamp(x) => spec::Literal::TimestampMicrosecond {
                 microseconds: Some(x),
-                timezone_info: spec::TimeZoneInfo::LocalTimeZone,
+                timestamp_type: spec::TimestampType::WithLocalTimeZone,
             },
             LiteralType::TimestampNtz(x) => spec::Literal::TimestampMicrosecond {
                 microseconds: Some(x),
-                timezone_info: spec::TimeZoneInfo::NoTimeZone,
+                timestamp_type: spec::TimestampType::WithLocalTimeZone,
```
shehabgamin: Why is this one `WithLocalTimeZone`?

linhr: Good catch! Fixed.

PR #440, opened 2025-04-22
Upgrade DataFusion to 47

linhr:


linhr:
The regression for the `last` function seems related to <https://github.com/apache/datafusion/issues/15676>. I do not see an easy workaround for this, so we probably have to accept the regression for now to unblock the upgrade. More investigation is needed in the future.

PR #443, opened 2025-04-23
Clean up literal value parser

linhr:
This PR removes the `LiteralValue` struct and various `TryFrom` trait implementation. The logic is now written in functions. This makes literal parsing logic in the SQL analyzer easier to understand.

### DIFF
--- a/crates/sail-sql-analyzer/src/literal.rs
+++ b/crates/sail-sql-analyzer/src/literal.rs
@@ -1,4 +1,3 @@
-use std::fmt::Debug;
 use std::iter::once;
 use std::ops::Neg;
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


crates/sail-sql-analyzer/src/literal.rs
```diff
@@ -57,187 +56,56 @@ lazy_static! {
...
-    type Error = SqlError;
-
-    fn try_from(literal: LiteralValue<chrono::NaiveDate>) -> SqlResult<spec::Literal> {
-        let value = chrono::NaiveDateTime::from(literal.0);
-        let days = (value - chrono::NaiveDateTime::UNIX_EPOCH).num_days();
-        let days = i32::try_from(days)
-            .map_err(|_| SqlError::invalid(format!("date literal: {:?}", literal.0)))?;
-        Ok(spec::Literal::Date32 { days: Some(days) })
-    }
-}
```
linhr, line 174: This is unused code, which becomes apparent during the clean up.

PR #447, opened 2025-04-29
Spark Coverage: Configs and Metadata

shehabgamin:


crates/sail-common/src/config/application.rs
```diff
@@ -117,6 +118,7 @@ pub enum RetryStrategy {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ExecutionConfig {
     pub batch_size: usize,
+    pub listing_table_ignore_subdirectory: bool,
```
linhr: I feel some configuration that we had hard-coded value should not be configurable, since they break the Spark behavior.

crates/sail-common/src/config/application.rs
```diff
@@ -136,10 +138,170 @@ pub struct KubernetesConfig {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ParquetConfig {
+    /// Doc Comments [Credit]: <https://github.com/apache/datafusion/blob/555fc2e24dd669e44ac23a9a1d8406f4ac58a9ed/datafusion/common/src/config.rs#L423-L592>
```
linhr: I'm not comfortable copying the docs here since there will be no way to detect out-of-date information. Instead, we could simply refer to the key in DataFusion. 

If we want to keep this for now, the `///` comment should be moved to `ParquetConfig` since it's not valid here. BTW `[Credit]` is not valid Markdown syntax but unfortunately we have been using it in many places.

crates/sail-common/src/config/default.toml
```diff
@@ -29,8 +29,40 @@ rpc_retry_strategy = { "fixed" = { "max_count" = 3, "delay_secs" = 5 } }
...
+
+[catalog]
+enable_information_schema = true
+has_header = false
 
 [parquet]
+enable_page_index = true
+pruning = true
+skip_metadata = true
+#metadata_size_hint =
```
linhr, line 40: Let's add a space after `#` for all the commented keys.

crates/sail-common/src/config/default.toml
```diff
@@ -29,8 +29,40 @@ rpc_retry_strategy = { "fixed" = { "max_count" = 3, "delay_secs" = 5 } }
...
+pushdown_filters = false
+reorder_filters = false
+schema_force_view_types = true
+binary_as_string = false
+coerce_int96 = "us"
```
linhr: `coerce_int96` should not be configurable.

crates/sail-plan/src/resolver/expression.rs
```diff
@@ -66,24 +66,45 @@ pub(super) struct NamedExpr {
...
                 "alias expected to create named expression",
             )),
         }
     }
 
     pub fn with_metadata(mut self, metadata: Vec<(String, String)>) -> Self {
-        self.metadata = metadata;
+        if self.metadata != metadata {
+            self.metadata = metadata;
+        }
```
linhr: This there a reason for this? I feel `self.metadata = metadata` would just work, and it avoids the overhead of comparing the metadata.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -26,13 +26,18 @@ impl TryFrom<adt::Field> for sdt::StructField {
...
         };
-        let metadata = field
+        let metadata = &field
             .metadata()
             .iter()
-            .filter(|(k, _)| k.starts_with("metadata."))
-            .map(|(k, v)| (k["metadata.".len()..].to_string(), v.clone()))
-            .collect::<HashMap<_, _>>();
-        let metadata = serde_json::to_string(&metadata)?;
+            .filter(|(k, _)| !k.starts_with("udt."))
```
linhr, line 33: `k.starts_with("metadata.")` is safer since we may have other key prefixes in the future. And we should not use `strip_prefix` below since any metadata key coming from the outside should be prefixed with `metadata.`.

crates/sail-common/src/config/default.toml
```diff
@@ -29,8 +29,40 @@ rpc_retry_strategy = { "fixed" = { "max_count" = 3, "delay_secs" = 5 } }
 
 [execution]
 batch_size = 8192
+listing_table_ignore_subdirectory = false
+
+[catalog]
+enable_information_schema = true
```
linhr: `enable_information_schema` should not be configurable.

crates/sail-common/src/config/application.rs
```diff
@@ -136,10 +138,170 @@ pub struct KubernetesConfig {
...
+    /// writing out already in-memory data, such as from a cached
+    /// data frame.
     pub maximum_buffered_record_batches_per_stream: usize,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CatalogConfig {
+    pub enable_information_schema: bool,
+    /// Spark defaults to false: https://spark.apache.org/docs/latest/sql-data-sources-csv.html
+    pub has_header: bool,
```
linhr, line 290: `has_header` is not a catalog configuration. It applies to CSV format only.

shehabgamin: @linhr In Datafusion it is. I guess the comment is misplaced, do you want me to remove the comment?

linhr: I feel the DataFusion way to handle that is rather confusing. IMO I'd simply leave out it in the Sail configuration.

```rust
/// Default value for `format.has_header` for `CREATE EXTERNAL TABLE`
/// if not specified explicitly in the statement.
pub has_header: bool, default = true
```

crates/sail-common/src/config/default.toml
shehabgamin: It's configurable in DataFusion and there are many reasons why someone may want to disable it

crates/sail-common/src/config/default.toml
shehabgamin: That will add annoying whitespace errors in git.

crates/sail-common/src/config/default.toml
linhr: Actually I was thinking we should not rely on `information_schema` at all and I had some thoughts to rewrite some of the catalog features without relying on `information_schema`. The schema of `information_schema` may not be compatible with Spark and I'd prefer not to expose this kind of details.

crates/sail-common/src/config/default.toml
shehabgamin: I think it should be because we may support other engines besides Spark in the future.

crates/sail-common/src/config/default.toml
linhr: For now, if we allow people to disable `information_schema`, it breaks a few catalog APIs.

crates/sail-plan/src/resolver/expression.rs
shehabgamin: No, nice catch.

crates/sail-common/src/config/default.toml
linhr: In general I agree that we will support behaviors outside Spark. But I feel it's premature to expose this. Right now there are many places where we assume timestamps have microsecond time unit. If people change this, many UDFs or data type conversion will simply break.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
shehabgamin: `k.starts_with("metadata.")` doesn't work and breaks the tests. Not all metadata keys start with `metadata.`. When inspecting the code, I found that `udt.` is internally set by us and should be skipped.

`strip_prefix` is the equivalent of the previous logic with the added benefit of allowing other keys that don't start with `metadata.`:
`.map(|(k, v)| (k["metadata.".len()..].to_string(), v.clone()))`

crates/sail-common/src/config/application.rs
shehabgamin: @linhr I'm not really sure how out-of-date information would arise here. It would be strange if the literal definition of a `key` changed. We need some sort of documentation for our docs, and it's not very clean to link users to another website entirely for some reference.

crates/sail-common/src/config/application.rs
shehabgamin: For this specific config, are you certain that Spark doesn't have an equivalent?

Regardless, I think it makes sense to keep because we may support other engines besides Spark in the future. We have the default set in the `.toml` file and if the user changes it, that's on them imo.

crates/sail-common/src/config/application.rs
shehabgamin: Let me know if you have another suggestion or would like to modify the comments in some way to better protect against your concern

crates/sail-spark-connect/src/proto/data_type_arrow.rs
linhr: > Not all metadata keys start with `metadata.`.

If this is the case, there is a bug somewhere else that should be fixed. The `metadata.` prefix is managed by Sail and the convention should be respected everywhere.

Let's at least add a FIXME here and I'll take a look.

crates/sail-common/src/config/application.rs
shehabgamin: @linhr The problem is that it's configurable in Spark

crates/sail-common/src/config/application.rs
linhr: Similar to `coerce_int96`, this is again a configuration that would simply **break** things without adding new features beyond Spark. There is likely more work to do in our listing table resolver besides simply exposing this configuration. For the resolver to function properly, it currently assumes that subdirectories are always included in the listing result.

crates/sail-common/src/config/application.rs
linhr: In Spark, the header option is configured at the reader and we already support it.

crates/sail-common/src/config/application.rs
shehabgamin: Done!

crates/sail-common/src/config/application.rs
shehabgamin: Done!

crates/sail-common/src/config/default.toml
shehabgamin: Done!

crates/sail-common/src/config/default.toml
shehabgamin: Done!

crates/sail-spark-connect/src/proto/data_type_arrow.rs
shehabgamin: Done!

PR #448, opened 2025-04-29
Improve datetime type casting

linhr:


crates/sail-plan/src/extension/analyzer/mod.rs
```diff
@@ -0,0 +1,14 @@
+use std::sync::Arc;
+
+use datafusion::optimizer::{Analyzer, AnalyzerRule};
+
+pub fn default_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
```
linhr, line 5: I tried to implement type casting logic as analyzers but it breaks cases when the logic should not be applied. So I ended up having the logic in the plan resolver. The setup for analyzers and optimizers is kept here in case we need them in the future.

PR #458, opened 2025-05-13
Add MCP endpoints to list local and S3 paths

linhr:


linhr:
The diff is a bit hard to read, but what I did was:
1. Removed all `create_xxx_views_for_subfolders` endpoints.
2. Added `list_local_directories` and `list_s3_directories` endpoints.
3. Added the `header` parameter for the `create_csv_view` endpoint.

PR #466, opened 2025-05-21
feat: CSV Reader Configs

shehabgamin:


crates/sail-common/src/config/data_source/csv_read.yaml
```diff
@@ -0,0 +1,333 @@
+# References:
+#   - [1] https://spark.apache.org/docs/4.0.0/sql-data-sources-csv.html#data-source-option
+#   - [2] https://github.com/apache/spark/blob/b0c2ba357bf080dd328b95e4a6402b134a641a1a/python/pyspark/sql/connect/readwriter.py#L254-L290
+#   - [3] https://github.com/apache/spark/blob/b0c2ba357bf080dd328b95e4a6402b134a641a1a/python/pyspark/sql/connect/readwriter.py#L767-L787
+
+# CHECK HERE: ADD CAMEL CASE ALIAS FOR ALL KEYS BEFORE MERGING!!!!
```
linhr: Remove this line?

crates/sail-plan/src/data_source/csv.rs
```diff
@@ -0,0 +1,123 @@
...
+        match (&null_value, &null_regex) {
+            (Some(null_value), Some(null_regex))
+                if !null_value.is_empty() && !null_regex.is_empty() =>
+            {
+                Err(PlanError::internal(
+                    "CSV `null_value` and `null_regex` cannot be both set",
+                ))
+            }
+            _ => Ok(()),
+        }?;
```
linhr: You don't need to detect conflicting options here since the resolver takes care of it.

The `CsvReadOptions` struct itself does not provide any consistency guarantee due to the presence of both `null_value` and `null_regex` as separate fields. As an alternative, if we define the null value as an enum inside `CsvReadOptions`, then the validation happens here and the resolver always gets a valid enum value.

The general idea is that we use the rich type system in Rust to express constraints whenever possible.

crates/sail-plan/src/data_source/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    supported: bool,
+}
+
+impl ConfigItem {
+    // user_options_normalized is expected to be a HashMap with all keys in lowercase
+    fn resolve_value(&self, user_options_normalized: &HashMap<String, String>) -> String {
+        if let Some(value) = user_options_normalized.get(&self.key.to_lowercase()) {
+            return value.clone();
+        }
+        for alias in &self.alias {
```
linhr, line 34: If both the key and the alias are present in the map, such duplication is silently ignored and the user does not know which one is picked without looking at the code. Let's add a TODO here for such situation.

crates/sail-plan/src/resolver/option.rs
```diff
@@ -1,39 +1,68 @@
...
+            {
+                Err(PlanError::internal(
+                    "CSV `null_value` and `null_regex` cannot be both set",
+                ))
             }
-        }
-        Ok(output)
+            (Some(null_value), _) if !null_value.is_empty() => {
+                // Convert null_value to regex by escaping special characters
+                Ok(Some(regex::escape(&null_value)))
```
linhr, line 35: Nice!

crates/sail-plan/src/resolver/option.rs
```diff
@@ -1,39 +1,68 @@
...
+            .with_delimiter(options.delimiter.parse().map_err(|e| {
+                PlanError::internal(format!("Invalid CSV `delimiter` read option: {e}"))
+            })?)
+            .with_quote(options.quote.parse().map_err(|e| {
+                PlanError::internal(format!("Invalid CSV `quote` read option: {e}"))
+            })?)
+            .with_terminator(
+                options
+                    .line_sep
+                    .map(|s| s.parse())
```
linhr: Could you try out the options manually and see if the usage of `.parse()` is correct? It seems to me that this is parsing digits into `u8` rather than converting a character to its ASCII code.

crates/sail-common/src/config/data_source/csv_read.yaml
shehabgamin: Oh lol nice catch.

crates/sail-plan/src/data_source/mod.rs
shehabgamin: Ooh yeah, nice catch!

crates/sail-plan/src/resolver/option.rs
shehabgamin: Very nice catch, yes I need to change the data type of line_sep to be a char.

shehabgamin: Actually I need to change multiple data types. For some reason I thought that the DataFusion config accepted `Vec<u8>`.

crates/sail-plan/src/resolver/option.rs
```diff
@@ -1,39 +1,46 @@
...
-        Ok(output)
+            (Some(null_value), _) if !null_value.is_empty() => {
+                // Convert null_value to regex by escaping special characters
+                Ok(Some(regex::escape(&null_value)))
+            }
+            (_, Some(null_regex)) if !null_regex.is_empty() => Ok(Some(null_regex)),
+            _ => Ok(None),
+        }?;
+
+        let file_format = CsvFormat::default()
```
linhr, line 41: Let's add a TODO here to return error if the `char` is not an ASCII character, otherwise `as u8` only returns the last byte.

linhr (approved):
🚀

linhr (approved):

PR #467, opened 2025-05-22
Update Spark Connect protocol for Spark 4.0

linhr:
This PR updates the Spark Connect protocol to support the upcoming Spark 4.0 release. Since the protocol is backward-compatible, tests from Spark 3.5 still runs without regression. The test suite will be updated for Spark 4.0 in a future PR.

The `.proto` files are copied from the `v4.0.0-rc7` tag (`fa33ea0`) in the Spark codebase.


shehabgamin:
@linhr Great work! Almost done reviewing.

We also need to update the supported Python versions:
https://github.com/apache/spark/blob/v4.0.0/python/packaging/connect/setup.py

linhr:
> We also need to update the supported Python versions

Good point! Let me take care of this as I work on the Python changes in the next PR. This PR is dedicated to protobuf upgrade to make sure the server is still compatible with Spark 3.5.

Issue #474, opened 2025-05-24
PySpark DataFrame Doctest: sample


shehabgamin:
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.sample`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.sample.html

davidlghellin:
I'm interested in trying to work on this. Could you please assign it to me?

Thanks in advance — I hope it’ll be good practice and that I can do a decent job! 😊

shehabgamin:
> I'm interested in trying to work on this. Could you please assign it to me?
> 
> Thanks in advance — I hope it’ll be good practice and that I can do a decent job! 😊

Done. Thank you so much!!

PR #496, opened 2025-05-24
feat: sample functions

davidlghellin:
Hi 🙂 This is a draft PR to check if I understood the issue correctly.

I wanted to confirm if these would be the right steps to follow.
If it makes sense, I’ll continue working in this direction.

Thanks for the feedback — and sorry if this wasn’t the right issue 😅

#474 

add  [https://spark.apache.org/docs/3.5.5/api/python/_modules/pyspark/sql/dataframe.html#DataFrame.sample](sample)


shehabgamin:
@davidlghellin Yep! You're on the right track. Check out how alias is dealt with in other places in the code base though. It is a little unintuitive.

davidlghellin:
I think I have made some progress, I do not know if the approach is correct, I thought about adding the random column and then make the filter with that column, what I do not know is then how I will raise the rest or if I have to refocus everything, but I'm watching and learning a lot.
For the moment there are some values to pinch.

Thanks
image

davidlghellin:
hello again

I think I have made some progress, but I still have some doubts:
- The random I use I don't know if it's the appropriate one and it's vectorized.
- I still have to try and see how to launch or use the tests for the function
- For the part with replacement, I have to see how to generate random with the Poisson function.

Thanks

shehabgamin:
@davidlghellin Nice progress! Yeah setting up the tests will be really important to get you going. 

These instructions can help get you setup with the Spark tests:
https://docs.lakesail.com/sail/latest/development/spark-tests/

davidlghellin:
Hi again, 
I am trying to set up the tests and I have not been able to compile, apart from the time it has failed.
image
I tried in devcontainer
I don't know if there is any option to download the jars and not have to compile, or that option is not viable.
I hope to find the time and try again
Thanks

linhr:
Would you mind trying on the host? Devcontainer is a new addition to our repo and we haven't validated the entire development process yet. It seems to me that your build takes much longer than we usually expect (> 3 hours vs 30 minutes). I suspect that the container may not have enough resources or there is some performance issue due to CPU architecture mismatch.

Or if you have more detailed error messages, we'd be happy to take a closer look.

davidlghellin:
ok, thanks, I will prepare a laptop for it, so I can retake debian with awesome or try nixOS.

shehabgamin:
@davidlghellin We’ve updated the GitHub workflow to automatically trigger tests whenever a pull request has the run spark tests label. I’ve applied the label to this PR, but you’ll need to merge from the main branch first!

davidlghellin:
> @davidlghellin We’ve updated the GitHub workflow to automatically trigger tests whenever a pull request has the run spark tests label. I’ve applied the label to this PR, but you’ll need to merge from the main branch first!

Thanks for this

davidlghellin:
> `doctest-dataframe`
> **After**	27	79	1	4	9.78
> **Before**	28	78	1	4	9.25
> `test-connect`	
> **After**	218	818	135	283	151.28
> **Before**	219	817	135	283	146.57


Oh thanks with these (good tools), we can see better update tests 🥹

davidlghellin:
Hi again
Honestly, I’m a bit lost right now. I’m unsure if what I’ve done so far is correct, especially around the Poisson random number generation, and I don’t really know how to move forward from here. Any help or guidance would be appreciated! 

(and sorry for the trouble!)

https://github.com/apache/spark/blob/357f4ade2deabdee9a50eab4673601da3c8c7d29/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L181

Edit: its for the random poisson, sampling with replacement

shehabgamin:
> Hi again Honestly, I’m a bit lost right now. I’m unsure if what I’ve done so far is correct, especially around the Poisson random number generation, and I don’t really know how to move forward from here. Any help or guidance would be appreciated!
> 
> (and sorry for the trouble!)
> 
> https://github.com/apache/spark/blob/357f4ade2deabdee9a50eab4673601da3c8c7d29/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L181
> 
> Edit: its for the random poisson, sampling with replacement

@davidlghellin Your approach with a `UDF` seems to be the right direction! What are you stuck on specifically?

davidlghellin:
> @davidlghellin Your approach with a `UDF` seems to be the right direction! What are you stuck on specifically?

Hi again
I feel like I’ve made some progress! I wasn’t sure how to integrate it at first, but I think I’ve found a way. Now I’m checking how to make the call correctly.
Thanks again — and sorry for all the extra trouble 😅

shehabgamin:
> > @davidlghellin Your approach with a `UDF` seems to be the right direction! What are you stuck on specifically?
> 
> 
> 
> Hi again
> 
> I feel like I’ve made some progress! I wasn’t sure how to integrate it at first, but I think I’ve found a way. Now I’m checking how to make the call correctly.
> 
> Thanks again — and sorry for all the extra trouble 😅


It's no trouble at all! Happy to help anytime and thank you so much for working on this!

davidlghellin:
I think is the good road
https://github.com/lakehq/sail/pull/496/commits/911cca6a1e535500d6357db3b5d01a67b0d437ed

image
now is
0 => drop file
n > 0 => duplicate n times

davidlghellin:
sorry for the spam
https://github.com/lakehq/sail/pull/496/commits/0d86f9c8e77b8c02bd6d0d12ba3e9c4132560fc2
image
```python
df = spark.range(10)
df.sample(True, 0.5, 3).show()

```

davidlghellin:
https://github.com/lakehq/sail/pull/496/commits/100b74c6042edb3b800afe158485c8fd97eadc14
image

davidlghellin:
While reviewing the execution plans, I noticed that rows with value 0 were being filtered out early in the process. I’ve removed that step to simplify the plans, assuming this is the correct approach — but if that’s not the case, please let me know and I’ll revert the change.
[42d4b98](https://github.com/lakehq/sail/pull/496/commits/42d4b9841ec685b4816bcb24b345226fb272eb80)
filter 0
image
drop filter 0 (commit)
image

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,74 @@
...
+impl Default for RandPoisson {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl RandPoisson {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::any(1, Volatility::Volatile),
```
shehabgamin: It looks like the `UDF` can either accept 0 or 1 input args. Perhaps we do something like this?
```
signature: Signature::one_of(vec![TypeSignature::Nullary, TypeSignature::Any(1)], Volatility::Volatile),
```

Or better, you could do:
```
signature:: Signature::user_defined(Volatility::Immutable)
```
And then have:
```
impl ScalarUDFImpl for RandPoisson
...
    fn coerce_types(&self, arg_types: &[DataType]) -> Result<Vec<DataType>> {
    ...
    }
...
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Int64)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let ScalarFunctionArgs {
+            args, number_rows, ..
+        } = args;
+        if args.is_empty() {
```
shehabgamin, line 56:
```suggestion
        if args.len() > 1 {
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,74 @@
...
+            );
+        }
+
+        let [_seed] = args.as_slice() else {
+            return exec_err!(
+                "random should be called with at most 1 argument, got {}",
+                args.len()
+            );
+        };
+        invoke_no_seed(number_rows)
```
shehabgamin: This line should go before line 58. Maybe we do something like this:
```
if args.is_empty() {
   invoke_no_seed(number_rows);
} else {
    let [_seed] = args.as_slice() else {
        return exec_err!(
            "random should be called with at most 1 argument, got {}",
            args.len()
        );
    };
}
```
And then let's make sure to handle the seed case.

shehabgamin:
@davidlghellin I noticed that the `rand` and `randn` functions didn't handle the `seed` properly. Here is the fix I did that you can apply similarly in this pr:
https://github.com/lakehq/sail/pull/530

crates/sail-plan/src/extension/function/math/rand_poisson.rs
davidlghellin: with these change
> ```
> signature:: Signature::user_defined(Volatility::Immutable)
> ```
The value not change, its alwais the same 
https://github.com/lakehq/sail/pull/496/commits/8cba54171b7e6f952919e32488d8c38d5016294d 
image
we want change the value not fixed, its with replacement

davidlghellin: I saw the comment https://github.com/lakehq/sail/pull/496#issuecomment-2973585554 now its right

davidlghellin:
> @davidlghellin I noticed that the `rand` and `randn` functions didn't handle the `seed` properly. Here is the fix I did that you can apply similarly in this pr: #530

Thanks to this PR, I can now better see that using a fixed seed actually returns the same values (I initially thought it might change). Really appreciate it!

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,127 @@
...
+                    ScalarValue::UInt64(Some(value)) => *value as i64,
+                    ScalarValue::Int64(None) | ScalarValue::UInt64(None) | ScalarValue::Null => {
+                        return invoke_no_seed(number_rows)
+                    }
+                    _ => {
+                        return exec_err!("`random_poisson` expects an integer seed, got {scalar}")
+                    }
+                };
+                let poisson = Poisson::new(1.0).unwrap();
+                let mut rng = StdRng::seed_from_u64(seed as u64);
```
davidlghellin: needs change fixed lamda in Poisson, this is percent parameter

davidlghellin:
Sorry for the last failure commits, I am watching how to works test and the actions

davidlghellin:
Hi, 
I have a question to help finalize the PR, @shehabgamin.
I was using for the random values, but I think using **randn** might be a better fit.
Which one do you think is more appropriate?

Thanks

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl RandPoisson {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature {
+                type_signature: TypeSignature::Exact(vec![DataType::Float64, DataType::Int64]),
```
shehabgamin: This will only accept signatures with two arguments which contradicts the logic in `invoke_with_args`.

Also, `coerce_types` is only called if the signature is `user_defined`.

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,147 @@
...
+                Ok(ColumnarValue::Array(Arc::new(array)))
+            }
+            _ => exec_err!(
+                "`random_poisson` expects a scalar seed argument, got {}",
+                seed.data_type()
+            ),
+        }
+    }
+
+    fn coerce_types(&self, arg_types: &[DataType]) -> Result<Vec<DataType>> {
```
shehabgamin, line 117: This function needs to be updated to handle 2 args

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,147 @@
...
+        if args.len() == 1 {
+            return invoke_no_seed(lambda, number_rows);
+        }
+        let seed = &args[1];
+
+        match seed {
+            ColumnarValue::Scalar(scalar) => {
+                let seed = match scalar {
+                    ScalarValue::Int64(Some(value)) => *value,
+                    ScalarValue::UInt64(Some(value)) => *value as i64,
```
shehabgamin:
```suggestion
                    ScalarValue::UInt64(Some(value)) => *value,
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+        if args.len() == 1 {
+            return invoke_no_seed(lambda, number_rows);
+        }
+        let seed = &args[1];
+
+        match seed {
+            ColumnarValue::Scalar(scalar) => {
+                let seed = match scalar {
+                    ScalarValue::Int64(Some(value)) => *value,
```
shehabgamin:
```suggestion
                    ScalarValue::Int64(Some(value)) => *value as u64,
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,147 @@
...
+                    ScalarValue::UInt64(Some(value)) => *value as i64,
+                    ScalarValue::Int64(None) | ScalarValue::UInt64(None) | ScalarValue::Null => {
+                        return invoke_no_seed(lambda, number_rows)
+                    }
+                    _ => {
+                        return exec_err!("`random_poisson` expects an integer seed, got {scalar}")
+                    }
+                };
+                let poisson = Poisson::new(lambda).unwrap();
+                let mut rng = StdRng::seed_from_u64(seed as u64);
```
shehabgamin:
```suggestion
                let mut rng = StdRng::seed_from_u64(seed);
```

davidlghellin:
Thanks again! I also wanted to mention that in one of the changes I didn’t realize the role of coerce_types, and that caused some confusion on my side. I appreciate the review — it helped me spot those details.
Hopefully it’ll feel more natural for me as I get more used to it!

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+        Ok(DataType::Int64)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let ScalarFunctionArgs {
+            args, number_rows, ..
+        } = args;
+        if args.is_empty() {
+            return exec_err!(
+                "`random_poisson` should be called with at most 2 argument, got {}",
```
shehabgamin:
```suggestion
                "`random_poisson` must be called with either 1 or 2 arguments, got {}",
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+                args.len()
+            );
+        }
+        if args.len() > 2 {
+            return exec_err!(
+                "`random_poisson` should be called with at most 2 argument, got {}",
```
shehabgamin:
```suggestion
                "`random_poisson` must be called with either 1 or 2 arguments, got {}",
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+                args.len()
+            );
+        }
+        let lambda = &args[0];
+        let lambda = match lambda {
+            ColumnarValue::Scalar(scalar) => match scalar {
+                ScalarValue::Float64(Some(value)) => *value,
+                ScalarValue::Float64(None) | ScalarValue::Null => 1.0,
+                _ => return exec_err!("`random_poisson` expects a float64 lambda, got {scalar}"),
```
shehabgamin:
```suggestion
                _ => return exec_err!("`random_poisson` expects a scalar Float64 for `lambda`, got {scalar}"),
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+            },
+            _ => return exec_err!("`random_poisson` expects an float64 seed, got {lambda}"),
```
shehabgamin:
```suggestion
            _ => return exec_err!("`random_poisson` expects a scalar Float64 for `lambda`, got {lambda}"),
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+                Err(unsupported_data_types_exec_err(
+                    "random_poisson",
+                    "Integer Type for seed",
+                    arg_types,
+                ))
+            }
+        } else {
+            Err(invalid_arg_count_exec_err(
+                "random_poisson",
+                (0, 1000),
```
shehabgamin:
```suggestion
                (1, 2),
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
```diff
@@ -0,0 +1,158 @@
...
+            }
+            _ => exec_err!(
+                "`random_poisson` expects a scalar seed argument, got {}",
+                seed.data_type()
+            ),
+        }
+    }
+
+    fn coerce_types(&self, arg_types: &[DataType]) -> Result<Vec<DataType>> {
+        if arg_types.len() == 2 {
```
shehabgamin: I think we need something like this:
```
if arg_types.len() == 1 {
...
} else if arg_types.len() == 2 {
...
} else {
...
}
```

crates/sail-plan/src/extension/function/math/rand_poisson.rs
davidlghellin: sorry forget change this :(

davidlghellin:
Thanks! And if needed, feel free to update the title of the PR to match the standard.
Best regards!

linhr (approved):
Thank you for your contribution!

PR #515, opened 2025-06-03
feat: improve s3 performance

shehabgamin:
1. use `DashMap`
2. use Use entry API for atomic get-or-insert

20% faster on average when testing on a 100 GB dataset


linhr (approved):
Great work!

To complement the PR description, this PR now caches all object stores for performance reasons. The assumption is that the object store instances are not expected to change upon changes to session configuration, or such changes should be handled internally to the object store.

The note above is for future reference.

Issue #523, opened 2025-06-09
Docs/script mismatch: local testing setup fails without specific git clones


davidlghellin:
Hi team,

I recently went back to running the tests, following the steps described in the documentation. However, I encountered some issues: the setup was failing unless I manually cloned the following repositories into specific paths:

```
git clone git@github.com:apache/spark.git opt/spark
git clone git@github.com:ibis-project/testing-data.git opt/ibis-testing-data
```

<img width="616" alt="Image" src="https://github.com/user-attachments/assets/3cbfece9-0917-4e32-b4e6-34c3d022a676" />

It might be a good idea to either update the documentation to mention these required steps or adjust the script to handle this automatically.

```
git clone --branch v3.5.5 --depth 1 git@github.com:apache/spark.git opt/spark
```

or change the script
```
scripts/spark-tests/build-pyspark.sh
```

Thanks!

linhr:
Thanks for raising the issue!

The "unknown revision" error can happen when the repo was cloned a while ago but the build script expects a newer tag. The docs should have mentioned how this error could be handled.

The docs fix in your PR makes sense to me, so let me merge it. We're going to support testing both Spark 3.5 and 4.0 soon, and I'll tweak the docs or scripts a bit more when I work on Spark 4.0 support.

PR #524, opened 2025-06-09
Docs set tag in Spark to test

davidlghellin:
Hi
Update doc to set tag in Spark repository #523 

if it is not necessary to close the PR, thank you


linhr (approved):
Thank you for the change!

Issue #517, opened 2025-06-04
Optimize Parquet metadata access patterns in query execution


linhr:
I'm under the impression that the physical plan node for Parquet reader may still read the file metadata before reading the data. We may want to optimize this to reduce execution overhead.
1. Confirm if this is really the case.
2. Explore if it is possible to pass Parquet metadata from the plan resolver to the execution plan, so that we do not need to repeat metadata reading and parsing during physical execution.

Issue #516, opened 2025-06-04
Cache file listing results and Parquet metadata


linhr:
Currently, the plan resolver lists all files and fetches metadata for each individual Parquet file when planning each query, even if the dataset has already been registered as a temporary view. This adds overhead especially when the data is remote (e.g. in an object store such as AWS S3) and when the query involves multiple datasets with large number of partitioned files.

We may want to cache file listing results and Parquet metadata in the plan resolver. The downside is that there is no way to detect staleness of the cache. This is acceptable though, since we usually assume the files would not change. If the dataset does change after being overwritten, as a workaround, the user can restart the session.

PR #525, opened 2025-06-11
chore: update object_store to 0.12.2

shehabgamin:


linhr:
As per discussion with @shehabgamin, we wanted to follow up with this PR by integrating `DefaultObjectStoreRegistry` from the latest object_store version. I thought it was a direct replacement for `DashMap<ObjectStoreKey, Arc<dyn ObjectStore>>` that we currently have. However, after reading the implementation of `DefaultObjectStoreRegistry` more carefully, I found it uses `parse_url_opts` to create the `ObjectStore` instances for S3 etc. This is not compatible with our setup, where we have to use our own `LazyObjectStore` since inferring S3 region is async while the registry interface is not async.

The `PathEntry` idea in `DefaultObjectStoreRegistry` seems interesting, but I'm not sure if registry is still a bottleneck after recent optimization in #515. So instead of spending time borrowing the `PathEntry` idea into our codebase, I'd prefer to look into #516 and #517 first. And we may come up with some telemetry setup along the way, which can help us understand the registry overhead more easily.

Issue #225, opened 2024-09-29
Spark SQL Functions Coverage and Parity - CSV


shehabgamin:
- [x] from_csv
- [ ] schema_of_csv
- [ ] to_csv

rafafrdz:
@shehabgamin ey! Can I get this issue? 😄

shehabgamin:
> [@shehabgamin](https://github.com/shehabgamin) ey! Can I get this issue? 😄

@rafafrdz Done! Thank you so much!!

shehabgamin:
Hey @rafafrdz, just a quick note: Sail has been collaborating closely with DataFusion (cc @alamb, @andygrove) on a shared [datafusion-spark crate](https://github.com/apache/datafusion/tree/main/datafusion/spark), with a current focus on function support.

Would you be open to contributing your work to the `datafusion-spark` crate? My thought is to keep the implementation in both repositories for now, so Sail users can access new functions immediately without having to wait for the next DataFusion release. Once the new DataFusion version is out, we can remove the duplicate code from the Sail codebase.

Development might also be easier on the Sail side first, since we already have a comprehensive test suite for these functions and for testing Spark parity in general. The idea is that the function implementation and tests can then be copied over to DataFusion with minimal friction.

If it’s easier, I’m also happy to handle the upstream contribution myself. I just want to make sure you get full credit for your work. Let me know what you think!

rafafrdz:
Yes sure! I don't mind. In fact, I love the idea of collaborating with both in some way 😃

anarefolio:
Hi, I'd like to work on implementing to_csv, Can I get this issue?

PR #529, opened 2025-06-14
feat: add `from_csv` udf function

rafafrdz:
This PR introduces the implementation of the `from_csv` functionality, closely replicating Spark's `from_csv` capabilities. The function enables parsing CSV data into structured formats using a provided schema definition, leveraging Arrow’s efficient in-memory data representations.

Ref: [`from_csv` spark udf documentation](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.from_csv.html)

In this PR, we:

- [x] Add basic support for the `from_csv` Spark SQL function  
- [x] Add support for the `options` argument in the `from_csv` function  
- [x] Add test coverage 

Regarding the configurable options (as listed under [Data Source Option](https://spark.apache.org/docs/latest/sql-data-sources-csv.html#data-source-option) in the Spark documentation) for this UDF, the following are currently supported:

- [x] `sep` / `delimiter`  
- [x] `timestampFormat`
- [ ] Rest

Part of #225


rafafrdz:
Should we consider supporting the remaining `from_csv` options listed in the Spark documentation, or is the current subset sufficient for our needs?

rafafrdz:
BTW: I'm not very experienced with Rust, so any suggestions on best practices or improvements to the code style are more than welcome. I really appreciate any feedback that can help me write more idiomatic Rust 😄

crates/sail-spark-connect/tests/gold_data/function/csv.json
```diff
@@ -35,7 +35,127 @@
         }
       },
       "output": {
-        "failure": "not implemented: function: from_csv"
+        "success": "ok"
+      }
+    },
+    {
+      "input": {
+        "query": "SELECT from_csv('2015-08-26 00:00:00', 'time TIMESTAMP');",
```
linhr: Nice tests!

Could you move newly added test cases to `python/pysail/tests/spark/function/` and write those tests as Python doctests?

The JSON gold data files are autogenerated so manually added tests will be overridden if someone generates the test files from scratch.

You can see how this was done in a previous PR: #613.

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
```diff
@@ -0,0 +1,771 @@
...
+        // We cannot know the final DataType result without knowing the schema input args
+        Ok(DataType::Struct(Fields::empty()))
+    }
+
+    /// Determines the return type of the function based on the schema string and separator
+    fn return_field_from_args(&self, args: ReturnFieldArgs) -> Result<FieldRef> {
+        let ReturnFieldArgs {
+            scalar_arguments, ..
+        } = args;
+        let schema: &String = match scalar_arguments[1] {
```
linhr: `scalar_arguments[1]` will panic if the index is out of bound. So you may need to validate the length before this.

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
```diff
@@ -0,0 +1,771 @@
...
+            Some(ScalarValue::Utf8(Some(schema)))
+            | Some(ScalarValue::LargeUtf8(Some(schema)))
+            | Some(ScalarValue::Utf8View(Some(schema))) => Ok(schema),
+            _ => internal_err!("Expected UTF-8 schema string"),
+        }?;
+
+        let schema: Result<DataType> = parse_fields(schema).map(DataType::Struct);
```
linhr: If you use the `?` operator, then you don't need `.map()`.

```rust
let dt = DataType::Struct(parse_fields(schema)?);
Ok(Arc::new(Field::new(self.name(), dt, true)))
```

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
```diff
@@ -0,0 +1,771 @@
...
+    if args.len() < 2 || args.len() > 3 {
+        return exec_err!(
+            "`{}` function requires 2 or 3 arguments, got {}",
+            SparkFromCSV::FROM_CSV_NAME,
+            args.len()
+        );
+    };
+
+    let array: &StringArray = downcast_arg!(&args[0], StringArray);
+    let schema_str: &str = downcast_arg!(&args[1], StringArray).value(0);
```
linhr: `.value(0)` may panic if the array is empty.

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
linhr: (This is only a style recommendation and does not impact correctness. Usually all local variables with `Result<T, E>` types can be replaced by `T` types using the `?` operator, and the code is often more readable.)

crates/sail-spark-connect/tests/gold_data/function/csv.json
rafafrdz: Got it! I didn't know, thanks! I'll do asap

rafafrdz: done!

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
rafafrdz: done!

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
rafafrdz: done!

crates/sail-plan/src/extension/function/csv/spark_from_csv.rs
rafafrdz: done!

rafafrdz:
@linhr I have a question regarding the output timestamp type.

When running the following query:

```python
spark.sql("SELECT from_csv('2015-08-26 00:00:00', 'time TIMESTAMP') AS result, typeof(from_csv('2015-08-26 00:00:00', 'time TIMESTAMP')) AS type").show(truncate=False)
```

Using vanilla PySpark, the result is of type `timestamp`:

image

However, when running the same query on Sail server, the output type is `timestamp_ntz`:

image

I’m not sure whether this difference is due to a global Spark session configuration or something related to implicit type casting. Technically, I’m not explicitly modifying the data type in any way.

That said, I'm not sure if it is better to use `timestamp_ntz` as the expected result in the tests, or to stick with the default `timestamp` type from vanilla PySpark

linhr:
The data type is `timestamp_ntz` because you used `None` time zone in the following lines:
```rust
        SQLType::Timestamp(_, _) | SQLType::Datetime(_) | SQLType::Datetime64(_, _) => {
            Ok(DataType::Timestamp(TimeUnit::Nanosecond, None))
        }
```

You can change the time zone to `Some(Arc::from("UTC"))` to get the `timestamp` type. (Strictly speaking, we should use the session time zone here, but UTC is Ok for now.) If we later use our existing utility functions to parse data types (tracked in #631), we won't need to worry about this tricky detail.

linhr (approved):
This looks nice!

I'll leave the PR open, if you'd like to change the timestamp type. Just let me know!

rafafrdz:
Got it! Yep, that was on me. I’ll update it to make it as close to vanilla PySpark as possible 😄

linhr:
Thanks!

Could you remove the gold tests from the JSON files as well? I just realized that manually added tests may cause a validation error in future PRs when the gold data validation workflow is triggered.

linhr (approved):
Nice! Thanks again for your contribution!

PR #536, opened 2025-06-17
test: refactor Spark remote fixture

linhr:
In Spark 4.0, the PySpark process has different Java classpath configured depending on whether Spark Connect is used ([code](https://github.com/apache/spark/blob/v4.0.0/launcher/src/main/java/org/apache/spark/launcher/AbstractCommandBuilder.java)). This means we can either run the Python tests against Sail or JVM Spark, but not both. This PR updates the pytest fixture to accommodate this change.

Being able to run Python tests against JVM Spark also helped us discover tests that do not accurately reflect Spark behaviors, or tests not resilient to output ordering.

### DIFF
--- a/python/pysail/tests/spark/test_dml.py
+++ b/python/pysail/tests/spark/test_dml.py
@@ -26,35 +26,35 @@ def person_table(spark):
 
 
 def test_insert_single_value(spark):
-    spark.sql("INSERT INTO person VALUES (101, 'Foo', 99)")
+    spark.sql("INSERT INTO person VALUES (101, 'Shehab', 99)")
     actual = spark.sql("SELECT * FROM person WHERE id = 101").toPandas()
     expected = pd.DataFrame(
-        {"id": [101], "name": ["Foo"], "age": [99]},
+        {"id": [101], "name": ["Shehab"], "age": [99]},
     ).astype({"id": "int32", "name": "str", "age": "int32"})
     assert_frame_equal(actual, expected)
 
 
 def test_insert_multiple_values(spark):
-    spark.sql("INSERT INTO person VALUES (201, 'Foo', 99), (202, 'Bar', 2), (203, 'Baz', 9000)")
+    spark.sql("INSERT INTO person VALUES (201, 'Shehab', 99), (202, 'Alice', 2), (203, 'Bob', 9000)")
     actual = spark.sql("SELECT * FROM person WHERE id IN (201, 202, 203)").toPandas()
     expected = pd.DataFrame(
-        {"id": [201, 202, 203], "name": ["Foo", "Bar", "Baz"], "age": [99, 2, 9000]},
+        {"id": [201, 202, 203], "name": ["Shehab", "Alice", "Bob"], "age": [99, 2, 9000]},
     ).astype({"id": "int32", "name": "str", "age": "int32"})
     assert_frame_equal(actual, expected)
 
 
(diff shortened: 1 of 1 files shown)


python/pysail/tests/spark/conftest.py
```diff
@@ -22,30 +24,15 @@ def remote():
...
-    sail.stop()
-
-
-@pytest.fixture(scope="module")
-def spark():
-    os.environ["SPARK_LOCAL_IP"] = "127.0.0.1"
-    os.environ["SPARK_LOCAL_REMOTE"] = "true"
-    sc = SparkContext("local", "Spark")
-    spark = SparkSession(sc)
+def spark(remote):
```
linhr, line 27: Instead of having two fixtures `sail` and `spark`, we now have only one fixture `spark` that can be either the Sail server, or JVM Spark server.

python/pysail/tests/spark/test_dml.py
```diff
@@ -1,192 +1,171 @@
-import random
-
 import pandas as pd
 import pytest
 from pandas.testing import assert_frame_equal
 
+from pysail.tests.spark.utils import is_jvm_spark
+
+# TODO: We should revisit all DML tests once we support lakehouse formats.
```
linhr, line 7: The DML tests are skipped for now since there is more work involved. The tests would become useful when we work on lakehouse formats.

python/pysail/tests/spark/test_dml.py
```diff
@@ -1,192 +1,171 @@
...
-    sail.sql("DROP TABLE person")
+    spark.sql("DROP TABLE person")
 
 
-def test_insert_single_value(sail):
-    p_id = random.randint(0, 1000000)  # noqa: S311
-    sail.sql(f"INSERT INTO person VALUES ({p_id}, 'Shehab', 99)")  # noqa: S608
-    actual = sail.sql(f"SELECT * FROM person where id = {p_id}").toPandas()  # noqa: S608
+def test_insert_single_value(spark):
+    spark.sql("INSERT INTO person VALUES (101, 'Foo', 99)")
```
linhr: 1. We no longer use random data. Using fixed values improve readability of the tests. And the determinism actually let me find the issue that the warehouse data is not properly cleaned up after the test.
2. We avoid using real names in the test.

python/pysail/tests/spark/test_tpcds.py
```diff
@@ -15,32 +14,22 @@ def duck():
...
+@pytest.mark.skipif(is_jvm_spark(), reason="slow tests in JVM Spark")
+def test_derived_tpcds_query_execution(spark, query):
+    # TODO: add tests for result parity
     for sql in read_sql(query):
-        sail.sql(sql).toPandas()
-
-
-@pytest.mark.parametrize("query", [f"q{x + 1}" for x in range(99)])
-@pytest.mark.skip(reason="Spark data loading is not reliable")
-def test_derived_tpcds_query_spark_parity(sail, spark, query):
```
linhr, line 38: I removed the parity tests since they cannot be made to work with Spark 4.0, and right now they are skipped anyway.

I understand that the parity tests would be useful. We can either move those to the Spark test suites, or implement those tests as snapshot tests. This would be future work.

python/pysail/tests/spark/test_dml.py
shehabgamin: I like keeping a personal touch in the codebase 🙏

PR #541, opened 2025-06-20
chore: update dependabot config (npm & major releases)

pimlie:
To help with maintaining dependencies I'm suggesting to update the dependabot configuration as follows:

- Update dependabot config to also update (p)npm package
- For both cargo & npm, also update major releases but don't group those (i.e. let each major dependency update be it's own PR)
- Increase limit number of PR's to 3 in total (per manager, so currently 6 possible PR's in total)
- Prefix dependabot commits with `deps(cargo|npm)` to also use conventional commit message style

Regarding the change of pnpm version (to include it in package.json and remove it from the docs-build flow), afaik its required for dependabot to work correctly to have a packageManager entry in the package.json because otherwise it doesnt know which lock file format it should use (f.e. that changed from pnpm v8 to v9)

LMKWYT!

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -29,8 +29,6 @@ jobs:
         uses: pypa/hatch@install
 
       - uses: pnpm/action-setup@v4
-        with:
-          version: latest
 
       - name: Setup Node.js
         uses: actions/setup-node@v4


linhr (approved):
This looks great. Thanks for having this!

PR #545, opened 2025-06-20
docs: support development without sphinx pages

pimlie:
Currently you cannot run the docs server for development if you have not generated the sphinx pages yet.

But as generating the sphinx pages for the python api is not a trivial task, let's improve DX around working on the docs and add:
- a placeholder page for the Python API which should only be shown when the `SPHINX_BUILD_OUTPUT` folder does not exist
- a Vite plugin that shows a nice warning (during development) or explicit error (in prod) when the sphinx build output folder does not exists

Screenshot of placeholder page:
image


linhr (commented):
I'm wondering if it's possible to have all the logic in `[page].paths.ts` and `[page].md`. I'm worried that having validation in `config.mts` may make things out of sync in case there are future changes to the Python docs generation logic.

Let me know if the following works:
* Add a field `placeholder: boolean` to `SphinxPageData`.
* In `loadSphinxPages()`, return a single index page with `placeholder` set to `true` if it is dev environment and the output directory is missing. In production, throw an error if the output directory is missing.
* In `[page].md`, assuming we have taken care of propagating `placeholder` to `$params`, we can use `v-if` to render the placeholder content. (In this way, you don't need the workaround to exclude the placeholder `index.md` in production.)

pimlie:
Hmm, sure that's possible but I'm not sure I agree about that being a better solution. Wdyt about the implementation I've added now:

- most/all logic is centralised in `theme/utils/sphinx` and `theme/utils/sphinx-plugins`
- Use a VitePress plugin to control all logic around loading sphinx pages with hooks where needed (for f.e. the sidebar)
  - Con is maybe a bit more code
  - Pro is better separation of concerns & less code duplication (no more multiple loadSphinxPages calls)
  - Pro is also that its more configurable now. If we'd ever decide to have multiple python libs we can easily add their api's with minor changes

- Use a cache file to store the found sphinx pages for use in `[page].data.ts`
  - Con of this approach is that we have to use the filesystem, unfortunately it is not possible to use a virtual vite module here yet because `[page].data.ts` is called during Vite(press) setup not during the build itself
  - Pro is that we are sure that sphinx files are only loaded once (as opposed to twice before)
  - It's safe because the `.vitepress/cache` folder is never committed, so it should not be possible for an old cache file to exist in prod. But even if it exists, it should always be overwritten anyway because the cache is written in the `withSphinxPages` vitepress plugin which is called before any build logic starts. So the build will fail if errors occur during overwriting
  - Also this is basically just a nom- virtual way of caching virtual files which vite(press) uses internally
- The approach of using a `v-if` is possible, but Im not sure if vite/rollup are smart enough to treeshake the placeholder page out in production builds. Hence why I prefer to keep the `srcExclude`
- I did realize there was a better way to include the placeholder page, it's now included as a parent page for the `python/` folder with it's default slot being the placeholder text. This means that if that page ever gets included in prod builds, the default slot content would get replaced by the actual page content

docs/reference/python.md
```diff
@@ -0,0 +1,35 @@
+<slot>
+
+<!-- Even if this page would be included in a production build, it would not show
+anything as the pages under pyton/* would fill this slot content -->
```
shehabgamin:
```suggestion
anything as the pages under python/* would fill this slot content -->
```

Issue #475, opened 2025-05-24
PySpark DataFrame Doctest: sampleBy


shehabgamin:
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.sampleBy`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.sampleBy.html

davidlghellin:
Hi again
Once my current PR is merged, I’d love to take this one on as well. It looks quite similar and I think it would be a great opportunity to reinforce what I’ve learned so far.

Thanks

shehabgamin:
> Hi again
> Once my current PR is merged, I’d love to take this one on as well. It looks quite similar and I think it would be a great opportunity to reinforce what I’ve learned so far.
> 
> Thanks

@davidlghellin Thank you so much!! Thrilled to have your contribution!

PR #547, opened 2025-06-21
feat: sampleBy functions

davidlghellin:
This PR is for the #475

### DIFF
--- a/crates/sail-plan/src/resolver/plan.rs
+++ b/crates/sail-plan/src/resolver/plan.rs
@@ -36,11 +36,7 @@ use datafusion_expr::utils::{
     columnize_expr, conjunction, expand_qualified_wildcard, expand_wildcard, expr_as_column_expr,
     find_aggregate_exprs,
 };
-use datafusion_expr::{
-    build_join_schema, col, expr, ident, lit, when, Aggregate, AggregateUDF, BinaryExpr,
-    ExplainFormat, ExprSchemable, LogicalPlanBuilder, Operator, Projection, ScalarUDF, TryCast,
-    WindowFrame, WindowFunctionDefinition,
-};
+use datafusion_expr::{build_join_schema, col, expr, ident, lit, or, when, Aggregate, AggregateUDF, BinaryExpr, ExplainFormat, ExprSchemable, LogicalPlanBuilder, Operator, Projection, ScalarUDF, TryCast, WindowFrame, WindowFunctionDefinition};
 use rand::{rng, Rng};
 use sail_common::spec;
 use sail_common::spec::{Literal, TableFileFormat};
@@ -3837,7 +3833,6 @@ impl PlanResolver<'_> {
         seed: Option<i64>,
         state: &mut PlanResolverState,
     ) -> PlanResult<LogicalPlan> {
-        
         let colname: &str = match &column {
             spec::Expr::UnresolvedAttribute { name, .. } => {
                 let colname = name
@@ -3890,41 +3885,40 @@ impl PlanResolver<'_> {
             .build()?;
         ////
         let mut lower_bound: f64 = 0.0;
-        let mut sampled_plans: Vec<LogicalPlan> = vec![];
+        let mut disjuncts: Vec<Expr> = vec![];
 
         for frac in &fractions {
             let upper_bound: f64 = lower_bound + frac.fraction;
 
-            let key_value = match &frac.stratum {
-                Literal::Int32 { value: f } => ScalarValue::Int32(*f),
-                Literal::Float32 { value: f } => ScalarValue::Float32(*f),
-                Literal::Float64 { value: f } => ScalarValue::Float64(*f),
-                other => return Err(PlanError::unsupported(format!("Unsupported literal: {other:?}"))),
+            let key_value: ScalarValue = match &frac.stratum {
+                Literal::Int32 { value } => ScalarValue::Int32(*value),
+                Literal::Float32 { value } => ScalarValue::Float32(*value),
+                Literal::Float64 { value } => ScalarValue::Float64(*value),
+                other => {
+                    return Err(PlanError::unsupported(format!("Unsupported literal: {other:?}")));
+                }
             };
 
-            let registered_colname: String = state.register_field_name(colname);
-            // Construye plan con el filtro correspondiente
-            let filtered: LogicalPlan = LogicalPlanBuilder::from(plan_with_rand.clone())
-                .filter(col("#1").eq(lit(key_value)))?
-                .filter(col(&rand_column_name).gt_eq(lit(lower_bound)))?
-                .filter(col(&rand_column_name).lt(lit(upper_bound)))?
-                .build()?;
+            let conj: Vec<Expr> = vec![
+                col("#1").eq(lit(key_value)),
+                col(&rand_column_name).gt_eq(lit(lower_bound)),
+                col(&rand_column_name).lt(lit(upper_bound)),
+            ];
 
-            sampled_plans.push(filtered);
+            let and_expr: Expr = conjunction(conj).expect("should not be empty");
+            disjuncts.push(and_expr);
 
             lower_bound = upper_bound;
         }
 
-        let mut combined = sampled_plans.into_iter();
-        let mut plan = combined.next().ok_or_else(|| PlanError::invalid("No sampling filters"))?;
+        let final_expr: Expr = disjuncts
+            .into_iter()
+            .reduce(|acc, expr| or(acc, expr))
+            .ok_or_else(|| PlanError::invalid("No valid fractions"))?;
 
-        for next_plan in combined {
-            plan = LogicalPlan::Union(plan::Union {
-                inputs: vec![Arc::new(plan), Arc::new(next_plan)],
-                schema: plan_with_rand.schema().clone(),
-            });
-        }
-        /////
+        let plan: LogicalPlan = LogicalPlanBuilder::from(plan_with_rand)
+            .filter(final_expr)?
+            .build()?;
 
         Ok(LogicalPlanBuilder::from(plan)
             // .project(
--- a/crates/sail-plan/src/resolver/plan.rs
+++ b/crates/sail-plan/src/resolver/plan.rs
@@ -43,7 +43,7 @@ use datafusion_expr::{
 };
 use rand::{rng, Rng};
 use sail_common::spec;
-use sail_common::spec::TableFileFormat;
+use sail_common::spec::{Literal, TableFileFormat};
 use sail_common_datafusion::utils::{cast_record_batch, read_record_batches, rename_logical_plan};
 use sail_python_udf::cereal::pyspark_udf::PySparkUdfPayload;
 use sail_python_udf::get_udf_name;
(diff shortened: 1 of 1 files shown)


crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3830,116 @@ impl PlanResolver<'_> {
...
+        debug!("Value of col_id_for_plan: {:?}", col_id_for_plan);
+
+        let mut lower_bound: f64 = 0.0;
+        let mut disjuncts: Vec<Expr> = vec![];
+        for frac in &fractions {
+            let upper_bound: f64 = lower_bound + frac.fraction;
+            let key_value: ScalarValue = match &frac.stratum {
+                Literal::Int32 { value } => ScalarValue::Int32(*value),
+                Literal::Float32 { value } => ScalarValue::Float32(*value),
+                Literal::Float64 { value } => ScalarValue::Float64(*value),
```
davidlghellin: add more types

crates/sail-plan/src/resolver/state.rs
```diff
@@ -219,6 +219,25 @@ impl PlanResolverState {
...
+
+    pub fn find_plan_id_by_field_name(&self, plan_id: i64, name: &str) -> Option<NameID> {
+        self.fields
+            .iter()
+            .find(|(_, info)| info.name() == name && info.plan_ids().contains(&plan_id))
+            .map(|(field_id, _)| NameID {
+                id: field_id.clone(),
+            })
+    }
+}
```
davidlghellin: I'm not sure, I couldn't find any other way to do it.

crates/sail-plan/src/resolver/state.rs
```diff
@@ -219,6 +219,25 @@ impl PlanResolverState {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct NameID {
+    id: String,
+}
+
+impl NameID {
+    pub fn as_str(&self) -> &str {
+        &self.id
+    }
```
davidlghellin: Maybe this is unnecessary, but if we can type this type of data, since it is a special String #n , we would have to make a refactor with many changes.

davidlghellin:
if it is 0.3 the value its 0.30000000000000004
image

crates/sail-plan/src/resolver/plan.rs
davidlghellin: https://github.com/apache/spark/blob/4688cfda16bd937aba9b6a87bf5edd7c3ab66c62/python/pyspark/sql/dataframe.py#L2031

davidlghellin:
@claude help me to fix find_plan_id_by_field_name and use the correct use to get the field '#n' for not use column name and check my code

davidlghellin:
> if it is 0.3 the value its 0.30000000000000004 image

I am looking this, because sample when its 0.3 
image

davidlghellin:
in fractions={0: 0.1, 1: 0.9}
image
fraction receive 
image

maybe its problem with precision

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+        fractions: Vec<spec::Fraction>,
+        seed: Option<i64>,
+        state: &mut PlanResolverState,
+    ) -> PlanResult<LogicalPlan> {
+        if fractions
+            .iter()
+            .any(|f| f.fraction < 0.0 || f.fraction > 1.0)
+        {
+            return Err(PlanError::invalid(
+                "All fraction values must be >= 0.0 or > 1.0",
```
linhr:
```suggestion
                "All fraction values must be >= 0.0 and <= 1.0",
```

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+            _ => {
+                return Err(PlanError::invalid("Expected UnresolvedAttribute"));
+            }
+        };
+
+        let init_exprs: Vec<Expr> = input
+            .schema()
+            .columns()
+            .iter()
+            .map(|col| Expr::Column(col.clone()))
```
linhr: I guess we can have `.into_iter()` to avoid `.clone()` for `col`?

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+            .collect();
+        let rand_column_name: String = state.register_hidden_field_name("rand_value");
+        let seed: i64 = seed.unwrap_or_else(|| {
+            let mut rng = rng();
+            rng.random::<i64>()
+        });
+        let rand_expr: Expr = Expr::ScalarFunction(ScalarFunction {
+            func: Arc::new(ScalarUDF::from(Random::new())),
+            args: vec![Expr::Literal(ScalarValue::Int64(Some(seed)), None)],
+        })
```
linhr: The `Random` function accepts `None` for the seed, so probably we don't need to handle the case here.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+            .any(|f| f.fraction < 0.0 || f.fraction > 1.0)
+        {
+            return Err(PlanError::invalid(
+                "All fraction values must be >= 0.0 or > 1.0",
+            ));
+        }
+        let total_fraction: f64 = fractions.iter().map(|f| f.fraction).sum();
+        if total_fraction > 1.0 {
+            return Err(PlanError::invalid("All fraction sum is more than 1"));
+        }
```
linhr: The fraction sum does not need to be smaller than one since each stratum is sampled independently.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+        all_exprs.push(rand_expr);
+        let plan_with_rand: LogicalPlan = LogicalPlanBuilder::from(input)
+            .project(all_exprs)?
+            .build()?;
+
+        let mut lower_bound: f64 = 0.0;
+        let mut acc_exprs: Vec<Expr> = vec![];
+        for frac in &fractions {
+            let upper_bound: f64 = lower_bound + frac.fraction;
+            let key_value: ScalarValue = match &frac.stratum {
```
linhr: Maybe we can simply call `resolve_literal` here. I feel it's OK to support all data types even if Spark has certain restrictions on the key type.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3823,7 +3829,125 @@ impl PlanResolver<'_> {
...
+                    return Err(PlanError::unsupported(format!(
+                        "Unsupported literal: {other:?}"
+                    )));
+                }
+            };
+            let conj: Vec<Expr> = vec![
+                col(&column_expr.name).eq(lit(key_value)),
+                col(&rand_column_name).gt_eq(lit(lower_bound)),
+                col(&rand_column_name).lt(lit(upper_bound)),
+            ];
```
linhr: The conditions need to be changed since the fractions are used independently. For example, the following seems to work in Spark.

```python
from pyspark.sql.functions import col

dataset = spark.range(0, 100).select((col("id") % 3).alias("key"))
sampled = dataset.sampleBy("key", fractions={0: 0.1, 1: 1.0, 2: 0.6}, seed=0)
sampled.groupBy("key").count().orderBy("key").show()
```

linhr (commented):
Nice progress so far! I feel the sampling logic is not correct unfortunately. Would you mind taking a further look?

linhr:
> maybe its problem with precision

Not all decimal fractions can be represented precisely as floating point numbers. So `0.1 == 0.10000000000000001` evaluates to `True` if you try it in Python.

crates/sail-plan/src/resolver/plan.rs
davidlghellin: ok, thanks I understand better into_iter()

davidlghellin:
> Nice progress so far! I feel the sampling logic is not correct unfortunately. Would you mind taking a further look?
> 
> It may also be helpful to have a few tests. It would also be great to have the existing doctests passing, though I'm not sure if the seed can be interpreted in the same way in JVM Spark and Sail.

yes of course, I make the changes and check again, and check the tests well.

Thanks

davidlghellin:
in spark 4

[connect](https://github.com/apache/spark/blob/04a1750cc8809322b1601dd41d035e5e276d013e/sql/connect/common/src/main/scala/org/apache/spark/sql/connect/DataFrameStatFunctions.scala#L122)
```scala
  /** @inheritdoc */
  def sampleBy[T](col: Column, fractions: Map[T, Double], seed: Long): DataFrame = {
    require(
      fractions.values.forall(p => p >= 0.0 && p <= 1.0),
      s"Fractions must be in [0, 1], but got $fractions.")
    sparkSession.newDataFrame { builder =>
      val sampleByBuilder = builder.getSampleByBuilder
        .setInput(root)
        .setCol(toExpr(col))
        .setSeed(seed)
      fractions.foreach { case (k, v) =>
        sampleByBuilder.addFractions(
          StatSampleBy.Fraction
            .newBuilder()
            .setStratum(toLiteral(k).getLiteral)
            .setFraction(v))
      }
    }
  }
````
[classic](https://github.com/apache/spark/blob/branch-4.0/sql/core/src/main/scala/org/apache/spark/sql/classic/DataFrameStatFunctions.scala#L108)
```scala
   /** @inheritdoc */
  def sampleBy[T](col: Column, fractions: Map[T, Double], seed: Long): DataFrame = withOrigin {
    require(fractions.values.forall(p => p >= 0.0 && p <= 1.0),
      s"Fractions must be in [0, 1], but got $fractions.")
    import org.apache.spark.sql.functions.{rand, udf}
    val r = rand(seed)
    val f = udf { (stratum: Any, x: Double) =>
      x < fractions.getOrElse(stratum.asInstanceOf[T], 0.0)
    }
    df.filter(f(col, r))
  }

```

davidlghellin:
I'm trying to see if the way the seed is generated is very different from ours, and if that's why the result changes.

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3908,7 +3914,92 @@ impl PlanResolver<'_> {
...
+        &self,
+        input: spec::QueryPlan,
+        column: spec::Expr,
+        fractions: Vec<spec::Fraction>,
+        seed: Option<i64>,
+        state: &mut PlanResolverState,
+    ) -> PlanResult<LogicalPlan> {
+        if fractions
+            .iter()
+            .any(|f| f.fraction < 0.0 || f.fraction > 1.0)
```
davidlghellin, line 3834:
```suggestion
            .any(|f| f.fraction <= 0.0 || f.fraction >= 1.0)
```

davidlghellin:
> It would also be great to have the existing doctests passing, though I'm not sure if the seed can be interpreted in the same way in JVM Spark and Sail.

I haven't found anything concrete yet on how to solve the difference in seed behavior between Rust and the JVM.

linhr:
> > It would also be great to have the existing doctests passing, though I'm not sure if the seed can be interpreted in the same way in JVM Spark and Sail.
> 
> I haven't found anything concrete yet on how to solve the difference in seed behavior between Rust and the JVM.

Got it. I think it's fine if the seed behavior is different. Likely the random number generators are different between these two languages.

Maybe we can add unit tests in future PRs, and for this PR, we can merge the change as long as the changes are tested manually in your local environments.

davidlghellin:
https://github.com/lakehq/sail/pull/547/commits/0291b593c78db3d4b1d920db1cd3d47303eec67f
spark 3.5
image
sail
image

PR #548, opened 2025-06-22
docs: support development without Sphinx pages

linhr:
This is an alternative to #545.

The solution is based on [this comment](https://github.com/lakehq/sail/pull/545#pullrequestreview-2947533837), but @pimlie is correct that `v-if` does not treeshake well in the build, so I put the placeholder message in the data loader to prevent it from showing up in the production build.


linhr:
We’ll go with #545 so that we won’t lose track of the Sphinx data loading issue that @pimlie pointed out. With the foundation in that PR, we can keep iterating on it for other aspects of the setup in the future.

PR #550, opened 2025-06-23
chore: update docs-only workflow

linhr:
#546 introduced a nice feature to skip Python package build when the PR contains docs-only change. I just found that this results in build failure for existing open PRs since the "baseline" Python package in `main` is no longer available. I feel this error can be recovered by building Python package when the baseline is missing.

This PR also consolidates the build workflow and uses [dorny/paths-filter](https://github.com/dorny/paths-filter) for change list glob. This seems more flexible and easier to maintain than using the built-in `paths-ignore` condition.


.github/workflows/build-docs.yml
```diff
@@ -1,54 +0,0 @@
...
-
-      - uses: ./.github/actions/commit-baseline
-        id: baseline
-
-      - uses: ./.github/actions/commit-workflow
-        id: baseline-workflow
-        with:
-          sha: ${{ steps.baseline.outputs.sha }}
-          workflow_name: Build
-          artifact_name: python-package
```
linhr, line 41: This is moved to `build.yml`.

PR #564, opened 2025-06-27
chore: update Arrow to 55.2.0

linhr:
This PR updates Arrow to 55.2.0 which introduces the `arrow-pyarrow` crate.

With this update, we no longer need the `pyarrow` feature for DataFusion, and DataFusion will no longer be rebuilt when PyO3 detects Python environment changes. This reduces build time from > 1 minute to ~15 seconds on my laptop when switching Hatch environments.

See also: <https://github.com/apache/arrow-rs/issues/7668>


crates/sail-common-datafusion/src/display.rs
```diff
@@ -622,14 +622,45 @@ macro_rules! duration_display {
...
 duration_display!(
     DurationSecondType,
-    duration_s_to_duration,
+    duration_s_to_maybe_duration,
     DurationSecondFormatter,
 );
 duration_display!(
     DurationMillisecondType,
-    duration_ms_to_duration,
+    duration_ms_to_maybe_duration,
```
linhr, line 663: This is to accommodate the deprecation of `duration_s_to_duration` and `duration_ms_to_duration` methods.

linhr:
This should preferably be merged after the 0.3.0 release since we'd like to avoid last-minute changes.

PR #565, opened 2025-06-27
chore: tweak cluster timeout configs

shehabgamin:
Try to align configs with the closest matching Spark config.
https://spark.apache.org/docs/latest/configuration.html

### DIFF
--- a/crates/sail-common/src/config/application.yaml
+++ b/crates/sail-common/src/config/application.yaml
@@ -95,12 +95,12 @@
 
 - key: cluster.worker_max_idle_time_secs
   type: number
-  default: "120"
+  default: "60"
   description: The maximum idle time in seconds before a worker is removed.
 
 - key: cluster.worker_heartbeat_interval_secs
   type: number
-  default: "30"
+  default: "10"
   description: The interval in seconds for worker heartbeats.
   experimental: true
 
@@ -112,7 +112,7 @@
 
 - key: cluster.worker_launch_timeout_secs
   type: number
-  default: "300"
+  default: "120"
   description: The timeout in seconds for launching a worker.
 
 - key: cluster.worker_task_slots
@@ -128,7 +128,7 @@
 
 - key: cluster.task_launch_timeout_secs
   type: number
-  default: "300"
+  default: "120"
   description: The timeout in seconds for launching a task.
 
 - key: cluster.job_output_buffer
@@ -496,5 +496,5 @@
 
 - key: spark.session_timeout_secs
   type: number
-  default: "300"
+  default: "120"
   description: The Spark session timeout in seconds.


crates/sail-common/src/config/application.yaml
```diff
@@ -95,12 +95,12 @@
 
 - key: cluster.worker_max_idle_time_secs
   type: number
-  default: "120"
+  default: "60"
```
shehabgamin, line 98: Most closely aligns with?
```
key: spark.dynamicAllocation.executorIdleTimeout

default: 60s

If dynamic allocation is enabled and an executor has been idle for more than this duration,
the executor will be removed. For more detail, see this
[description](https://spark.apache.org/docs/latest/job-scheduling.html#resource-allocation-policy).
```

crates/sail-common/src/config/application.yaml
```diff
@@ -95,12 +95,12 @@
...
   description: The maximum idle time in seconds before a worker is removed.
 
 - key: cluster.worker_heartbeat_interval_secs
   type: number
-  default: "30"
+  default: "10"
```
shehabgamin, line 103: Most closely aligns with?
```
key: spark.executor.heartbeatInterval

default: 10s

Interval between each executor's heartbeats to the driver. Heartbeats let the driver know
that the executor is still alive and update it with metrics for in-progress tasks.
spark.executor.heartbeatInterval should be significantly less than spark.network.timeout 
```

crates/sail-common/src/config/application.yaml
```diff
@@ -112,7 +112,7 @@
 
 - key: cluster.worker_launch_timeout_secs
   type: number
-  default: "300"
+  default: "120"
```
shehabgamin, line 115: Most closely aligns with?
```
key: spark.network.timeout

default: 120s

Default timeout for all network interactions.
```

crates/sail-common/src/config/application.yaml
```diff
@@ -128,7 +128,7 @@
 
 - key: cluster.task_launch_timeout_secs
   type: number
-  default: "300"
+  default: "120"
```
shehabgamin, line 131: Most closely aligns with?
```
key: spark.network.timeout

default: 120s

Default timeout for all network interactions.
```

crates/sail-common/src/config/application.yaml
```diff
@@ -496,5 +496,5 @@
 
 - key: spark.session_timeout_secs
   type: number
-  default: "300"
+  default: "120"
```
shehabgamin, line 499: Most closely aligns with?
```
key: spark.network.timeout

default: 120s

Default timeout for all network interactions.
```

linhr: This is not related to network. This is related to how long the user can be idle for the session.

crates/sail-common/src/config/application.yaml
linhr: In cloud settings, there may be time needed to provision resources, and that's a separate factor from network.

linhr (approved):

Issue #171, opened 2024-09-08
[Epic] Full Delta Lake Integration


shehabgamin:
### Milestone 1: Core Table Operations
- [x] Basic Read (Snapshot Isolation)
- [x] Basic Write (Append & Overwrite modes)
- [x] Partitioned Writes
- [x] Data Skipping (Partition Pruning & File Statistics)
- [x] Schema Handling (Validation & Evolution)
- [x] Time Travel (By Version & Timestamp)
### Milestone 2: Core Log & File Types
- [x] JSON Commits
- [x] Parquet Checkpoint (V1)
- [x] V2 Checkpoint 
- [ ] Multi-part Checkpoints (Deprecated)
- [x] Log Compaction Files
- [x] Last Checkpoint File
- [x] Version Checksum File

### Milestone 3: DML Operations
- [x] DELETE (Copy-on-Write)
- [x] MERGE INTO (Copy-on-Write)
- [ ] UPDATE
- [ ] Merge-on-Read (Writing Deletion Vectors)

### Milestone 4: Table Management & Protocol
- [ ] Transaction Control (Conflict Detection & ACID)
- [x] Log Cleanup (Expired Metadata Removal)
- [x] Column Mapping (Id & Name modes)
- [ ] Constraints (Invariants & Check Constraints)
- [ ] VACUUM (Physical Data File Removal)
- [ ] OPTIMIZE (Bin-packing & Z-Ordering)
- [ ] RESTORE
- [ ] Change Data Feed
- [ ] Identity Columns & Generated Columns & Default Columns
- [ ] Variant Data Type
- [x] TimeStampNtz
- [x] In-Commit Timestamps
- [x] Append-Only Tables
- [ ] Delection Vectors
- [ ] Row Tracking
- [x] Type Widening
- [ ] Clustered Table
- [ ] VACUUM Protocol Check

### Milestone 6: Ecosystem & Streaming
- [x] Unity Catalog Integration
- [ ] AWS Glue / Hive Metastore Integration
- [ ] Catalog-managed Table 
- [ ] Iceberg Compat V1/V2
- [ ] Streaming Read (Tail Log)
- [ ] Streaming Write (Sink)

Other Milestones to be mentioned later.

## Related Work
- https://github.com/lakehq/sail/pull/578
- https://github.com/lakehq/sail/pull/634
- https://github.com/lakehq/sail/pull/677
- https://github.com/lakehq/sail/pull/680
- https://github.com/lakehq/sail/pull/717
[...]

PR #578, opened 2025-06-30
feat: basic read/write operations for Delta Lake

lonless9:
part of #171 

Delta table read/write operations in Spark SQL and DataFrame APIs.


Cargo.toml
```diff
@@ -136,6 +138,13 @@ serde_arrow = { version = "0.13.4", features = ["arrow-55"] }
 object_store = { version = "0.12.2", features = ["aws", "gcp", "azure", "http"] }
 hdfs-native-object-store = "0.14.2"
 
+# Lakehouse
+deltalake = { git = "https://github.com/lakehq/delta-rs.git", rev = "3607c31", default-features = false, features = ["rustls"] }
```
linhr: We could simply use the upstream repo since our fork does not have custom logic.

crates/sail-execution/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = { workspace = true }
 workspace = true
 
 [dependencies]
+deltalake = { workspace = true }
```
linhr: Is this needed?

crates/sail-delta-lake/src/delta_datafusion/schema_adapter.rs
```diff
@@ -0,0 +1,87 @@
+use std::fmt::Debug;
+use std::sync::Arc;
+
+use datafusion::arrow::array::RecordBatch;
+use datafusion::arrow::datatypes::{Schema, SchemaRef};
+use datafusion::common::DataFusionError;
+use datafusion::datasource::schema_adapter::{SchemaAdapter, SchemaAdapterFactory, SchemaMapper};
+use deltalake::kernel::schema::cast::cast_record_batch;
+
+/// A Schema Adapter Factory which provides casting record batches from parquet to meet
```
linhr, line 11: Let's add credit for this file as well.

crates/sail-delta-lake/src/operations/load.rs
```diff
@@ -0,0 +1,117 @@
...
+use deltalake::kernel::transaction::PROTOCOL;
+use deltalake::logstore::LogStoreRef;
+use deltalake::table::state::DeltaTableState;
+use deltalake::{DeltaTable, DeltaTableConfig};
+use futures::future::BoxFuture;
+
+use crate::delta_datafusion::{DataFusionMixins, DeltaScanConfig, DeltaTableProvider};
+
+#[derive(Debug, Clone)]
+pub struct LoadBuilder {
```
linhr: Is this `LoadBuilder` and `SailDeltaOps` used anywhere so far?

(This is just a question for my own understanding. No need to change anything.)

crates/sail-delta-lake/src/delta_datafusion/mod.rs
```diff
@@ -0,0 +1,1439 @@
...
+            InsertOp::Overwrite => deltalake::protocol::SaveMode::Overwrite,
+            InsertOp::Replace => deltalake::protocol::SaveMode::Overwrite, // Use Overwrite for Replace
+        };
+
+        // Execute the input plan to collect the data and count rows
+        let task_ctx = Arc::new(TaskContext::from(session_state));
+        let mut all_batches = Vec::new();
+        let mut total_rows = 0u64;
+
+        // Execute all partitions and collect the data
```
linhr: This is how the existing implementation collects all data in memory and writes all data using a single process. It would be much more scalable if the writer tasks are distributed and ingest data in a streaming fashion.

(This is just a note to explain the future work.)

crates/sail-delta-lake/src/delta_datafusion/mod.rs
```diff
@@ -0,0 +1,1439 @@
...
+    /// Define which files to consider while building a scan, for advanced usecases
+    #[allow(dead_code)]
+    pub fn with_files(mut self, files: Vec<Add>) -> DeltaTableProvider {
+        self.files = Some(files);
+        self
+    }
+}
+
+#[async_trait]
+impl TableProvider for DeltaTableProvider {
```
linhr, line 1222: `WriterBuilder` is used here in `insert_into()`, but we have similar writer logic in `DeltaDataSink` while we use `TableProvider` only for reading. Is my understanding correct?

python/pysail/tests/spark/test_deltalake.py
```diff
@@ -0,0 +1,335 @@
...
+    @pytest.fixture(scope="class")
+    def expected_pandas_df(self):
+        """Expected pandas DataFrame"""
+        return pd.DataFrame({"id": [10, 11, 12], "event": ["A", "B", "A"], "score": [0.98, 0.54, 0.76]}).astype(
+            {"id": "int32", "event": "string", "score": "float64"}
+        )
+
+    def test_delta_write_and_read_basic(self, spark, delta_test_data, expected_pandas_df):
+        """Test basic Delta Lake write and read operations"""
+        with tempfile.TemporaryDirectory() as tmpdir:
```
linhr: Could we use the pytest `tmp_path` fixture here?

python/pysail/tests/spark/test_deltalake.py
```diff
@@ -0,0 +1,335 @@
...
+import pytest
+from pandas.testing import assert_frame_equal
+from pyspark.sql.types import Row
+
+from pysail.tests.spark.utils import is_jvm_spark
+
+# Test constants
+YEAR_2025 = 2025
+YEAR_2026 = 2026
+EXPECTED_RESULT_COUNT = 2
```
linhr: I think it's Ok to not define constants here. Using the literal values directly in the tests could make them more readable.

linhr (approved):
This is amazing!! Great work!!! 🚀

crates/sail-delta-lake/src/operations/load.rs
lonless9: Nice Catch! These structures are all designed to conveniently expose table APIs externally or for writing Rust tests, which we currently don't need.

python/pysail/tests/spark/test_deltalake.py
lonless9: Hatch fmt complains about it, I'll fix this.

linhr: Oh I see. The Python linter can get annoying sometimes, especially in tests. You can simply bypass a certain rule for a particular line via `# noqa: <RULE>` comments (where `<RULE>` is the rule in violation).

PR #579, opened 2025-06-30
perf: improve data source listing

linhr:


crates/sail-object-store/src/layers/logging.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl fmt::Display for LoggingObjectStore {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "LoggingObjectStore({})", self.inner)
+    }
+}
+
+#[async_trait::async_trait]
+impl ObjectStore for LoggingObjectStore {
```
shehabgamin, line 32: Love how simple and clever this is!

PR #582, opened 2025-06-30
feat: string split to array by delim function

### LINKED ISSUE #572: Split function parity in case of just split by delimiter
The overall definition of `split` function in spark is really hard to implement:
https://spark.apache.org/docs/latest/api/sql/index.html#split

But one basic case of split function is super easy to implement
This is when we provide a simple string as delimiter and not regex
And just want to split all text to array of elements by delimiter

As a user, I think this is also the most used one. Never used any other cases of this function

This case can work for a long time while the whole solution is not done yet
Please let me know if it is OK, then I will do a pull request

```python
spark = SparkSession.builder.appName("split_example").getOrCreate()

one = spark.sql("SELECT split('a,b,c,d,e,', ',')").toArrow()
two = spark.sql("SELECT split('a12b12c12d1e12', '12')").toArrow()
print("one:", one)
print("two:", two)
```

```
one: pyarrow.Table
split(a,b,c,d,e,, ,, -1): list<element: string not null> not null
  child 0, element: string not null
----
split(a,b,c,d,e,, ,, -1): [[["a","b","c","d","e",""]]]

two: pyarrow.Table
split(a12b12c12d1e12, 12, -1): list<element: string not null> not null
  child 0, element: string not null
----
split(a12b12c12d1e12, 12, -1): [[["a","b","c","d1e",""]]]
```

SparkApplicationMaster:
Implementation of basic spark split function usecase:
Given str1 and str2, split str1 to array of strings using str2 as delimiter

Doesn't provide full feature parity with https://spark.apache.org/docs/latest/api/sql/index.html#split
And doesn't pass any doctests
But is still useful for splitting string by delim to array of strings, and this works like in vanilla spark

Usage:
```python

server = SparkConnectServer()
server.start()

one = spark.sql("SELECT split('a,b,c,d,e,', ',')").toArrow()
two = spark.sql("SELECT split('a12b12c12d1e12', '12')").toArrow()
print("one:", one)
print("two:", two)
```
```
one: pyarrow.Table
split(a,b,c,d,e,, ,): list<element: string>
  child 0, element: string
----
split(a,b,c,d,e,, ,): [[["a","b","c","d","e",""]]]
two: pyarrow.Table
split(a12b12c12d1e12, 12): list<element: string>
  child 0, element: string
----
split(a12b12c12d1e12, 12): [[["a","b","c","d1e",""]]]
```

Closes #572

### DIFF
--- a/crates/sail-plan/src/function/scalar/string.rs
+++ b/crates/sail-plan/src/function/scalar/string.rs
@@ -5,6 +5,7 @@ use datafusion::functions;
 use datafusion::functions::expr_fn;
 use datafusion::functions::regex::regexpcount::RegexpCountFunc;
 use datafusion::functions::string::contains::ContainsFunc;
+use datafusion::functions_array::string::string_to_array;
 use datafusion_common::ScalarValue;
 use datafusion_expr::{expr, lit, ExprSchemable, ScalarUDF};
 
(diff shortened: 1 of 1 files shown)


linhr (approved):
This is nice! Thanks!

Issue #309, opened 2024-11-22
 Spark SQL Functions Coverage and Parity - Aggregate


shehabgamin:
Gotta go through this huge list of implemented aggregate functions and check if any of them lack full parity:
https://github.com/lakehq/sail/blob/main/crates/sail-plan/src/function/aggregate.rs

While that gets done, here are the aggregate functions that haven't been implemented yet:
- [x] bitmap_construct_agg
- [x] bitmap_or_agg
- [x] collect_set (partial https://github.com/lakehq/sail/pull/585)
- [x] count_if
- [ ] count_min_sketch
- [ ] grouping_id
- [x] histogram_numeric
- [ ] hll_sketch_agg
- [ ] hll_union_agg
- [x] listagg
- [x] percentile
- [x] percentile_cont
- [x] percentile_disc
- [x] string_agg
- [x] try_avg
- [x] try_sum

keen85:
@shehabgamin 
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html#aggregate-functions
- `percentile_cont`
- `percentile_disc`
these are no valid functions in PySpark

linhr:
> * `percentile_cont`
> * `percentile_disc`

These are SQL functions defined here: <https://spark.apache.org/docs/latest/api/sql/index.html>

In PySpark these can be used via `SparkSession.sql()` but yeah these do not have corresponding PySpark APIs.

PR #585, opened 2025-07-01
feat: collect_set function

SparkApplicationMaster:
collect_set function implementation
https://spark.apache.org/docs/latest/api/sql/index.html#collect_set

What it does is applies three datafusion functions:
1) collect_list
2) array_distinct
3) remove null (because vanilla spark result is without null)

Passes doctest

Caveats:
1)  Spark does not sort the result string set by default string comparator, it applies some kind of ordering that cannot be understood
This implementation returns ordered set due to datafusion array_distinct implementation
But in maths, set is an unordered structure, so the result is not byte-to-byte equal but mathematically equivalent

This is more representative example than in doctest:
```python

spark = SparkSession.builder.appName("example").getOrCreate()
spark.sql("select array('11','3','2','3',null,'1') as a, 'b' as b")\
.withColumn("c", F.explode("a"))\
.createOrReplaceTempView("v")

spark.sql("select b, (collect_set(c)) as d from v group by b").toArrow()
```
```
# vanilla spark:
pyarrow.Table
b: string not null
d: list<element: string not null> not null
  child 0, element: string not null
----
b: [["b"]]
d: [[["1","3","2","11"]]]

# this implementation
pyarrow.Table
b: string not null
d: list<element: string>
  child 0, element: string
----
b: [["b"]]
d: [[["1","11","2","3"]]]
```

2) This collect_set version cannot work with Decimal because array_remove datafusion function doesn't work with Decimal
This case is very rare I suppose
Reproducible example:
```python

server = SparkConnectServer()
server.start()

spark.sql("select array(11.1,3.3,2.2,3.3,null,1.5) as a, 'b' as b")\
.withColumn("c", F.explode("a"))\
.createOrReplaceTempView("v")

spark.sql("select b, (collect_set(c)) as d from v group by b").toArrow()
```
```
AnalysisException: Failed to coerce arguments to satisfy a call to 'array_remove' function: coercion from [List(Field { name: "item", data_type: Decimal128(3, 1), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), Null] to the signature ArraySignature(Array { arguments: [Array, Element], array_coercion: Some(FixedSizedListToList) }) failed No function matches the given name and argument types 'array_remove(List(Field { name: "item", data_type: Decimal128(3, 1), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), Null)'. You might need to add explicit type casts.
	Candidate functions:
	array_remove(array, element)
```

part of #309


shehabgamin:
@SparkApplicationMaster Thanks, this makes sense to me! Future work can improve on the parity

PR #586, opened 2025-07-02
feat: count_if function

SparkApplicationMaster:
count_if function implementation
https://spark.apache.org/docs/latest/api/sql/index.html#count_if

```
> SELECT count_if(col % 2 = 0) FROM VALUES (NULL), (0), (1), (2), (3) AS tab(col);
 2
> SELECT count_if(col IS NULL) FROM VALUES (NULL), (0), (1), (2), (3) AS tab(col);
 1
```

passes doctests, no caveats found yet
part of #309


linhr (approved):
Looks nice! Thanks!

PR #587, opened 2025-07-02
refactor: data source options

linhr:
1. Use `build.rs` script to generate code for data source options.
2. Clean up data source option resolver logic so that reader/writer options can properly fallback to session defaults.
3. Remove code related to `CREATE TABLE ... AS SELECT ...` (CTAS) statements since this was never working. (See the various `FIXME` comments in the code.)


crates/sail-data-source/build.rs
```diff
@@ -0,0 +1,137 @@
...
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=build.rs");
+    build_options("CsvReadOptions", "csv_read")?;
+    build_options("CsvWriteOptions", "csv_write")?;
+    build_options("JsonReadOptions", "json_read")?;
+    build_options("JsonWriteOptions", "json_write")?;
+    build_options("ParquetReadOptions", "parquet_read")?;
+    build_options("ParquetWriteOptions", "parquet_write")?;
+    Ok(())
+}
```
linhr, line 137: Here is a sample of generated code.

```rust
#[derive(Debug, Clone, serde::Deserialize)]
#[serde(deny_unknown_fields)]
pub struct JsonReadOptions {
    #[serde(default)]
    #[serde(deserialize_with = "crate::options::serde::deserialize_usize")]
    #[serde(alias = "schemainfermaxrecords")]
    pub schema_infer_max_records: Option<usize>,
    #[serde(default)]
    #[serde(deserialize_with = "crate::options::serde::deserialize_string")]
    #[serde(alias = "codec")]
    pub compression: Option<String>,
}
impl crate::options::DataSourceOptions for JsonReadOptions {
    const ALLOWED_KEYS: &'static [&'static str] = &[
        "schema_infer_max_records",
        "schemainfermaxrecords",
        "compression",
        "codec",
    ];
    const DEFAULT_VALUES: &'static [(&'static str, &'static str)] = &[
        ("schema_infer_max_records", "1000"),
        ("compression", "UNCOMPRESSED"),
    ];
}
```

crates/sail-data-source/src/options/data/csv_read.yaml
```diff
@@ -93,17 +107,11 @@
     Parsing newlines in quoted values may be affected by execution behavior such as parallel file scanning.
     Setting this to `true` ensures that newlines in values are parsed successfully, which may reduce performance.
   supported: true
-
-- key: file_extension
```
shehabgamin, line 97: This is a useful configuration for specifying which files should be filtered.

crates/sail-data-source/src/options/data/parquet_read.yaml
```diff
@@ -1,96 +1,114 @@
...
   supported: true
+  rust_type: bool
+  rust_deserialize_with: crate::options::serde::deserialize_bool
 
 - key: coerce_int96
-  alias:
+  aliases:
     - coerceInt96
-  default: "us"
+  default: "ms"
```
shehabgamin: In `SessionManager` we have `"us"`:
https://github.com/lakehq/sail/blob/78f0651e77e2efe4d135bd94975fe3120cad6f93/crates/sail-spark-connect/src/session_manager.rs#L131

crates/sail-data-source/src/options/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+mod internal {
+    include!(concat!(env!("OUT_DIR"), "/options/csv_read.rs"));
+    include!(concat!(env!("OUT_DIR"), "/options/csv_write.rs"));
+    include!(concat!(env!("OUT_DIR"), "/options/json_read.rs"));
+    include!(concat!(env!("OUT_DIR"), "/options/json_write.rs"));
+    include!(concat!(env!("OUT_DIR"), "/options/parquet_read.rs"));
+    include!(concat!(env!("OUT_DIR"), "/options/parquet_write.rs"));
+}
+
+pub trait DataSourceOptions: for<'de> serde::Deserialize<'de> {
```
shehabgamin, line 19: Nice!

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -851,44 +847,44 @@ impl PlanResolver<'_> {
...
             }
             other => {
                 return Err(PlanError::unsupported(format!(
                     "unsupported data source format: {other:?}"
                 )))
             }
-        }
-        .with_session_config_options(&self.ctx.copied_config());
+        };
+        let options = options.with_session_config_options(&self.ctx.copied_config());
```
shehabgamin, line 887: I think we need to specify `with_file_extension` otherwise all files in the path will be selected?
https://github.com/apache/datafusion/blob/6870cc180f3dd72583919d01bf983fabb700434c/datafusion/core/src/datasource/listing/table.rs#L315-L317

crates/sail-plan/src/extension/logical/catalog.rs
```diff
@@ -78,6 +78,74 @@ pub(crate) enum CatalogTableFunction {
...
+    table_partition_cols: &'a Vec<String>,
+    file_sort_order: &'a Vec<Vec<Sort>>,
+    if_not_exists: bool,
+    or_replace: bool,
+    unbounded: bool,
+    options: &'a Vec<(String, String)>,
+    definition: &'a Option<String>,
+}
+
+impl<'a> From<&'a CatalogTableDefinition> for CatalogTableDefinitionOrd<'a> {
```
shehabgamin, line 113: Very nice!

crates/sail-plan/src/resolver/plan.rs
```diff
@@ -3160,95 +3127,23 @@ impl PlanResolver<'_> {
...
+                column_defaults,
+                constraints,
+                location,
+                file_format,
+                table_partition_cols,
+                file_sort_order,
+                if_not_exists,
+                or_replace,
+                unbounded,
+                options,
```
shehabgamin, line 3144: Please correct me if I’m misunderstanding the code, but I don’t think this approach will work as-is. There isn’t a one-to-one mapping between the Sail and DataFusion config options. For example, in Sail `meow` might correspond to DataFusion's `bark`. The naming and structure don’t align directly. Also, don't we need to prefix all of the configs with `format.`?

crates/sail-plan/src/resolver/option.rs
```diff
@@ -1,335 +1,378 @@
...
+    // TODO: support no comment
+    if let Some(Some(comment)) = comment {
+        to.comment = Some(char_to_u8(comment, "comment")?);
+    }
+    if let Some(header) = header {
+        to.has_header = Some(header);
+    }
+    // TODO: support default line separator
+    //   The default line separator is any of `\n`, `\r`, or `\r\n` in Spark,
+    //   but `\r\n` in DataFusion.
```
shehabgamin: In DataFusion, the default line separator is `CRLF` which I believe is equal to `\r, \r\n and \n`?
https://github.com/apache/datafusion/blob/6870cc180f3dd72583919d01bf983fabb700434c/datafusion/core/src/datasource/file_format/options.rs#L64-L65

crates/sail-data-source/src/options/data/parquet_read.yaml
linhr: Oh, nice catch!

crates/sail-plan/src/resolver/plan.rs
linhr: I removed that since by default the extension of `FileFormat` is used so there is no change to the behavior. There is also more work related to compressed files so I added a TODO in the code.

crates/sail-data-source/src/options/data/csv_read.yaml
linhr: I don't see this mentioned in other data source options so I removed this for consistency. I feel the user likely will get confused by this option since the usual way is to filter files using wildcard in the path.

crates/sail-plan/src/resolver/plan.rs
linhr: Yes this is intended. The Sail options is an abstraction layer and is not supposed to have one-to-one mapping with DataFusion options. And we do not have the `format.` prefix. This is still working because we set each individual config values directly. The core logic is in `crates/sail-plan/src/resolver/option.rs` but unfortunately this large diff is not rendered by default.

crates/sail-plan/src/resolver/plan.rs
shehabgamin: Ahh kk makes sense!

crates/sail-data-source/src/options/data/csv_read.yaml
shehabgamin: That's fair!

crates/sail-plan/src/resolver/option.rs
linhr: You're right. We don't need the TODO then.

<https://docs.rs/csv/latest/csv/enum.Terminator.html#variant.CRLF>

crates/sail-plan/src/resolver/plan.rs
linhr: Oh it seems you're referring to options in the `CREATE TABLE` statement. Good catch!

As per offline discussion, I'm going have an immediate follow-up PR to fix this issue.

PR #588, opened 2025-07-03
feat: support SQL `SHOW` statements

linhr:
I've tested the changes locally but we could add some tests in the future once we have better setup around catalog features.

### DIFF
--- a/crates/sail-plan/src/catalog/table.rs
+++ b/crates/sail-plan/src/catalog/table.rs
@@ -234,42 +234,6 @@ impl CatalogManager<'_> {
             .map(TableMetadata::from_table_object))
     }
 
-    async fn list_global_temporary_views(
-        &self,
-        pattern: Option<&str>,
-    ) -> Result<Vec<TableMetadata>> {
-        manage_temporary_views(self.ctx, true, |views| {
-            views.list_views(pattern).map(|views| {
-                views
-                    .into_iter()
-                    .map(|(name, plan)| {
-                        TableMetadata::from_table_object(TableObject::GlobalTemporaryView {
-                            database_name: self.config.global_temp_database.clone(),
-                            table_name: name,
-                            plan,
-                        })
-                    })
-                    .collect()
-            })
-        })
-    }
-
-    async fn list_temporary_views(&self, pattern: Option<&str>) -> Result<Vec<TableMetadata>> {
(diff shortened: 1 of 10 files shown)


crates/sail-plan/src/catalog/table.rs
```diff
@@ -234,42 +234,6 @@ impl CatalogManager<'_> {
             .map(TableMetadata::from_table_object))
     }
 
-    async fn list_global_temporary_views(
```
linhr, line 237: These two methods are moved to `view.rs`.

crates/sail-plan/src/catalog/view.rs
```diff
@@ -1,12 +1,77 @@
...
+    pub(crate) fn is_global_temporary_view_database(
+        &self,
+        database: &Option<SchemaReference>,
+    ) -> bool {
+        database.as_ref().is_some_and(|x| match x {
+            SchemaReference::Bare { schema } => schema.as_ref() == self.config.global_temp_database,
+            SchemaReference::Full { .. } => false,
+        })
+    }
+
```
lonless9: I wonder if we can put this function in utils.rs as table.rs using it.

It's also fine to keep it here.

lonless9:
NICE!

crates/sail-plan/src/catalog/view.rs
linhr: I just looked at `utils.rs` and it doesn't have methods for `impl CatalogManager<'_>` yet, so it seems `utils.rs` is for more general utilities. So maybe we keep it in `view.rs` since it's related to temporary views?

linhr: As per offline discussion, I'm going to move it to `mod.rs`.

crates/sail-spark-connect/src/session_manager.rs
```diff
@@ -98,7 +98,9 @@ impl SessionManager {
         let mut session_config = SessionConfig::new()
             .with_create_default_catalog_and_schema(true)
             .with_default_catalog_and_schema(DEFAULT_SPARK_CATALOG, DEFAULT_SPARK_SCHEMA)
-            .with_information_schema(true)
+            // We do not use the information schema since we use the catalog/schema/table providers
+            // directly for catalog operations.
+            .with_information_schema(false)
```
lonless9, line 103: Great

PR #590, opened 2025-07-04
chore: update deps

shehabgamin:


Cargo.toml
```diff
@@ -32,7 +30,7 @@ async-stream = "0.3.6"
 lazy_static = "1.5.0"
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.140"
-serde_yaml = "0.9.34"
```
shehabgamin, line 35: `serde_yaml` has been deprecated for over a year. `serde_yml` is a fork.

linhr: I've looked at this problem before. Let's keep `serde_yaml` since the fork does not have recent activities either, and the original version is more widely used and tested in the past. (It's still used in projects such as [kube-rs/kube](https://github.com/kube-rs/kube).)

I'd prefer to move to the fork when it receives wider community adoption.

linhr (approved):

PR #592, opened 2025-07-04
fix: resolve options for external table read

linhr:


crates/sail-data-source/src/factory.rs
```diff
@@ -0,0 +1,275 @@
...
+    ListingOptions, ListingTable, ListingTableConfig, ListingTableUrl,
+};
+use datafusion::prelude::SessionContext;
+use datafusion_common::{internal_err, plan_err, Result};
+use futures::{StreamExt, TryStreamExt};
+
+use crate::options::DataSourceOptionsResolver;
+use crate::url::{rewrite_directory_url, GlobUrl};
+
+pub struct TableProviderFactory<'a> {
```
linhr, line 22: The logic in this file is mostly moved from the `sail-plan` crate.

There is also a trait of the same name `TableProviderFactory` in DataFusion. We can choose a better name in the future to avoid confusion, after we get better ideas about the responsibility of this file.

crates/sail-plan/src/catalog/table.rs
```diff
@@ -115,67 +112,45 @@ impl TableMetadata {
...
+        if exist {
+            return match if_not_exists {
+                true => Ok(()),
+                false => exec_err!("table already exists: {table}"),
+            };
+        }
+        let factory = TableProviderFactory::new(self.ctx);
+        // TODO: This only registers the table for read and we need to support write as well.
+        let table_provider = factory
+            .read_table(
```
linhr, line 146: Here is the bug fix. We use our own factory to create tables to make sure we handle the options the way we want.

linhr:
As discussed previously, I'm also reverting #598 in this PR.

crates/sail-data-source/src/factory.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    pub async fn write_table(
+        &self,
+        source: &str,
+        options: Vec<(String, String)>,
+    ) -> Result<Arc<dyn FileFormatFactory>> {
+        let options: HashMap<String, String> = options.into_iter().collect();
+        let resolver = DataSourceOptionsResolver::new(self.ctx);
+        let format_factory: Arc<dyn FileFormatFactory> = match source {
```
lonless9: Please get lowercase string here.

linhr: Nice catch!

lonless9 (approved):

lonless9:
Nice!

Issue #505, opened 2025-05-28
Spark SQL Functions Coverage and Parity - Datetime


linhr:
- [x] add_years
- [x] add_months
- [x] add_days
- [x] convert_timezone
- [x] curdate
- [x] current_date
- [x] current_timestamp
- [x] current_timezone
- [x] date_add
- [x] date_diff
- [x] date_format
- [x] date_from_unix_date
- [x] date_part
- [x] date_sub
- [x] date_trunc
- [x] dateadd
- [x] datediff
- [x] datepart
- [x] day
- [x] dayofmonth
- [x] dayofweek
- [x] dayofyear
- [x] extract
- [x] from_unixtime
- [x] from_utc_timestamp
- [x] hour
- [x] last_day
- [x] localtimestamp
- [x] make_date
- [x] make_dt_interval
- [x] make_interval
- [x] make_timestamp
- [x] make_timestamp_ltz
- [x] make_timestamp_ntz
- [x] make_ym_interval
- [x] minute
- [x] month
- [x] months_between
- [x] monthname
- [x] next_day
- [x] now
- [x] quarter
- [x] second
- [ ] session_window
- [x] timestamp_micros
- [x] timestamp_millis
- [x] timestamp_seconds
- [x] to_date
- [x] to_timestamp
- [x] to_timestamp_ltz
- [x] to_timestamp_ntz
- [x] to_unix_timestamp
- [x] to_utc_timestamp
- [x] trunc
- [x] try_to_timestamp
- [x] unix_date
- [x] unix_micros
- [x] unix_millis
- [x] unix_seconds
- [x] unix_timestamp
- [x] weekday
- [x] weekofyear
- [x] width_bucket
- [ ] window
- [ ] window_time
- [x] year

radmirnovii:
For session_window: sessions merge across rows, so I'm choosing between (a) a window UDF that sessionizes each sorted partition in one pass, or (b) a Spark-style UpdatingSessionsExec custom operator. Any preferences?

shehabgamin:
> For session_window: sessions merge across rows, so I'm choosing between (a) a window UDF that sessionizes each sorted partition in one pass, or (b) a Spark-style UpdatingSessionsExec custom operator. Any preferences?

@radmirnovii A UDF seems like it would be a lot less code, although I'm not sure what the tradeoffs would be (besides the lack of optimization) without digging in. What are your thoughts?

Also cc @linhr

linhr:
[...]

PR #596, opened 2025-07-05
fix: from_utc_timestamp function

### LINKED ISSUE #595: from_utc_timestamp inconsistent behaviour
In vanilla spark, `from_utc_timestamp` does this:
treats input timestamp as UTC
adds input timezone offset
treats adjusted timestamp as if it is in user local timezone
converts it to UTC, but saves local timezone information

so spark does not store the input timezone in timestamp column, but adds the difference between timezones to it
this is more complicated than just treat timestamp as input timezone
and this affects toArrow(), show() and toPandas() results

for example, my timezone is +03:00
so toArrow() should return result adjusted to -03:00 in UTC
then toPandas and show both will be correct

the vector implementation is missing for now too, but this is known: https://github.com/lakehq/sail/blob/8977cc51694feabb84e649b6ac5eb1bc67278aae/crates/sail-plan/src/extension/function/datetime/spark_from_utc_timestamp.rs#L157

reproducible example:

```python

server = SparkConnectServer()
server.start()

def get_spark(what):
    if what == "vanilla":
        return SparkSession.builder.appName("vanilla_spark").getOrCreate()
    elif what == "sail":
        return SparkSession.builder.remote(f"sc://localhost:{port}").getOrCreate()

#spark = get_spark("vanilla")
spark = get_spark("sail")

def print_query_result(sql):
    print(sql.toArrow(), "\n")
    sql.show()
    print(sql.toPandas(), "\n")

spark.createDataFrame([[
    '2016-08-31 00:00:00', 'Asia/Seoul'], 
    ['2016-08-31 00:00:00', 'America/New_York'], 
    ['2016-08-31 00:00:00', 'UTC'], 
    ['2016-08-31 00:00:00', None], 
    [None, 'America/New_York'], 
    [None, None]
    ], 
    schema="ts string, tz string"
).createOrReplaceTempView("v")

print("scalar, scalar:")
print_query_result(spark.sql("select from_utc_timestamp('2016-08-31 00:00:00', 'Asia/Seoul') as a"))
print("vector, scalar:")
print_query_result(spark.sql("select from_utc_timestamp(ts, 'Asia/Seoul') as a from v"))
print("vector, vector:")
print_query_result(spark.sql("select from_utc_timestamp(ts, tz) as a from v"))
```

vanilla spark result:
```
scalar, scalar:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
+-------------------+

                    a
0 2016-08-31 09:00:00 

vector, scalar:                                                     
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z],[2016-08-31 06:00:00.000000Z],...,[null],[null]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|               NULL|
|               NULL|
+-------------------+

                    a
0 2016-08-31 09:00:00
1 2016-08-31 09:00:00
2 2016-08-31 09:00:00
3 2016-08-31 09:00:00
4                 NaT
5                 NaT 

vector, vector:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z],[2016-08-30 17:00:00.000000Z],...,[null],[null]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
|2016-08-30 20:00:00|
|2016-08-31 00:00:00|
|               NULL|
|               NULL|
|               NULL|
+-------------------+

                    a
0 2016-08-31 09:00:00
1 2016-08-30 20:00:00
2 2016-08-31 00:00:00
3                 NaT
4                 NaT
5                 NaT 

```

sail 0.3.0 result:
```
scalar, scalar:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 00:00:00.000000Z]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
+-------------------+

                    a
0 2016-08-31 03:00:00 

vector, scalar:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 00:00:00.000000Z,2016-08-31 00:00:00.000000Z,2016-08-31 00:00:00.000000Z,2016-08-31 00:00:00.000000Z,null,null]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|               NULL|
|               NULL|
+-------------------+

                    a
0 2016-08-31 03:00:00
1 2016-08-31 03:00:00
2 2016-08-31 03:00:00
3 2016-08-31 03:00:00
4                 NaT
5                 NaT 

vector, vector:
AnalysisException: Second argument for `from_utc_timestamp` must be string, received None
# this is not implemented yet
```

I plan to provide a solution to this soon

SparkApplicationMaster:
1) fix the inconsistent behaviour
2) now working with both scalar and array input arguments

example:
```python

server = SparkConnectServer()
server.start()

def get_spark(what):
    if what == "vanilla":
        return SparkSession.builder.appName("vanilla_spark").getOrCreate()
    elif what == "sail":
        return SparkSession.builder.remote(f"sc://localhost:{port}").getOrCreate()

#spark = get_spark("vanilla")
spark = get_spark("sail")

def print_query_result(sql):
    print(sql.toArrow(), "\n")
    sql.show()
    print(sql.toPandas(), "\n")

spark.createDataFrame([[
    '2016-08-31 00:00:00', 'Asia/Seoul'], 
    ['2016-08-31 00:00:00', 'America/New_York'], 
    ['2016-08-31 00:00:00', 'UTC'], 
    ['2016-08-31 00:00:00', None], 
    [None, 'America/New_York'], 
    [None, None]
    ], 
    schema="ts string, tz string"
).createOrReplaceTempView("v")

print("scalar, scalar:")
print_query_result(spark.sql("select from_utc_timestamp('2016-08-31 00:00:00', 'Asia/Seoul') as a"))
print("vector, scalar:")
print_query_result(spark.sql("select from_utc_timestamp(ts, 'Asia/Seoul') as a from v"))
print("vector, vector:")
print_query_result(spark.sql("select from_utc_timestamp(ts, tz) as a from v"))
```
output:
```
scalar, scalar:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
+-------------------+

                    a
0 2016-08-31 09:00:00 

vector, scalar:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z,2016-08-31 06:00:00.000000Z,2016-08-31 06:00:00.000000Z,2016-08-31 06:00:00.000000Z,null,null]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|2016-08-31 09:00:00|
|               NULL|
|               NULL|
+-------------------+

                    a
0 2016-08-31 09:00:00
1 2016-08-31 09:00:00
2 2016-08-31 09:00:00
3 2016-08-31 09:00:00
4                 NaT
5                 NaT 

vector, vector:
pyarrow.Table
a: timestamp[us, tz=UTC]
----
a: [[2016-08-31 06:00:00.000000Z,2016-08-30 17:00:00.000000Z,2016-08-30 21:00:00.000000Z,null,null,null]] 

+-------------------+
|                  a|
+-------------------+
|2016-08-31 09:00:00|
|2016-08-30 20:00:00|
|2016-08-31 00:00:00|
|               NULL|
|               NULL|
|               NULL|
+-------------------+

                    a
0 2016-08-31 09:00:00
1 2016-08-30 20:00:00
2 2016-08-31 00:00:00
3                 NaT
4                 NaT
5                 NaT 
```

closes #595
part of #505

### DIFF
--- a/crates/sail-plan/src/extension/function/datetime/spark_from_utc_timestamp.rs
+++ b/crates/sail-plan/src/extension/function/datetime/spark_from_utc_timestamp.rs
@@ -1,15 +1,17 @@
 use std::any::Any;
 use std::sync::Arc;
 
+use arrow::array::{Array, ArrayRef, Int64Array, PrimitiveArray};
+use arrow::datatypes::TimestampMicrosecondType;
+use chrono::{Local, TimeZone};
+use chrono_tz::Tz;
+use datafusion::arrow::array::AsArray;
 use datafusion::arrow::datatypes::{DataType, Field, FieldRef, TimeUnit};
 use datafusion_common::{exec_err, internal_err, Result, ScalarValue};
-use datafusion_expr::simplify::{ExprSimplifyResult, SimplifyInfo};
 use datafusion_expr::{
-    ColumnarValue, Expr, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Volatility,
+    ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Volatility,
 };
 use datafusion_expr_common::signature::{Signature, TypeSignature, TIMEZONE_WILDCARD};
-
-use crate::utils::ItemTaker;
 
 #[derive(Debug)]
 pub struct SparkFromUtcTimestamp {
@@ -154,15 +156,12 @@ impl ScalarUDFImpl for SparkFromUtcTimestamp {
                 args.arg_fields.len()
             );
         }
-        // FIXME: Second arg can be ColumnarValue::Array, but DataFusion doesn't support that.
-        match &args.scalar_arguments[1] {
-            Some(ScalarValue::Utf8(tz))
-            | Some(ScalarValue::Utf8View(tz))
-            | Some(ScalarValue::LargeUtf8(tz)) => Ok(Arc::new(Field::new(
+        match &args.arg_fields[1].data_type() {
+            DataType::Utf8 | DataType::Utf8View | DataType::LargeUtf8 => Ok(Arc::new(Field::new(
                 self.name(),
                 DataType::Timestamp(
                     *self.time_unit(),
-                    tz.as_ref().map(|tz| Arc::from(tz.to_string())),
+                    Some(Arc::from(Local::now().offset().to_string())),
                 ),
                 true,
             ))),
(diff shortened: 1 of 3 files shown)


crates/sail-plan/src/extension/function/datetime/spark_from_utc_timestamp.rs
```diff
@@ -172,35 +171,102 @@ impl ScalarUDFImpl for SparkFromUtcTimestamp {
...
-                });
-                Ok(ExprSimplifyResult::Simplified(expr))
+        let from_utc_timestamp_func = |inputs: (Option<i64>, Option<&str>)| match inputs {
+            (Some(ts_nanos), Some(tz_str)) => {
+                let to_zone: Tz = tz_str.parse().unwrap();
+                let to_dt = to_zone.timestamp_nanos(ts_nanos);
+                let result_ts = to_dt
+                    .naive_local()
+                    .and_local_timezone(*local_offset)
+                    .unwrap()
```
shehabgamin: Instead of `.unwrap()`, it would be great to use `.map_err` throughout the file.

SparkApplicationMaster: Thanks for the review!
mapped all the errors to DataFusionError
added better error messages for all cases

SparkApplicationMaster:
The error message in raw test results about this function not taking array argument has vanished
But the test is still failing, and I can't see the reason
Sorry I haven't managed to setup full local spark-tests yet, will try again
Please don't merge this MR yet
Will it be possible to rerun CI tests after I fix it locally, to check if the test doesn't fail?

SparkApplicationMaster:
found the problem
spark does not follow it's own doc

it says everywhere that the tz should be either offset or area/city:
https://github.com/apache/spark/blob/d1d29f9dd9ae8a9dcee3147e7276dd4cffdb541d/sql/api/src/main/scala/org/apache/spark/sql/functions.scala#L5850
https://github.com/apache/spark/blob/d1d29f9dd9ae8a9dcee3147e7276dd4cffdb541d/python/pyspark/sql/functions/builtin.py#L12216

but the examples use old three-letter time zone identifiers:
https://github.com/apache/spark/blob/d1d29f9dd9ae8a9dcee3147e7276dd4cffdb541d/python/pyspark/sql/functions/builtin.py#L12240 

which are deprecated long time ago:
https://errorprone.info/bugpattern/ThreeLetterTimeZoneID
https://bugs.openjdk.org/browse/JDK-8149811


this is the query that fails the test suite:
```python
df.select(from_utc_timestamp(df.ts, "PST").alias('local_time')).collect()
->
UNEXPECTED EXCEPTION: AnalysisException('[INVALID_TIMEZONE] The timezone: "PST" is invalid
```
I will implement some logic that matches this case and not wastes compute power if not nesessary

linhr:
@SparkApplicationMaster Thanks for helping with this!

You can also take a look at `python/pysail/tests/spark/literal/test_utc_timestamp.txt` for some tests I added in the past. You can remove `# doctest: +SKIP` there and run `pytest` to verify the bug fix. Also feel free to add more doctests there. This may make your testing easier since setting up full Spark tests locally is a bit involved.

Here is the doc for running `pytest`: <https://docs.lakesail.com/sail/main/development/build/python.html#running-tests>

SparkApplicationMaster:
Thanks!
I managed to set up the full tests, for my pc it was nesessary to override the .remote("local") variable:
https://github.com/lakehq/sail/blob/8977cc51694feabb84e649b6ac5eb1bc67278aae/scripts/spark-tests/plugins/spark.py#L72
hardcoded sail server host-port to this because vanilla spark just intercepted all the tests
with local tests it will be a lot better

SparkApplicationMaster:
I just checked locally for doctests in here: `python/pysail/tests/spark/literal/test_utc_timestamp.txt`
UPD: the first test is failing, gives 10:30 instead of 09:30, the second is working
The daylight saving time change moment is shifted in my implementation. 
it applies to the result date. needs to be applied to utc date
this is more complicated

but spark-test finally passed, so if you are satisfied with the code, now it can be merged

crates/sail-plan/src/extension/function/datetime/spark_from_utc_timestamp.rs
```diff
@@ -172,35 +174,168 @@ impl ScalarUDFImpl for SparkFromUtcTimestamp {
...
-    // TODO: Implement this method after the FIXME above is fixed so we can accept array input.
-    fn invoke_with_args(&self, _args: ScalarFunctionArgs) -> Result<ColumnarValue> {
-        internal_err!("`invoke` should not be called on a simplified `from_utc_timestamp` function")
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let args = &args.args;
+        let now = Local::now();
+        let local_offset = now.offset();
+        let local_offset_arc = Some(Arc::from(local_offset.to_string()));
+
+        let legacy_timezones = HashMap::from([
```
linhr, line 183: This is nice! I was puzzled by the three-letter time zones in the past as well. Glad to see this handled via such a mapping.

According to the [Java documentation](https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/time/ZoneId.html#SHORT_IDS), "EST", "MST", and "HST" do not include daylight savings, so this may need to be adjusted. Also I'm curious whether there are other datetime functions that would need this. So maybe the mapping logic can be moved to some utility functions. Of course, these could all be future work. I've created an issue to track this: <https://github.com/lakehq/sail/issues/600>.

linhr (approved):
Thanks for this fix!

I think we can merge this since it's a great improvement over the current broken state. The daylight savings issue is a bit tricky and can be looked at separately in the future.

linhr:
> I managed to set up the full tests, for my pc it was nesessary to override the .remote("local") variable:

In the test setup, a patched version of Spark is built to intercept session creation. I'd be curious to see why that didn't work in your setup. Feel free to create an issue and I'd be happy to look into this.

PR #597, opened 2025-07-06
chore: clippy deny unwrap, expect, and panic

shehabgamin:


crates/sail-execution/src/worker_manager/kubernetes.rs
```diff
@@ -44,6 +44,7 @@ impl KubernetesWorkerManager {
     }
 
     pub fn generate_name() -> String {
```
shehabgamin, line 46: Not sure what to do here.

crates/sail-plan/src/literal.rs
```diff
@@ -19,6 +19,7 @@ impl Default for LiteralEvaluator {
 impl LiteralEvaluator {
     pub fn new() -> Self {
```
shehabgamin, line 20: Not sure what to do here.

crates/sail-sql-analyzer/src/literal/datetime.rs
```diff
@@ -84,6 +85,7 @@ where
         .map(|s: &str| s.parse::<T>().unwrap())
```
shehabgamin, line 85: @linhr I spent some time trying to `map_err` here with Chumsky, but I wasn't sure how.

crates/sail-sql-analyzer/src/literal/datetime.rs
```diff
@@ -98,6 +100,7 @@ where
...
```
shehabgamin, line 100: @linhr I spent some time trying to `map_err` here with Chumsky, but I wasn't sure how.

crates/sail-sql-analyzer/src/literal/datetime.rs
```diff
@@ -98,6 +100,7 @@ where
...
 }
 
+#[allow(clippy::unwrap_used)]
 fn fraction<'a, T, E>(min_digits: usize, max_digits: usize) -> impl Parser<'a, &'a str, T, E>
```
shehabgamin, line 104: @linhr I spent some time trying to `map_err` here with Chumsky, but I wasn't sure how.

crates/sail-object-store/src/hugging_face.rs
```diff
@@ -20,10 +20,9 @@ use reqwest::StatusCode;
 use tonic::codegen::http;
 
 lazy_static! {
-    static ref HF_PATH_PATTERN: Regex = Regex::new(
+    static ref HF_PATH_PATTERN: Result<Regex, regex::Error> = Regex::new(
```
linhr: I don't think it makes much sense to store `Result` in static since we *should* panic when the process starts if the regex is invalid. I like your earlier `create_regex` idea and I feel we should stick to that pattern instead.

linhr (commented):
Regarding the few places you pointed out, I don't have better ideas either and it should be fine to use `#[allow(clippy::unwrap_used)]` in those places.

crates/sail-object-store/src/hugging_face.rs
shehabgamin: Done!

linhr (approved):

PR #606, opened 2025-07-07
refactor: clean `project` in sample, because use hidden fields

davidlghellin:


davidlghellin:
I'm closing this for now. I'll investigate more carefully why it failed locally — it's possible I was missing some generated artifacts or setup steps. Once I confirm the cause, I'll open a proper fix if needed.

Issue #574, opened 2025-06-30
Type casting parity for null functions


linhr:
This is a follow-up for #571. We may need to look into the Spark source code and replicate the type casting logic as UDFs.

It's also worth understanding whether this casting behavior is relevant to other scenarios as well.

jamesfricker:
missed the keyword arg in my pr, I believe this is closed by https://github.com/lakehq/sail/pull/613

PR #613, opened 2025-07-09
support `nullifzero` and `zeroifnull`

jamesfricker:
ref: https://github.com/lakehq/sail/issues/574


linhr:
Thanks for your contribution!

Would you mind moving your tests to `python/pysail/tests/spark/function/` and write those tests as Python doctests?

The JSON gold data files are autogenerated so manually added tests will be overridden if someone generates the test files from scratch.

linhr:
BTW I think those tests themselves look pretty good!

PR #617, opened 2025-07-12
fix: inconsistent size, array_size and cardinality logic

### LINKED ISSUE #614: cardinality, array_size, size functions inconsistent type and result
to reproduce:
```python

server = SparkConnectServer()
server.start()

df = spark.createDataFrame([([[2, 1], [3, 4]],)], ['data'])
df.select(sf.cardinality(df.data), sf.array_size(df.data), sf.size(df.data)).toArrow()
```
spark:
```
pyarrow.Table
cardinality(data): int32
array_size(data): int32
size(data): int32
----
cardinality(data): [[2]]
array_size(data): [[2]]
size(data): [[2]]
```

sail 0.3.1:
```
pyarrow.Table
cardinality(data): int64
array_size(data): int64
size(data): int64
----
cardinality(data): [[4]]
array_size(data): [[4]]
size(data): [[4]]
```

the Size function just returns the number of elements in outer collection
not going down to calculate and sum up all the sizes of child arrays:
https://github.com/apache/spark/blob/2571992aabb43064de5f7a70041c1aaa53b15c9c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L113
The return type is Int32, because spark is not able to contain arrays of total size more than 2^31 elements

all three functions  are using this Size function under the hood:
https://github.com/apache/spark/blob/2571992aabb43064de5f7a70041c1aaa53b15c9c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L726
https://github.com/apache/spark/blob/2571992aabb43064de5f7a70041c1aaa53b15c9c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L728
https://github.com/apache/spark/blob/2571992aabb43064de5f7a70041c1aaa53b15c9c/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L166

but there are two key differences in these functions:
`size` is just normal length of input array/map

`array_size` is `size` but works only on array_type

`cardinality` is `coalesce(size, -1) iIF the input value is NULL array/map AND some config flag is set: https://spark.apache.org/docs/latest/api/sql/index.html#cardinality

I'll take it

SparkApplicationMaster:
fix return type for input types:
For List | ListView | FixedSizeList | Map -> `Int32` (Spark-compartible)
For LargeList | LargeListView -> `Int64` (Beyond spark functionality)

```python

server = SparkConnectServer()
server.start()

df = spark.createDataFrame([([[2, 1], [3, 4]],)], ['data'])
df.select(sf.cardinality(df.data), sf.array_size(df.data), sf.size(df.data)).toArrow()
```
```
pyarrow.Table
cardinality(data): int32
array_size(data): int32
size(data): int32
----
cardinality(data): [[2]]
array_size(data): [[2]]
size(data): [[2]]
```

`array_size` for `Map` input now returns exec_error like in spark
also introduce parameter for supporting legacy mode, but this needs external spark configuration to be applied
https://spark.apache.org/docs/latest/api/sql/index.html#cardinality
the legacy configuration logic is not implemented for now

Closes #614

Also move existing code that calcs nested lengths sum to deep_size function
so it can be used like this:
```python
df = spark.createDataFrame([([[2, 1], [3, 4]],)], ['data']).createOrReplaceTempView("v")
spark.sql("select deep_size(data) from v").show()
```
```
+---------------+
|deep_size(data)|
+---------------+
|              4|
+---------------+
```

### DIFF
--- a/crates/sail-execution/src/codec.rs
+++ b/crates/sail-execution/src/codec.rs
@@ -48,6 +48,7 @@ use sail_plan::extension::function::array::spark_array_item_with_position::Array
 use sail_plan::extension::function::array::spark_array_min_max::{ArrayMax, ArrayMin};
 use sail_plan::extension::function::array::spark_map_to_array::MapToArray;
 use sail_plan::extension::function::array::spark_sequence::SparkSequence;
+use sail_plan::extension::function::collection::deep_size::DeepSize;
 use sail_plan::extension::function::collection::spark_concat::SparkConcat;
 use sail_plan::extension::function::collection::spark_reverse::SparkReverse;
 use sail_plan::extension::function::collection::spark_size::SparkSize;
@@ -763,8 +764,12 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "randn" => Ok(Arc::new(ScalarUDF::from(Randn::new()))),
             "random" | "rand" => Ok(Arc::new(ScalarUDF::from(Random::new()))),
             "spark_size" | "size" | "spark_cardinality" | "cardinality" => {
-                Ok(Arc::new(ScalarUDF::from(SparkSize::new())))
+                Ok(Arc::new(ScalarUDF::from(SparkSize::new(false, false))))
             }
+            "spark_array_size" | "array_size" => {
+                Ok(Arc::new(ScalarUDF::from(SparkSize::new(true, false))))
+            }
+            "deep_size" => Ok(Arc::new(ScalarUDF::from(DeepSize::new()))),
             "spark_array" | "spark_make_array" | "array" => {
                 Ok(Arc::new(ScalarUDF::from(SparkArray::new())))
             }
--- a/crates/sail-plan/src/extension/function/collection/spark_size.rs
+++ b/crates/sail-plan/src/extension/function/collection/spark_size.rs
@@ -2,19 +2,16 @@
 use std::any::Any;
 use std::sync::Arc;
 
-use datafusion::arrow::array::{Array, ArrayRef, GenericListArray, OffsetSizeTrait, UInt64Array};
+use arrow::array::{Int32Array, Int64Array};
+use datafusion::arrow::array::{Array, ArrayRef};
 use datafusion::arrow::datatypes::DataType;
 use datafusion_common::cast::{as_large_list_array, as_list_array, as_map_array};
-use datafusion_common::{exec_err, plan_datafusion_err, plan_err, Result};
+use datafusion_common::{exec_err, plan_err, Result, ScalarValue};
 use datafusion_expr::{
     ArrayFunctionSignature, ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature,
     TypeSignature, Volatility,
 };
 use datafusion_expr_common::signature::ArrayFunctionArgument;
-
-use crate::extension::function::functions_nested_utils::{
-    compute_array_dims, make_scalar_function,
-};
 
 // expr_fn::cardinality doesn't fully match expected behavior.
 // Spark's cardinality function seems to be the same as the size function.
@@ -23,16 +20,18 @@ use crate::extension::function::functions_nested_utils::{
 #[derive(Debug)]
 pub struct SparkSize {
     signature: Signature,
+    is_array_size: bool,
+    is_legacy_cardinality: bool,
 }
 
 impl Default for SparkSize {
     fn default() -> Self {
-        Self::new()
+        Self::new(false, false)
     }
 }
 
 impl SparkSize {
-    pub fn new() -> Self {
+    pub fn new(is_array_size: bool, is_legacy_cardinality: bool) -> Self {
         Self {
             signature: Signature::one_of(
                 vec![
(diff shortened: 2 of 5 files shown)


crates/sail-execution/src/codec.rs
```diff
@@ -763,8 +764,12 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "randn" => Ok(Arc::new(ScalarUDF::from(Randn::new()))),
             "random" | "rand" => Ok(Arc::new(ScalarUDF::from(Random::new()))),
             "spark_size" | "size" | "spark_cardinality" | "cardinality" => {
-                Ok(Arc::new(ScalarUDF::from(SparkSize::new())))
+                Ok(Arc::new(ScalarUDF::from(SparkSize::new(false, false))))
             }
+            "spark_array_size" | "array_size" => {
+                Ok(Arc::new(ScalarUDF::from(SparkSize::new(true, false))))
+            }
+            "deep_size" => Ok(Arc::new(ScalarUDF::from(DeepSize::new()))),
```
shehabgamin, line 773: Whenever there is a new UDF we need to add it to both `try_encode_udf` and `try_decode_udf`

crates/sail-execution/src/codec.rs
```diff
@@ -763,8 +764,12 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "randn" => Ok(Arc::new(ScalarUDF::from(Randn::new()))),
             "random" | "rand" => Ok(Arc::new(ScalarUDF::from(Random::new()))),
             "spark_size" | "size" | "spark_cardinality" | "cardinality" => {
-                Ok(Arc::new(ScalarUDF::from(SparkSize::new())))
+                Ok(Arc::new(ScalarUDF::from(SparkSize::new(false, false))))
```
shehabgamin: In `try_encode_udf`, we can extract the struct values (see, for example, `SparkWeekOfYear`) and store them in variables. Then, in `try_decode_udf`, we can use these variables instead of hardcoding the values.

SparkApplicationMaster: thanks!
this was a bit tricky to me, will try to implement the logic tomorrow

shehabgamin: Relevant file you will need to modify!
https://github.com/lakehq/sail/blob/main/crates/sail-execution/proto/sail/plan/physical.proto#L43

SparkApplicationMaster:
provided serialization
please check if this is correct, and the naming and all is OK to you

also there is some caveat:
DeepSize (previously SparkSize) returns UInt64

spark was written for JVM, so there is no UInt64, only Int64
so the problem is that spark cannot df.collect() this Uint64 and fails
toArrow, toPandas and show are casting to Int64, so they are working

it's not the scope of this MR, I suppose, because moved logic stays as is

shehabgamin:
> provided serialization please check if this is correct, and the naming and all is OK to you
> 
> also there is some caveat: DeepSize (previously SparkSize) returns UInt64
> 
> spark was written for JVM, so there is no UInt64, only Int64 so the problem is that spark cannot df.collect() this Uint64 and fails toArrow, toPandas and show are casting to Int64, so they are working
> 
> it's not the scope of this MR, I suppose, because moved logic stays as is

@SparkApplicationMaster Should we change it to `Int64` in the UDF?

Also no need to worry about naming in the `.proto` file. It's not user-facing, changes can be made with no impact.

crates/sail-execution/src/codec.rs
```diff
@@ -986,6 +993,11 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             let time_unit: gen_datafusion_common::TimeUnit = func.time_unit().into();
             let time_unit = time_unit.as_str_name().to_string();
             UdfKind::SparkFromUtcTimestamp(gen::SparkFromUtcTimestampUdf { time_unit })
+        } else if let Some(func) = node.inner().as_any().downcast_ref::<SparkSize>() {
```
shehabgamin, line 996: Nice!

crates/sail-plan/src/extension/function/collection/spark_size.rs
```diff
@@ -2,37 +2,36 @@
...
 // `cardinality`: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.cardinality.html
 // `size`: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.size.html
 #[derive(Debug)]
 pub struct SparkSize {
     signature: Signature,
+    is_array_size: bool,
+    is_legacy_cardinality: bool,
 }
 
 impl Default for SparkSize {
```
shehabgamin, line 27: nitpick: I think we can get rid of this entirely.

shehabgamin:
> LGTM! Let me know if you wanna adjust the return type of `DeepSize`. If not, then this is good to merge in.

@SparkApplicationMaster I went ahead and merged this in. Follow up work can be done if needed!

SparkApplicationMaster:
@shehabgamin sorry, was working hard this week and missed this, thanks for merging!

PR #619, opened 2025-07-12
fix: array_contains inconsistent null logic

### LINKED ISSUE #618: array_contains should return NULL on the NULL input
the logic is:
```
NULL -> NULL
[] -> False
some_array_that_doesn't_contain -> False
some_array_that_contains -> True
```
https://github.com/apache/spark/blob/6c6ca1f0b6f43a2034f4d6bcf1af1f07e6b7ccc7/python/pyspark/sql/functions/builtin.py#L18083

to reproduce:
```python
df = spark.createDataFrame([(None,), ([],), (["a", "b", "c"],)], ['data'])
df.select(sf.array_contains(df.data, "a")).show()
```
spark:
```
+-----------------------+
|array_contains(data, a)|
+-----------------------+
|                   NULL|
|                  false|
|                   true|
+-----------------------+
```
sail 0.3.1:
```
+-----------------------+
|array_contains(data, a)|
+-----------------------+
|                  false|
|                  false|
|                   true|
+-----------------------+
```

I'll take it

not sure if `array_contains_all` needs to implement the same logic or not
because it is not the part of spark api

SparkApplicationMaster:
implement NULL logic like in spark:

```
This function returns a boolean indicating whether the array
contains the given value, returning null if the array is null, true if the array
contains the given value, and false otherwise.
```
https://github.com/apache/spark/blob/6c6ca1f0b6f43a2034f4d6bcf1af1f07e6b7ccc7/python/pyspark/sql/functions/builtin.py#L18030

```python
df = spark.createDataFrame([(None,), ([],), (["a", "b", "c"],)], ['data'])
df.select(sf.array_contains(df.data, "a")).show()
```
```
+-----------------------+
|array_contains(data, a)|
+-----------------------+
|                   NULL|
|                  false|
|                   true|
+-----------------------+
```

passes spark-tests

closes #618

### DIFF
--- a/crates/sail-plan/src/function/scalar/array.rs
+++ b/crates/sail-plan/src/function/scalar/array.rs
@@ -1,5 +1,5 @@
 use datafusion::arrow::datatypes::DataType;
-use datafusion::functions::expr_fn::nvl;
+use datafusion::functions::expr_fn::{coalesce, nvl};
 use datafusion::functions_nested::expr_fn;
 use datafusion::functions_nested::position::array_position_udf;
 use datafusion_common::ScalarValue;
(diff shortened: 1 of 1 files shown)


shehabgamin:
LGTM!

PR #628, opened 2025-07-16
refactor: data source format

lonless9:


crates/sail-data-source/src/factory.rs
```diff
@@ -1,268 +0,0 @@
...
-        paths: Vec<String>,
-        schema: Option<Schema>,
-        options: Vec<(String, String)>,
-    ) -> Result<Arc<dyn TableProvider>> {
-        if paths.is_empty() {
-            return plan_err!("empty data source paths");
-        }
-        let urls = self.resolve_listing_urls(paths).await?;
-        // TODO: infer compression type from file extension
-        // TODO: support global configuration to ignore file extension (by setting it to empty)
```
linhr, line 43: Could you add back these TODOs somewhere in the code?

lonless9: Nice catch! Do you think it's reasonable to place this TODO near the ListingFormat trait?

linhr: Sounds good to me!

crates/sail-data-source/src/lib.rs
```diff
@@ -1,5 +1,7 @@
-mod factory;
+pub mod formats;
+pub mod listing;
 pub mod options;
-mod url;
+pub mod registry;
+pub mod url;
 
-pub use factory::TableProviderFactory;
+pub use registry::default_registry;
```
lonless9: I seem to have overlooked this part. Actually we only need to export `default_registry`

Issue #508, opened 2025-05-28
Spark SQL Functions Coverage and Parity - String


linhr:
- [x] ascii
- [x] base64
- [x] bit_length
- [x] btrim
- [x] char
- [x] char_length
- [x] character_length
- [x] chr
- [x] concat_ws
- [x] contains
- [x] decode
- [x] elt
- [x] encode
- [x] endswith
- [x] find_in_set
- [x] format_number (https://github.com/lakehq/sail/pull/1324)
- [x] format_string (https://github.com/lakehq/sail/pull/1347)
- [x] initcap
- [x] instr
- [x] lcase
- [x] left
- [x] len
- [x] length
- [x] levenshtein
- [x] locate
- [x] lower
- [x] lpad
- [x] ltrim
- [x] luhn_check
- [x] make_valid_utf8
- [x] mask
- [x] octet_length
- [x] overlay
- [x] position
- [x] printf
- [x] randstr (https://github.com/lakehq/sail/pull/1263)
- [x] regexp_count
- [x] regexp_extract
- [x] regexp_extract_all
- [x] regexp_instr
- [x] regexp_replace
- [x] regexp_substr (https://github.com/lakehq/sail/pull/1323)
- [x] repeat
- [x] replace
- [x] right
- [x] rpad
- [x] rtrim
- [x] sentences
- [x] soundex (https://github.com/lakehq/sail/pull/1263)
- [x] space
- [x] split
- [x] split_part
- [x] startswith
- [x] substr
- [x] substring
- [x] substring_index
- [x] to_binary
- [ ] to_char
- [x] to_number
- [ ] to_varchar
- [x] translate
- [x] trim
- [x] try_to_binary
- [x] try_to_number
- [x] ucase
- [x] unbase64
- [x] upper
- [x] strpos
- [x] try_validate_utf8
- [x] validate_utf8
- [x] is_valid_utf8 
- [x] make_valid_utf8

rafafrdz:
I'll get `to_number` function 😄

SparkApplicationMaster:
`like`, `ilike` and `luhn_check` functions are now ported from comet to datafusion
and will be available in one of the future releases:
https://github.com/apache/datafusion/blob/c62d67186fd034eca461304bedda5d2313ca62d3/datafusion/spark/src/function/string/mod.rs#L65

wudihero2:
The format_string function has been implemented in #1347.
Please let me know if you have any suggestions or feedback. Thanks!

[...]

PR #629, opened 2025-07-16
feat: add `to_number` function

rafafrdz:
This PR introduces the implementation of the `to_number` functionality, closely replicating Spark's `to_number` capabilities. The function enables safe conversion of string columns into numeric types by parsing formatted strings according to locale-aware number patterns, returning a decimal representation or null on invalid input.

Spark Ref: [`to_number` doc](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_number.html#pyspark-sql-functions-to-number)
Databricks Ref: [to_number` doc](https://docs.databricks.com/aws/en/sql/language-manual/functions/to_number#syntax)

In this PR, we:

- [x] Add support for the `to_number` Spark SQL function  
- [x] Add doc tests

part of #508

### DIFF
--- a/crates/sail-plan/src/extension/function/string/spark_to_number.rs
+++ b/crates/sail-plan/src/extension/function/string/spark_to_number.rs
@@ -167,7 +167,7 @@ macro_rules! get_capture_group {
 /// Parses a numeric value from a string using a specified format string.
 pub fn parse_number(value: &str, format: &Captures) -> Result<ParsedNumber> {
     // Getting the precision and scale
-    let (_, _, f_precision, f_scale) = extract_numbers_and_decimals(&format)?;
+    let (_, _, f_precision, f_scale) = extract_numbers_and_decimals(format)?;
 
     // Second we need to match the value
     // Matching the raw value pattern weather it's correct or not is not
@@ -175,7 +175,7 @@ pub fn parse_number(value: &str, format: &Captures) -> Result<ParsedNumber> {
     let factor: i8 = get_sign_factor(&value_captures);
 
     // Check if the numbers groupings match the format
-    match_grouping(&value_captures, &format)?;
+    match_grouping(&value_captures, format)?;
 
     // Getting the numbers, decimals, precision and scale from the value captures
     let (v_numbers, v_decimals, v_precision, v_scale): (String, Option<String>, u8, i8) =
@@ -250,11 +250,11 @@ impl PatternExpression {
         }
     }
     pub fn try_from(format: &Captures) -> Result<Self> {
-        let expr: PatternExpression = handle_number(&format)?;
-        let expr: PatternExpression = handle_decimal(&format)?.prepend(expr)?;
-        let expr: PatternExpression = handle_currency(&format, &expr)?;
-        let expr: PatternExpression = handle_sign(&format, &expr)?;
-        handle_brackets(&format, &expr)
+        let expr: PatternExpression = handle_number(format)?;
+        let expr: PatternExpression = handle_decimal(format)?.prepend(expr)?;
+        let expr: PatternExpression = handle_currency(format, &expr)?;
+        let expr: PatternExpression = handle_sign(format, &expr)?;
+        handle_brackets(format, &expr)
     }
 }
 


rafafrdz:
The Apache Spark implementation for this function is pure madness:

- [to_number](https://github.com/apache/spark/blob/e888e37ee2eb3c9e970476ad1a254e661bcd3d21/sql/api/src/main/scala/org/apache/spark/sql/functions.scala#L4739)
-  [ToNumberParser](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ToNumberParser.scala)

To simulate the same behavior in a structured and deterministic way, I use an expression-constructor approach that dynamically builds regex patterns based on the specification defined in the [Databricks `to_number` doc](https://docs.databricks.com/aws/en/sql/language-manual/functions/to_number#syntax). So here is an example:

**Example**

Let’s say we use the format `S$999G099.99` for the value `-$12,345G67` (which conforms to the format specification):

- This results in the following expression: `Group([Sign(false), Group([Currency("$"), Group([Number, Dot("."), Decimal])])])`
- Its corresponding regex pattern becomes: `^(?<sign>[+-])?(?<currency>[$])(?<numbers>[0-9,G]+)(?<dot>[G])?(?<decimals>[0-9]+)?$`
- And now, we check if the value matches that regex, in this case, yes:
  -  `^(?<sign>[+-])?` : `-`
  - `(?<currency>[$])`: `$`
  - `(?<numbers>[0-9,G]+)`: `12,345`
  - `(?<dot>[G])?`: `G`
  - `(?<decimals>[0-9]+)?`: `67`

lonless9:
Thanks for your contribution, it looks great!

You can check how to handle `crates/sail-execution/src/codec.rs` in #529

crates/sail-spark-connect/tests/gold_data/function/string.json
```diff
@@ -3713,7 +3713,7 @@
         }
       },
       "output": {
-        "failure": "not implemented: function: to_number"
+        "failure": "error in DataFusion: Execution error: Execution error: String '12,454.8-' does not match the expected regex pattern ^(?<sign>[+-])?(?<numbers>[0-9,G]+)(?<dot>[.])?(?<decimals>[0-9]+)?$"
```
lonless9, line 3716: Is this by design? Or could it be considered as a follow-up task?

rafafrdz:
> Thanks for your contribution, it looks great!
> 
> You can check how to handle `crates/sail-execution/src/codec.rs` in #529

Sorry, I forgot to push it! :sweat_smile:

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -0,0 +1,653 @@
...
+/// Features:
+/// - Accurate extraction of number structure based on format.
+/// - Support for locale-aware grouping and currency parsing.
+/// - Flexible placement of optional signs and symbols (prefix/suffix).
+/// - Output conforms to a DECIMAL type with precision and scale inferred from `fmt`.
+///
+/// The implementation uses Arrow arrays and DataFusion error handling conventions.
+/// Errors are returned when `expr` includes invalid characters, or when `expr`
+/// does not conform to the structure defined by `fmt`.
+///
```
lonless9: Could we make the comments like this in this entire file shorter or remove them to clearly show the function's behavior?

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -0,0 +1,653 @@
...
+
+/// Modifies the `PatternExpression` to include a sign based on captured sign information.
+///
+/// # Parameters
+/// - `captures`: A `Captures` object that contains the regex capture groups.
+/// - `expr`: The `PatternExpression` potentially modified to append sign information.
+///
+/// # Returns
+/// A `PatternExpression` with sign processing, or an error if signs are imbalanced.
+fn handle_sign(captures: &Captures, expr: &PatternExpression) -> Result<PatternExpression> {
```
lonless9, line 303: Could we find a way to distinguish between prefix and suffix symbols here? 
Some related lines may also need to add logic related to suffixes.

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -0,0 +1,653 @@
...
+    let all_character_same = [',', 'G']
+        .iter()
+        .any(|c| format_positions.iter().all(|(_, d)| *d == *c));
+    let _ = if all_character_same {
+        Ok(true)
+    } else {
+        exec_err!(
+            "Malformed integer format related groupings: {format}. Use only ',' or 'G', not both"
+        )
+    };
```
lonless9: I'm not sure about the intention of dropping the return values in similar places. Perhaps could we properly set the number of return values, or adopt the following approach? 

`if !condition { return Err(...) }`

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -0,0 +1,653 @@
...
+    }
+    let values: &StringArray = downcast_arg!(&args[0], StringArray);
+    let format: &str = downcast_arg!(&args[1], StringArray).value(0);
+    let (precision, scale) = get_precision_and_scale(values, format);
+
+    let scalars: Result<Vec<ScalarValue>> = values
+        .iter()
+        .map(|value| match value {
+            None => Ok(ScalarValue::Decimal256(None, precision, scale)),
+            Some(value) => ParsedNumber::try_from(value, format)
```
lonless9: During each iteration, the process involves passing through `try_from` to `match_regex` and executing `Regex::new`. Can we complete the related logic outside the loop and reuse the regex?

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -0,0 +1,653 @@
...
+/// error if the group is not found.
+macro_rules! get_capture_group {
+    ($captures:expr, $group_name:expr) => {
+        get_opt_capture_group!($captures, $group_name).ok_or_else(|| {
+            exec_datafusion_err!("Missing '{}' group in the format regex", $group_name)
+        })
+    };
+}
+
+pub fn get_precision_and_scale(values: &StringArray, format: &str) -> (u8, i8) {
```
lonless9: I'm just curious, no need to change. Could we get this from other places like format?

lonless9:
If no other changes are expected, I think just reducing the size of the comments would be good. The rest can be tracked by opening another issue!

python/pysail/tests/spark/function/test_to_number.txt
```diff
@@ -0,0 +1,35 @@
+>>> spark.sql("SELECT to_number('<-$12,345.67>', 'S$999,099.99PR') AS result, typeof(to_number('<-$12,345.67>', 'S$999,099.99PR')) AS type").show(truncate=False)
```
lonless9, line 1: Great, these tests are exactly what we need!

crates/sail-spark-connect/tests/gold_data/function/string.json
rafafrdz: I think it could be considered as a follow-up task. It changed when I tested the code; it wasn't manual.

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: sure! sorry my mistake 😄

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: I think that approach is better, sorry, I came from a functional programming mindset and I'm not properly used to coding Rust yet hahaha

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Sorry, I’m not sure I follow. Could you give an example or a situation where distinguishing between prefix and suffix would make a difference? Maybe I’m missing something 😅

For instance (from my point of view), it doesn't matter to me whether `PR` or `L` is allocated as a prefix or suffix, I just capture them and add the corresponding regex expression in the appropriate position.

What I mean is,
- If I found a `PR` (sign spec) on either the left side or right side, I would always set the `(?<sign>[+-])` expression (sign regex expression) at the beginning.
- If I found an `L` (currency spec) on either the left side or right side, I will set the `(?<currency>[{currency}])` expression (currency regex expression) on either the left side or right side (depends on the spec expr position).

On the other hand, the expression-building logic follows the order (from inner to outer spec expressions), so it always places currency first, then sign:  `(?<sign_left>MI|S)?(?<currency_left>L|\$)? ... (?<currency_right>L|\$)?(?<sign_right>PR|MI|S)?$`

Not sure if this helps clarify things or if I’m totally off track here hahaha 😅

lonless9: Sorry, my previous comment here could not be reasonable, maybe it should have been placed elsewhere. I was referring to test case like the failure in the gold test that is relevant here, but there are other places involved as well . We can keep it as a follow-up work.

crates/sail-spark-connect/tests/gold_data/function/string.json
lonless9: Sure! Besides, the gold data is of course not manually generated. You can use the following command to update and verify it locally. 

```cargo +nightly fmt && \
  cargo clippy --all-targets --all-features && \
  cargo build && \
  env SAIL_UPDATE_GOLD_DATA=1 cargo test```

crates/sail-plan/src/extension/function/string/spark_to_number.rs
lonless9: No worries, pleasantly surprised by your deep understanding!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
lonless9: This is just a code style and doesn't have much impact!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Sure! In fact, it must be gotten from the format, not the values. Done!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Done! I think it's already better now. I moved the Regex type (Capture) to the top of the function (before the loop) and reused it inside the loop, updating the signatures of the other functions accordingly.

Actually, I hadn’t realized it before, but what I really needed all along was the Captures, not the raw format string. Thank you!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: As you wish, I can define more gold tests if you want, or keep it as follow-up work

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Never mind, I prefer to do it hahah. Done!

crates/sail-spark-connect/tests/gold_data/function/string.json
rafafrdz: done!

lonless9: Yeah, when you use this command, you can see the effect of the gold test, and none of the text within this dir should be modified manually.

crates/sail-plan/src/extension/function/string/spark_to_number.rs
lonless9: The gold tests are not manually added; they are the content that the [spark tests] needs to pass.

For manually adding tests, we can use either pytest or doc test approach. As you've used in this PR with the txt file, it is a doc test.

lonless9 (commented):
Great job! Once it passes the clippy check, I'll merge this PR and open an issue to track the follow-up tasks.

PR #632, opened 2025-07-17
feat: catalog manager and provider redesign

linhr:
1. Add a `CatalogProvider` trait that supports multi-level database names and async catalog operations.
2. Update `CatalogManager` to manage catalogs via a `HashMap` internally, instead of relying on DataFusion's catalog/schema providers. `CatalogManager` is now an extension to the DataFusion `SessionContext`.
3. Update application configuration to support defining a list of catalogs for the session. The default catalog is now `sail` instead of `spark_catalog`.
4. Update spec for various write/insert operations and implement a unified resolver method for all these operations. This also fixes a known issue that `saveAsTable` was incorrectly implemented as an in-memory view.
5. Define our own logical node and physical planning logic for the file write operation. We no longer need to work with `FileFormatFactory`, which makes the `TableFormat` trait (introduced in #628) more intuitive, and makes the Delta Lake integration simpler.


crates/sail-common-datafusion/src/datasource.rs
```diff
@@ -3,36 +3,75 @@ use std::sync::Arc;
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct BucketBy {
+    pub columns: Vec<String>,
+    pub num_buckets: usize,
+}
 
 /// Information required to create a data source.
-pub struct SourceInfo<'a> {
-    pub ctx: &'a SessionContext,
```
linhr, line 14: `ctx` is moved out of `SourceInfo` and `SinkInfo` so that we can derive `Debug` for these structs.

crates/sail-common-datafusion/src/datasource.rs
```diff
@@ -3,36 +3,75 @@ use std::sync::Arc;
...
+    ) -> Result<Arc<dyn TableProvider>>;
 
-    /// Creates a `FileFormatFactory` for the format.
-    fn create_writer(&self, info: SinkInfo<'_>) -> Result<Arc<dyn FileFormatFactory>>;
+    /// Creates a `ExecutionPlan` for write.
+    async fn create_writer(
+        &self,
+        ctx: &dyn Session,
+        info: SinkInfo,
+    ) -> Result<Arc<dyn ExecutionPlan>>;
```
linhr, line 81: Since we now have our own logical plan node for the write operation, we can return an `ExecutionPlan` directly, instead of returning a `FileFormatFactory` required by the `CopyTo` struct in DataFusion.

`DeltaFormatFactory` and `DeltaFileFormat` are no longer necessary. The Delta Lake format now integrates directly with the `TableFormat` trait. The `FileFormat` trait is for listing tables, and is no longer related to Delta Lake. This is more intuitive since Delta Lake is not a "file format" and does not need to work with listing tables.

crates/sail-plan/src/resolver/command/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+use crate::error::{PlanError, PlanResult};
+use crate::resolver::state::PlanResolverState;
+use crate::resolver::PlanResolver;
+
+mod catalog;
+mod insert;
+mod write;
+
+impl PlanResolver<'_> {
+    pub(super) async fn resolve_command_plan(
```
linhr, line 15: We start moving resolver methods out of the giant `plan.rs` file.

crates/sail-common/src/spec/plan.rs
```diff
@@ -384,13 +391,24 @@ pub enum CommandNode {
...
-        kind: Option<ViewKind>,
+        if_exists: bool,
+    },
+    CreateTemporaryView {
+        view: Identifier,
+        is_global: bool,
+        #[serde(flatten)]
+        definition: TemporaryViewDefinition,
+    },
+    DropTemporaryView {
```
linhr, line 405: We consolidated the spec for (global) temporary views and views when we first supported (global) temporary views in #91. But now it turns out that it's better to define spec for (global) temporary views and views separately since their definitions are very different. For (global) temporary views, we keep track of query plans in memory. For views, we need a SQL query string that can be stored in remote catalogs.

crates/sail-catalog/src/provider/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+pub use status::*;
+
+use crate::error::CatalogResult;
+
+/// A trait that defines the interface for a catalog.
+/// A catalog contains *databases*, where each database has a multi-level name
+/// that represents a *namespace*.
+/// A database contains *objects* such as *tables* and *views*.
+#[async_trait::async_trait]
+pub trait CatalogProvider: Send + Sync {
```
lonless9, line 16: Is this where we will implement the `alter_table` logic in the future?

crates/sail-catalog-memory/src/provider.rs
```diff
@@ -0,0 +1,378 @@
...
+    status: DatabaseStatus,
+    tables: HashMap<String, TableStatus>,
+    views: HashMap<String, TableStatus>,
+}
+
+/// An in-memory catalog provider.
+pub struct MemoryCatalogProvider {
+    name: String,
+    databases: Arc<Mutex<HashMap<Namespace, MemoryDatabase>>>,
+}
```
lonless9: Could we push the lock down or replace the related Hashmap with Dashmap? Maybe RwLock could help as well.

crates/sail-catalog-memory/src/provider.rs
```diff
@@ -0,0 +1,378 @@
...
+    async fn drop_table(
+        &self,
+        database: &Namespace,
+        table: &str,
+        options: DropTableOptions,
+    ) -> CatalogResult<()> {
+        let DropTableOptions {
+            if_exists,
+            purge: _,
+        } = options;
```
lonless9, line 251: I'm curious if the `MemoryCatalogProvider` is responsible for triggering data clean-up. Does it means that it's sufficient to just remove it from the catalog, without triggering the actual deletion? Or is this a TODO?

lonless9 (commented):
Very exciting PR! I haven't gone through all the files yet today, but I think I've reviewed the parts that concern me and left some comments. I probably won't have any further adjustments to make when I look at the remaining parts tomorrow!

crates/sail-catalog/src/provider/mod.rs
linhr: Yes I think so!

crates/sail-catalog-memory/src/provider.rs
linhr: Oh good idea! Let me use `DashMap` here since we only need to manage a map as the state for `MemoryCatalogProvider`.

crates/sail-catalog-memory/src/provider.rs
linhr: Oh yeah I think the catalog provider is supposed to clean up data. Let me add a TODO here for `purge`.

lonless9 (approved):

crates/sail-catalog-memory/src/provider.rs
linhr: I integrated `DashMap` in [82911d2](https://github.com/lakehq/sail/pull/632/commits/82911d2a6b15ac31afb933adcc294aa87091cf65). This allows for fine-grained locking for each namespace. I probably wouldn't go further to separate locks for tables and views separately within each namespace, since I feel it's a bit hard to reason about, and may cause atomicity bugs if we're not careful enough when the logic becomes more complex in the future.

PR #633, opened 2025-07-17
fix: codec for `from_csv`

davidlghellin:
When I merged the main branch into mine earlier, I didn’t see this line and it caught my attention. After reviewing it again later, I confirmed it wasn’t there before.
If this line isn't necessary, feel free to close the PR. Thanks!


rafafrdz:
Ey @davidlghellin, thanks for the fix. I didn't know I had to do that. Just to know, why/what is that?

davidlghellin:
> Ey @davidlghellin, thanks for the fix. I didn't know I had to do that. Just to know, why/what is that?

Not 100% sure, but I think it’s needed so that when the plan is deserialized on the workers, we can still detect that the node is a SparkFromCSV. Otherwise, it wouldn’t be recognized during analysis or execution, maybe @linhr can explain it better.

PR #634, opened 2025-07-18
feat: partitioned writes and data skipping for Delta Lake

lonless9:
part of #171


crates/sail-delta-lake/src/kernel/log_data.rs
```diff
@@ -0,0 +1,446 @@
...
+        is_min: bool,
+    ) -> Precision<ScalarValue> {
+        let mut accumulator: Option<Box<dyn Accumulator>> = None;
+        let mut mask_offset = 0;
+
+        for batch in &self.data {
+            let num_rows = batch.num_rows();
+            let mut path_iter = path_step.split('.');
+            let array = if let Ok(array) =
+                ex::extract_column(batch, path_iter.next().unwrap(), &mut path_iter)
```
linhr: It seems that `sail-delta-lake` does not use the workspace linter configuration, otherwise `.unwrap()` would probably trigger a linter error.

We could add the following to `Cargo.toml`.

```toml
[lints]
workspace = true
```

For this file, we could add `#[allow(clippy::unwrap_used)]` if it's too much work to rewrite the log data handler logic.

crates/sail-delta-lake/src/operations/write/writer.rs
```diff
@@ -505,3 +483,97 @@ fn arrow_schema_without_partitions(
             .collect::<Vec<_>>(),
     ))
 }
+
+/// Partition a RecordBatch along partition columns
+pub(crate) fn divide_by_partition_values(
```
linhr, line 489: We could add the credit for `crates/core/src/writer/record_batch.rs` as well.

crates/sail-delta-lake/src/delta_datafusion/mod.rs
```diff
@@ -920,102 +961,67 @@ fn prune_file_statistics(
 }
 
 pub(crate) fn get_null_of_arrow_type(t: &ArrowDataType) -> DeltaResult<ScalarValue> {
```
linhr: There is a `ScalarValue::try_new_null()` method in DataFusion. It's interesting to see a similar function being maintained in `delta-rs`.

python/pysail/tests/spark/test_deltalake.py
```diff
@@ -233,6 +321,507 @@ def test_delta_with_different_data_types(self, spark, tmp_path):
...
+
+        # Test mixed type conditions
+        filtered_df = (
+            spark.read.format("delta").load(delta_table_path).filter("int_col IN (100, 300) AND bool_col = true")
+        )
+        result_count = filtered_df.count()
+        assert result_count == 2, f"Expected 2 rows for mixed type condition, got {result_count}"  # noqa: PLR2004
+
+    def test_delta_partition_pruning_performance_validation(self, spark, tmp_path):
+        """Test partition pruning effectiveness with performance validation"""
```
linhr: I feel this test still validates correctness. I'm not sure if performance issues can be surfaced by this test. (Let me know if I missed anything, or if I misunderstood the intention of this test.)

We could probably look at `.explain()` and assert if the filters are pushed down. We could have doctests for the physical plan, though the plan information (e.g. attribute IDs) may change over time, so it's a bit inconvenient to maintain. This could be future work.

python/pysail/tests/spark/test_deltalake.py
```diff
@@ -233,6 +321,507 @@ def test_delta_with_different_data_types(self, spark, tmp_path):
...
+        expected_count = 2 * 10  # 2 categories * 10 active records each
+        assert result_count == expected_count, f"Expected {expected_count} rows, got {result_count}"
+
+        filtered_df = spark.read.format("delta").load(delta_table_path).filter("score = 7")
+        result_count = filtered_df.count()
+        expected_count = 2 * 3 * 2  # 2 years * 3 categories * 2 records with score = 7
+        assert result_count == expected_count, f"Expected {expected_count} rows, got {result_count}"
+
+    def test_delta_partition_pruning_optimization_verification(self, spark, tmp_path):
+        """Test to verify partition pruning optimization is working by comparing filtered vs unfiltered scans"""
```
linhr: I'm not sure how this compares filtered vs unfiltered scans. (Again, let me know if I missed anything. :) )

lonless9: I will handle the relevant content clearly. Currently, push down is enabled by default.

python/pysail/tests/spark/test_deltalake.py
lonless9: While fixing the issue with null values, I added some `explain` statements, then removed them all during the final merge . I will add validation for `explain` in the doc tests.

crates/sail-delta-lake/src/delta_datafusion/mod.rs
lonless9: See https://github.com/delta-io/delta-rs/commit/89742b2a 
But I'm not sure if `ScalarValue::try_new_null()` will work, I will take a try.

lonless9: I used `ScalarValue::try_new_null()`, which should not cause issues.

lonless9:
@linhr  I think we need to find an appropriate way to handle the validation for `explain`, so is it reasonable to consider this PR completed at this stage?

linhr:
> @linhr I think we need to find an appropriate way to handle the validation for `explain`, so is it reasonable to consider this PR completed at this stage?

Yes, `explain` could be future work. Let me take a quick look at the recent commit and then it should be ready to be merged!

linhr (approved):

Issue #220, opened 2024-09-29
Spark SQL Functions Coverage and Parity - Array


shehabgamin:
- [x] array 
- [x] array_append
- [x] array_compact
- [x] array_contains
- [x] array_contains_all
- [x] array_distinct
- [x] array_except
- [x] array_insert
- [x] array_intersect
- [ ] array_join (Partial parity)
- [x] array_max
- [x] array_min
- [x] array_position
- [x] array_prepend
- [x] array_remove
- [x] array_repeat
- [x] array_union
- [x] arrays_overlap
- [x] arrays_zip
- [x] flatten
- [x] get
- [ ] sequence (Partial parity)
- [ ] shuffle
- [x] slice
- [x] sort_array

keen85:
@shehabgamin, @SparkApplicationMaster 
I was not able to find anything about `array_contains_all` as [PySpark function](https://spark.apache.org/docs/latest/api/python/search.html?q=array_contains_all#) nor as [Spark SQL function](https://spark.apache.org/docs/latest/api/sql/search.html?q=array_contains_all).
So, is this some proprietary function of `Sail`?

shehabgamin:
@keen85 You're right, not sure where I got that function from when creating this issue. Other engines seem to have this function but Spark doesn't.

PR #638, opened 2025-07-20
feat: add array_insert function

SparkApplicationMaster:
Add `array_insert` function implementation:
https://spark.apache.org/docs/latest/api/sql/index.html#array_insert
using only datafusion expressions

passes spark-tests
please tell if there is duplicate expression elimination in Sail, because the plan creates a lot of them

raises exception if position is zero, like spark does
but the exception message is not informative in this case
because cannot override datafusion runtime exceptions at the plan level
so I just found some datafusion method that is always raising exception and call it in case of 0

part of #220

### DIFF
--- a/crates/sail-plan/src/function/common.rs
+++ b/crates/sail-plan/src/function/common.rs
@@ -332,3 +332,19 @@ pub(crate) fn get_null_treatment(ignore_nulls: Option<bool>) -> Option<NullTreat
         None => None,
     }
 }
+
+#[macro_export]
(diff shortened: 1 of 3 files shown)


SparkApplicationMaster:
found that there is already a helper for when_then:
https://github.com/apache/datafusion/blob/386985788a35e474b829d46c9dccdfd7c5117d98/datafusion/expr/src/expr_fn.rs#L364
will rewrite to this

SparkApplicationMaster:
maybe `raise_error` function with more descriptive comment can help with zero position case?
just found that function, it sounds like what is needed

PR #643, opened 2025-07-20
fix: arrays_overlap function behaviour on inputs containing NULL

SparkApplicationMaster:
the logic is not consistent with https://spark.apache.org/docs/latest/api/sql/index.html#arrays_overlap:
arrays_overlap(a1, a2) - Returns true if a1 contains at least a non-null element present also in a2.
If the arrays have no common element and they are both non-empty and either of them contains a null element null is returned, false otherwise.

the meaning is that NULL value is undefined
so if NULL exists in one of arrays, the result is undefined too - maybe arrays overlap, maybe not, so NULL is returned

```python
server = SparkConnectServer()
server.start()

df = spark.createDataFrame([
    (["a", None], ["b", "c", None]),
    (["a", "b"], ["b", "c"]), 
    (["a"], ["b", "c"]), 
    ], ['x', 'y']
)
df.select(sf.arrays_overlap(df.x, df.y)).show()
```

```
+--------------------+
|arrays_overlap(x, y)|
+--------------------+
|                NULL|
|                true|
|               false|
+--------------------+
```

### DIFF
--- a/crates/sail-plan/src/function/scalar/array.rs
+++ b/crates/sail-plan/src/function/scalar/array.rs
@@ -3,7 +3,8 @@ use datafusion::functions::expr_fn::{coalesce, nvl};
 use datafusion::functions_nested::expr_fn;
 use datafusion::functions_nested::position::array_position as datafusion_array_position;
 use datafusion_common::ScalarValue;
-use datafusion_expr::{expr, lit, BinaryExpr, Operator};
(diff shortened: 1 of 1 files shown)


linhr (approved):
This looks nice!

Could you resolve the merge conflict? Otherwise I think this is ready to go!

PR #647, opened 2025-07-21
fix: concat function behaviour on some subset of input types

### LINKED ISSUE #646: concat function should be able to take binary strings and other types
in spark, concat behaviour is like this:
1) strings -> string
2) binary strings -> binary string
3) primitive types -> cast to strings and concat as string
4) arrays of the same type -> array
5) arrays of different type -> array of coerced type

1 and 4 are already in the code
2 and 3 can be solved relatively easy
5 is worth it's own issue I suppose

also if any of the arguments is NULL, concat must return NULL, now it just eats NULL and concats other

SparkApplicationMaster:
fix concat behaviour for various types:
binary -> binary (casts to strings and uses string concat but casts result to binary)
primitive (coerce to string) -> string 
string and primitive -> string
any NULL input -> NULL output

passes spark-tests
closes #646 (if it's ok to create another issue about passing input arrays with different but coercible types)

### DIFF
--- a/crates/sail-plan/src/extension/function/collection/spark_concat.rs
+++ b/crates/sail-plan/src/extension/function/collection/spark_concat.rs
@@ -1,6 +1,9 @@
 use std::any::Any;
 use std::cmp::Ordering;
 
+use arrow::array::BooleanArray;
+use arrow::compute::kernels::boolean::{is_null, or};
+use arrow::compute::kernels::nullif::nullif;
 use datafusion::arrow::datatypes::DataType;
 use datafusion::functions::string::concat::ConcatFunc;
 use datafusion_common::utils::list_ndims;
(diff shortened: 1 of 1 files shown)


crates/sail-plan/src/extension/function/collection/spark_concat.rs
```diff
@@ -104,14 +117,79 @@ impl ScalarUDFImpl for SparkConcat {
...
+                number_rows: args.number_rows,
+                return_field: args.return_field,
+            };
+
+            ConcatFunc::new().invoke_with_args(casted_args)
+        }?;
+
+        let concatenated_array = match concatenated {
+            ColumnarValue::Array(arr) => arr,
+            ColumnarValue::Scalar(s) => s.to_array()?, // преобразуем скаляр в массив
```
linhr: I feel this line of code is likely self-explanatory without comment.

```suggestion
            ColumnarValue::Scalar(s) => s.to_array()?,
```

linhr (approved):
This looks nice!

linhr:
Thanks for having this fix! I applied a minor edit to the code and created #650 for future work.

PR #649, opened 2025-07-21
fix: case when else expr inconsistent eager evaluation

### LINKED ISSUE #648: CASE expression: ELSE part should be lazy evaluated
spark lazy evaluates all the expressions
but datafusion evaluates `CASE ... ELSE this` eagerly

I think we should also be more careful when using datafusion CASE expression:
```rust
expr::Expr::Case(expr::Case {
        expr: None,
        when_then_expr,
        else_expr: Some(else_expr) // this else_expr is always evaluated and it's not good
    })
```

to reproduce:
```python

server = SparkConnectServer()
server.start()

df = spark.createDataFrame([(2, "Alice"), (5, "Bob")], ["age", "name"])
result = df.select(
    df.name, sf.when(sf.lit(True), 1).otherwise(
        sf.raise_error("unreachable")).alias("when"))
result.show()
```
spark:
```
+-----+----+
| name|when|
+-----+----+
|Alice|   1|
|  Bob|   1|
+-----+----+
```
sail 0.3.1:
```
AnalysisException: unreachable
```

SparkApplicationMaster:
now `case when ... else this` or `when(...).otherwise(this)`  is lazy evaluated like in spark:
```python

server = SparkConnectServer()
server.start()

df = spark.createDataFrame([(2, "Alice"), (5, "Bob")], ["age", "name"])
result = df.select(
    df.name, sf.when(sf.lit(True), 1).otherwise(
        sf.raise_error("unreachable")).alias("when"))
result.show()
```
```
+-----+----+
| name|when|
+-----+----+
|Alice|   1|
|  Bob|   1|
+-----+----+
```

passes spark-test
closes #648

### DIFF
--- a/crates/sail-plan/src/function/scalar/conditional.rs
+++ b/crates/sail-plan/src/function/scalar/conditional.rs
@@ -11,19 +11,18 @@ fn case(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     let ScalarFunctionInput { arguments, .. } = input;
     let mut when_then_expr = Vec::new();
     let mut iter = arguments.into_iter();
-    let mut else_expr: Option<Box<expr::Expr>> = None;
     while let Some(condition) = iter.next() {
         if let Some(result) = iter.next() {
             when_then_expr.push((Box::new(condition), Box::new(result)));
         } else {
-            else_expr = Some(Box::new(condition));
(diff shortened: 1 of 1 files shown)


linhr (approved):
Nice fix!

PR #662, opened 2025-07-28
fix: avoid repartitioning child for PySpark UDTF

### LINKED ISSUE #659: inconsistent behaviour between local and local-cluster modes
```python
from pyspark.sql import SparkSession, Row
from pyspark.sql.functions import udtf, lit

import os
os.environ["SAIL_MODE"]="local"
os.environ["SAIL_MODE"]="local-cluster"

server = SparkConnectServer()
server.start()

@udtf(returnType="key: string, value: float")
class TestUDTF:
    def __init__(self):
        self._count = 0
        self._sum = 0

    def eval(self, x: int):
        self._count += 1
        self._sum += x
        yield "input", float(x)

    def terminate(self):
        yield "count", float(self._count)
        yield "avg", self._sum / self._count

spark.udtf.register("test_udtf", TestUDTF)

df = spark.sql(
    """
    SELECT id, key, value 
    FROM range(0, 10, 1, 2), LATERAL test_udtf(id)
    WHERE key != 'input'
    """
)

df.collect()
```

LOCAL mode:
```
  File "/tmp/ipykernel_50135/3584556853.py", line 28, in terminate
    yield "avg", self._sum / self._count
                 ~~~~~~~~~~^~~~~~~~~~~~~
ZeroDivisionError: division by zero
```

LOCAL-CLUSTER mode
```
[Row(id=9, key='count', value=5.0),
 Row(id=9, key='avg', value=7.0),
 Row(id=4, key='count', value=5.0),
 Row(id=4, key='avg', value=2.0)]
```

the exception on empty partitions will be fixed in https://github.com/lakehq/sail/pull/658
but what causes sail to create this empty partitions in local mode?

this affects the test: pyspark/sql/tests/connect/test_parity_udtf.py::UDTFParityTests::test_udtf_terminate
in local-cluster mode the result is correct and consistent with spark
but spark-tests are running in local mode

linhr:
Fixes #659.

1. Correct the partitioning requirement for `MapPartitionsExec`.
2. Ensure explicit repartitioning for `DataFrame.repartition()` and `DataFrame.coalesce()`.

### DIFF
--- a/crates/sail-physical-optimizer/src/lib.rs
+++ b/crates/sail-physical-optimizer/src/lib.rs
@@ -68,8 +68,8 @@ mod tests {
             .filter(|name| datafusion_optimizer_names.contains(name))
             .collect();
         assert_eq!(
-            actual_datafusion_optimizer_names,
             datafusion_optimizer_names,
+            actual_datafusion_optimizer_names,
             "the custom physical optimizer rules should include all the default DataFusion optimizer rules in the same order"
         );
 
--- a/crates/sail-plan/src/resolver/query/repartition.rs
+++ b/crates/sail-plan/src/resolver/query/repartition.rs
@@ -18,12 +18,12 @@ impl PlanResolver<'_> {
         let input = self
             .resolve_query_plan_with_hidden_fields(input, state)
             .await?;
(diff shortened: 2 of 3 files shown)


lonless9 (approved):
Wonderful!

PR #670, opened 2025-07-29
feat: add `try_to_number` function

rafafrdz:
Following the implementation of `to_number` (merged in #629 ).

This PR introduces the implementation of the `try_to_number` functionality, closely replicating Spark's `try_to_number` capabilities. The function enables safe conversion of string columns into numeric types by parsing formatted strings according to locale-aware number patterns, returning a decimal representation or null on invalid input.

Spark Ref: [`try_to_number` doc](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.try_to_number.html)
Databricks Ref: [`try_to_number` doc](https://docs.databricks.com/aws/en/sql/language-manual/functions/try_to_number)

In this PR, we:

- [x] Add support for the `try_to_number` Spark SQL function  
- [x] Add doc tests
- [x] Fix grouping comma condition in `to_number`
- [x] Fix the angled brackets in the format string (they must be both or none)

part of #508

### DIFF
--- a/python/pysail/tests/spark/function/test_try_to_number.txt
+++ b/python/pysail/tests/spark/function/test_try_to_number.txt
@@ -125,11 +125,11 @@
 +------+------------+
 
 >>> spark.sql("SELECT try_to_number('-$123,456.32', 'S$999,999.999999PR') AS result, typeof(try_to_number('-$123,456.32', 'S$999,999.999999PR')) AS type").show(truncate=False)
-+------+-------------+
-|result|type         |
-+------+-------------+
-|NULL  |decimal(12,6)|
-+------+-------------+
++--------------+-------------+
+|result        |type         |
++--------------+-------------+
+|-123456.320000|decimal(12,6)|
++--------------+-------------+
 
 >>> spark.sql("SELECT try_to_number('-123,456.32-', 'MI$999,999.99S') AS result, typeof(try_to_number('-123,456.32-', 'MI$999,999.99S')) AS type").show(truncate=False)
 +------+------------+


rafafrdz:
In this PR, I extracted the logic that handles Array values from to_number to avoid code duplication. This allows me to reuse the SparkToNumber behavior by simply wrapping the value handler function to return Null whenever SparkToNumber throws an exception.

crates/sail-plan/src/extension/function/string/spark_try_to_number.rs
```diff
@@ -0,0 +1,78 @@
...
+impl Default for SparkTryToNumber {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ScalarUDFImpl for SparkTryToNumber {
+    fn as_any(&self) -> &dyn Any {
+        self.inner.as_any()
+    }
```
lonless9: Could we return `self` instead?

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -98,49 +99,72 @@ impl ScalarUDFImpl for SparkToNumber {
...
-            SparkToNumber::NAME,
-            args.len()
-        );
+pub trait SparkToNumberInnerImpl {
+    fn value_to_number(
+        value: Option<&str>,
+        number_components: &NumberComponents,
+        value_regex: &Regex,
+        format_spec: &RegexSpec,
+    ) -> Result<ScalarValue>;
```
lonless9: Maybe we can follow `SparkTryToBinary` instead?

python/pysail/tests/spark/function/test_try_to_number.txt
```diff
@@ -0,0 +1,139 @@
...
++----------+------------+
+|-123456.32|decimal(8,2)|
++----------+------------+
+
+>>> spark.sql("SELECT try_to_number('-$123,456.32-', 'MI$999,999.99S') AS result, typeof(try_to_number('-$123,456.32-', 'MI$999,999.99S')) AS type").show(truncate=False)
++---------+------------+
+|result   |type        |
++---------+------------+
+|123456.32|decimal(8,2)|
++---------+------------+
```
lonless9, line 69: I'm curious whether Spark would return a positive or negative number in this case?

crates/sail-plan/src/extension/function/string/spark_to_number.rs
```diff
@@ -510,18 +534,27 @@ fn match_grouping(value_captures: &Captures, format_spec: &RegexSpec) -> Result<
     let format_positions = get_grouping_positions(&format_numbers);
     let number_positions = get_grouping_positions(&numbers);
 
+    // Check if the positions of the number groupings are in the format groupings
+    let grouping_cond = number_positions.is_subset(&format_positions);
+
```
lonless9: What will format `9,999` with value `1000` look like?

lonless9 (commented):
This looks great! I've left some comments.

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Got it!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: Yeah, you're right! The code below does check if it's a subset, but also ensures that the grouping separator characters are at the expected positions, not just that the separators exist. As you said, just checking for a subset isn't enough. Thanks!

```rust
    for (pos, sep) in &format_positions {
        if pos < &numbers.len() && numbers.chars().rev().nth(*pos) != Some(*sep) {
            return exec_err!("Malformed integer format related groupings: {format_numbers}.");
        }
    }
```

I'll delete it then!

python/pysail/tests/spark/function/test_try_to_number.txt
rafafrdz: The way `to_number` and `try_to_number` work under the hood is kind of black magic, you think you understand it until you run into some weird formatting edge case 😅

It's not really obvious how they thought /implemented their behaviour... and the documentation as well...

crates/sail-plan/src/extension/function/string/spark_try_to_number.rs
rafafrdz: done!

crates/sail-plan/src/extension/function/string/spark_to_number.rs
rafafrdz: done! I re-did it using that approach 😄

rafafrdz:
Hey @lonless9 do you know anything about the clippy error that the GitHub action threw? it has nothing to do with `try_to_number`. Besides, I ran in my local `cargo clippy --all-targets --all-features -- -D warnings` and nothing happened 😅

lonless9:
> Hey @lonless9 do you know anything about the clippy error that the GitHub action threw? it has nothing to do with `try_to_number`. Besides, I ran in my local `cargo clippy --all-targets --all-features -- -D warnings` and nothing happened 😅

You may need to update your rust toolchain version via `rustup update`.

rafafrdz:
> > Hey @lonless9 do you know anything about the clippy error that the GitHub action threw? it has nothing to do with `try_to_number`. Besides, I ran in my local `cargo clippy --all-targets --all-features -- -D warnings` and nothing happened 😅
> 
> You may need to update your rust toolchain version via `rustup update`.

Sorry, I didn't explain myself well 😅 What I wanted to say is that there is an error when the GitHub action runs clippy that which has nothing to do with any of the code I did.

I show you in this PR: https://github.com/lakehq/sail/pull/715
Build error: https://github.com/lakehq/sail/actions/runs/16804418669/job/47593139862?pr=715

lonless9:
> Sorry, I didn't explain myself well 😅 What I wanted to say is that there is an error when the GitHub action runs clippy that which has nothing to do with any of the code I did.

Sorry for the inconvenience! I'll fix this.

linhr:
This is probably fixed in #720.

This happens since we use the latest Rust version in GitHub Actions, so whenever a new version is available, there may be new lints that cause our build to fail.

lonless9 (approved):
This is great! Thanks for your contribution!

PR #672, opened 2025-07-29
fix: position of the sign in `to_number` function

### LINKED ISSUE #669: Follow-up work for `to_number` function
This is a follow-up for #629 

* Pass the gold_data test case.
* Avoid repeated call of `Regex::new`

rafafrdz:
In this PR, I fixed the position of the sign in the regex expression (both prefix and suffix), and also addressed the right-padding of zeros when the format spec includes more 9s or 0s in the decimal part than are present in the input string.

On the other hand, I added more tests to cover those cases

Close #669


rafafrdz:
> Thank you for your contribution! This perfectly addresses the previous issue!
> 
> Could you please continue fixing the issue regarding the repeated `Regex::new` ? It seems that reusing `Capture` doesn't prevent this from happening. We should completely move the related logic to the outer before the iter loop.

I will try a different approach for that 😄

lonless9:
> I will try a different approach for that 😄

Thanks! I hope we can call `Regex::new` just once.

rafafrdz:
Hey @lonless9, I think I got it now! Check it out when you can and let me know if I’m still missing something.

BTW, just a quick thought. I'm not sure if the code is well organized in terms of struct definitions, impls, macros and all that. If you think it could be cleaner or better structured, feel free to say it 😅

Thank you in advance!

lonless9:
Sorry!  For the dynamic generated Regex, new is surely needed.

For the `format_spec`, we could follow this as a reference. 

https://github.com/lakehq/sail/blob/main/crates/sail-catalog/src/utils.rs#L6-L11

rafafrdz:
> Sorry! For the dynamic generated Regex, new is surely needed.
> 
> For the `format_spec`, we could follow this as a reference.
> 
> https://github.com/lakehq/sail/blob/main/crates/sail-catalog/src/utils.rs#L6-L11

Done! Yep, that's right, I think I cannot move the dynamically generated regex any further out. What do you think?

On the other hand, the format regex is now created as static by using `lazy_static` 😄

lonless9:
Thank you for your contribution!!!

PR #680, opened 2025-07-31
feat: schema evolution for Delta Lake

lonless9:
part of #171

This PR implements the following Delta Lake schema options
* replaceWhere
* mergeSchema
* overwriteSchema

Additionally, the following write options are also supported
* targetFileSize
* writeBatchSize


crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -164,9 +164,13 @@ impl TryFrom<adt::DataType> for DataType {
...
-            adt::DataType::Union { .. }
-            | adt::DataType::Dictionary { .. }
-            | adt::DataType::RunEndEncoded(_, _) => return Err(error(&data_type)),
+            adt::DataType::Dictionary(_, value_type) => {
+                let inner_data_type: DataType = value_type.as_ref().clone().try_into()?;
+                return Ok(inner_data_type);
+            }
+            adt::DataType::Union { .. } | adt::DataType::RunEndEncoded(_, _) => {
+                return Err(error(&data_type))
+            }
```
lonless9: The change here is related to partition (a schema related optimize in delta to use Dictionary), and I'm not sure if this is correct

linhr: This probably does not work since the Spark client cannot handle the data even if the data type is converted here. I worked around it for listing tables like [this](https://github.com/lakehq/sail/blob/44a76ce89f4293c267744fd05d012529dd7f77e5/crates/sail-data-source/src/listing.rs#L80-L95). You can try to bring similar logic to Delta.

python/pysail/tests/spark/delta/test_delta_advanced.py
```diff
@@ -0,0 +1,59 @@
...
+        v0_df = spark.read.format("delta").option("versionAsOf", "0").load(delta_table_path)
+        assert v0_df.collect() == [Row(id=1, value="v0")]
+
+        # Read version 1
+        v1_df = spark.read.format("delta").option("versionAsOf", "1").load(delta_table_path).sort("id")
+        expected_v1 = [Row(id=1, value="v0"), Row(id=2, value="v1")]
+        assert v1_df.collect() == expected_v1
+
+    @pytest.mark.skipif(is_jvm_spark(), reason="Sail only - Delta Lake error handling")
+    def test_delta_feature_error_handling(self, spark, tmp_path):
```
linhr: This seems a duplicate of `test_delta_io_error_on_read_nonexistent_table`.

python/pysail/tests/spark/delta/test_delta_schema_handling.py
```diff
@@ -0,0 +1,171 @@
...
+        delta_path = tmp_path / "delta_merge_type_changes"
+
+        # Create initial table with integer column
+        initial_data = [Row(id=1, value=100)]
+        df1 = spark.createDataFrame(initial_data)
+        df1.write.format("delta").mode("overwrite").save(str(delta_path))
+
+        # Append data with same column but potentially different precision
+        # This tests the schema merging logic for compatible types
+        new_data = [Row(id=2, value=200)]
```
linhr: How is the schema different for new data?

crates/sail-delta-lake/src/delta_format/sink.rs
```diff
@@ -136,158 +269,150 @@ impl DataSink for DeltaDataSink {
...
-            }
+        use crate::delta_datafusion::create_object_store_url;
+        let object_store_url = create_object_store_url(&self.table_url);
+        context
+            .runtime_env()
+            .register_object_store(object_store_url.as_ref(), object_store.clone());
+
+        // Collect all batches first
+        let mut batches = Vec::new();
+        let mut total_rows = 0u64;
```
linhr: Is there a reason to collect all batches? I feel it may cause memory issue if the input stream is large.

crates/sail-delta-lake/src/operations/write/execution.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub fn new(num_indexed_cols: i32, stats_columns: Option<Vec<String>>) -> Self {
+        Self {
+            num_indexed_cols,
+            stats_columns,
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn execute_non_empty_expr(
```
linhr, line 43: Ok now I understand the write logic better. This is part of "planning" but requires some "execution" to gather information for the planning. There is actually an opportunity to make this distributed, if this function has access to the job runner. (Spark can actually do distributed file listing when planning to read a directory of files.) (Another example we already have is the `DataFrame.tail()` method, which requires getting the dataset size to determine the offset. Right now this is implemented as a query execution within the plan resolver, which is kind of hacky.)

I also created #711 for a use case that we can consider in the future.

We now have many examples that SQL parsing, planning, and query execution can all interleave. Let's think about this holistically and see if we can make our query planning architecture cleaner in the future!

linhr (commented):
This is exciting! Overall it looks great!

I have a few comments/questions, and I'd be happy to discuss!

crates/sail-spark-connect/src/proto/data_type_arrow.rs
lonless9: I've turned off this option, and we can consider removing the entire related logic later

python/pysail/tests/spark/delta/test_delta_advanced.py
lonless9: Done!

python/pysail/tests/spark/delta/test_delta_schema_handling.py
lonless9: Done! And I realized I missed the promotion logic, which is now fixed.

crates/sail-delta-lake/src/operations/write/execution.rs
lonless9: Sailing here we are!

crates/sail-delta-lake/src/delta_format/sink.rs
lonless9: Done!

PR #688, opened 2025-08-01
feat: consolidate write plan resolver

linhr:


crates/sail-plan/src/resolver/command/insert.rs
```diff
@@ -1,55 +1,128 @@
...
+use crate::resolver::command::write::{WriteColumnMatch, WritePlanBuilder, WriteTarget};
 use crate::resolver::state::PlanResolverState;
 use crate::resolver::PlanResolver;
 
 impl PlanResolver<'_> {
-    #[allow(clippy::too_many_arguments)]
+    pub(super) async fn resolve_command_insert_overwrite_directory(
+        &self,
+        input: spec::QueryPlan,
+        _local: bool,
```
lonless9: We can add explanation for ignoring `local`

lonless9 (approved):
Wonderful!

crates/sail-plan/src/resolver/command/insert.rs
linhr: Done!

PR #694, opened 2025-08-02
feat: try_cast column method

SparkApplicationMaster:
implement `try_cast` column method:
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.try_cast.html#pyspark.sql.Column.try_cast

the overriden cast cases need to be further checked for correctness in terms of try_cast and adjusted accordingly
but it passes the spark-test, so basic functionality is provided

### DIFF
--- a/crates/sail-common/src/spec/expression.rs
+++ b/crates/sail-common/src/spec/expression.rs
@@ -32,6 +32,7 @@ pub enum Expr {
         cast_to_type: DataType,
         /// Whether to rename the expression to `CAST(... AS ...)`.
         rename: bool,
+        is_try: bool,
     },
     UnresolvedRegex {
         /// The regular expression to match column names.
--- a/crates/sail-plan/src/resolver/expression.rs
+++ b/crates/sail-plan/src/resolver/expression.rs
@@ -16,8 +16,8 @@ use datafusion::sql::unparser::expr_to_sql;
 use datafusion_common::{Column, DFSchemaRef, DataFusionError, TableReference};
 use datafusion_expr::expr::{FieldMetadata, ScalarFunction, WindowFunctionParams};
 use datafusion_expr::{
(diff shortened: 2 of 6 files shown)


SparkApplicationMaster:
there is also a function with the same name
it's not even in the SQL docs yet, so no gold_data tests:
https://spark.apache.org/docs/latest/api/sql/index.html

but it works in vanilla spark 4.0.0:
```python
spark.sql("select try_cast(1 as string)").show()
+---------------------+
|TRY_CAST(1 AS STRING)|
+---------------------+
|                    1|
+---------------------+

```
this doesn't work in my solution, and I don't know yet how to implement it because I haven't done any SQL grammar parsing

linhr (approved):
This is nice!

If you're curious how the SQL syntax can be supported, you can look at `AtomExpr::Cast` in the `sail-sql-parser` crate. Once the grammar is defined, the parsing works automatically. Then you can update `from_ast_atom_expression` to convert SQL AST to the Sail `spec::Expr`. (Adding doctests to `python/pysail/tests/spark` could be helpful to ensure everything works end to end.)

SparkApplicationMaster:
yep, first I tried to add AtomExpr::TryСast and the whole new parsing branch with proto messages copy, but failed
and then in proto generated files I found that TRY is already parsed in original Cast if it is called as a Column method)
after some time I'll come back to this if you don't mind

linhr:
The SQL parser and the proto are two independent inputs to Sail, and they are both converted to the Sail "spec" before logical analysis. You can add `AtomExpr::TryСast` without worrying about how the operation is modeled in proto (which is used for the DataFrame API). Then in `from_ast_atom_expression` you convert `AtomExpr` (part of the SQL AST) to the Sail spec and set the `is_try` flag that you defined in this PR.

Yeah feel free to explore this further and I'm happy to answer questions! And feel free to create a new issue for this in case I miss future messages in this merged PR.

PR #698, opened 2025-08-03
feat: arrays_zip

SparkApplicationMaster:
implement arrays_zip function:
https://spark.apache.org/docs/latest/api/sql/index.html#arrays_zip

reuses existing `fn to_struct_array` to build row elements
passes spark-tests
part of #220


crates/sail-plan/src/extension/function/array/arrays_zip.rs
```diff
@@ -0,0 +1,162 @@
...
+use arrow::compute::{cast, concat};
+use arrow::datatypes::{DataType, Field};
+use datafusion_common::{arrow_err, exec_err, plan_err, DataFusionError, Result};
+use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility};
+use datafusion_functions::utils::make_scalar_function;
+
+use crate::extension::function::struct_function::to_struct_array;
+
+#[derive(Debug)]
+pub struct ArraysZip {
```
linhr, line 18: Could you register this UDF to `crates/sail-execution/src/codec.rs`? (We need to update both `try_encode_udf` and `try_decode_udf`.)

crates/sail-plan/src/extension/function/array/arrays_zip.rs
```diff
@@ -0,0 +1,162 @@
...
+    Arc::new(Field::new_struct("item", fields, true))
+}
+
+fn arrays_zip_inner(args: &[ArrayRef]) -> Result<ArrayRef> {
+    let casted_lists = args
+        .iter()
+        .map(|arr| match arr.data_type() {
+            DataType::List(_field) => Ok(arr.clone()),
+            DataType::LargeList(field) | DataType::FixedSizeList(field, _) => {
+                cast(arr, &DataType::List(field.clone())).map_or_else(|err| arrow_err!(err), Ok)
```
linhr, line 216: Could we add a TODO here so that we can properly support `LargeList` and `FixedSizeList` in the future?

linhr (commented):
This looks nice!

I have a few minor comments, but otherwise I think this PR is ready to go!

crates/sail-plan/src/extension/function/array/arrays_zip.rs
SparkApplicationMaster: done

crates/sail-plan/src/extension/function/array/arrays_zip.rs
SparkApplicationMaster: finally worked out how this OffsetSizeTraits work and implemented both LargeList and FixedSizeList cases
FixedSizeList works here only if sizes of all nested lists are equal, otherwise casts to List

linhr: Nice!

Issue #506, opened 2025-05-28
Spark SQL Functions Coverage and Parity - Math


linhr:
- [x] %
- [x] *
- [x] +
- [x] -
- [x] /
- [x] abs
- [x] acos
- [x] acosh
- [x] asin
- [x] asinh
- [x] atan
- [x] atan2
- [x] atanh
- [x] bin
- [ ] bround
- [x] cbrt
- [x] ceil
- [x] ceiling
- [ ] conv
- [x] cos
- [x] cosh
- [x] cot
- [ ] csc
- [x] degrees
- [x] div
- [x] e
- [x] exp
- [x] expm1
- [x] factorial
- [x] floor
- [x] greatest
- [x] hex
- [x] hypot
- [x] least
- [x] ln
- [x] log
- [x] log10
- [x] log1p
- [x] log2
- [x] mod
- [x] negative
- [x] pi
- [x] pmod
- [x] positive
- [x] pow
- [x] power
- [x] radians
- [x] rand
- [x] randn
- [x] random
- [x] rint
- [x] round
- [ ] sec
- [x] shiftleft
- [x] sign
- [x] signum
- [x] sin
- [x] sinh
- [x] sqrt
- [x] tan
- [x] tanh
- [ ] try_add
- [ ] try_divide
- [ ] try_multiply
- [ ] try_subtract
- [x] unhex
- [ ] width_bucket

davidlghellin:
Hi! I started working on this before asking, sorry about that 😅  
If possible, could this issue  be assigned to me?  
Thanks!

shehabgamin:
> Hi! I started working on this before asking, sorry about that 😅 If possible, could this PR be assigned to me? Thanks!

of course!

PR #699, opened 2025-08-03
chore: math functions refactor

SparkApplicationMaster:
LN func: now it is not calculating expr_fn::ln case branch if input is zero(#648 for more info)
make other math functions more readable with less code, no logic changed

fix output column naming for `positive` and `sec` funcs so they pass the spark-tests, no logic changed
part of #506


SparkApplicationMaster:
sorry, my linters tricked me somehow

SparkApplicationMaster:
btw, is metadata here so important?
if not, lower literals can be simplified too
https://github.com/lakehq/sail/blob/cd7422f6f4af7cf8680dad921a8d4fbd6c9f3c43/crates/sail-plan/src/function/scalar/math.rs#L479

linhr:
> btw, is metadata here so important?
if not, lower literals can be simplified too

It seems there is a `lit_with_metadata` function we can use.

PR #709, opened 2025-08-05
feat: file listing and statistics cache

shehabgamin:
Subset of https://github.com/lakehq/sail/pull/687 that we can merge in


linhr (approved):
🚀

PR #712, opened 2025-08-06
feat: update cache config

linhr:
1. Update configuration design for file listing cache and file statistics cache.
2. Support global and per-session cache.


crates/sail-cache/src/file_listing_cache.rs
```diff
@@ -0,0 +1,114 @@
...
+            builder = builder.max_capacity(max_entries);
+        }
+
+        Self {
+            objects: builder.build(),
+        }
+    }
+}
+
+impl CacheAccessor<Path, Arc<Vec<ObjectMeta>>> for MokaFileListingCache {
```
linhr, line 38: This file is shown as a new file due to renaming, but I didn't change the implementation.

crates/sail-object-store/src/s3.rs
```diff
@@ -108,7 +108,6 @@ pub async fn get_s3_object_store(url: &Url) -> object_store::Result<AmazonS3> {
     let config = DEFAULT_AWS_CONFIG
         .get_or_init(|| aws_config::defaults(BehaviorVersion::latest()).load())
         .await;
-    debug!("Using AWS config: {config:#?}");
```
linhr, line 111: This is removed since the AWS config may contain sensitive information.

crates/sail-spark-connect/src/session_manager.rs
```diff
@@ -164,22 +168,58 @@ impl SessionManager {
...
+                match &options.config.parquet.file_statistics_cache.r#type {
+                    CacheType::None => {
+                        debug!("Not using files statistics cache");
+                        None
+                    }
+                    CacheType::Global => {
+                        debug!("Using global files statistics cache");
+                        Some(
+                            self.global_file_statistics_cache
+                                .get_or_insert_with(|| {
```
linhr, line 183: No lock is needed since we can take `&mut self` thanks to the actor mode.

The global cache is stored in the `SessionManagerActor` now and we can now access cache config from `AppConfig`. The cache is still "global" since a single instance of the actor is created when the server starts.

PR #722, opened 2025-08-08
feat: `split` function

rafafrdz:
Implement `split` function: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.split.html

Part of #508

### DIFF
--- a/crates/sail-plan/src/extension/function/string/spark_split.rs
+++ b/crates/sail-plan/src/extension/function/string/spark_split.rs
@@ -2,7 +2,8 @@ use std::any::Any;
 use std::sync::Arc;
 
 use arrow::array::{
-    Array, ArrayRef, Int32Array, ListArray, ListBuilder, StringArray, StringBuilder,
+    Array, ArrayRef, GenericStringArray, Int32Array, ListArray, ListBuilder, OffsetSizeTrait,
+    StringBuilder,
 };
 use arrow::datatypes::{DataType, Field};
 use datafusion_common::Result;
@@ -97,11 +98,38 @@ pub fn spark_split_inner(args: &[ArrayRef]) -> Result<ArrayRef> {
             "requires 2 or 3 arguments",
         ));
     }
+
+    match (args[0].data_type(), args[1].data_type()) {
+        (DataType::LargeUtf8, DataType::LargeUtf8) => spark_split_inner_downcast::<i64, i64>(args),
+        (_, DataType::LargeUtf8) => spark_split_inner_downcast::<i32, i64>(args),
+        (DataType::LargeUtf8, _) => spark_split_inner_downcast::<i64, i32>(args),
+        _ => spark_split_inner_downcast::<i32, i32>(args),
+    }
+}
+pub fn spark_split_inner_downcast<FirstOffset, SecondOffset>(args: &[ArrayRef]) -> Result<ArrayRef>
+where
+    FirstOffset: OffsetSizeTrait,
+    SecondOffset: OffsetSizeTrait,
+{
+    if args.len() < 2 || args.len() > 3 {
+        return Err(generic_exec_err(
+            SparkSplit::NAME,
+            "requires 2 or 3 arguments",
+        ));
+    }
     let len: usize = args[0].len();
 
     // Getting the arrays
-    let values: Arc<Option<&StringArray>> = Arc::new(opt_downcast_arg!(&args[0], StringArray));
-    let format: Arc<Option<&StringArray>> = Arc::new(opt_downcast_arg!(&args[1], StringArray));
+    let values: Arc<Option<&GenericStringArray<FirstOffset>>> = Arc::new(
+        args[0]
+            .as_any()
+            .downcast_ref::<GenericStringArray<FirstOffset>>(),
+    );
+    let format: Arc<Option<&GenericStringArray<SecondOffset>>> = Arc::new(
+        args[1]
+            .as_any()
+            .downcast_ref::<GenericStringArray<SecondOffset>>(),
+    );
 
     let limit_arg_is_none = args.get(2).is_none();
     let limit = Arc::new(


crates/sail-plan/src/extension/function/functions_nested_utils.rs
```diff
@@ -9,7 +9,14 @@ macro_rules! downcast_arg {
         })?
     }};
 }
-pub(crate) use downcast_arg;
+
+macro_rules! opt_downcast_arg {
```
rafafrdz, line 13: Hey! I consider creating a dowcast_arg but with an Option result in case of dealing with the error instead of throwing it. Is it okay? 😅

rafafrdz, line 13: Hey! I considered creating a dowcast_arg but with an Option result in case of dealing with the error instead of throwing it. Is it okay? 😅

crates/sail-plan/src/extension/function/string/spark_split.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+    let len: usize = args[0].len();
+    // Initialize default values in case of nulls
+    let default_nulls: Vec<Option<&str>> = vec![None; len];
+    let default_zeros: Vec<i32> = vec![0; len];
+    let default_nulls: StringArray = StringArray::from(default_nulls);
+    let default_limit_nulls: NullBuffer = NullBuffer::from(vec![false; len]);
+    let default_limit_nulls: Int32Array =
+        Int32Array::new(default_zeros.into(), Some(default_limit_nulls));
+    let default_limit: Int32Array = Int32Array::from(vec![0; len]);
```
rafafrdz: I was wondering if there’s a better way to “thunk” or lazily initialize those default values. The straightforward approach would be to handle the arrays by downcasting with pattern matching and declaring those defaults on the fly. What do you think? 😄 

Maybe using `lazy_static` would make sense here, or is it just not worth the extra complexity?

crates/sail-execution/src/codec.rs
```diff
@@ -831,9 +832,10 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_concat" | "concat" => Ok(Arc::new(ScalarUDF::from(SparkConcat::new()))),
             "spark_from_csv" | "from_csv" => Ok(Arc::new(ScalarUDF::from(SparkFromCSV::new()))),
             "spark_to_number" | "to_number" => Ok(Arc::new(ScalarUDF::from(SparkToNumber::new()))),
-            "spark_try_to_number" | "rey_to_number" => {
+            "spark_try_to_number" | "try_to_number" => {
```
lonless9, line 835: Oh nice catch!

crates/sail-execution/src/codec.rs
```diff
@@ -831,9 +832,10 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_concat" | "concat" => Ok(Arc::new(ScalarUDF::from(SparkConcat::new()))),
             "spark_from_csv" | "from_csv" => Ok(Arc::new(ScalarUDF::from(SparkFromCSV::new()))),
             "spark_to_number" | "to_number" => Ok(Arc::new(ScalarUDF::from(SparkToNumber::new()))),
-            "spark_try_to_number" | "rey_to_number" => {
```
lonless9, line 834: Oh nice catch!

crates/sail-plan/src/extension/function/functions_nested_utils.rs
lonless9: I think this make sense!
But we should properly handle the `None` rather than using `unwrap_or` and ignoring it.

crates/sail-plan/src/extension/function/string/spark_split.rs
lonless9: I guess we can determine the number of rows and then loop through them. In each iteration, we fetch the value for that specific row from each arg. In all this is `append null` for Null and `append value` for Value.

I think `lazy_static` may not work here, for the reason that we actually care about the size rather than eagerly or lazy.

crates/sail-plan/src/extension/function/string/spark_split.rs
```diff
@@ -0,0 +1,149 @@
...
+            ] => {
+                Ok(vec![
+                    arg_types[0].clone(),
+                    arg_types[1].clone()
+                ])
+            }
+            [
+                DataType::Utf8 | DataType::Utf8View | DataType::LargeUtf8 | DataType::Null ,
+                DataType::Utf8 | DataType::Utf8View | DataType::LargeUtf8 | DataType::Null ,
+                DataType::Int32 | DataType::Int64 | DataType::UInt32 | DataType::UInt64 | DataType::Null | DataType::Utf8
```
lonless9: Does the usage of `limit` in `invoke_with_args` matches the `DataType::Utf8` here?

rafafrdz: Sorry, my mistake!

crates/sail-plan/src/extension/function/string/spark_split.rs
rafafrdz: Done!

crates/sail-plan/src/extension/function/functions_nested_utils.rs
rafafrdz: Got it!

crates/sail-plan/src/extension/function/string/spark_split.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    let array: ListArray = builder.finish();
+    Ok(Arc::new(array))
+}
+
+pub fn split_to_array(value: &str, format: &str, limit: i32) -> Result<Vec<Option<String>>> {
+    let format: Regex =
+        Regex::new(format).map_err(|_| generic_exec_err(SparkSplit::NAME, "Invalid regex"))?;
```
lonless9, line 172: It seems that we are looping with `Regex` here?

crates/sail-plan/src/extension/function/string/spark_split.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::List(Arc::new(Field::new_list_field(
+            DataType::Utf8,
+            true,
+        ))))
+    }
+
+    fn coerce_types(&self, arg_types: &[DataType]) -> Result<Vec<DataType>> {
```
lonless9, line 58: How about `LargeStringArray` or `StringViewArray`  as `LargeUtf8` mentioned here?

crates/sail-plan/src/extension/function/string/spark_split.rs
rafafrdz: It does... the regex depends on the format value from each row...

lonless9: Ohh this doesn't matter!

crates/sail-plan/src/extension/function/string/spark_split.rs
rafafrdz: Sorry, I didn't catch you 😅 . Could you explain to me what you mean? 😃

lonless9: Sorry, what I actually meant is that the parameters of `coerce_types` and `invoke_with_args` could correspond. We are currently handling `StringArray` and `Int32Array`. Will the coresponding type of `LargeUtf8` be handled as well?

rafafrdz: Ahh got it, now I understand, sorry for my mistake 🙏

What happened is that I thought `StringArray` was the only physical array type covering all the logical _string_ types, meaning that both `Utf8` and `LargeUtf8` would converge to `StringArray`.

So, if that works for you, I can just assume the longer string case and downcast to `LargeStringArray` instead of `StringArray`. Is this ok? 😄

lonless9: Sounds good to me!

rafafrdz: I updated the implementation to properly handle both `Utf8` and `LargeUtf8` inputs.

The `spark_split_inner` function now dispatches based on the logical types of the two string arguments and selects the appropriate offset sizes (`i32` or `i64`). This way, `spark_split_inner_downcast` is invoked with the correct `OffsetSizeTrait` combination, covering all cases and downcasting to the corresponding array type (`StringArray` or `LargeStringArray`)

PR #723, opened 2025-08-08
feat: support Delta codec

linhr:
1. Support cluster mode codec for `DeltaDataSink`. (`DeltaScan` does not seem part of the physical plan so the codec is not added.)
2. Remove the logic for `delta-rs://` URLs since it breaks the reader when running in cluster mode. (The registered object store for `delta-rs://` URLs is not available in the worker.) I've adjusted the file group logic in `DeltaScanBuilder::build()` so that we do not need to rely on `delta-rs://` URLs (which is IMO a hack that we inherited from the Delta library).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5628,32 +5628,23 @@ dependencies = [
 name = "sail-delta-lake"
 version = "0.3.2"
 dependencies = [
- "arrow-buffer",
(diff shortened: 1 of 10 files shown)


linhr:
All Delta Lake Python tests passed when running in `local-cluster` mode.

PR #727, opened 2025-08-09
feat: add `try_mod` function Spark 4.0.0

davidlghellin:
This should be part of #506 
[try_mod](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.try_mod.html) is new in version 4.0.0


crates/sail-plan/src/extension/function/math/spark_try_mod.rs
```diff
@@ -0,0 +1,170 @@
...
+                let r = right_arr.as_primitive::<Int64Type>();
+                let result = try_binary_op_primitive::<Int64Type, _>(l, r, i64::checked_rem);
+
+                binary_op_scalar_or_array(left, right, result)
+            }
+            (DataType::Decimal128(_, _), DataType::Decimal128(_, _)) => {
+                let l = left_arr.as_primitive::<Decimal128Type>();
+                let r = right_arr.as_primitive::<Decimal128Type>();
+                let result =
+                    try_binary_op_primitive::<Decimal128Type, _>(l, r, |a: i128, b: i128| {
```
linhr: I think there is more work to handle decimals of different precision and scale. In Arrow, some transformation is needed on the raw `i128` data, as you can see here: <https://github.com/apache/arrow-rs/blob/876585c1cd986dbaee0c26d52b55a4186a2f68c8/arrow-arith/src/numeric.rs#L916-L933>

linhr (commented):
Thanks for the contribution! I have a comment regarding decimals, but otherwise it looks good!

crates/sail-plan/src/extension/function/math/spark_try_mod.rs
davidlghellin: Thanks for the pointer! I dug into it and you’re right—handling Decimal128 with differing precision/scale needs proper rescaling of the raw i128. Didn’t realize it was that involved; learned a lot here. Appreciate it!

crates/sail-plan/src/extension/function/math/spark_try_mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        "try_mod"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, arg_types: &[DataType]) -> Result<DataType> {
+        if let [DataType::Decimal128(pl, sl), DataType::Decimal128(pr, sr)] = arg_types {
+            return Ok(DataType::Decimal128((*pl).max(*pr), (*sl).max(*sr)));
```
linhr: The precision and scale need to be adjusted here as well.

crates/sail-plan/src/extension/function/math/spark_try_mod.rs
```diff
@@ -0,0 +1,175 @@
...
+                let r = right_arr.as_primitive::<Decimal128Type>();
+
+                let result_scale: i8 = (*s1).max(*s2);
+                let int_digits_l = *p1 as i8 - *s1;
+                let int_digits_r = *p2 as i8 - *s2;
+                let result_precision: u8 =
+                    (result_scale.saturating_add(int_digits_l.min(int_digits_r)) as u8)
+                        .min(Decimal128Type::MAX_PRECISION);
+
+                let raw = try_binary_op_primitive::<Decimal128Type, _>(l, r, i128::checked_rem);
```
linhr: I think you need to transform the `l` and `r` arrays before doing the binary op. In the same Arrow code snippet I shared, there are two lines (which calls `pow_wrapping`) that does this.

Maybe we can add some doctests in `python/pysail/tests/spark/function` as well?

linhr (commented):
Nice progress! I think it's getting closer!

linhr (approved):
This looks great! And thanks for adding all the tests!

PR #737, opened 2025-08-10
feat: text file format

shehabgamin:
Also fixes issues with reading and writing compressed files.

### DIFF
--- a/crates/sail-execution/src/codec.rs
+++ b/crates/sail-execution/src/codec.rs
@@ -44,6 +44,7 @@ use prost::bytes::BytesMut;
 use prost::Message;
 use sail_common_datafusion::udf::StreamUDF;
 use sail_common_datafusion::utils::{read_record_batches, write_record_batches};
+use sail_data_source::formats::text::source::TextSource;
 use sail_delta_lake::delta_format::DeltaDataSink;
 use sail_plan::extension::function::array::arrays_zip::ArraysZip;
 use sail_plan::extension::function::array::spark_array::SparkArray;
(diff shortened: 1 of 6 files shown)


scripts/spark-tests/generate-test-report.sh
```diff
@@ -80,8 +80,9 @@ function show_test_summary() {
 }
 
 function show_code_block() {
-  # A GitHub comment has a maximum length of 65536 characters.
-  # So we need to truncate the content if it is too long.
+  # Need to truncate the content if it's too long since a GitHub comment has a maximum length of 65536 characters.
+  # Although, the max character limit may be based off of the gzipped size:
+  #   https://github.com/orgs/community/discussions/41331#discussioncomment-9276173
```
linhr, line 85: Nice find!

crates/sail-execution/src/codec.rs
```diff
@@ -615,7 +645,27 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
...
+                    let file_compression_type =
+                        self.try_encode_file_compression_type(file_scan.file_compression_type)?;
+                    let text_source = file_source
+                        .as_any()
+                        .downcast_ref::<TextSource>()
+                        .ok_or_else(|| {
+                            plan_datafusion_err!(
+                                "try_encode: TextSource not found in FileScanConfig"
+                            )
+                        })?;
```
linhr, line 660: You could just do this:
```rust
if let Some(text_source) = file_source.as_any().downcast_ref::<TextSource>() {
    ...
} else if file_source.as_any().is::<JsonSource>() {
    ...
}
```

PR #751, opened 2025-08-13
feat: support Spark Connect streaming API

linhr:


crates/sail-data-source/src/formats/rate/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+                let batch_size = rows_per_second / batches_per_second;
+                let interval = Duration::from_secs(1) / (batches_per_second as u32);
+                let generator = BatchGenerator::try_new(
+                    Arc::clone(&self.time_zone),
+                    &self.projection,
+                    self.schema(),
+                )?;
+                let output = futures::stream::unfold(generator, move |mut generator| async move {
+                    // The interval does not take into account the time it takes to generate data,
+                    // but the sleep itself is inaccurate anyway.
```
lonless9, line 232: Maybe we can look into something like the token bucket algorithm later.

crates/sail-data-source/src/formats/socket/reader.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+        if self.options.timeout_sec == 0 {
+            return plan_err!("timeout must be greater than 0 for reading from socket");
+        }
+        let options = self.options.clone();
+        let schema = self.schema();
+        let output = futures::stream::once(async move {
+            let stream = tokio::net::TcpStream::connect((options.host.as_str(), options.port))
+                .await
+                .map_err(|e| exec_datafusion_err!("failed to connect to socket: {e}"))?;
```
lonless9: Do we need to provide the host and port when throwing the error?

crates/sail-data-source/src/formats/rate/reader.rs
```diff
@@ -0,0 +1,327 @@
...
+                });
+                Box::pin(output)
+            } else {
+                let rows_per_second =
+                    (self.options.rows_per_second / self.options.num_partitions).max(1);
+                // We generate at most 1000 batches per second
+                // since the sleep function only has millisecond accuracy.
+                let batches_per_second = rows_per_second.min(1_000);
+                let batch_size = rows_per_second / batches_per_second;
+                let interval = Duration::from_secs(1) / (batches_per_second as u32);
```
lonless9, line 224: Maybe we can provide some configurable session options for related algorithms in the future.

crates/sail-data-source/src/formats/console/writer.rs
```diff
@@ -0,0 +1,108 @@
...
+        let output = futures::stream::once(async move {
+            stream
+                .enumerate()
+                .for_each(|(i, batch)| async move {
+                    println!("partition {partition} batch {i}");
+                    match batch {
+                        Ok(batch) => match print_batches(&[batch]) {
+                            Ok(_) => {}
+                            Err(e) => {
+                                println!("error printing batch: {e}");
```
lonless9: Is there an issue with stdout lock?

crates/sail-data-source/src/formats/socket/reader.rs
lonless9: And should we have a retry logic if the first connection failed? To be precise, we may divide errors into recoverable and non-recoverable.

linhr: > Do we need to provide the host and port when throwing the error?

Good point!

> And should we have a retry logic if the first connection failed? To be precise, we may divide errors into recoverable and non-recoverable.

Yeah for long running streaming queries, retry logic is important. Since this socket source is for testing purpose, I only had the basic logic. Let me add a comment in the code.

crates/sail-data-source/src/formats/console/writer.rs
linhr: Yes indeed! Multiple partitions write to stdout concurrently so the lines here may interleave. Actually let me use the stdout lock to make things look better.

crates/sail-spark-connect/src/streaming.rs
```diff
@@ -0,0 +1,317 @@
...
+        stopped: watch::Sender<bool>,
+        mut stream: SendableRecordBatchStream,
+    ) {
+        let task = async move {
+            while let Some(x) = stream.next().await {
+                match x {
+                    Ok(_) => {}
+                    Err(e) => {
+                        let cause = CommonErrorCause::new(&e);
+                        let _ = error.send(Some(cause.into()));
```
lonless9, line 75: Oh, this might be where we'll add fault tolerance logic.

lonless9 (approved):

crates/sail-data-source/src/formats/rate/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::formats::rate::options::resolve_rate_read_options;
+pub use crate::formats::rate::options::TableRateOptions;
+pub use crate::formats::rate::reader::RateSourceExec;
+use crate::formats::rate::reader::RateTableProvider;
+
+#[derive(Debug)]
+pub struct RateTableFormat;
+
+#[async_trait]
+impl TableFormat for RateTableFormat {
```
shehabgamin, line 25: nitpick: it would be helpful to add a short comment for each new table format describing what it's for.

crates/sail-data-source/src/formats/socket/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+        if !partition_by.is_empty() {
+            return plan_err!("the socket table format does not support partitioning");
+        }
+        if bucket_by.is_some() || !sort_order.is_empty() {
+            return plan_err!("the socket table format does not support bucketing");
+        }
+        let schema = match schema {
+            Some(schema) if !schema.fields.is_empty() => schema,
+            _ => Schema::new(vec![Arc::new(Field::new("value", DataType::Utf8, false))]),
```
shehabgamin, line 55: `value` should never be null?

linhr: Yes, so `nullable` is `false`.

crates/sail-spark-connect/src/streaming.rs
linhr: When there is an error during stream processing, the job runner needs to schedule another job with the correct checkpoint loaded. So there is more work outside this function. (It may not be possible to retry with the same `SendableRecordBatchStream`.)

crates/sail-data-source/src/formats/rate/reader.rs
linhr: Nice idea! I added a TODO for this.

crates/sail-data-source/src/formats/rate/reader.rs
linhr: I added a TODO for configurable algorithms.

crates/sail-data-source/src/formats/rate/mod.rs
linhr: Done!

PR #758, opened 2025-08-14
chore: code coverage

lonless9:


python/pysail/tests/spark/test_execution.py
```diff
@@ -0,0 +1,237 @@
...
+        spark.catalog.dropTempView("sales")
+
+    def test_multiple_concurrent_operations(self, spark):
+        """Test that multiple operations can run concurrently in local-cluster mode."""
+        df1 = spark.range(100).select(F.col("id").alias("id1"))
+        df2 = spark.range(100).select(F.col("id").alias("id2"))
+
+        result1 = df1.groupBy().sum("id1").collect()[0][0]
+        result2 = df2.groupBy().sum("id2").collect()[0][0]
+        result3 = df1.join(df2, df1.id1 == df2.id2).count()
```
linhr, line 214: These operations do not run concurrently though. Maybe we could call this `test_multiple_operations`. Testing concurrent execution is probably not easy and we can leave it as future work.

.github/actions/setup-rust/action.yml
```diff
@@ -16,10 +16,20 @@ runs:
       run: |
         rustup toolchain install stable
         rustup toolchain install nightly
-        rustup component add --toolchain stable rustfmt clippy
+        rustup component add --toolchain stable rustfmt clippy llvm-tools-preview
```
linhr, line 19: So `llvm-tools-preview` works without requiring explicit LLVM path for `grcov`?

.github/workflows/build.yml
```diff
@@ -74,6 +78,26 @@ jobs:
       - name: Run Cargo Test
         run: cargo test
 
+      # Generate coverage report after tests
+      - name: Generate coverage report
+        # Only run if tests succeed
+        if: success()
```
linhr: I think the default condition is `success()` so we can just remove such `if` conditions.

<https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions>

.github/actions/setup-rust/action.yml
lonless9: Yes there are somehow difference between `llvm-tools` and `llvm-tools-preview`.

python/pysail/tests/spark/test_execution.py
lonless9: Done!

.github/workflows/build.yml

Issue #748, opened 2025-08-12
csv not found if the csv has a gz extensions


djouallah:
this generate an error

```
import pandas as pd
import os
data = [("Alice", 1), ("Bob", 2), ("Charlie", 3)]
columns = ["Name", "ID"]
dummy_df = pd.DataFrame(data, columns=columns)
os.makedirs("/content/test/", exist_ok=True)
dummy_df.to_csv("/content/test/dummy.csv.gz", index=False, compression='gzip')

df = spark.read.format("csv")\
     .option("header","true") \
     .option("compression", "gzip")\
     .load("/content/test")
df.show()
```

```
---------------------------------------------------------------------------
AnalysisException                         Traceback (most recent call last)
[/tmp/ipython-input-1712069872.py](https://localhost:8080/#) in <cell line: 0>()
     11      .option("compression", "gzip")\
     12      .load("/content/test")
---> 13 df.show()

7 frames
[/usr/local/lib/python3.11/dist-packages/pyspark/sql/connect/client/core.py](https://localhost:8080/#) in _handle_rpc_error(self, rpc_error)
   1880                         self._closed = True
   1881 
-> 1882                     raise convert_exception(
   1883                         info,
   1884                         status.message,

AnalysisException: No files found in the specified paths: file:///content/test/


```

linhr:
Thanks for reporting this!

Yeah we do have some issues listing compressed files. We have a few TODOs in our code and we'll have an update here once we implement the related logic.

shehabgamin:
@djouallah We just merged in https://github.com/lakehq/sail/pull/760!

PR #760, opened 2025-08-14
fix: reading and writing compressed files

shehabgamin:
Closes: https://github.com/lakehq/sail/issues/748

### DIFF
--- a/python/pysail/tests/spark/test_datasources.py
+++ b/python/pysail/tests/spark/test_datasources.py
@@ -1,5 +1,6 @@
 import glob
 
+import pandas as pd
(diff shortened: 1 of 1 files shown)


shehabgamin:
I think we can easily fix this if we just rewrite DataFusion's `list_all_files`.  Specifically, this line needs to change:
https://github.com/apache/datafusion/blob/bb2d90ed1543b56deec834d0de8df0a98d068291/datafusion/datasource/src/url.rs#L267

For example, the code needs to instead check for both `.gz.csv` and `.csv.gz`.

cc @linhr

Issue #307, opened 2024-11-22
Spark SQL Functions Coverage and Parity - Map


shehabgamin:
- [x] element_at
- [x] map
- [x] map_concat
- [x] map_contains_key
- [x] map_entries
- [x] map_from_arrays
- [x] map_from_entries
- [x] map_keys
- [x] map_values
- [x] str_to_map
- [x] try_element_at

SparkApplicationMaster:
map_entries is merged a week ago, so may be available in the next datafusion release:
https://github.com/apache/datafusion/commit/0143b20de3f9e25ab8d87ef4b96b78c28e462bee

SparkApplicationMaster:
taking `map_concat` next, should be relatively easy with new unified logic

SparkApplicationMaster:
please assign this to me for history

shehabgamin:
> please assign this to me for history

@SparkApplicationMaster Well done!

The entire map collection is complete thanks to you, what a huge milestone!

P.S. I sent you a DM on Slack :)

PR #777, opened 2025-08-20
feat: map_from_entries, unify all ways to create a map in MapFunction

### LINKED ISSUE #667: MapType keys should be unique
In Arrow and Datafusion MapType is MultiMap
in Spark, MapType is Map (with unique keys)

There are two options (`spark.sql.mapKeyDedupPolicy`):
1) EXCEPTION (default) - don't accept duplicate keys
2) LAST_WINS - insert values by key in order they go in map creation function

https://github.com/apache/spark/blob/cf3a34e19dfcf70e2d679217ff1ba21302212472/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L4961

SparkApplicationMaster:
implement `map_from_entries`
https://spark.apache.org/docs/latest/api/sql/index.html#map_from_entries

unify all ways to create a map in MapFunction:
- `map(k1,v1,k2,v2,..)`  (now done with some additional lightweight expression logic)
- `map_from_arrays`
- `map_from_entries`

remove usage of datafusion `map_udf` because it is not consistent with spark in many ways

all types of keys and values are coerced with datafusion `make_array` logic which works well for spark-tests
no more need to implicitly cast anything

now all maps are deduplicated on creation
and the code can be easily adjusted to change behavior when spark config will be implemented
closes #667
part of #307

### DIFF
--- a/crates/sail-plan/src/function/scalar/map.rs
+++ b/crates/sail-plan/src/function/scalar/map.rs
@@ -1,10 +1,33 @@
 use datafusion::functions_nested::expr_fn;
 use datafusion_expr::expr;
-use datafusion_functions_nested::map::map_udf;
 
+use crate::error::PlanResult;
 use crate::extension::function::map::map_function::MapFunction;
 use crate::extension::function::map::spark_element_at::{SparkElementAt, SparkTryElementAt};
-use crate::function::common::ScalarFunction;
+use crate::function::common::{ScalarFunction, ScalarFunctionInput};
+
+fn map(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
+    use crate::function::common::ScalarFunctionBuilder as F;
+
+    let keys = input
+        .arguments
+        .chunks(2)
+        .map(|key_value| key_value[0].clone())
+        .collect();
+
+    let values = input
+        .arguments
+        .chunks(2)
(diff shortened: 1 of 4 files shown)


crates/sail-plan/src/function/scalar/map.rs
```diff
@@ -1,10 +1,33 @@
...
+    let keys = input
+        .arguments
+        .chunks(2)
+        .map(|key_value| key_value[0].clone())
+        .collect();
+
+    let values = input
+        .arguments
+        .chunks(2)
+        .map(|key_value| key_value[1].clone())
```
linhr, line 21: I guess this will panic if the input has an odd number of arguments. Maybe we should check the length of `input.arguments` at the beginning?

SparkApplicationMaster: Thanks for review!
I'll wrap it into PlanError if number of arguments is incorrect

SparkApplicationMaster: done in this PR:
https://github.com/lakehq/sail/pull/779/files#diff-f037076e248b370392eae8a00c94f944b48a45896b92b872a9703f3efd31432dR12

PR #781, opened 2025-08-22
refactor: split data source options resolver

linhr:
As I work on streaming data sources in #751, I found it's cleaner to have option resolver logic closer to the table format. This PR makes the change for all existing formats. The all-in-one `DataSourceOptionsResolver` is now removed.


crates/sail-data-source/src/formats/delta.rs
```diff
@@ -17,11 +18,12 @@ use sail_delta_lake::delta_format::DeltaDataSink;
...
+use sail_delta_lake::options::TableDeltaOptions;
 use sail_delta_lake::table::open_table_with_object_store;
 use url::Url;
 use uuid::Uuid;
 
-use crate::options::DataSourceOptionsResolver;
+use crate::options::{load_default_options, load_options, DeltaReadOptions, DeltaWriteOptions};
 
 #[derive(Debug, Default)]
 pub struct DeltaTableFormat;
```
linhr, line 29: I haven't split `delta.rs` into multiple files, since there is ongoing work for Delta. But this can be done in future PRs.

lonless9 (approved):
Well done!!
Besides we can slove the Codecov CI problem later.

PR #782, opened 2025-08-22
fix: upload codecov report after tests finished

lonless9:


.github/workflows/spark-tests.yml
```diff
@@ -182,14 +178,12 @@ jobs:
       - name: Record Rust version
         run: echo "RUST=$(rustc --version)" >> "$GITHUB_ENV"
```
linhr: This is no longer needed.

.github/workflows/python-build.yml
```diff
@@ -85,11 +81,9 @@ jobs:
       - name: Record Rust version
         run: echo "RUST=$(rustc --version)" >> "$GITHUB_ENV"
```
linhr: This is no longer needed.

.github/codecov.yml
```diff
@@ -2,35 +2,16 @@ coverage:
...
         target: 0
         informational: true
 
 codecov:
   notify:
-    after_n_builds: 4
+    after_n_builds: 1
 
 comment:
   layout: "files"
```
linhr: Let's try `"condensed_header, condensed_files"` and see how it looks. And we can adjust the layout further in future PRs.

.github/workflows/build.yml
```diff
@@ -85,14 +85,12 @@ jobs:
       - name: Record Rust version
         run: echo "RUST=$(rustc --version)" >> "$GITHUB_ENV"
```
linhr: This needs to be set as a job output and retrieved by `upload-coverage`.

(We cannot run this command directly before coverage upload, since different jobs run on different host and they may have different Rust versions.)

.github/workflows/build.yml
```diff
@@ -145,6 +139,37 @@ jobs:
...
+        with:
+          path: coverage-reports
+          pattern: "*-coverage-report*"
+          merge-multiple: true
+
+      - name: List downloaded files
+        run: |
+          echo "Downloaded coverage reports:"
+          find coverage-reports -type f -name "*.info" || echo "No .info files found"
+          ls -la coverage-reports/ || echo "No coverage-reports directory"
```
linhr: We could store the number of files in the step output and conditionally upload when the count is 4 or more.

.github/workflows/build.yml
```diff
@@ -145,6 +139,37 @@ jobs:
...
 
+  upload-coverage:
+    name: Upload Coverage Reports
+    runs-on: ubuntu-latest
+    needs:
+      - build
+      - python-build
+      - spark-tests
+    steps:
+      - uses: actions/checkout@v4
```
linhr, line 149: We don't need to checkout the repository for this job.

And then it might be good to use absolute path for the coverage reports (e.g. `/tmp/coverage-reports`).

linhr (commented):
🚀

.github/workflows/build.yml
lonless9: The upload action is not only a process of uploading the info file, it also carries the entire repository's file tree, which was a problem that troubled me.

.github/workflows/build.yml

.github/workflows/build.yml
lonless9: I think it could be fine to directly delete this version record, as I've only seen this in a single reference.

.github/codecov.yml
lonless9: Done!

.github/workflows/python-build.yml
lonless9: Done!

.github/workflows/spark-tests.yml
lonless9: Done!

.github/workflows/build.yml
linhr: Oh I see!

PR #785, opened 2025-08-23
fix: `element_at` and `try_element_at` inconsistent error treatment, chore: simplify, unify

### LINKED ISSUE #784: `element_at` and `try_element_at` inconsistent error treatment
### element_at
according to the docs:
element_at(array, index) - 
Returns element of array at given (1-based) index. 
**If Index is 0, Spark will throw an error.** 
If index < 0, accesses elements from the last to the first. 
The function returns NULL if the index exceeds the length of the array and spark.sql.ansi.enabled is set to false. 
If spark.sql.ansi.enabled is set to true, it throws ArrayIndexOutOfBoundsException for invalid indices.
https://spark.apache.org/docs/latest/api/sql/index.html#element_at

```python

server = SparkConnectServer()
server.start()

#spark = SparkSession.builder.remote(f"sc://localhost:{port}").getOrCreate()
spark = SparkSession.builder.appName("test").master("local[4]").getOrCreate()

spark.sql("SELECT element_at(array(1,2,3), 0)").show()
spark.sql("SELECT element_at(array(1,2,3), 100)").show()
```
spark:
```
ArrayIndexOutOfBoundsException: [INVALID_ARRAY_INDEX_IN_ELEMENT_AT] The index 100 is out of bounds. The array has 3 elements. Use `try_element_at` to tolerate accessing element at invalid index and return NULL instead. SQLSTATE: 22003
== SQL (line 1, position 8) ==
SELECT element_at(array(1,2,3), 100)

SparkRuntimeException: [INVALID_INDEX_OF_ZERO] The index 0 is invalid. An index shall be either < 0 or > 0 (the first element has index 1). SQLSTATE: 22003
== SQL (line 1, position 8) ==
SELECT element_at(array(1,2,3), 0)
```
sail 0.3.3:
```
+-------------------------------+
|element_at(array(1, 2, 3), 100)|
+-------------------------------+
|                           NULL|
+-------------------------------+
+-----------------------------+
|element_at(array(1, 2, 3), 0)|
+-----------------------------+
|                         NULL|
+-----------------------------+
```

### try_element_at
`try_element_at` is also not as permissive as it can seem from the naming
it doesn't hide all the errors
it only checks for ArrayOutOfBounds exception and does NULL then

```python
spark.sql("SELECT try_element_at(array(1,2,3), 100)").show()
spark.sql("SELECT try_element_at(array(1,2,3), 0)").show()
spark.sql("SELECT try_element_at(array(1,2,3), 'a')").show()
```
```
+-----------------------------------+
|try_element_at(array(1, 2, 3), 100)|
+-----------------------------------+
|                               NULL|
+-----------------------------------+

SparkRuntimeException: [INVALID_INDEX_OF_ZERO] The index 0 is invalid. An index shall be either < 0 or > 0 (the first element has index 1). SQLSTATE: 22003

AnalysisException: [DATATYPE_MISMATCH.UNEXPECTED_INPUT_TYPE] Cannot resolve "element_at(array(1, 2, 3), a)" due to data type mismatch: The second parameter requires the "INT" type, however "a" has the type "STRING". SQLSTATE: 42K09; line 1 pos 7;
'Project [unresolvedalias(try_element_at(array(1, 2, 3), a))]
```

sail 0.3.3:
```
+-----------------------------------+
|try_element_at(array(1, 2, 3), 100)|
+-----------------------------------+
|                               NULL|
+-----------------------------------+

+---------------------------------+
|try_element_at(array(1, 2, 3), 0)|
+---------------------------------+
|                             NULL|
+---------------------------------+

+---------------------------------+
|try_element_at(array(1, 2, 3), a)|
+---------------------------------+
|                             NULL|
+---------------------------------+
```

I suppose `ArrayIndexOutOfBoundsException` can be correctly raised or not raised only when `spark.sql.ansi.enabled` config can be set and processed well by all the functions, so it won't be in current scope, can only set one value at this time
But other cases can be solved relatively easy

SparkApplicationMaster:
`element_at` and `try_element_at` now act more like in spark:
- `element_at` raises if index is out of bounds (like if spark.sql.ansi.enabled=true)
- `try_element_at` returns NULL in that case
- if index == 0 or types are incorrect, both raise exceptions

simplify both functions by using only datafusion expressions

move both to the collection module because they belong here
gold-tests for them are already in collection

introduce a little udf_expr helper to remove boilerplate when using Sail custom UDFs in other UDFs
this will work when no config or special error treatment is needed
some functions in other modules can reuse it on the next refactoring

closes #784


crates/sail-plan/src/function/common.rs
```diff
@@ -337,3 +337,12 @@ pub(crate) fn get_null_treatment(ignore_nulls: Option<bool>) -> Option<NullTreat
...
 }
+
+pub(crate) fn udf_expr<UDFType: ScalarUDFImpl + Default + 'static>(
+    args: Vec<expr::Expr>,
+) -> expr::Expr {
+    expr::Expr::ScalarFunction(expr::ScalarFunction {
+        func: Arc::new(ScalarUDF::from(UDFType::default())),
+        args,
+    })
+}
```
linhr, line 348: As a side note, we could also do `ScalarUDF::from(UDFType::default()).call(args)`. This is not as succinct as this helpful function though.

SparkApplicationMaster: Oh I just didn't find it surfing through the code of other functions
And there were a lot of `Expr::ScalarFunction(ScalarFunction {` calls which I intended to replace

Thanks for highlihting this! 
Default functionality is better than handwritten
I may provide a refactoring some day with `Expr::ScalarFunction(ScalarFunction {` changed to `scalarUDF::from(UDFType::new()).call(args)` which is nice and short enough

SparkApplicationMaster: refactored this for array and collection in #786

PR #789, opened 2025-08-25
fix: position and locate funcs with startpos arg

SparkApplicationMaster:
`position` and `locate` functions now support third argument 
which is start 1-based position to search substring:
```
> SELECT position('bar', 'foobarbar', 5);
 7
```

https://spark.apache.org/docs/latest/api/sql/index.html#position
https://spark.apache.org/docs/latest/api/sql/index.html#locate

part of #508

### DIFF
--- a/crates/sail-plan/src/function/scalar/string.rs
+++ b/crates/sail-plan/src/function/scalar/string.rs
@@ -6,7 +6,7 @@ use datafusion::functions::expr_fn;
 use datafusion::functions::regex::regexpcount::RegexpCountFunc;
 use datafusion::functions::string::contains::ContainsFunc;
 use datafusion_common::ScalarValue;
-use datafusion_expr::{cast, expr, lit, try_cast, ExprSchemable, ScalarUDF};
+use datafusion_expr::{cast, expr, lit, try_cast, when, ExprSchemable, ScalarUDF};
 
 use crate::error::{PlanError, PlanResult};
 use crate::extension::function::string::levenshtein::Levenshtein;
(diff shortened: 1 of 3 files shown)


crates/sail-plan/src/function/scalar/string.rs
```diff
@@ -133,19 +133,18 @@ fn overlay(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
...
-        return Ok(expr_fn::strpos(str, substr));
-    }
-    if arguments.len() == 3 {
-        // TODO: optional third argument
-        return Err(PlanError::todo(
-            "position with 3 arguments is not supported yet",
-        ));
+fn position(args: Vec<expr::Expr>) -> PlanResult<expr::Expr> {
+    match args.as_slice() {
+        [substr, str] => Ok(expr_fn::strpos(str.clone(), substr.clone())),
```
linhr, line 138: As a side note, in the previous implementation, the weird `if arguments.len() == 2 { ... arguments.two()? }` is meant to avoid clone, since there isn't a clean way to conditionally deconstruct a `Vec` in Rust. But matching on `as_slice()` and then clone is probably fine since the performance is not a big concern here.

linhr (approved):
This looks nice! Thanks!

crates/sail-plan/src/function/scalar/string.rs
SparkApplicationMaster: Thanks for the note!
Refactored with clone avoiding here: 
https://github.com/lakehq/sail/pull/798/files#diff-1ca8d2416f4a5633e7876d4a0a4312439665fd6fbda8670c4f76c4e2d78f264cR68

PR #790, opened 2025-08-26
fix: plan optimization for show string

linhr:
We need to implement the `UserDefinedLogicalNodeCore::necessary_children_exprs()` method for `ShowStringNode` so that the `optimize_projections` logical optimization rule can work properly.


shehabgamin:
@linhr Nice find! Do we need to add this to the other extension nodes that we have?

linhr:
> Do we need to add this to the other extension nodes that we have?

Good call! Done!

PR #810, opened 2025-08-30
chore: set up config for join reorder optimizer

shehabgamin:
https://15721.courses.cs.cmu.edu/spring2020/papers/20-optimizer2/p539-moerkotte.pdf


crates/sail-optimizer/src/error.rs
```diff
@@ -0,0 +1,44 @@
+use sail_common::error::CommonError;
+use thiserror::Error;
+pub type OptimizerResult<T> = Result<T, OptimizerError>;
+
+#[derive(Debug, Error)]
+pub enum OptimizerError {
```
linhr: We don't need the error enum. We don't need to define custom errors for all the crates in Sail. For this crate, using DataFusion error enum would be enough.

crates/sail-optimizer/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+
+use crate::dphyp::DPhyp;
+
+mod dphyp;
+pub mod error;
+
+pub fn get_physical_optimizers() -> Vec<Arc<dyn PhysicalOptimizerRule + Send + Sync>> {
+    vec![
+        // If there is a output requirement of the query, make sure that
```
linhr: I'd prefer not to copy the comments here since they will go outdated over time. We could simply add a test to ensure that our order is in sync with the default order.

crates/sail-optimizer/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+use datafusion::physical_optimizer::output_requirements::OutputRequirements;
+use datafusion::physical_optimizer::projection_pushdown::ProjectionPushdown;
+use datafusion::physical_optimizer::sanity_checker::SanityCheckPlan;
+use datafusion::physical_optimizer::topk_aggregation::TopKAggregation;
+use datafusion::physical_optimizer::update_aggr_exprs::OptimizeAggregateOrder;
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+
+use crate::dphyp::DPhyp;
+
+mod dphyp;
```
linhr: I feel we need to consider both logical and physical optimizers. So I'd suggest the following directory layout.

```
logical/ (not needed for now)
physical/
  mod.rs
  join_reorder/
    mod.rs
lib.rs

File: crates/sail-optimizer/src/dphyp.rs [hunk 4]
```diff
@@ -0,0 +1,40 @@
+use std::sync::Arc;
+
+use datafusion::config::ConfigOptions;
+use datafusion::physical_optimizer::PhysicalOptimizerRule;
+use datafusion::physical_plan::ExecutionPlan;
+
+#[derive(Debug, Clone)]
+pub struct DPhyp {}
```
linhr: We could name this `JoinReorder` to make the rule more obvious. In the comment we can say that it's based on the DPhyp algorithm.

lonless9 (approved):
🚀

linhr:
The work for join reorder algorithm implementation is moved to #917.

PR #813, opened 2025-08-31
fix: data sink exec

shehabgamin:


crates/sail-execution/src/codec.rs
```diff
@@ -1848,11 +1956,28 @@ impl RemoteExecutionCodec {
         Ok(file_compression_type.into())
```
linhr: This can now use `try_decode_compression_type_variant`.

linhr (approved):

crates/sail-execution/src/codec.rs
shehabgamin: done!

PR #815, opened 2025-09-01
fix: Handle harmless OSError in socket test

lonless9:


python/pysail/tests/spark/datasource/test_socket.py
```diff
@@ -38,7 +39,8 @@ def stop(self):
         if self._socket is not None:
             # We need to explicitly shut down the socket to unblock `accept()`,
             # otherwise the thread may hang, and we will wait indefinitely in `join()`.
-            self._socket.shutdown(socket.SHUT_RDWR)
+            with contextlib.suppress(OSError):
```
linhr, line 42: `contextlib` is nice!

linhr (approved):

PR #820, opened 2025-09-02
feat: add `elt` function

davidlghellin:
Part #508

### DIFF
--- a/crates/sail-plan/src/extension/function/string/spark_elt.rs
+++ b/crates/sail-plan/src/extension/function/string/spark_elt.rs
@@ -1,0 +1,271 @@
+use arrow::array::{ArrayRef, Int64Array, Int64Builder, StringArray, StringBuilder, Array, Int32Array, Int32Builder};
+use arrow::datatypes::{DataType};
+use arrow::datatypes::DataType::{Utf8,Int32,Int64};
+use datafusion_common::{exec_err, DataFusionError, Result};
+use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature};
+use datafusion_expr_common::signature::Volatility::Immutable;
+use std::any::Any;
+use std::sync::Arc;
+
+use arrow::compute::cast;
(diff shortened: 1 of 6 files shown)


python/pysail/tests/spark/function/test_elt.txt
```diff
@@ -0,0 +1,34 @@
...
+|           NULL|
++---------------+
+
+>>> spark.sql("SELECT elt(1, 'scala', 3) AS col UNION ALL SELECT elt(2, 'scala', 3)").show()
++-----+
+|  col|
++-----+
+|    3|
+|scala|
++-----+
```
davidlghellin: drop o put order

davidlghellin:
Spark
default in pyspark shell
image

```python
spark.conf.set("spark.sql.ansi.enabled","false")
spark.sql("SELECT elt(111, 10, 20)").show()
```
image

```python
spark.conf.set("spark.sql.ansi.enabled","true")
spark.sql("SELECT elt(111, 10, 20)").show()
```
image

crates/sail-plan/src/extension/function/string/spark_elt.rs
```diff
@@ -0,0 +1,268 @@
...
+                )
+            }
+        };
+
+        let Some(n) = n_opt else {
+            builder.append_null();
+            continue;
+        };
+        // If spark.sql.ansi.enabled is set to true, it throws ArrayIndexOutOfBoundsException for invalid indices.
+        let ansi_enable: bool = false; // I need get value -> spark.sql.ansi.enabled
```
davidlghellin, line 120: I couldn’t find a way to retrieve the value of the configuration `spark.sql.ansi.enabled` from the current context.

linhr: Yeah this is not supported yet. So we have to expose one of the behaviors for now.

linhr (approved):
Thanks for the contribution!

This looks solid and well tested!

Issue #230, opened 2024-09-29
Spark SQL Functions Coverage and Parity - URL


shehabgamin:
- [x] parse_url
- [x] url_decode
- [x] url_encode
- [x] try_parse_url
- [x] try_url_decode

anhvdq:
@shehabgamin Hi, can I get this issue?
I'm implementing these 3 functions

shehabgamin:
@anhvdq Done! Thank you so much!!

anhvdq:
A note on this issue is that the `url_encode`, `url_decode` and `try_url_decode` functions have been merged into datafusion on this PR: https://github.com/apache/datafusion/pull/17399

Waiting for them to be released

PR #824, opened 2025-09-03
feat: add `try_parse_url` function

rafafrdz:
Part of #230 

This PR introduces the implementation of the `try_parse_url` functionality, closely replicating Spark's `try_parse_url` capabilities. This is a special version of `parse_url` that performs the same operation, but returns a `NULL` value instead of raising an error if the parsing cannot be performed.

Spark Ref: [`try_parse_url` doc](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.try_parse_url.html)

- [x] Add support for the `try_parse_url` Spark SQL function  
- [x] Add tests for both `parse_url` and `try_parse_url`
- [x] Fix getting `PATH` in `parse_url` when those cases path is _empty_

### DIFF
--- a/python/pysail/tests/spark/function/test_parse_url.txt
+++ b/python/pysail/tests/spark/function/test_parse_url.txt
@@ -1,69 +1,41 @@
->>> spark.sql("SELECT try_parse_url('https://example.com/a?x=1', 'QUERY', 'x') AS result, typeof(try_parse_url('https://example.com/a?x=1', 'QUERY', 'x')) AS type").show(truncate=False)
+>>> spark.sql("SELECT parse_url('https://example.com/a?x=1', 'QUERY', 'x') AS result, typeof(parse_url('https://example.com/a?x=1', 'QUERY', 'x')) AS type").show(truncate=False)
 +------+------+
 |result|type  |
 +------+------+
 |1     |string|
 +------+------+
 
->>> spark.sql("SELECT try_parse_url('www.example.com/path?x=1', 'HOST') AS result, typeof(try_parse_url('www.example.com/path?x=1', 'HOST')) AS type").show(truncate=False)
-+------+------+
-|result|type  |
-+------+------+
-|NULL  |string|
-+------+------+
-
->>> spark.sql("SELECT try_parse_url('https://example.com/?a=1', 'QUERY', 'b') AS result, typeof(try_parse_url('https://example.com/?a=1', 'QUERY', 'b')) AS type").show(truncate=False)
-+------+------+
-|result|type  |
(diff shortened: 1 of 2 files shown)


crates/sail-plan/src/extension/function/url/spark_try_parse_url.rs
```diff
@@ -0,0 +1,61 @@
...
+        parse_url.coerce_types(arg_types)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let parse_url: ParseUrl = ParseUrl::new();
+        match parse_url.invoke_with_args(args) {
+            Err(_) => Ok(ColumnarValue::Scalar(ScalarValue::Utf8(None))),
+            result => result,
+        }
+    }
```
linhr: This does not work if the input has multiple rows. In general, `try_` functions need to iterate through the input array and return either the value or null depending on whether the function is successful or not, and the result would also be an array.

You can take a look at existing `try_` functions and get an idea how the iteration works.

rafafrdz: You're right! I modified the `parse_url` logic to use a handler error function. Now we can re-use the same implementation, but return `None` when url parser returns errors in the array iteration by using a different handler function than identity

python/pysail/tests/spark/function/test_parse_url.txt
```diff
@@ -0,0 +1,69 @@
+>>> spark.sql("SELECT try_parse_url('https://example.com/a?x=1', 'QUERY', 'x') AS result, typeof(try_parse_url('https://example.com/a?x=1', 'QUERY', 'x')) AS type").show(truncate=False)
```
linhr: This file needs to be named `test_try_parse_url.txt`.

python/pysail/tests/spark/function/test_try_parse_url.txt
```diff
@@ -0,0 +1,42 @@
+>>> spark.sql("SELECT parse_url('https://example.com/a?x=1', 'QUERY', 'x') AS result, typeof(parse_url('https://example.com/a?x=1', 'QUERY', 'x')) AS type").show(truncate=False)
```
linhr: This file needs to be named `test_parse_url.txt`.

linhr (approved):
This looks nice! Thanks for your contribution!

I happened to notice that we need to swap the names of the two `.txt` test files. Other than this minor thing, this PR is ready to go!

PR #825, opened 2025-09-03
fix: rename `array_sort` function

davidlghellin:
https://spark.apache.org/docs/4.0.0/api/python/reference/pyspark.sql/api/pyspark.sql.functions.array_sort.html

### DIFF
--- a/crates/sail-plan/src/function/scalar/array.rs
+++ b/crates/sail-plan/src/function/scalar/array.rs
@@ -31,7 +31,7 @@ fn slice(array: expr::Expr, start: expr::Expr, length: expr::Expr) -> expr::Expr
     expr_fn::array_slice(array, start, end, None)
 }
 
-fn sort_array(array: expr::Expr, asc: expr::Expr) -> PlanResult<expr::Expr> {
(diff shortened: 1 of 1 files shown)


davidlghellin:
hi 
the doc tell `array_sort` but the gold_test said `sort_array`

SparkApplicationMaster:
it's a bit confusing, but there are both functions in Spark SQL, and they have different behaviour https://spark.apache.org/docs/latest/api/sql/index.html#sort_array
`sort_array` does what Sail does now, just sorts array lexicographically, asc or desc
`array_sort` uses lambda function to get sorting key, so it's more complicated
there are some missing parts in spark api docs, this is one of them I suppose

davidlghellin:
> it's a bit confusing, but there are both functions in Spark SQL, and they have different behaviour https://spark.apache.org/docs/latest/api/sql/index.html#sort_array `sort_array` does what Sail does now, just sorts array lexicographically, asc or desc `array_sort` uses lambda function to get sorting key, so it's more complicated there are some missing parts in spark api docs, this is one of them I suppose

I understand now well
thanks

PR #826, opened 2025-09-03
feat: add `luhn_check` function

rafafrdz:
Part of #508 

This PR introduces the implementation of the `luhn_check` functionality, closely replicating Spark's `luhn_check` capabilities. If the num value contains any character that is not a digit, the result is `false`. If the input's last digit matches the algorithm, the result is `true`.

Spark Ref: [`luhn_check` doc](https://docs.databricks.com/gcp/en/sql/language-manual/functions/luhn_check)

- [ ] Add support for the `luhn_check` Spark SQL function

### DIFF
--- a/crates/sail-plan/src/extension/function/string/mod.rs
+++ b/crates/sail-plan/src/extension/function/string/mod.rs
@@ -7,3 +7,4 @@ pub mod spark_split;
 pub mod spark_to_binary;
 pub mod spark_to_number;
 pub mod spark_try_to_number;
+pub mod spark_luhn_check;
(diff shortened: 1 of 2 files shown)


linhr:
Thanks for the contribution!

DataFusion recently added this function (<https://github.com/apache/datafusion/pull/16848>), which will be available once DataFusion 50 is released. Would you like to take a look and see if covers the desired functionality?

(We haven't started the process of moving all Spark functions upstream yet, so contributions to Sail are still welcomed and much appreciated!)

rafafrdz:
> Thanks for the contribution!
> 
> 
> 
> DataFusion recently added this function (<https://github.com/apache/datafusion/pull/16848>), which will be available once DataFusion 50 is released. Would you like to take a look and see if covers the desired functionality?
> 
> 
> 
> (We haven't started the process of moving all Spark functions upstream yet, so contributions to Sail are still welcomed and much appreciated!)

Oh! it seems so.. 😅 I close this PR then

PR #832, opened 2025-09-04
feat: set up streaming events

linhr:


crates/sail-common-datafusion/src/streaming/event/marker.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// will be emitted in the future.
+    /// Downstream operators can use the watermark to discard state
+    /// that is no longer needed for processing late data.
+    Watermark {
+        source: String,
+        timestamp: DateTime<Utc>,
+    },
+    /// A checkpoint marker emitted by each data source to start a new
+    /// checkpoint.
+    Checkpoint { id: u64 },
```
lonless9, line 30: Is it reasonable that we can later introduce states like `Start`, `NotifyCommit`, `Abort`, etc., and differentiate between sync and async checkpoints?

crates/sail-common-datafusion/src/streaming/event/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Debug, Clone)]
+pub enum FlowEvent {
+    /// A retractable batch of data.
+    Data {
+        /// The data batch.
+        batch: RecordBatch,
+        /// A boolean array indicating whether each row is retracted.
+        /// The array has the same length as the data batch.
+        retracted: BooleanArray,
+    },
```
lonless9, line 23: Would it help to include an optional `DataMetadata` field?

crates/sail-common-datafusion/src/streaming/event/marker.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// An indicator that the flow event stream will contain no more data events.
+    /// A bounded source cannot close the stream to indicate end of data, since
+    /// it is supposed to still wait for another checkpoint marker before finishing.
+    /// When a sink receives this marker from all its input, it should wait for
+    /// the next checkpoint marker and then the job can finish.
+    EndOfData,
```
lonless9, line 36: Do we need to keep idle sources alive to unblock the advance of watermark? For example:
`IdleSource { source: String, duration: Duration }`

crates/sail-common-datafusion/src/streaming/event/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// A marker injected in the data flow.
+    Marker(marker::FlowMarker),
```
lonless9, line 25: Does it make sense that we could add a field here in the future to handle schema evolution?

crates/sail-common-datafusion/src/array/placeholder.rs
```diff
@@ -0,0 +1,606 @@
...
+    TimestampMicrosecondType, TimestampMillisecondType, TimestampNanosecondType,
+    TimestampSecondType, UInt16Type, UInt32Type, UInt64Type, UInt8Type, UnionMode,
+};
+use datafusion_common::internal_err;
+
+/// Create a non-null placeholder array of the specified data type and length.
+pub fn placeholder_array(
+    data_type: &DataType,
+    length: usize,
+) -> datafusion_common::Result<ArrayRef> {
```
lonless9, line 26: Maybe we can do some cache operations on the placeholder of `Maker`, for example.
`DashMap<DataType, ArrayRef>`
I don't have a good idea yet.

crates/sail-common-datafusion/src/streaming/event/schema.rs
```diff
@@ -0,0 +1,74 @@
+//! Utilities for working with the schema of encoded flow event streams.
+//! A flow event stream can contain both data or markers, and the flow event stream
+//! can be encoded as a stream of Arrow record batches with additional fields
+//! that carry the information about the marker and data retraction.
```
lonless9, line 4: Perhaps we can package the methods in this file into a struct like this
`struct FlowSchema(& Schema);`
`impl FlowSchema {...}`

crates/sail-common-datafusion/src/streaming/event/marker.rs
```diff
@@ -0,0 +1,151 @@
...
+    },
+    /// A watermark marker emitted by a time-based source to indicate
+    /// that no data with a timestamp older than the watermark timestamp
+    /// will be emitted in the future.
+    /// Downstream operators can use the watermark to discard state
+    /// that is no longer needed for processing late data.
+    Watermark {
+        source: String,
+        timestamp: DateTime<Utc>,
+    },
```
lonless9, line 27: How should watermarks be aligned with multiple input streams like `UnionExec`?

lonless9 (approved):
🚀Awesome!
I've left some comments, mainly about the design issue, which should have no effect on the current set up, and some of these issues may be unconsidered.

crates/sail-common-datafusion/src/streaming/event/marker.rs
linhr: Yeah we'll need to expand the marker for various checkpoint status. Right now the marker is more of a placeholder and it does not distinguish e.g. the start/completion of checkpoints.

crates/sail-common-datafusion/src/streaming/event/mod.rs
linhr: I feel the schema metadata of `RecordBatch` can serve the purpose to some extent.

crates/sail-common-datafusion/src/streaming/event/marker.rs
linhr: Yeah Flink has this concept. We will need to figure out this when we implement the actual watermark logic.

crates/sail-common-datafusion/src/streaming/event/mod.rs
linhr: I feel the marker can be extended to pass schema change events to operators and data sinks.

crates/sail-common-datafusion/src/array/placeholder.rs
linhr: A global map could be used to achieve this. We could consider this when we find performance benefit of caching in this case.

crates/sail-common-datafusion/src/streaming/event/schema.rs
linhr: Yeah the "newtype" pattern seems viable. What I found is that I still need to work with Arrow schema directly in most situations when integrating with DataFusion, so the wrapped type does not provide type safety across function calls. So I simply have all the logics in functions.

crates/sail-common-datafusion/src/streaming/event/marker.rs
linhr: Good question! Flink has certain strategies for this case (e.g. taking the minimum of the values). We will need to take care of this when implementing the watermark logic.

linhr:
@lonless9 Thanks for the review! These are very good design questions!

PR #833, opened 2025-09-04
feat: kustomize + support pod template for Sail workers

shehabgamin:
- Kustomize setup for local testing.
- Supporting pod template for Sail workers by merging pod template spec.
- Fixed specifying `path`/`location` option when writing table.

### DIFF
--- a/python/pysail/tests/spark/test_write_table.py
+++ b/python/pysail/tests/spark/test_write_table.py
@@ -18,207 +18,214 @@ def tables(spark, tmp_path):
     spark.sql("DROP TABLE IF EXISTS t5")
 
 
-class TestWrite:
-    def test_insert_into_with_values(self, spark):
-        spark.sql("INSERT INTO t1 VALUES (101, 'Alice', 20)")
-        df = spark.createDataFrame([(102, "Bob", 30)])
-        df.write.insertInto("t1")
+def test_insert_into_with_values(spark):
+    spark.sql("INSERT INTO t1 VALUES (101, 'Alice', 20)")
+    df = spark.createDataFrame([(102, "Bob", 30)])
+    df.write.insertInto("t1")
 
-        actual = spark.sql("SELECT * FROM t1 WHERE id IN (101, 102) ORDER BY id").toPandas()
-        expected = pd.DataFrame(
-            {"id": [101, 102], "name": ["Alice", "Bob"], "age": [20, 30]},
+    actual = spark.sql("SELECT * FROM t1 WHERE id IN (101, 102) ORDER BY id").toPandas()
+    expected = pd.DataFrame(
+        {"id": [101, 102], "name": ["Alice", "Bob"], "age": [20, 30]},
+    )
+    assert_frame_equal(actual, expected)
+
+
+def test_insert_into_by_position(spark):
+    # The field names are ignored.
+    spark.sql("INSERT INTO t1 SELECT 201 AS age, 'Alice' AS name, 21 AS id")
+    df = spark.createDataFrame([(202, "Bob", 31)], schema=["age", "name", "id"])
+    df.write.insertInto("t1")
+
+    actual = spark.sql("SELECT * FROM t1 WHERE id IN (201, 202) ORDER BY id").toPandas()
+    expected = pd.DataFrame(
+        {"id": [201, 202], "name": ["Alice", "Bob"], "age": [21, 31]},
+    )
+    assert_frame_equal(actual, expected)
+
+
+def test_insert_into_by_name(spark):
+    spark.sql("INSERT INTO t1 (age, name, id) VALUES (21, 'Alice', 301)")
+    spark.sql("INSERT INTO t1 BY NAME SELECT 31 AS age, 'Bob' AS name, 302 AS id")
+
+    actual = spark.sql("SELECT * FROM t1 WHERE id IN (301, 302) ORDER BY id").toPandas()
+    expected = pd.DataFrame(
+        {"id": [301, 302], "name": ["Alice", "Bob"], "age": [21, 31]},
+    )
+    assert_frame_equal(actual, expected)
+
+
+def test_insert_into_with_invalid_options(spark):
+    df = spark.createDataFrame([(101, "Alice", 20)])
+
+    with pytest.raises(Exception, match="partition"):
+        df.write.partitionBy("id").insertInto("t1")
+
+
+@pytest.mark.skipif(not is_jvm_spark(), reason="`INSERT OVERWRITE` is not supported in Sail yet")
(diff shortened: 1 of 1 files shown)


k8s/README.md
```diff
@@ -0,0 +1,32 @@
+# Kubernetes Deployment
```
linhr: This should be part of our docs. The README file is less discoverable.

crates/sail-execution/src/worker_manager/kubernetes.rs
```diff
@@ -209,27 +210,35 @@ impl WorkerManager for KubernetesWorkerManager {
...
+                image_pull_policy: Some(self.options.image_pull_policy.clone()),
+                ..Default::default()
+            }],
+            restart_policy: Some("Never".to_string()),
+            service_account_name: Some(self.options.worker_service_account_name.clone()),
+            ..Default::default()
+        };
+        let template: PodTemplateSpec = serde_json::from_str(&self.options.worker_pod_template)?;
+        let spec = if let Some(mut s) = template.spec {
+            s.merge_from(spec);
```
linhr: I realized that the merge should be swapped. For example, we want the user to override the pod spec with their image, rather than the opposite.

k8s/local-volume-patch.yaml
```diff
@@ -0,0 +1,44 @@
...
+            path: /tmp/sail
+            type: DirectoryOrCreate
+      containers:
+        - name: server
+          volumeMounts:
+            - name: data
+              mountPath: /tmp/sail
+          env:
+            - name: SAIL_KUBERNETES__WORKER_POD_TEMPLATE
+              value: |
```
linhr: The environment variable value is parsed with TOML syntax, so to be safe we should use multi-line literal string (triple single-quote) here to avoid any escape sequence inside the JSON string to be interpreted by the config parser.

```
'''
{
  "spec": {}
}
'''
```

linhr: https://toml.io/en/v1.0.0#string

shehabgamin: Nice catch!

shehabgamin: Can't get this to work with `'''`, I keep getting a parsing error.

python/pysail/tests/conftest.py
```diff
@@ -14,5 +17,6 @@ def pytest_configure(config):
     # of pytest.
     # In common cases, plugins only access the configuration for the first time after this hook,
     # so the cache is not a problem.
-
     config.inicfg["doctest_optionflags"] = "ELLIPSIS NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL"
+    # In macOS, this ends up being /private/tmp/sail
+    os.environ["PYTEST_DEBUG_TEMPROOT"] = "/tmp/sail"  # noqa: S108
```
linhr: Let's revert this since this doesn't work on Windows so the release workflow would fail. In general, we'd like the user to run Sail Python tests in their installation, and the user may have different system settings or preferences for the temporary directory, so we cannot enforce a value here.

We can add a note in the docs so that the user knows how to change the temporary directory via the environment variable if they want.

k8s/test-volume-patch.yaml
```diff
@@ -0,0 +1,61 @@
...
+  template:
+    spec:
+      volumes:
+        - name: data
+          hostPath:
+            path: /tmp/sail
+            type: DirectoryOrCreate
+        - name: pytest-data
+          hostPath:
+            path: /private/tmp/sail
```
linhr, line 15: I'm not sure if `/private` works on Linux for pytest. This is not a blocker for this PR though.

python/pysail/tests/conftest.py
linhr: Actually this also causes CI to fail.

python/pysail/tests/spark/test_write_table.py
```diff
@@ -17,212 +18,207 @@ def tables(spark):
...
-    spark.sql("INSERT INTO t1 VALUES (101, 'Alice', 20)")
-    df = spark.createDataFrame([(102, "Bob", 30)])
-    df.write.insertInto("t1")
-
-    actual = spark.sql("SELECT * FROM t1 WHERE id IN (101, 102) ORDER BY id").toPandas()
-    expected = pd.DataFrame(
-        {"id": [101, 102], "name": ["Alice", "Bob"], "age": [20, 30]},
-    )
-    assert_frame_equal(actual, expected)
+class TestWrite:
```
linhr: Ahh I actually wanted to discourage the use of `class` for tests. (We have this style for Delta but I haven't suggested to make the change yet.) We already group tests by files (Python modules). The `class` style could be due to the `unittest` framework but it adds unnecessary indentation for pytest.

If possible, would you mind reverting the style change here? (For this PR this would make the diff much easier to read.)

k8s/local-volume-patch.yaml
linhr: Interesting. Maybe the config parser does not always consider the TOML syntax then. We can figure it out later.

python/pysail/tests/spark/test_write_table.py
```diff
@@ -17,212 +18,207 @@ def tables(spark):
...
+            )
 
-    df = spark.createDataFrame([(404, "Bob", 35)])
-    df.write.mode("overwrite").insertInto("t1")
-    actual = spark.sql("SELECT * FROM t1").toPandas()
-    expected = pd.DataFrame(
-        {"id": [404], "name": ["Bob"], "age": [35]},
-    )
-    assert_frame_equal(actual, expected)
+        df.write.option("path", location).saveAsTable("t2")
```
linhr: ~~Let's remove `.option("path", location)`. This is ignored right now, and in the future we may return an error since the user is not supposed to specify the location when writing to the table. (The location is already specified when creating the table.)~~

We need to consider how conflicting location is handled when writing to an existing table.

k8s/test-volume-patch.yaml
shehabgamin: Yeah I don't think it does, this is a macOS specific thing. `PYTEST_DEBUG_TEMPROOT` is supposed to override the root directory ([see here)](https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention), but with macOS it doesn't do that for some reason.  So, I'm not sure if the `pytest-data` mount is even needed for Linux but that is easy to verify, I just haven't done it.

python/pysail/tests/spark/test_write_table.py
shehabgamin: Done!

linhr (approved):

PR #834, opened 2025-09-04
feat: support pod template for Sail workers

linhr:


linhr:
The idea has been brought into #833 and the work will continue there.

PR #682, opened 2025-07-31
feat: add `make_interval` function

davidlghellin:
This is part of https://github.com/lakehq/sail/issues/505

and later try solved #679

PR #835, opened 2025-09-04
feat: add `make_dt_interval` function

davidlghellin:
Part #505 
~~Reuses code of make_interval #682~~

https://spark.apache.org/docs/3.5.0/api/python/reference/pyspark.sql/api/pyspark.sql.functions.make_dt_interval.html
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.make_dt_interval.html


crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,182 @@
...
+    fn name(&self) -> &str {
+        "make_dt_interval"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Interval(MonthDayNano))
```
linhr: Working with interval types is tricky. Here is a table for the expected return types for various functions.

| Function | Spark return type | Arrow return type |
| :-- | :-- | :-- |
| `make_interval` | `interval` | `DataType::Interval(IntervalUnit::MonthDayNano)` |
| `make_dt_interval` | `interval day to second` | `DataType::Duration(TimeUnit::Microsecond)` |
| `make_ym_interval` | `interval year to month` | `DataType::Interval(IntervalUnit::YearMonth)` |

(Note: In Arrow, there is also `DataType::Interval(IntervalUnit::DayTime)`, but it does not map cleanly to the Spark `interval day to second` time.)

So I think we cannot directly use `make_interval_month_day_nano` and need a separate function for the expected return type.

Would you mind adding a few doctests to ensure both the value and the type are working as expected?

davidlghellin: Of course, I’ll add that as well and investigate it further — thanks a lot for pointing it out! 🙏

I also thought it seemed a bit too easy, haha 😅

davidlghellin:
🤦‍♂️ one more time 🤦‍♂️  sorry

crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,400 @@
...
+            }
+        }
+    }
+
+    Ok(Arc::new(builder.finish()))
+}
+pub fn make_interval_dt_nano(day: i32, hour: i32, min: i32, sec: f64) -> Result<Option<i64>> {
+    if !sec.is_finite() {
+        return Ok(None);
+    }
```
davidlghellin: drop, checked any_null_present

crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,400 @@
...
+                    DataType::Int32
+                }
+            })
+            .collect())
+    }
+}
+
+fn make_dt_interval_kernel(args: &[ArrayRef]) -> Result<ArrayRef, DataFusionError> {
+    // 0 args is in invoke_with_args
+    if args.is_empty() || args.len() > 4 {
```
davidlghellin:
```suggestion
    if args.len() > 4 {
```

crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,413 @@
...
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        if args.args.is_empty() {
+            let n: usize = std::cmp::max(args.number_rows, 1);
+            let mut b: PrimitiveBuilder<DurationMicrosecondType> =
+                DurationMicrosecondBuilder::with_capacity(n);
+            for _ in 0..n {
+                b.append_value(0_i64);
+            }
+            return Ok(ColumnarValue::Array(Arc::new(b.finish())));
+        }
```
linhr: We need to return an empty array if `args.number_rows` is `0`.

Alternatively, you could just return `ColumnarValue::Scalar(ScalarValue::DurationMicrosecond(Some(0)))` since the value will be the same across rows.

crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,413 @@
...
+    if args.len() > 4 {
+        return Err(invalid_arg_count_exec_err(
+            "make_dt_interval",
+            (0, 4),
+            args.len(),
+        ));
+    }
+
+    let n_rows = args[0].len();
+    debug_assert!(args.iter().all(|a| a.len() == n_rows));
```
linhr: We should not use `debug_assert!` since the code is not supposed to panic when the user provides invalid data. We should return `Err` in this case.

crates/sail-plan/src/extension/function/datetime/spark_make_dt_interval.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    let days = args.first().map(|a| a.as_primitive::<Int32Type>());
+    let hours = args.get(1).map(|a| a.as_primitive::<Int32Type>());
+    let mins = args.get(2).map(|a| a.as_primitive::<Int32Type>());
+    let secs = args.get(3).map(|a| a.as_primitive::<Float64Type>());
```
linhr: `as_primitive` will panic if the data type mismatches. We can use `as_primitive_opt` and returns an error if the result is `None`.

Here is an example:

```rust
let days = args.first().map(|a| a.as_primitive_opt::<Int32Type>().ok_or_else(|| plan_datafusion_err!("..."))).transpose()?;
```

davidlghellin: image

I see, thanks

image

linhr (approved):
This looks nice! Thank you for the contribution!

PR #837, opened 2025-09-05
fix: writing Delta table with partitioned inputs

### LINKED ISSUE #836: bug: DeltaWriterExec requires exactly one input partition, got 32
I'm using Sail 0.3.4 and something as simple as the below fails when the folder being referenced has multiple parquet files:

```python
spark.read.parquet('.../sf10_parquet/catalog_returns/') \
            .write.format("delta") \
            .mode("overwrite") \
            .save(".../sail_tpcds_sf10/catalog_returns")
```

`DeltaWriterExec requires exactly one input partition, got 32`

If there's a single file in the parquet folder it works fine. Pretty sure this wasn't an issue in 0.3.3. thx!

linhr:
Fixes #836.

### DIFF
--- a/crates/sail-delta-lake/src/delta_format/plan_builder.rs
+++ b/crates/sail-delta-lake/src/delta_format/plan_builder.rs
@@ -1,7 +1,6 @@
 use std::sync::Arc;
 
 use datafusion::physical_expr::LexRequirement;
-use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion_common::Result;
 use sail_common_datafusion::datasource::PhysicalSinkMode;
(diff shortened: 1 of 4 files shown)


crates/sail-delta-lake/src/delta_format/plan_builder.rs
```diff
@@ -45,27 +44,14 @@ impl DeltaPlanBuilder {
...
-        let current_plan = self.add_projection_node(current_plan)?;
-
-        // 2. Repartition Node
-        let current_plan = self.add_repartition_node(current_plan)?;
-
-        // 3. Sort Node
-        let current_plan = self.add_sort_node(current_plan)?;
-
-        // 4. Coalesce Partitions Node
-        let current_plan = self.add_coalesce_partitions_node(current_plan)?;
```
linhr, line 60: I guess explicitly adding the coalesce node may not work since the optimizer could decide to remove it. So we have to define the input distribution requirement in `ExecutionPlan`.

PR #845, opened 2025-09-07
fix: robots meta tag for docs

shehabgamin:
We would like search engines to index everything under `/latest`

Also, hoping to fix the fact that Claude will not visit `docs.lakesail.com`:
```
Unfortunately, I cannot access docs.lakesail.com because the site's robots.txt file blocks automated access.
```


linhr:
Search engine is not supposed to index the dev guide from `latest`. The dev guide needs to come from `main`.

PR #848, opened 2025-09-08
refactor: handle side-effect of replaceWhere for Delta Lake

### LINKED ISSUE #711: Ensure no side-effect for Delta table operation planning
If we ever want to support something like `EXPLAIN INSERT ... REPLACE WHERE ...`, we need to make sure no data is written when creating the logical/physical plan for the table operation.

In general, we need to ensure that the plan resolver and physical planner have no side-effect to external systems.

lonless9:
closes #711


linhr (approved):
Great work! 🚀

PR #849, opened 2025-09-08
fix: session key without user ID

### LINKED ISSUE #844: bug: division by decimal zero raises generic `IllegalArgumentException` instead of `ArithmeticException: divide by zero`
Hello,

Division by **integer** zero raises meaningful exception
```python
>>> spark.sql("SELECT 1/0").collect()
ArithmeticException: divide by zero
```

Whereas division by **decimal** zero raises generic incomprehensible `IllegalArgumentException`
```python
>>> spark.sql("SELECT 1/0.0").collect()
IllegalArgumentException: invalid argument: operation not found: af0960ca-9a6f-4703-bbf7-53ddbcb31dd6
```

linhr:
Fixes #844.

If the `USER` or `SPARK_USER` environment variables are not set, the PySpark client does not consistently send the same user context to the server across gRPC calls. For some endpoints, the client sends `None`, while for other endpoints, the client sends `Some(UserContext { user_id: "", user_name: "", extensions: [] })`. Currently, the server generates different session keys for these two cases, so two sessions are created, resulting in the confusing "operation not found" error for some operations.

The "operation not found" error is observed on Windows since the `USER` environment variable is not set. I can also reproduce this on macOS after running `unset USER` in the shell.

This PR fixes the issue by handling missing or empty user ID in the same way.

### DIFF
--- a/crates/sail-plan/src/config.rs
+++ b/crates/sail-plan/src/config.rs
@@ -54,7 +54,7 @@ impl Default for PlanConfig {
             pyspark_udf_config: Arc::new(PySparkUdfConfig::default()),
             default_table_file_format: "PARQUET".to_string(),
             default_warehouse_directory: "spark-warehouse".to_string(),
-            session_user_id: "sail".to_string(),
+            session_user_id: "".to_string(),
             ansi_mode: false,
         }
     }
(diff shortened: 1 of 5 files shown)


lonless9 (approved):
Great!

PR #853, opened 2025-09-10
feat: binary file reader

shehabgamin:
Support any modality

### DIFF
--- a/crates/sail-data-source/src/formats/binary/file_format.rs
+++ b/crates/sail-data-source/src/formats/binary/file_format.rs
@@ -17,20 +17,23 @@ use datafusion_datasource::source::DataSourceExec;
 use object_store::{ObjectMeta, ObjectStore};
 
 use crate::formats::binary::source::BinarySource;
-use crate::formats::binary::DEFAULT_BINARY_EXTENSION;
+use crate::formats::binary::TableBinaryOptions;
 
 #[derive(Debug)]
-pub struct BinaryFileFormat;
-
-impl BinaryFileFormat {
-    pub fn new() -> Self {
-        Self
-    }
+pub struct BinaryFileFormat {
+    options: TableBinaryOptions,
 }
 
-impl Default for BinaryFileFormat {
-    fn default() -> Self {
-        Self::new()
+impl BinaryFileFormat {
+    pub fn new(table_binary_options: TableBinaryOptions) -> Self {
+        Self {
+            options: table_binary_options,
+        }
+    }
+
+    #[allow(unused)]
+    pub fn options(&self) -> &TableBinaryOptions {
+        &self.options
     }
 }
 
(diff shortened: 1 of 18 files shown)


crates/sail-data-source/src/formats/binary/file_format.rs
```diff
@@ -0,0 +1,120 @@
...
+        &self,
+        _state: &dyn Session,
+        _store: &Arc<dyn ObjectStore>,
+        _objects: &[ObjectMeta],
+    ) -> Result<SchemaRef> {
+        let schema = SchemaRef::new(Schema::new(vec![
+            Field::new("path", DataType::Utf8, false),
+            Field::new(
+                "modificationTime",
+                DataType::Timestamp(datafusion::arrow::datatypes::TimeUnit::Microsecond, None),
```
linhr: This is `timestamp` instead of `timestamp_ntz`, so we need an arbitrary time zone here (e.g. `Some(Arc::from("UTC"))`.

docs/introduction/getting-started/index.md
```diff
@@ -81,7 +81,21 @@ _, port = server.listening_address
...
+spark.read.format("binaryFile").option("pathGlobFilter", "*.png").load("/path/to/data").show()
+
+# Use Spark SQL to create a table that refers to multimodal data in Object Storage
+sql_query = """
+CREATE TABLE pdfs
+USING binaryFile
+OPTIONS (pathGlobFilter '*.pdf')
+LOCATION 's3://my-bucket/path/to/data'
+"""
+spark.sql(sql_query)
```
linhr:
```suggestion
# Use Spark SQL to create a table that refers to multimodal data in an object storage
spark.sql("""
CREATE TABLE pdfs
USING binaryFile
OPTIONS (pathGlobFilter '*.pdf')
LOCATION 's3://my-bucket/path/to/data'
""")
```

docs/guide/formats/index.md
```diff
@@ -18,7 +18,7 @@ Here is a summary of the supported (:white_check_mark:) and unsupported (:x:) da
 | JSON                  | :white_check_mark: | :white_check_mark: |
 | Parquet               | :white_check_mark: | :white_check_mark: |
 | Text                  | :white_check_mark: | :white_check_mark: |
-| Binary                | :construction:     | :construction:     |
+| Binary                | :white_check_mark: | :construction:     |
```
linhr: I guess writing binary is not a meaningful operation, so we probably won't have a plan to support that.

```suggestion
| Binary                | :white_check_mark: | :x:                |
```

crates/sail-data-source/src/formats/binary/file_format.rs
linhr: Let me fix this by using the session time zone.

Issue #507, opened 2025-05-28
Spark SQL Functions Coverage and Parity - Misc


linhr:
- [x] aes_decrypt
- [x] aes_encrypt
- [x] assert_true
- [ ] bitmap_bit_position
- [ ] bitmap_bucket_number
- [ ] bitmap_count
- [x] current_catalog
- [x] current_database
- [x] current_schema
- [x] current_user
- [x] equal_null
- [ ] hll_sketch_estimate
- [ ] hll_union
- [ ] input_file_block_length
- [ ] input_file_block_start
- [ ] input_file_name
- [ ] java_method
- [ ] monotonically_increasing_id
- [x] raise_error
- [ ] reflect
- [ ] spark_partition_id
- [x] try_aes_encrypt
- [x] try_aes_decrypt
- [x] typeof
- [ ] user
- [x] uuid
- [ ] version

keen85:
@linhr, this list suggests that `java_method` and `reflect` would be something that you are working on and that might be supported someday in the future.

But those are JVM specific and these will never work in Sail, right? Maybe remove those from the list or highlight that these as some "known limitation"?

linhr:
Thanks for pointing it out! I've updated the issue accordingly.

PR #855, opened 2025-09-10
feat: add `version` function

rafafrdz:
Part of #507 

This PR introduces the implementation of the `version` functionality.

The original (PySpark) `version` function returns the PySpark version used in that project. In this case, I'm assuming the `version` functions should return the Sail version.

Spark Ref: [`version` doc](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.version.html?highlight=version#pyspark.sql.functions.version)

- [x] Add support for the `version` Spark SQL function  

**Disclaimer**: I'm not defining any tests, since the query results will vary with each commit and every Sail release.

### DIFF
--- a/crates/sail-plan/src/extension/function/misc/version.rs
+++ b/crates/sail-plan/src/extension/function/misc/version.rs
@@ -55,5 +55,5 @@ impl ScalarUDFImpl for SparkVersion {
 }
 
 fn spark_version(_args: &[ArrayRef]) -> Result<ArrayRef> {
-    Ok(Arc::new(StringArray::from(vec![Some( SparkVersion::SAIL_VERSION)])) as ArrayRef)
+    Ok(Arc::new(StringArray::from(vec![Some(SparkVersion::SAIL_VERSION)])) as ArrayRef)
 }


rafafrdz:
> Thanks for the contribution! I have a few concerns with this change.
> 
>    I'm not sure if the `version` function should return the Sail version. It may be helpful if we return the PySpark version in the current Python environment where Sail runs.

Got it! I'll try to check if there is a way to get PySpark version

>    I'm concerned about the dependency on the `git-version` crate. This assumes that the `git` binary is available, which is not always the case. For example, this does not work when the user builds the Python wheel from the packaged Sail source code distributed on PyPI.

mmm I see.. So, I'll try to check it too. Can I assume that I have to return the git revision for the last Spark release as well? or the last Sail release? @linhr

linhr:
Getting PySpark version can be done via PyO3. But moving forward we're planning to move all functions to a separate crate, and the functions crate will not depend on `PyO3` (to improve build time), so it will not be straightforward to get the PySpark version. We can revisit this in the future.

Maybe for now we only return Sail version without the Git commit hash. We don't have to follow the exact format as the Spark `version()` function since this function is for information only. I feel returning the Sail version obtained from `CARGO_PKG_VERSION` is good enough.

rafafrdz:
> Maybe for now we only return Sail version without the Git commit hash. We don't have to follow the exact format as the Spark `version()` function since this function is for information only. I feel returning the Sail version obtained from `CARGO_PKG_VERSION` is good enough.

YES! I think the same hahaha alright, I'll apply the changes

PR #856, opened 2025-09-10
feat: delete operation for Delta Lake

### LINKED ISSUE #802: add support for delete from tbl
when using delta, delete from tbl where field = 'something" does not work, will be nice to support it :)

lonless9:
closes #802 

1. Added delete operation support for Delta Lake.
2. Refactored the physical plan path related to `find files`.
3. Fixed an issue where predicates were not being correctly written to the log.

### DIFF
--- a/crates/sail-delta-lake/src/delta_format/delete_exec.rs
+++ b/crates/sail-delta-lake/src/delta_format/delete_exec.rs
@@ -151,6 +151,7 @@ impl DeltaDeleteExec {
             .build();
         let adapter_factory = Arc::new(DeltaPhysicalExprAdapterFactory {});
 
+        // TODO: Turn `prepare_predicate_actions_physical` into an `ExecutionPlan` node.
         let (predicate_actions, _cdf_df) = prepare_predicate_actions_physical(
             self.condition.clone(),
             table.log_store(),
--- a/crates/sail-physical-plan/src/file_delete.rs
+++ b/crates/sail-physical-plan/src/file_delete.rs
@@ -1,10 +1,10 @@
 use std::sync::Arc;
 
+use datafusion::common::DFSchemaRef;
 use datafusion::execution::SessionState;
 use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_planner::PhysicalPlanner;
 use datafusion_common::Result;
-use datafusion_expr::LogicalPlan;
 use sail_common_datafusion::datasource::DeleteInfo;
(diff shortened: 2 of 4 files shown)


crates/sail-physical-plan/src/file_delete.rs
```diff
@@ -0,0 +1,45 @@
...
+use datafusion_common::Result;
+use datafusion_expr::LogicalPlan;
+use sail_common_datafusion::datasource::DeleteInfo;
+use sail_data_source::default_registry;
+use sail_logical_plan::file_delete::FileDeleteOptions;
+
+pub async fn create_file_delete_physical_plan(
+    ctx: &SessionState,
+    planner: &dyn PhysicalPlanner,
+    logical_plan: &LogicalPlan,
```
linhr: I feel we could just pass `schema` as an argument here?

crates/sail-delta-lake/src/delta_format/delete_exec.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        // File-level scan: need to rewrite files with filtered data
+        let writer_stats_config = WriterStatsConfig::new(32, None);
+        let partition_columns = snapshot.metadata().partition_columns().clone();
+        let session_state = SessionStateBuilder::new()
+            .with_runtime_env(context.runtime_env().clone())
+            .build();
+        let adapter_factory = Arc::new(DeltaPhysicalExprAdapterFactory {});
+
+        let (predicate_actions, _cdf_df) = prepare_predicate_actions_physical(
```
linhr, line 155: `prepare_predicate_actions_physical` calls `execute_non_empty_expr_physical`, which internally constructs a DataFusion logical/physical plan and executes all partitions. This needs to be turned into an `ExecutionPlan` node as well, to make it compatible with distributed processing.

This could be out of scope for this PR.

crates/sail-plan/src/resolver/command/delete.rs
```diff
@@ -0,0 +1,157 @@
...
+            .map(|(field, field_id)| {
+                datafusion_common::arrow::datatypes::Field::new(
+                    field_id,
+                    field.data_type().clone(),
+                    field.is_nullable(),
+                )
+            })
+            .collect();
+        let schema_for_resolution =
+            datafusion_common::arrow::datatypes::Schema::new(fields_with_ids);
```
linhr: You could use the `rename_schema` utility function in the `sail-common-datafusion` crate.

Issue #805, opened 2025-08-29
aggregate functions should work on window frame


SparkApplicationMaster:
basically all the aggregate functions except some (in the end of issue) can be called as window functions in spark
maybe there is some simple solution to provide functionality without duplicating all that function calls to the window module?

```python
from pysail.spark import SparkConnectServer
from pyspark.sql import SparkSession
import pyspark.sql.functions as F

server = SparkConnectServer()
server.start()
_, port = server.listening_address

print(port)
spark = SparkSession.builder.appName("test").master("local[4]").getOrCreate()

spark.sql(
"""
    select 
        part,  
        any(f1 == 1) over(partition by part order by ord) as any,
        any_value(f1) over(partition by part order by ord) as any_value,
        approx_count_distinct(f1) over(partition by part order by ord) as approx_count_distinct,
        approx_percentile(f1, 0.5) over(partition by part order by ord) as approx_percentile,
        array_agg(f1) over(partition by part order by ord) as array_agg,
        bitmap_construct_agg(f1) over(partition by part order by ord) as bitmap_construct_agg,
        bitmap_or_agg(binary(hex(f1))) over(partition by part order by ord) as bitmap_or_agg,
        bit_and(f1) over(partition by part order by ord) as bit_and,
        bit_or(f1) over(partition by part order by ord) as bit_or,
        bit_xor(f1) over(partition by part order by ord) as bit_xor,
        bool_and(f1 == 1) over(partition by part order by ord) as bool_and,
        bool_or(f1 == 1) over(partition by part order by ord) as bool_or,
        collect_list(f1) over(partition by part order by ord) as collect_list,
        collect_set(f1) over(partition by part order by ord) as collect_set,
        corr(f1, f2) over(partition by part order by ord) as corr,
        count(f1) over(partition by part order by ord) as count,
        count_if(f1 == 1) over(partition by part order by ord) as count_if,
        count_min_sketch(f1, double(0.01), double(0.01), 1) over(partition by part order by ord) as count_min_sketch,
        covar_pop(f1, f2) over(partition by part order by ord) as covar_pop,
        covar_samp(f1, f2) over(partition by part order by ord) as covar_samp,
        every(f1 == 1) over(partition by part order by ord) as every,
        first(f1) over(partition by part order by ord) as first,
        first_value(f1) over(partition by part order by ord) as first_value,
        --grouping_id(f1) over(partition by part order by ord) as grouping_id,
[...]

PR #861, opened 2025-09-11
chore: add basic function docs, missing functions, and aggregate functions to window list

shehabgamin:
1. Added basic documentation listing supported functions and expressions. Future work will focus on defining a clear process for generating these docs along with their function signatures.

2. Spark 4.0 introduces several new functions. These have been added and marked as unimplemented in their respective lists.

3. All aggregate functions should now work on window frames:
https://github.com/lakehq/sail/issues/805

### DIFF
--- a/docs/guide/functions/scalar.md
+++ b/docs/guide/functions/scalar.md
@@ -491,6 +491,9 @@ rank: 1
 
 | Function        | Supported      |
 | --------------- | -------------- |
+| `from_xml`      | :construction: |
+| `schema_of_xml` | :construction: |
+| `to_xml`        | :construction: |
 | `xpath`         | :construction: |
 | `xpath_boolean` | :construction: |
 | `xpath_double`  | :construction: |
@@ -500,6 +503,3 @@ rank: 1
 | `xpath_number`  | :construction: |
 | `xpath_short`   | :construction: |
 | `xpath_string`  | :construction: |
-| `from_xml`      | :construction: |
-| `schema_of_xml` | :construction: |
-| `to_xml`        | :construction: |


crates/sail-plan/src/extension/function/url/try_url_decode.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let result = UrlDecode::new().invoke_with_args(args);
+        match result {
+            Ok(result) => Ok(result),
+            Err(_) => Ok(ColumnarValue::Scalar(ScalarValue::Utf8(None))),
+        }
+    }
```
linhr: This does not work. See my earlier comment in another PR: <https://github.com/lakehq/sail/pull/824#discussion_r2320705617>.

I'd appreciate if new features can be in separate PRs. Given the increased volumes of commits in the past few releases, it would make it easier for me to draft changelog without missing important updates.

docs/guide/functions/aggregate.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Aggregate Functions
+rank: 2
+---
+
+# Built-in Spark Aggregate Functions
+
+| Function                | Supported          |
+|-------------------------|--------------------|
+| `any`                   | :white_check_mark: |
```
linhr, line 10: I'd prefer to sort all tables solely by function names, regardless of whether they are supported or not. This would make it easier for the user to look up the content.

There are some text editors that can sort lines easily. If you are not sure how it can be done, I'd be happy to help a bit later.

shehabgamin: I get your point, but IMO it’s significantly easier to look up content when unsupported functions are grouped at the bottom. That way, I can scan through the supported ones without distraction, which feels less of an eye strain (at least for me).

Maybe we can find a middle ground: I made sure the functions are still sorted by name within each section (supported vs. unsupported).

linhr: Works with me!

crates/sail-plan/src/extension/function/url/try_url_decode.rs
shehabgamin: Ahh nice catch

PR #863, opened 2025-09-12
docs: sql features - literals (update)

everettroeth:


everettroeth:
i'm having trouble getting the failed check to pass, @shehabgamin .

the second .md commit is merely the first .md commit but put through prettier.io and repasted

PR #873, opened 2025-09-14
fix: exponential type parsing

shehabgamin:
Fixes https://github.com/lakehq/sail/issues/870.


linhr:
I feel `NumberSuffix::E` is confusing since `E` is not a valid suffix. I actually think what Claude suggested in #870 is a good approach.

PR #874, opened 2025-09-14
feat: to_char expression

SubhamSinghal:


crates/sail-plan/src/extension/function/string/spark_to_char.rs
```diff
@@ -0,0 +1,264 @@
...
+        match (&args[0], &args[1]) {
+            (ColumnarValue::Scalar(v), ColumnarValue::Scalar(ScalarValue::Utf8(Some(fmt)))) => {
+                if v.is_null() {
+                    Ok(ColumnarValue::Scalar(ScalarValue::Utf8(None)))
+                } else {
+                    let s = format_scalar_value(v, fmt)?;
+                    Ok(ColumnarValue::Scalar(ScalarValue::Utf8(Some(s))))
+                }
+            }
+            _ => not_impl_err!("to_char currently supports only scalar values with literal format string"),
```
shehabgamin: We also need to account for array input:
```
(ColumnarValue::Array(array), ColumnarValue::Array(array))
(ColumnarValue::Array(array), ColumnarValue::Scalar(ScalarValue::Utf8(Some(fmt))))
```

shehabgamin (commented):
@SubhamSinghal Thank you so much for this contribution! Exciting progress

crates/sail-plan/src/extension/function/string/spark_to_char.rs
SubhamSinghal: @shehabgamin need some help in fixing rust error

linhr:
Thanks for the contribution!

You can run the following commands locally for formatting and linting:
* `cargo +nightly fmt`
* `cargo clippy --all-targets --all-features`

linhr:
Nice progress!

Could you merge from `main` and resolve the conflict in `codec.rs`? Then we can run the workflow and see how it goes.

linhr:
The `to_char` function has recently been implemented in #2054. Thanks for your interest in contributing to Sail!

PR #883, opened 2025-09-16
docs: sql - features (add data types)

everettroeth:


everettroeth:
unsure if i did lines 712-724 on Floating Point Values mapping to Arrow types correctly. @shehabgamin

PR #886, opened 2025-09-17
feat: applyInArrow

SparkApplicationMaster:
implement `groupBy(cols).applyInArrow(func, return_type)`:
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.GroupedData.applyInArrow.html

reusing the code that implements applyInPandas 
and changing only input and output conversion

### DIFF
--- a/crates/sail-common/src/spec/expression.rs
+++ b/crates/sail-common/src/spec/expression.rs
@@ -358,6 +358,7 @@ pub enum PySparkUdfType {
     CogroupedMapPandas = 206,
     MapArrowIter = 207,
     GroupedMapPandasWithState = 208,
+    GroupedMapArrow = 209,
     Table = 300,
     ArrowTable = 301,
 }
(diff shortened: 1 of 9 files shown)


SparkApplicationMaster:
the code is not really great
but I don't know how to improve it
so review is appreciated

linhr (approved):
Thanks for having this!

I feel this looks good, so there is no further change needed at the moment!

SparkApplicationMaster:
@linhr thanks for the response!
This need to add `# noqa: FBT001` was confusing me but no simple way to get it away
I'll try to do co-grouped mapInArrow in the same style then, should be relatively easy

linhr:
I see. Good point!

The boolean flag is not too bad since its an internal function, so it should be fine to ignore `FBT001` in this particular case.

We could have separate Python functions for each Pandas/Arrow UDF type so that we don't need the flag. This makes the code structure more consistent across all UDF types. This may introduce some boilerplate code though. I'm open to this exploration and refactoring as well.

PR #887, opened 2025-09-18
feat: cogrouped applyInArrow

SparkApplicationMaster:
implement `left.cogroup(right).applyInArrow(func, return_type)`:
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.PandasCogroupedOps.applyInArrow.html

reusing the code that implements cogrouped `applyInPandas`
and changing only input and output conversion

### DIFF
--- a/crates/sail-common/src/spec/expression.rs
+++ b/crates/sail-common/src/spec/expression.rs
@@ -359,6 +359,7 @@ pub enum PySparkUdfType {
     MapArrowIter = 207,
     GroupedMapPandasWithState = 208,
     GroupedMapArrow = 209,
+    CogroupedMapArrow = 210,
     Table = 300,
     ArrowTable = 301,
 }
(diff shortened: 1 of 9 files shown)


SparkApplicationMaster:
last test in this test group `test_with_local_data` is failing:
https://github.com/apache/spark/blob/552effccce9afb7875ad4fc6cf020d457dd0c267/python/pyspark/sql/tests/arrow/test_arrow_cogrouped_map.py#L267
because if depends on the order of groups after groupBy
which is non-deterministic by nature and implementation-dependent I suppose
and in spark itself it can also can depend on spark.sql.shuffle.partitions and other params

linhr (approved):
This looks nice! Thanks for having this!

Yeah some Spark tests do not take into account the non-deterministic nature of result row order. So it's fine if they fail in our implementation.

Issue #862, opened 2025-09-12
Spark SQL Functions Coverage and Parity - Variant


shehabgamin:
- [x] is_variant_null
- [x] parse_json
- [x] schema_of_variant
- [ ] schema_of_variant_agg
- [ ] to_variant_object
- [ ] try_parse_json
- [x] try_variant_get
- [ ] variant_explode
- [ ] variant_explode_outer
- [x] variant_get

davidlghellin:
https://github.com/datafusion-contrib/datafusion-variant

shehabgamin:
@davidlghellin Because DataFusion is tested on Sail before each release, we can’t depend on crates that themselves depend on DataFusion unless we’re willing to maintain a fork.

The only exception today is `datafusion-functions-json`, which was added before DataFusion pre-releases were tested on Sail. We plan to remove this dependency eventually, since maintaining a fork is a bit of a hassle.

If you’d like, you can copy the variant functions from `datafusion-variant`. Just make sure to include proper attribution in the code.

twsl:
@davidlghellin whats the status on `try_parse_json` and the other missing functions? are you actively working on the pr?

davidlghellin:
> [@davidlghellin](https://github.com/davidlghellin) whats the status on `try_parse_json` and the other missing functions? are you actively working on the pr?

Yes, actively working on it. I have the implementation ready — just need to resolve some merge conflicts and clean up before it's PR-ready.

PR #890, opened 2025-09-19
feat: add `parse_json` function [POC]

davidlghellin:
Part of #862 

Try to start Variant Type


davidlghellin:
https://spark.apache.org/docs/4.0.0/api/python/reference/pyspark.sql/api/pyspark.sql.functions.parse_json.html

need to see VariantType
https://github.com/lakehq/sail/blob/1e1a475416a7c3236bbd3b7e7f4818b2b9f9bdfb/crates/sail-common/src/spec/data_type.rs#L27
not datatype some like Variant

PR #891, opened 2025-09-19
fix: df.replace() method

SparkApplicationMaster:
make df.replace() more consistent with spark:
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.replace.html

now strings don't mess with numerics and booleans
and exception is raised if a subset col doesn't exist


SparkApplicationMaster:
`CogroupedMapInArrowParityTests::test_with_local_data` is a nondeterministic test from previous PR 
and has nothing to do with this new PR
the explanation is here:
https://github.com/lakehq/sail/pull/887#issuecomment-3309913779

Issue #907, opened 2025-09-23
New function in Datafusion 50 to migrate Sail


davidlghellin:
I'm trying to figure out which functions we need to add for the DataFusion 50 upgrade. I'd still need to review everything, but this could be a good starting point for the idea I had.

In Datafusion Proyect
```
git checkout 49.0.2
❯ cd datafusion/spark
for f in **/mod.rs(.N); do
  # empty vec
  if rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[\s*\S' "$f" >/dev/null; then
    echo "---- $f ----"
    rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[([\s\S]*?\S[\s\S]*?)\]' -or '$1' "$f"
    echo
  fi
done > datafusion-49
```

```
❯ git checkout 50.0.0
M	datafusion-testing
Previous HEAD position was f43df3f2a [branch-49] Prepare `49.0.2` version and changelog (#17277)
HEAD is now at 10343c182 Revert #17295 (Support from-first SQL syntax) (#17520) (#17544)
❯ cd ../..
❯ cd datafusion/spark
for f in **/mod.rs(.N); do
  # empty vec
  if rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[\s*\S' "$f" >/dev/null; then
    echo "---- $f ----"
    rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[([\s\S]*?\S[\s\S]*?)\]' -or '$1' "$f"
    echo
  fi
done > datafusion-50
```

```
❯ diff datafusion-49 datafusion-50
3a4
> 32:array()
4a6,8
> ---- src/function/bitmap/mod.rs ----
> 36:bitmap_count()
> 
5a10
> 39:bit_get(), bit_count()
9a15
> 32:r#if()
15a22
> 59:date_add(), date_sub(), last_day(), next_day()
20c27
< 31:sha2()
---
> 39:crc32(), sha1(), sha2()
29c36,43
< 42:expm1(), factorial(), hex()
---
> 54:        expm1(),
> 55:        factorial(),
> 56:        hex(),
> 57:        modulus(),
> 58:        pmod(),
> 59:        rint(),
> 60:        width_bucket(),
> 61:    
36c50
< 43:ascii(), char()
---
> 64:ascii(), char(), ilike(), like(), luhn_check()
42a57
> 32:parse_url()
```

[...]

PR #909, opened 2025-09-23
refactor: use datafusion 50 func

davidlghellin:
This is an attempt to migrate the functions crate to datafusion-spark.
Some functions will be straightforward (we will only need to update the use statements and remove the old code), but others may require more work once we update the gold tests, for example.

part #907


crates/sail-execution/src/codec.rs
```diff
@@ -1269,9 +1268,6 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_try_subtract" | "try_subtract" => {
                 Ok(Arc::new(ScalarUDF::from(SparkTrySubtract::new())))
             }
-            "spark_width_bucket" | "width_bucket" => {
-                Ok(Arc::new(ScalarUDF::from(SparkWidthBucket::new())))
-            }
```
davidlghellin, line 1424: I have a quick question — I believe this is correct since DataFusion handles the encode/decode, but I’d like to double-check to be sure. Thanks!

crates/sail-plan/src/function/scalar/string.rs
```diff
@@ -233,7 +234,7 @@ pub(super) fn list_built_in_string_functions() -> Vec<(&'static str, ScalarFunct
         ("lower", F::custom(lower)),
         ("lpad", F::var_arg(expr_fn::lpad)),
         ("ltrim", F::var_arg(rev_args(expr_fn::ltrim))),
-        ("luhn_check", F::unknown("luhn_check")),
+        ("luhn_check", F::udf(SparkLuhnCheck::new())),
```
davidlghellin: necesary encode/decode?
like https://github.com/lakehq/sail/pull/909/files#r2372083426

linhr:
Thanks for having this!

Let's keep the encode/decode logic since Spark functions may or may not be listed in the DataFusion's default function registry. So it's safer to have explicit logic inside Sail.

crates/sail-plan/src/function/scalar/misc.rs
```diff
@@ -3,11 +3,11 @@ use std::sync::Arc;
 use datafusion::functions::expr_fn;
 use datafusion_common::ScalarValue;
 use datafusion_expr::{expr, lit, when, ExprSchemable, Operator, ScalarUDF};
+use datafusion_spark::function::bitmap::bitmap_count::BitmapCount;
```
SparkApplicationMaster: if we use the exported functions like this:
```rust
use datafusion_spark::function::bitmap::expr_fn as bitmap_fn;
...
("bitmap_count", F::unary(bitmap_fn::bitmap_count)),
```
we can get rid of the codec overrides for them and simplify the logic a bit

davidlghellin: thanks, I will check

davidlghellin:
Other functions needs review to update in datafusion

shehabgamin:
Would love to get @SparkApplicationMaster's eyes on this as well!

crates/sail-execution/src/codec.rs
```diff
@@ -37,7 +37,15 @@ use datafusion_proto::physical_plan::{AsExecutionPlan, PhysicalExtensionCodec};
...
 };
+use datafusion_spark::function::bitmap::bitmap_count::BitmapCount;
+use datafusion_spark::function::bitwise::bit_count::SparkBitCount;
+use datafusion_spark::function::bitwise::bit_get::SparkBitGet;
+use datafusion_spark::function::hash::crc32::SparkCrc32;
+use datafusion_spark::function::hash::sha1::SparkSha1;
 use datafusion_spark::function::math::expm1::SparkExpm1;
+use datafusion_spark::function::math::modulus::SparkPmod;
+use datafusion_spark::function::math::width_bucket::SparkWidthBucket;
+use datafusion_spark::function::string::luhn_check::SparkLuhnCheck;
```
SparkApplicationMaster, line 48: now, when the exported functions are used, I suppose these all functions above can be safely removed. except SparkWidthBucket
SparkWidthBucket needs to stay like this, because it has 4 arguments, but there is no `F::fournary` common function to use the exported datafusion function 
but maybe to add  `F::fournary` [here](https://github.com/lakehq/sail/blob/a84748e4d07539b6b56e42c49a69559e8821fb79/crates/sail-plan/src/function/common.rs#L86) and reuse [this](https://github.com/apache/datafusion/blob/daeb6597a0c7344735460bb2dce13879fd89d7bd/datafusion/spark/src/function/math/mod.rs#L50) will be better

linhr: The exported functions still internally use the scalar UDF implementations so we need to explicitly handle them in the codec.

davidlghellin:
SparkBitmapCount does not support dictionary encoded input  -> https://github.com/apache/datafusion/issues/18058

shehabgamin:
Will get to this by end of week!

crates/sail-plan/src/function/scalar/bitwise.rs
```diff
@@ -65,8 +44,8 @@ pub(super) fn list_built_in_bitwise_functions() -> Vec<(&'static str, ScalarFunc
     vec![
         ("&", F::binary_op(Operator::BitwiseAnd)),
         ("^", F::binary_op(Operator::BitwiseXor)),
-        ("bit_count", F::unary(bit_count)),
-        ("bit_get", F::binary(bit_get)),
+        ("bit_count", F::unary(bitwise_fn::bit_count)),
+        ("bit_get", F::binary(bitwise_fn::bit_get)),
         ("getbit", F::binary(bit_get)),
```
linhr: The `getbit` function can be migrated as well. And then we can remove the `bit_get` function in this file.

crates/sail-execution/src/codec.rs
```diff
@@ -1318,9 +1321,14 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
                 Ok(Arc::new(ScalarUDF::from(SparkTryToBinary::new())))
             }
             "spark_abs" | "abs" => Ok(Arc::new(ScalarUDF::from(SparkAbs::new()))),
+            "spark_bit_count" | "bit_count" => Ok(Arc::new(ScalarUDF::from(SparkBitCount::new()))),
+            "spark_bit_get" | "bit_get" => Ok(Arc::new(ScalarUDF::from(SparkBitGet::new()))),
```
linhr: When migrating `getbit` as suggested in the other comment, we need to make a change here as well.

linhr (approved):
@davidlghellin Thanks for the contribution! Sorry for not following up earlier for this PR.

The changes look good to me! Would you mind taking care of the minor comments I have, and merging from `main` to resolve the conflicts? Thanks a lot!

linhr (approved):

PR #913, opened 2025-09-23
feat: check function replace [POC] [NOT-MERGE]

davidlghellin:
This PR aims to demonstrate and provide evidence that the change cannot always be applied directly without adjusting our own code to match our data types. Additional changes might be required.
Should we  consider changing DataFusion to provide the right extension points or type support.? In short, we’ll first try to fit our types via composition/conversions; if the gap remains, we’ll propose upstream changes to DataFusion.

cc: @linhr @shehabgamin @lonless9


linhr:
I believe all the functions in the `datafusion-spark` crate is supposed to fully support Spark's behavior. So if there is a discrepancy in type support etc., the change need to happen upstream.

PR #916, opened 2025-09-25
fix: handle side-effect of delete operation for Delta Lake

lonless9:


crates/sail-delta-lake/src/delta_format/scan_by_adds_exec.rs
```diff
@@ -0,0 +1,235 @@
...
+use datafusion_physical_expr::{Distribution, EquivalenceProperties};
+use deltalake::logstore::StorageConfig;
+use futures::stream::{self, StreamExt, TryStreamExt};
+use url::Url;
+
+use crate::delta_datafusion::{DeltaScanConfigBuilder, DeltaTableProvider};
+use crate::table::open_table_with_object_store;
+
+/// An ExecutionPlan that scans Delta files based on a stream of Add actions from its input.
+/// This node acts as a bridge, consuming metadata (file list) and producing data.
```
linhr, line 29: This is great!

PR #917, opened 2025-09-25
feat: implement join reorder physical optimizer with the DPHyp algorithm

lonless9:
#759

### DIFF
--- a/crates/sail-physical-optimizer/src/join_reorder/builder.rs
+++ b/crates/sail-physical-optimizer/src/join_reorder/builder.rs
@@ -93,6 +93,7 @@ impl GraphBuilder {
         trace!("Visiting plan: {}", plan.name());
         let any_plan = plan.as_any();
 
+        // TODO: Extend to support `SortMergeJoinExec`.
         if let Some(join_plan) = any_plan.downcast_ref::<HashJoinExec>() {
             if join_plan.join_type() == &JoinType::Inner {
                 trace!("Visiting inner join: {}", join_plan.name());
@@ -178,6 +179,7 @@ impl GraphBuilder {
                 &left_map,
                 &right_map,
             )?;
+            // TODO: Separating the equi-join conditions from the non-equi filter expression at the source
             filter_expr = Arc::new(BinaryExpr::new(filter_expr, Operator::And, extra))
                 as Arc<dyn PhysicalExpr>;
         }
@@ -222,9 +224,11 @@ impl GraphBuilder {
         let transformed = expr_arc.transform(|node| {
             if let Some(col) = node.as_any().downcast_ref::<Column>() {
                 let i = col.index();
-                // Safeguard: if index out of bounds, leave as-is
+                // Safeguard: if index out of bounds, return error
                 if i >= indices.len() {
-                    return Ok(Transformed::no(node));
+                    return Err(DataFusionError::Internal(
+                        "Column index out of bounds in join filter rewrite".to_string(),
+                    ));
                 }
                 let ci = &indices[i];
                 // Determine which stable column this refers to
@@ -240,6 +244,7 @@ impl GraphBuilder {
                 }) = stable_entry_opt.cloned()
                 {
                     // Build a stable name like R{relation_id}.C{column_index}
+                    // TODO: Consider implement PhysicalExpr trait for StableColumn.
                     let stable_name = format!("R{}.C{}", relation_id, column_index);
                     // TODO: Reconstructor will retarget indices to its compact schema
                     let new_col = Column::new(&stable_name, 0);
@@ -393,35 +398,21 @@ impl GraphBuilder {
         &self,
         on_conditions: &JoinConditionPairs,
     ) -> Result<Arc<dyn PhysicalExpr>> {
-        use datafusion::logical_expr::Operator;
-        use datafusion::physical_expr::expressions::BinaryExpr;
-
-        if on_conditions.is_empty() {
-            return Err(DataFusionError::Internal(
-                "Join must have at least one ON condition".to_string(),
-            ));
-        }
-
-        // Start with the first equality condition
-        let mut result_expr = Arc::new(BinaryExpr::new(
-            on_conditions[0].0.clone(),
-            Operator::Eq,
-            on_conditions[0].1.clone(),
-        )) as Arc<dyn PhysicalExpr>;
-
-        // Chain additional conditions with AND
-        for (left_expr, right_expr) in on_conditions.iter().skip(1) {
-            let eq_expr = Arc::new(BinaryExpr::new(
-                left_expr.clone(),
-                Operator::Eq,
-                right_expr.clone(),
-            )) as Arc<dyn PhysicalExpr>;
-
-            result_expr = Arc::new(BinaryExpr::new(result_expr, Operator::And, eq_expr))
-                as Arc<dyn PhysicalExpr>;
-        }
-
-        Ok(result_expr)
+        on_conditions
+            .iter()
+            .map(|(a, b)| -> Arc<dyn PhysicalExpr> {
+                Arc::new(BinaryExpr::new(a.clone(), Operator::Eq, b.clone()))
+            })
+            .fold(None, |acc, expr| -> Option<Arc<dyn PhysicalExpr>> {
+                if let Some(acc_expr) = acc {
+                    Some(Arc::new(BinaryExpr::new(acc_expr, Operator::And, expr)))
+                } else {
+                    Some(expr)
+                }
+            })
+            .ok_or_else(|| {
+                DataFusionError::Internal("Join must have at least one ON condition".to_string())
+            })
     }
 }
 
--- a/crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
+++ b/crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
@@ -6,7 +6,7 @@ use datafusion::error::{DataFusionError, Result};
 use crate::join_reorder::cardinality_estimator::CardinalityEstimator;
 use crate::join_reorder::cost_model::CostModel;
 use crate::join_reorder::dp_plan::DPPlan;
-use crate::join_reorder::graph::{JoinEdge, QueryGraph};
+use crate::join_reorder::graph::QueryGraph;
 use crate::join_reorder::join_set::JoinSet;
 
 /// Plan enumerator that implements dynamic programming algorithm to find optimal join order.
@@ -33,10 +33,10 @@ impl PlanEnumerator {
             return vec![];
         }
         let mut subsets = Vec::new();
-        let total = 1usize << n;
-        for mask in 1..total {
+        let last = 1usize.unbounded_shl(n as u32).wrapping_sub(1);
+        for mask in 1..=last {
             let mut subset = Vec::new();
-            for (i, &elem) in elems.iter().enumerate().take(n) {
+            for (i, &elem) in elems.iter().enumerate() {
                 if (mask & (1usize << i)) != 0 {
                     subset.push(elem);
                 }
@@ -57,30 +57,6 @@ impl PlanEnumerator {
             cost_model,
             emit_count: 0,
         }
-    }
-
-    /// Helper to get indices of connecting edges for two disjoint sets using the optimized structure.
-    fn get_connecting_edge_indices(&self, left: JoinSet, right: JoinSet) -> Vec<usize> {
-        if !left.is_disjoint(&right) {
-            return vec![];
-        }
-
-        let connecting_edges = self.query_graph.get_connecting_edges(left, right);
-        let mut edge_indices = Vec::new();
-
-        for edge in connecting_edges {
-            // Find the index of this edge in the original edges vector
-            if let Some(index) = self
-                .query_graph
-                .edges
-                .iter()
-                .position(|graph_edge| std::ptr::eq(edge, graph_edge))
-            {
-                edge_indices.push(index);
-            }
-        }
-
-        edge_indices
     }
 
     /// Main method that solves for the optimal join order using DPhyp-style enumeration.
@@ -198,7 +174,7 @@ impl PlanEnumerator {
 
         for &nbr in neighbors.iter().rev() {
             let nbr_set = JoinSet::new_singleton(nbr)?;
-            let edge_indices = self.get_connecting_edge_indices(nodes, nbr_set);
+            let edge_indices = self.query_graph.get_connecting_edge_indices(nodes, nbr_set);
 
             if !edge_indices.is_empty()
                 && !self.try_emit_csg_cmp(nodes, nbr_set, edge_indices.clone())?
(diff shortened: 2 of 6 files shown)


crates/sail-physical-optimizer/src/join_reorder/graph.rs
```diff
@@ -0,0 +1,664 @@
...
+        // Test neighbor lookup for relation 0
+        let set_0 = JoinSet::new_singleton(0).unwrap();
+        let neighbors = graph.get_neighbors(set_0);
+        assert_eq!(neighbors, vec![1]); // Relation 0 connected to relation 1
+
+        // Test neighbor lookup for relation 1
+        let set_1 = JoinSet::new_singleton(1).unwrap();
+        let neighbors = graph.get_neighbors(set_1);
+        let mut expected = vec![0, 2];
+        expected.sort();
```
linhr:
```suggestion
        let mut neighbors = graph.get_neighbors(set_1);
        neighbors.sort();
        let expected = vec![0, 2];
```

crates/sail-physical-optimizer/src/join_reorder/builder.rs
```diff
@@ -0,0 +1,646 @@
...
+        }
+    }
+
+    /// Recursively traverses the execution plan, building the query graph.
+    /// Returns a map of the plan's output columns to our stable IDs.
+    fn visit_plan(&mut self, plan: Arc<dyn ExecutionPlan>) -> Result<ColumnMap> {
+        trace!("Visiting plan: {}", plan.name());
+        let any_plan = plan.as_any();
+
+        if let Some(join_plan) = any_plan.downcast_ref::<HashJoinExec>() {
```
linhr, line 97: We can add a TODO for `SortMergeJoinExec`.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
```diff
@@ -0,0 +1,646 @@
...
+        if let Some(proj_plan) = any_plan.downcast_ref::<ProjectionExec>() {
+            trace!("Visiting projection: {}", proj_plan.name());
+            return self.visit_projection(proj_plan);
+        }
+
+        // If it's not a reorderable join or a projection we can see through,
+        // it's either a boundary (leaf of our graph) or not part of a reorderable region.
+        // AggregateExec and other transformation nodes are not part of reorderable regions.
+
+        if any_plan.is::<AggregateExec>() {
```
linhr, line 119: Why is `AggregateExec` a special case? Is it because of this comment?

```rust
                        // This column comes from a complex expression (e.g., aggregate output)
                        // We cannot resolve it to a *specific base relation* for join condition purposes.
                        // If a join condition relies on a column that is an aggregate output,
                        // that join condition cannot be directly mapped to base relations.
```

crates/sail-physical-optimizer/src/join_reorder/builder.rs
```diff
@@ -0,0 +1,646 @@
...
+    /// Helper function to build a conjunction expression from join ON conditions.
+    /// Converts (left_expr, right_expr) pairs into a single AND expression.
+    fn build_conjunction_from_on(
+        &self,
+        on_conditions: &JoinConditionPairs,
+    ) -> Result<Arc<dyn PhysicalExpr>> {
+        use datafusion::logical_expr::Operator;
+        use datafusion::physical_expr::expressions::BinaryExpr;
+
+        if on_conditions.is_empty() {
```
linhr: This function could be implemented purely via iterator methods. (Just wanted to share a cool trick that could be useful in general.)

```rust
        on_conditions
            .iter()
            .map(|(a, b)| -> Arc<dyn PhysicalExpr> {
                Arc::new(BinaryExpr::new(a.clone(), Operator::Eq, b.clone()))
            })
            .fold(None, |acc, expr| -> Option<Arc<dyn PhysicalExpr>> {
                if let Some(acc_expr) = acc {
                    Some(Arc::new(BinaryExpr::new(acc_expr, Operator::And, expr)))
                } else {
                    Some(expr)
                }
            })
            .ok_or_else(|| {
                DataFusionError::Internal("Join must have at least one ON condition".to_string())
            })
````

File: crates/sail-physical-optimizer/src/join_reorder/builder.rs [hunk 5]
```diff
@@ -0,0 +1,646 @@
...
+        let indices = filter.column_indices();
+
+        let expr_arc = Arc::clone(expr);
+        let transformed = expr_arc.transform(|node| {
+            if let Some(col) = node.as_any().downcast_ref::<Column>() {
+                let i = col.index();
+                // Safeguard: if index out of bounds, leave as-is
+                if i >= indices.len() {
+                    return Ok(Transformed::no(node));
+                }
```
linhr: Should we return an error instead?

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
```diff
@@ -0,0 +1,903 @@
...
+                }
+            }
+
+            // Final sanity check: warn if there are still pending filters after reconstruction
+            if !self.pending_filters.is_empty() && dp_plan.join_set.cardinality() > 1u32 {
+                log::warn!(
+                    "JoinReorder: {} pending filters remain after reconstructing join set {:?}. These filters may not be applied.",
+                    self.pending_filters.len(),
+                    dp_plan.join_set.iter().collect::<Vec<_>>()
+                );
```
linhr: Should we return an error here?

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
```diff
@@ -0,0 +1,903 @@
...
+        // Create HashJoinExec
+        let join_plan = Arc::new(HashJoinExec::try_new(
+            left_plan,
+            right_plan,
+            on_conditions,
+            join_filter,         // Use JoinEdge.filter for non-equi conditions
+            &join_type,          // Use determined join type
+            None,                // projection
+            PartitionMode::Auto, // partition_mode
+            NullEquality::NullEqualsNothing, // null_equality
```
linhr: We may need to add a TODO here. I guess `NullEquality` needs to have the same value before and after join reorder, so we should probably skip the optimizer completely if `NullEquality` is something else in the input region.

crates/sail-physical-optimizer/src/join_reorder/mod.rs
```diff
@@ -33,8 +58,596 @@ impl PhysicalOptimizerRule for JoinReorder {
...
+                trace!(
+                    "JoinReorder: Optimization successful at current level. Returning new plan."
+                );
+                trace!(
+                    "JoinReorder: Optimized plan:\n{}",
+                    displayable(final_plan.as_ref()).indent(true)
+                );
+
+                // The entire region has been optimized and replaced. Return the new plan.
+                return Ok(final_plan);
```
linhr, line 129: We can add a TODO here for recursively optimizing descendant regions.

crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
```diff
@@ -0,0 +1,665 @@
...
+    fn generate_all_nonempty_subsets(&self, elems: &[usize]) -> Vec<Vec<usize>> {
+        let n = elems.len();
+        if n == 0 {
+            return vec![];
+        }
+        let mut subsets = Vec::new();
+        let total = 1usize << n;
+        for mask in 1..total {
+            let mut subset = Vec::new();
+            for (i, &elem) in elems.iter().enumerate().take(n) {
```
linhr: I guess `take(n)` is not necessary?

crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
```diff
@@ -0,0 +1,665 @@
...
+
+impl PlanEnumerator {
+    /// Generate all non-empty subsets of the given neighbor list.
+    fn generate_all_nonempty_subsets(&self, elems: &[usize]) -> Vec<Vec<usize>> {
+        let n = elems.len();
+        if n == 0 {
+            return vec![];
+        }
+        let mut subsets = Vec::new();
+        let total = 1usize << n;
```
linhr: This doesn't work for 64 elements, but I guess it could be fixed by having `let last = 1usize.unbounded_shl(n as u32).wrapping_sub(1);` and `for mask in 1..=last`.

Not a big issue since we probably won't support enumerating all the `2 ^ 64 - 1` subsets anyway. 😆

crates/sail-physical-optimizer/src/join_reorder/graph.rs
```diff
@@ -0,0 +1,664 @@
...
+        // Collect all neighbors from this position
+        for neighbor_info in &current.neighbors {
+            for rel in neighbor_info.neighbor.iter() {
+                neighbors.insert(rel);
+            }
+        }
+    }
+
+    /// Gets all edges connecting two disjoint subsets `left` and `right` using the optimized structure.
+    pub fn get_connecting_edges(&self, left: JoinSet, right: JoinSet) -> Vec<&JoinEdge> {
```
linhr: I wonder if `get_connecting_edges` should just return edge indices. This way we don't need the fragile pointer operation `std::ptr::eq` to get edge index from edge in a few places.

`CardinalityEstimator::estimate_join_cardinality()` can accept edge indices directly as well, since it has access to the graph.

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
```diff
@@ -0,0 +1,903 @@
...
+    /// Clear plan cache.
+    #[cfg(test)]
+    pub fn clear_cache(&mut self) {
+        self.plan_cache.clear();
+        self.pending_filters.clear();
+    }
+}
+
+/// Helper to find the physical index of a stable column in a ColumnMap.
+fn find_physical_index(stable_col: &StableColumn, map: &ColumnMap) -> Option<usize> {
```
linhr: This function is duplicated.

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
```diff
@@ -0,0 +1,903 @@
...
+
+    fn analyze_predicate_dependencies(
+        &self,
+        predicate: &Arc<dyn PhysicalExpr>,
+        left_map: &ColumnMap,
+        right_map: &ColumnMap,
+        left_plan: &Arc<dyn ExecutionPlan>,
+        right_plan: &Arc<dyn ExecutionPlan>,
+    ) -> Result<JoinSet> {
+        let parse_stable = |name: &str| -> Option<(usize, usize)> {
```
linhr: This function is duplicated.

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
```diff
@@ -0,0 +1,903 @@
...
+            })?
+        } else {
+            self.combine_filters_with_and(non_equi_filters)?
+        };
+
+        use datafusion::common::tree_node::{Transformed, TreeNode};
+        use datafusion::physical_expr::expressions::Column;
+
+        // Helper: parse stable name like "R{rel}.C{col}" -> (rel, col)
+        let parse_stable = |name: &str| -> Option<(usize, usize)> {
```
linhr: This function is duplicated.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
```diff
@@ -0,0 +1,646 @@
...
+                    _ => None,
+                };
+
+                if let Some(ColumnMapEntry::Stable {
+                    relation_id,
+                    column_index,
+                }) = stable_entry_opt.cloned()
+                {
+                    // Build a stable name like R{relation_id}.C{column_index}
+                    let stable_name = format!("R{}.C{}", relation_id, column_index);
```
linhr, line 248: I realized that something like `PlaceholderColumn` in your earlier prototype actually makes sense, so that we don't have to encode relation and column index in the name.

This could be future work.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
```diff
@@ -0,0 +1,646 @@
...
+        if let Some(join_filter) = join_plan.filter() {
+            // Rewrite filter expressions to use stable column names (R{relation}.C{index})
+            // instead of ephemeral projection names and join-local indices.
+            let extra = self.rewrite_join_filter_to_stable(
+                join_plan,
+                join_filter.expression(),
+                &left_map,
+                &right_map,
+            )?;
+            filter_expr = Arc::new(BinaryExpr::new(filter_expr, Operator::And, extra))
```
linhr, line 183: Do we have to put equi join conditions into the filter expression? Since we have `equi_pairs` in the `JoinEdge`, if the filter expression only represents non-equi join conditions, would it make the reconstructor easier?

This is just my rough thought so I'm not fully sure if it makes sense. And this is future work for sure.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: Correct! Thus we need to fix build_final_projection and other related issues of complex and nested expressions. We can add this as a todo.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: Yes, we can differentiate between the two and store them separately to avoid the current duplication problem. I've added todo.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: I've added a todo and we can switch in following PR.

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/graph.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/builder.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/reconstructor.rs
lonless9: I have added a todo to track this issue.

crates/sail-physical-optimizer/src/join_reorder/mod.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/enumerator.rs
lonless9: Done!

crates/sail-physical-optimizer/src/join_reorder/graph.rs
lonless9: Done!

keen85:
awesome!
I'm really looking forward to trying this out.

PR #921, opened 2025-09-26
refactor: share scan utilities for Delta Lake

lonless9:


crates/sail-delta-lake/src/physical_plan/scan_by_adds_exec.rs
```diff
@@ -96,23 +95,40 @@ impl DeltaScanByAddsExec {
...
+            },
+            &session_state,
+            file_schema,
+        )
+        .map_err(|e| DataFusionError::External(Box::new(e)))?;
 
-        scan_plan.execute(0, context)
+        let scan_exec =
+            datafusion::catalog::memory::DataSourceExec::from_data_source(file_scan_config);
+        scan_exec.execute(0, context)
```
lonless9, line 131: Check here.

linhr: This is fine for now. Let me think about it and see how we can improve the physical planning for this case.

PR #926, opened 2025-09-29
fix: map_from_entries error on lists with null entries

### LINKED ISSUE #925: map_from_entries(entries_list) should return NULL if any of entries in entries_list is NULL
this case was not covered in previous versions of my code:

```python

server = SparkConnectServer()
server.start()

#spark = SparkSession.builder.appName("test").master("local[4]").getOrCreate()

spark.sql("""
select map_from_entries(
    array(
        struct(4 as a, "d" as b),
        cast(NULL as struct<a: int, b: string>),
        struct(5 as a, "e" as b),
        struct(6 as a, "f" as b)
    )
) as m;
""").show()
```
spark:
```
+----+
|   m|
+----+
|NULL|
+----+
```
sail 0.3.5:
```
AnalysisException: create_map: expected array<struct<key, value>> or (array<key>, array<value>), got Null
```

SparkApplicationMaster:
map_from_entries now returns NULL for the row if any entry in list is null:
(was failing)
```python

server = SparkConnectServer()
server.start()

#spark = SparkSession.builder.appName("test").master("local[4]").getOrCreate()

spark.sql("""
SELECT map_from_entries(data) as result, typeof(map_from_entries(data)) as type
from values
    (array(
        struct(1 as a, "a" as b), 
        cast(NULL as struct<a: int, b: string>), 
        cast(NULL as struct<a: int, b: string>)
    )),
    (NULL),
    (array(
        struct(2 as a, "b" as b), 
        struct(3 as a, "c" as b)
    )),
    (array(
        struct(4 as a, "d" as b),
        cast(NULL as struct<a: int, b: string>),
        struct(5 as a, "e" as b),
        struct(6 as a, "f" as b)
    ))
as tab(data);
""").show(truncate=False)
```
```
+----------------+---------------+
|result          |type           |
+----------------+---------------+
|NULL            |map<int,string>|
|NULL            |map<int,string>|
|{2 -> b, 3 -> c}|map<int,string>|
|NULL            |map<int,string>|
+----------------+---------------+
```
added a test for this

slightly tuned up the offsets logic
not sure if the first offset can be not 0, in list slices maybe
but now it will be correct if it can

closes #925

### DIFF
--- a/crates/sail-function/src/scalar/map/map_from_entries.rs
+++ b/crates/sail-function/src/scalar/map/map_from_entries.rs
@@ -1,6 +1,7 @@
 use std::any::Any;
 
-use datafusion::arrow::array::{Array, ArrayRef, StructArray};
+use datafusion::arrow::array::{Array, ArrayRef, NullBufferBuilder, StructArray};
+use datafusion::arrow::buffer::NullBuffer;
 use datafusion::arrow::datatypes::DataType;
 use datafusion_common::utils::take_function_args;
 use datafusion_common::{exec_err, Result};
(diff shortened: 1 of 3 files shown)


linhr (approved):
Nice fix! Thanks!

PR #928, opened 2025-09-29
feat: file metadata cache for Parquet

### LINKED ISSUE #516: Cache file listing results and Parquet metadata

linhr:
Closes #516.

This is a follow-up of #687.

Note: `parquet.cache_metadata` is removed in <https://github.com/apache/datafusion/pull/17062> so we can no longer configure metadata cache for each data source.

### DIFF
--- a/crates/sail-cache/src/file_metadata_cache.rs
+++ b/crates/sail-cache/src/file_metadata_cache.rs
@@ -1,0 +1,224 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use datafusion::execution::cache::cache_manager::{
+    FileMetadata, FileMetadataCache, FileMetadataCacheEntry,
+};
+use datafusion::execution::cache::CacheAccessor;
+use log::debug;
+use moka::policy::EvictionPolicy;
+use moka::sync::Cache;
+use object_store::path::Path;
+use object_store::ObjectMeta;
+
+pub struct MokaFileMetadataCache {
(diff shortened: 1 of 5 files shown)


crates/sail-cache/src/file_metadata_cache.rs
```diff
@@ -0,0 +1,224 @@
...
+
+pub struct MokaFileMetadataCache {
+    size_limit: Option<u64>,
+    metadata: Cache<Path, (ObjectMeta, Arc<dyn FileMetadata>)>,
+}
+
+impl MokaFileMetadataCache {
+    const NAME: &'static str = "MokaFileMetadataCache";
+
+    pub fn new(size_limit: Option<u64>) -> Self {
```
shehabgamin: should we add a `ttl`?

linhr: Done!

PR #941, opened 2025-09-30
refactor: item taker

linhr:


crates/sail-common-datafusion/src/utils/items.rs
```diff
@@ -12,8 +12,6 @@ pub trait ItemTaker {
     fn one(self) -> Result<Self::Item>;
     fn two(self) -> Result<(Self::Item, Self::Item)>;
     fn three(self) -> Result<(Self::Item, Self::Item, Self::Item)>;
-    #[allow(clippy::type_complexity, dead_code)]
-    fn four(self) -> Result<(Self::Item, Self::Item, Self::Item, Self::Item)>;
```
linhr, line 16: We need to add back this due to #909.

PR #960, opened 2025-10-10
feat: add `try_sum Aggregate` function

davidlghellin:
Part of #309 

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.try_sum.html?highlight=try_sum


davidlghellin:
It’s my first accumulator function—I hope I learn a lot and do it right! 😅

crates/sail-function/src/aggregate/try_sum.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+}
+
+impl TrySumFunction {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::one_of(
+                vec![
+                    TypeSignature::Exact(vec![DataType::Int64]),
+                    TypeSignature::Exact(vec![DataType::Float64]),
```
davidlghellin: add Decimal128

linhr:
Thanks for your contribution! @lonless9 will take a look at this PR in the next few days.

python/pysail/tests/spark/aggregate/test_agg_try_sum.txt
```diff
@@ -0,0 +1,138 @@
...
+|  7.0|
++-----+
+
+>>> df = spark.createDataFrame([(1e308,), (1e308,)], ["x"])
+>>> df.select(sf.try_sum("x").alias("sum_x")).show()
++--------+
+|   sum_x|
++--------+
+|Infinity|
++--------+
```
lonless9, line 45: `Returns the sum calculated from values of a group and the result is null on overflow.`
Does this case match the doc?

lonless9 (commented):
Great work! 🚀 
Here is my only question.

python/pysail/tests/spark/aggregate/test_agg_try_sum.txt
davidlghellin: I try these in pyspark (because not remember hehe), and these is the ouput 
image

lonless9: Looks like there is some mismatch in the documentation. 
Then I'll merge this PR. Well done!

PR #969, opened 2025-10-13
fix: correct subset logic in get_edges_contained_in_set

davidlghellin:
I'm not entirely sure I’ve fully understood this part, so I’ve added some tests to verify and make sure my understanding is correct.

@lonless9 could you please take a look to confirm this change makes sense?

### DIFF
--- a/crates/sail-physical-optimizer/src/join_reorder/cardinality_estimator.rs
+++ b/crates/sail-physical-optimizer/src/join_reorder/cardinality_estimator.rs
@@ -511,12 +511,14 @@ mod tests {
         use datafusion::physical_plan::empty::EmptyExec;
 
         let mut graph: QueryGraph = QueryGraph::new();
-        let schema: Arc<Schema> = Arc::new(Schema::new(vec![Field::new("id", DataType::Int32, false)]));
+        let schema: Arc<Schema> =
+            Arc::new(Schema::new(vec![Field::new("id", DataType::Int32, false)]));
 
         // 3 relations: R0, R1, R2
         for (id, rows) in [(0, 1000.0), (1, 2000.0), (2, 3000.0)] {
             let plan: Arc<EmptyExec> = Arc::new(EmptyExec::new(schema.clone()));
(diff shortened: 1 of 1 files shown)


crates/sail-physical-optimizer/src/join_reorder/cardinality_estimator.rs
```diff
@@ -498,4 +498,96 @@ mod tests {
...
+                    column_index: 0,
+                    name: "id".into(),
+                },
+            )],
+        );
+        let _ = graph.add_edge(edge12);
+
+        let estimator: CardinalityEstimator = CardinalityEstimator::new(graph);
+
+        // {0,1} → edge 0–1
```
davidlghellin, line 579:
```sql
SELECT *
FROM R0
JOIN R1 ON R0.id = R1.id;
```

crates/sail-physical-optimizer/src/join_reorder/cardinality_estimator.rs
```diff
@@ -498,4 +498,96 @@ mod tests {
...
+        let s01: JoinSet = JoinSet::new_singleton(0)
+            .unwrap()
+            .union(&JoinSet::new_singleton(1).unwrap());
+        assert_eq!(estimator.get_edges_contained_in_set(s01).len(), 1);
+
+        // {0,1,2} → both edges
```
davidlghellin, line 585:
```sql
SELECT *
FROM R0
JOIN R1 ON R0.id = R1.id
JOIN R2 ON R1.id = R2.id;
```

crates/sail-physical-optimizer/src/join_reorder/cardinality_estimator.rs
```diff
@@ -498,4 +498,96 @@ mod tests {
...
+        let s012 = JoinSet::from_iter([0, 1, 2]).unwrap();
+        assert_eq!(estimator.get_edges_contained_in_set(s012).len(), 2);
+
+        // {0,2} → any
```
davidlghellin:
```suggestion
        // {0,2} → none
```

PR #974, opened 2025-10-17
feat: iceberg rest catalog provider

shehabgamin:
Follow up PR(s):
- Integration tests for Apache Polaris in `crates/sail-catalog-iceberg/tests/rest_integration_test.rs`.
  - [Testcontainers ref](https://github.com/apache/polaris/blob/main/getting-started/minio/docker-compose.yml)

- Address the various `FIXME` comments in `crates/sail-catalog-iceberg/src/provider.rs`.

- Docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,16 +127,16 @@ prost = "0.13.5"
 # The `axum` version must match the one used in `tonic` (replace `RELEASE` with the release we are using):
 #   https://github.com/hyperium/tonic/blob/vRELEASE/tonic/Cargo.toml
 axum = "0.8.4"
-datafusion = { version = "50.1.0", features = ["serde", "avro"] }
-datafusion-common = { version = "50.1.0", features = ["object_store", "avro"] }
-datafusion-datasource = { version = "50.1.0", features = ["compression", "parquet"] }
-datafusion-expr = { version = "50.1.0" }
-datafusion-expr-common = { version = "50.1.0" }
-datafusion-proto = { version = "50.1.0" }
-datafusion-functions = { version = "50.1.0" }
-datafusion-functions-nested = { version = "50.1.0" }
-datafusion-physical-expr = { version = "50.1.0" }
-datafusion-spark = { version = "50.1.0" }
+datafusion = { version = "50.3.0", features = ["serde", "avro"] }
+datafusion-common = { version = "50.3.0", features = ["object_store", "avro"] }
+datafusion-datasource = { version = "50.3.0", features = ["compression", "parquet"] }
+datafusion-expr = { version = "50.3.0" }
+datafusion-expr-common = { version = "50.3.0" }
+datafusion-proto = { version = "50.3.0" }
+datafusion-functions = { version = "50.3.0" }
+datafusion-functions-nested = { version = "50.3.0" }
+datafusion-physical-expr = { version = "50.3.0" }
+datafusion-spark = { version = "50.3.0" }
 datafusion-functions-json = { git = "https://github.com/lakehq/datafusion-functions-json.git", rev = "db8e357" }
 # The `pyo3` version must match the one used in `arrow-pyarrow` (replace `RELEASE` with the release we are using):
 #   https://github.com/apache/arrow-rs/blob/RELEASE/arrow-pyarrow/Cargo.toml


crates/sail-iceberg/src/arrow_conversion.rs
```diff
@@ -1,11 +1,30 @@
...
+use crate::spec::{ListType, MapType, NestedField, PrimitiveType, Schema, StructType, Type};
+
+pub const ICEBERG_ARROW_FIELD_DOC_KEY: &str = "doc";
+
+fn get_field_id(field: &ArrowField) -> i32 {
+    field
+        .metadata()
+        .get(PARQUET_FIELD_ID_META_KEY)
+        .and_then(|x| x.parse().ok())
+        .unwrap_or(0)
```
shehabgamin: Should we throw error here?

crates/sail-object-store/src/layers/runtime.rs
```diff
@@ -177,12 +176,16 @@ impl ObjectStore for RuntimeAwareObjectStore {
...
-        // while tasks spawned by the runtime handle must be `'static`.
-        once(Err(object_store::Error::NotImplemented)).boxed()
+        // FIXME: We cannot run `delete_stream` in a runtime-aware manner because
+        //  the input and output streams are expected to have the lifetime `'a`,
+        //  while tasks spawned by the runtime handle must be `'static`.
+        //  NOTE:
+        //  DataFusion doesn't use `delete_stream`, so we bypass runtime isolation (no spawn).
+        //  This means stream polling occurs on the caller's runtime while the inner store's
+        //  I/O still uses the secondary runtime.
+        self.inner.delete_stream(locations)
```
linhr, line 188: As discussed offline, we need to keep `NotImplemented` since delegating `delete_stream` may cause undefined behaviors. And we need to update the comment here.

crates/sail-common/src/config/application.yaml
```diff
@@ -15,10 +15,12 @@
   default: "8388608"
   description: The stack size in bytes for each thread.
 
-- key: runtime.enable_secondary
+- key: runtime.io_runtime_for_object_store
```
linhr, line 18: It seems the upstream may be able to update `ObjectStore::delete_stream` so that we can fully support runtime-aware object store in the near future. Once that's done, we may be able to remove this configuration option entirely.

I feel the new name is indeed better, but if we were going to remove the option anyway, we could just leave the option as is to avoid the extra migration in the meantime. What do you think?

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -0,0 +1,2258 @@
...
+        let (client, catalog_config) = self.load_client_and_merged_config().await?;
+
+        let CreateDatabaseOptions {
+            if_not_exists,
+            comment,
+            location,
+            properties,
+        } = options;
+
+        let exists = match client
```
linhr, line 561: Are we able to tell if the database exists already from the status of the `create_namespace` API? This additional validation can suffer from race condition.

crates/sail-catalog-iceberg/src/runtime.rs
```diff
@@ -0,0 +1,805 @@
...
+    NamespaceExistsError, PlanTableScanError, RegisterTableError, RenameTableError,
+    RenameViewError, ReplaceViewError, ReportMetricsError, TableExistsError, UpdatePropertiesError,
+    UpdateTableError, ViewExistsError,
+};
+use crate::apis::configuration_api_api::GetConfigError;
+use crate::apis::o_auth2_api_api::GetTokenError;
+use crate::apis::{self, Api, ApiClient, Error};
+use crate::models;
+
+pub struct RuntimeAwareApiClient {
```
linhr, line 20: Ok for now, but I guess the runtime-aware layer can later be implemented for the `CatalogProvider` trait so that all catalogs can benefit from it.

crates/sail-catalog-iceberg/tests/rest_integration_test.rs
```diff
@@ -0,0 +1,1773 @@
...
+    DropDatabaseOptions, DropTableOptions, DropViewOptions, Namespace, TableColumnStatus,
+    TableKind,
+};
+use sail_catalog_iceberg::{IcebergRestCatalogProvider, REST_CATALOG_PROP_URI};
+use sail_common::runtime::RuntimeHandle;
+use testcontainers::core::{ContainerPort, WaitFor};
+use testcontainers::runners::AsyncRunner;
+use testcontainers::{ContainerAsync, GenericImage, ImageExt};
+
+async fn setup_catalog() -> (
```
linhr, line 18: Ok for now, but I feel we can later implement integration tests in Python so that it fully covers the features end-to-end.

Since `setup_catalog` is called on every tests, the overhead of each test is large, and Rust doesn't seem to have good support of conditionally running subsets of tests on demand. For Python tests, there are better support for fixtures, markers etc.

I feel the best approach for now is to add `#[ignore]` for these tests so that they can be triggered on demand with additional flags to `cargo test`. By default, these expensive tests are not run to avoid impacting developer experience when not working on catalog code changes.

crates/sail-common/src/config/application.yaml
shehabgamin: Works with me!

crates/sail-catalog-iceberg/src/provider.rs
shehabgamin: Oh nice catch. I meant to remove this, as it's not necessary at all. The match on `result` after calling `create_namespace` already handles it.

crates/sail-catalog-iceberg/src/runtime.rs
shehabgamin: Good call!

crates/sail-catalog-iceberg/tests/rest_integration_test.rs
shehabgamin: Whatever works for me. We can ignore by default so that local development is not slowed down and then make sure to run the tests in Github actions.

PR #975, opened 2025-10-18
chore: execution configs

shehabgamin:
- `execution.use_row_number_estimates_to_optimize_partitioning` now exposed
- `execution.batch_size` has a high default and could be contributing to OOMs
- `parquet.maximum_buffered_record_batches_per_stream` has a high default and could be contributing to OOMs

### DIFF
--- a/crates/sail-common/src/config/application.yaml
+++ b/crates/sail-common/src/config/application.yaml
@@ -195,7 +195,7 @@
 
 - key: execution.batch_size
   type: number
-  default: "16384"
+  default: "8192"
   description: The batch size for physical plan execution.
   experimental: true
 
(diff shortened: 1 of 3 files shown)


linhr (approved):

crates/sail-common/src/config/application.yaml
```diff
@@ -195,7 +195,7 @@
 
 - key: execution.batch_size
   type: number
-  default: "16384"
+  default: "8192"
```
davidlghellin, line 198: interesting number, like clickhouse https://github.com/search?q=repo%3AClickHouse%2FClickHouse%208192&type=code

Issue #172, opened 2024-09-08
Iceberg Integration


linhr:
https://github.com/apache/iceberg-rust

shehabgamin:
https://iceberg.apache.org/docs/latest/spark-getting-started/

fafacao86:
> We plan to maintain our in-house Iceberg implementation based on existing libraries such as https://github.com/apache/iceberg-rust.
> 
@linhr  I'm curious why  Lake Sail plans to maintain an in-house iceberg implementation, could you please share some motivation? I'm trying to build an OLAP database on top of iceberg, and I find it hard to extend iceberg often times.

linhr:
> I'm curious why Lake Sail plans to maintain an in-house iceberg implementation, could you please share some motivation? I'm trying to build an OLAP database on top of iceberg, and I find it hard to extend iceberg often times.

We are working on deep integration of Iceberg and other lakehouse formats into Sail's query planning and distributed execution mechanisms. We found it's currently best to maintain our own Iceberg implementation as the existing ones lack integration points to the internals for our needs. That being said, we are still part of the `iceberg-rust` community and we've been involved in a few design discussions. We do hope that the official implementation would become more feature-complete and modular over time, so that it can provide shared building blocks for downstream projects like Sail.

fafacao86:
> > I'm curious why Lake Sail plans to maintain an in-house iceberg implementation, could you please share some motivation? I'm trying to build an OLAP database on top of iceberg, and I find it hard to extend iceberg often times.
> 
> We are working on deep integration of Iceberg and other lakehouse formats into Sail's query planning and distributed execution mechanisms. We found it's currently best to maintain our own Iceberg implementation as the existing ones lack integration points to the internals for our needs. That being said, we are still part of the `iceberg-rust` community and we've been involved in a few design discussions. We do hope that the official implementation would become more feature-complete and modular over time, so that it can provide shared building blocks for downstream projects like Sail.

Got it, thanks for your reply!

atakanyenel:
I'm testing sail to read from iceberg-on-hms tables. I built from https://github.com/lakehq/sail/pull/1929 commit and reading iceberg tables work. But **writing doesn't.** Especially *writeTo/createTable* functions are failing. What is the timeline for te write ability for iceberg tables ?

linhr:
[...]

PR #976, opened 2025-10-20
feat: basic write operation for Iceberg

lonless9:
#172

### DIFF
--- a/python/pysail/tests/spark/iceberg/test_iceberg_partitioned_reads.py
+++ b/python/pysail/tests/spark/iceberg/test_iceberg_partitioned_reads.py
@@ -73,7 +73,7 @@ def _append_sample_data(table):
     ("table_name", "spec", "predicate_column", "predicate_value", "expected_numbers"),
     [
         pytest.param(
-            # FIXME: this test is failing in `local-cluster` mode because of 
+            # FIXME: this test is failing in `local-cluster` mode because of
             # the proto issue related to `PartitionedFile`.
             "default.test_partitioned_by_identity",
             PartitionSpec(PartitionField(3, 1001, IdentityTransform(), "ts")),


crates/sail-iceberg/src/datasource/provider.rs
```diff
@@ -128,11 +114,14 @@ impl IcebergTableProvider {
...
+        let table_url_parsed = Url::parse(&self.table_uri)
+            .map_err(|e| datafusion::common::DataFusionError::External(Box::new(e)))?;
+
         let manifest_list_path = if let Ok(url) = Url::parse(manifest_list_str) {
             log::trace!("Parsed manifest list as URL, path: {}", url.path());
-            ObjectPath::from(url.path())
+            ObjectPath::from(url.path().strip_prefix('/').unwrap_or(url.path()))
         } else {
-            ObjectPath::from(manifest_list_str)
+            ObjectPath::from(table_url_parsed.path()).child(manifest_list_str)
```
lonless9: I'm not confident in parts related to parsing `manifest list`. Don't konw if there's a better way rather than all the changes in `provider.rs`.

crates/sail-iceberg/src/physical_plan/commit/commit_exec.rs
```diff
@@ -0,0 +1,651 @@
...
+        let input_stream = self.input.execute(0, Arc::clone(&context))?;
+
+        let table_url = self.table_url.clone();
+        let schema = self.schema();
+        let future = async move {
+            let object_store = get_object_store_from_context(&context, &table_url)?;
+            let table_path = table_url
+                .path()
+                .strip_prefix('/')
+                .unwrap_or(table_url.path());
```
lonless9, line 126: Check here

crates/sail-iceberg/src/io/mod.rs
```diff
@@ -0,0 +1,49 @@
+use std::sync::Arc;
+
+use bytes::Bytes;
+use object_store::path::Path as ObjectPath;
+
+#[derive(Clone)]
+pub struct IcebergObjectStore {
+    pub object_store: Arc<dyn object_store::ObjectStore>,
+    pub root: ObjectPath,
+}
```
lonless9, line 14: I'm not sure if we need this abstraction, it's a very simplified version of storage, and maybe help with subsequent implementations related to cache.

crates/sail-iceberg/src/writer/partition.rs
```diff
@@ -0,0 +1,264 @@
...
+    let dir = build_partition_dir(spec, iceberg_schema, &values);
+    Ok((values, dir))
+}
+
+/// Group a RecordBatch by partition values according to the given spec and schema.
+pub fn group_by_partition(
+    batch: &RecordBatch,
+    spec: &PartitionSpec,
+    iceberg_schema: &IcebergSchema,
+) -> Result<Vec<PartitionBatchResult>, String> {
```
lonless9: This method could be redundant.

crates/sail-iceberg/src/writer/table_writer.rs
```diff
@@ -0,0 +1,189 @@
...
+        Ok(out)
+    }
+
+    pub async fn write(&mut self, batch: &RecordBatch) -> Result<(), String> {
+        let groups = group_by_partition(
+            batch,
+            &self.config.partition_spec,
+            &self.config.iceberg_schema,
+        )?;
+        for g in groups {
```
lonless9: Maybe we need to assume input batches are already partitioned and extract partition values only from the first row of each batch.

crates/sail-iceberg/src/physical_plan/commit/commit_exec.rs
```diff
@@ -0,0 +1,651 @@
...
+                .map_err(|e| DataFusionError::External(Box::new(e)))?
+                .bytes()
+                .await
+                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+            let mut table_meta: TableMetadata = TableMetadata::from_json(&bytes)
+                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+            log::trace!(
+                "commit_exec: loaded table metadata with snapshot id: {:?}",
+                table_meta.current_snapshot_id
+            );
```
lonless9: Maybe we don't need to reload here.

crates/sail-iceberg/src/io/mod.rs
linhr: Delta Lake uses `PrefixStore`. Would that be helpful for Iceberg as well?

python/pysail/tests/spark/iceberg/utils.py
```diff
@@ -1,7 +1,14 @@
-from pathlib import Path
+from __future__ import annotations
```
linhr: nitpick: We probably don't need this line since we only support recent Python versions.

python/pysail/tests/spark/iceberg/test_iceberg_write.py
```diff
@@ -0,0 +1,89 @@
...
+            NestedField(field_id=2, name="event", field_type=StringType(), required=False),
+            NestedField(field_id=3, name="score", field_type=DoubleType(), required=False),
+        ),
+    )
+    try:
+        df = spark.createDataFrame(
+            [(10, "A", 0.98), (11, "B", 0.54), (12, "A", 0.76)],
+            schema="id LONG, event STRING, score DOUBLE",
+        )
+        df.write.format("iceberg").mode("overwrite").save(table.location())
```
linhr, line 25: We could also test overwrite for a table with existing data.

python/pysail/tests/spark/iceberg/test_iceberg_cross_compat.py
```diff
@@ -0,0 +1,104 @@
...
+            pd.DataFrame({"id": [1, 2], "event": ["a", "b"]})
+            .astype({"id": "int64"})
+            .assign(event=lambda d: d["event"].astype("object"))
+        )
+        assert_frame_equal(actual_py, expected_py)
+    finally:
+        catalog.drop_table(identifier)
+
+
+def test_pyiceberg_read_after_sail_multiple_files(spark, tmp_path):
```
linhr: It seems this test has the same logic as the previous one.

python/pysail/tests/spark/iceberg/test_iceberg_cross_compat.py
```diff
@@ -0,0 +1,104 @@
...
+        expected = (
+            pd.DataFrame({"id": [1, 2, 3, 4], "event": ["a", "b", "c", "d"]})
+            .astype({"id": "int64"})
+            .assign(event=lambda d: d["event"].astype("object"))
+        )
+        spark_df = spark.read.format("iceberg").load(table.location()).sort("id").toPandas()
+        assert_frame_equal(spark_df, expected)
+
+        expected_py = (
+            pd.DataFrame({"id": [1, 2], "event": ["a", "b"]})
```
linhr, line 69: Why is the appended data missing for the PyIceberg reader?

crates/sail-iceberg/src/writer/config.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub write_batch_size: usize,
+    pub num_indexed_cols: i32,
+    pub stats_columns: Option<Vec<String>>,
+    pub iceberg_schema: Arc<IcebergSchema>,
+    pub partition_spec: UnboundPartitionSpec,
+}
+
+impl WriterConfig {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
linhr: We probably don't need this method since all the fields are `pub` so we can construct the struct directly.

crates/sail-iceberg/src/writer/table_writer.rs
linhr: Yeah data partitioning seems something that should be done by a separate physical node. Doing this via `group_by_partition` is probably not scalable...

We can add a TODO for this. This is probably not a quick change though. I feel this requires some architecture design since this is a very common functionality applicable for Delta Lake and raw Parquet tables as well.

crates/sail-iceberg/src/writer/table_writer.rs
```diff
@@ -0,0 +1,189 @@
...
+            store,
+            config,
+            writers: HashMap::new(),
+            partition_values_map: HashMap::new(),
+            written: Vec::new(),
+            partition_spec_id,
+        }
+    }
+
+    fn cast_batch_to_schema(
```
linhr, line 53: This seems a general utility function that can be moved to `sail-common-datafusion` (or maybe we already have it somewhere). No need to do it now, but let's add a TODO for it.

crates/sail-iceberg/src/datasource/provider.rs
linhr: Is `manifest_list_str` an absolute path? Should paths in `self.snapshot` be relative so that the table can be moved to a different location?

linhr (approved):
Great work!! 🚀

crates/sail-iceberg/src/io/mod.rs
lonless9: Great proposal, I've added relevant content.

crates/sail-iceberg/src/writer/table_writer.rs
lonless9: I removed relevant logic, and I agree surely we can do that in the future when we optimize the physics plan.

crates/sail-iceberg/src/physical_plan/commit/commit_exec.rs
lonless9: Got this loading removed.

python/pysail/tests/spark/iceberg/utils.py
lonless9: Done!

python/pysail/tests/spark/iceberg/test_iceberg_write.py
lonless9: Done!

python/pysail/tests/spark/iceberg/test_iceberg_cross_compat.py
lonless9: This could be things related to catalog.

crates/sail-iceberg/src/writer/config.rs
lonless9: Done!

python/pysail/tests/spark/iceberg/test_iceberg_cross_compat.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+    table = catalog.create_table(identifier=identifier, schema=schema)
+    try:
+        df = spark.createDataFrame(
+            [(10, "A", 0.98), (11, "B", 0.54), (12, "A", 0.76)],
+            schema="id LONG, event STRING, score DOUBLE",
+        )
+        df.write.format("iceberg").mode("overwrite").save(table.location())
+
+        static_table = StaticTable.from_metadata(table.location())
```
lonless9, line 94: Added some tests related to `StaticTable`

crates/sail-iceberg/src/io/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+                other => Err(format!("object_store head error: {other}")),
+            })
+    }
+}
+
+#[derive(Clone)]
+pub struct StoreContext {
+    pub base: Arc<dyn object_store::ObjectStore>,
+    pub prefixed: Arc<dyn object_store::ObjectStore>,
+}
```
lonless9, line 59: Added this abstraction.

crates/sail-iceberg/src/physical_plan/commit/commit_exec.rs
```diff
@@ -0,0 +1,656 @@
...
+            .await
+            .map_err(DataFusionError::Execution)?;
+            log::trace!("Metadata written successfully");
+
+            // Update version-hint with the metadata filename (not version number)
+            let metadata_filename = if let Some(fname) = new_meta_rel.rsplit('/').next() {
+                fname.to_string()
+            } else {
+                new_meta_rel.clone()
+            };
```
lonless9, line 621: Fixed `version-hint` issue

PR #986, opened 2025-10-24
fix: add file group when creating writer for listing table format

shehabgamin:


linhr:
I reviewed all usages of `file_group` of `FileSinkConfig` in the DataFusion codebase but couldn’t find anything meaningful (other than being used in `DisplayAs`). I looked at the history and it was first introduced [over two years ago](https://github.com/apache/datafusion/commit/36292f63e3c7dfbd5bca77f5e36c2aa43048d2be) for appending to the same file for existing CSV external tables. This is clearly not how we handle the append operation since we always write new files. And the logic has been removed and I cannot find reference to `file_group` in current version of `CsvSink` anymore. The file group construction logic in `ListingTable::insert_into` seems a leftover.

Conceptually, `file_group` for sink does not make sense to me either. I don't see why the existing file paths and statistics can be made useful for writing new data.

Due to the overhead introduced in the listing operation and unclear benefits, I'd suggest closing this without merging.

PR #1005, opened 2025-10-31
feat: unity catalog

shehabgamin:
- Initial Unity Catalog implementation.
- Implement General runtime-aware layer for all catalog providers .
- Fix inconsistent field naming of `List` data types across code base (noticed when integrating Unity).

Part of https://github.com/lakehq/sail/issues/171

### DIFF
--- a/crates/sail-catalog-unity/spec/unity-catalog-all.yaml
+++ b/crates/sail-catalog-unity/spec/unity-catalog-all.yaml
@@ -1,23 +1,4 @@
-#
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
-#
-#   http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an
-# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-# KIND, either express or implied.  See the License for the
-# specific language governing permissions and limitations
-# under the License.
-#
-
-# SOURCE: https://github.com/unitycatalog/unitycatalog/blob/b4369440ab250f4d6fbacf2d0e1b25273c4ff0d6/api/all.yaml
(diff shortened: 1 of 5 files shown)


crates/sail-catalog-unity/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(clippy::unwrap_used)]
```
shehabgamin: Generated code uses unwrap, which I don't like. Not sure if there is a way around this...

crates/sail-catalog-unity/src/config.rs
```diff
@@ -0,0 +1,426 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
linhr: The license header does not seem correct. The Delta Lake project is not owned by ASF.

crates/sail-catalog-unity/spec/unity-catalog-all.yaml
```diff
@@ -0,0 +1,2852 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
```
linhr: The license header does not seem correct. The Unity Catalog project is not owned by ASF.

crates/sail-plan/src/resolver/tree/explode.rs
```diff
@@ -116,14 +117,24 @@ impl TreeNodeRewriter for ExplodeRewriter<'_> {
             }?,
             (ExplodeDataType::Map, false, _) => {
                 vec![
-                    ident(&name).field("key").alias("key"),
-                    ident(&name).field("value").alias("value"),
+                    ident(&name)
+                        .field(SAIL_MAP_KEY_FIELD_NAME)
+                        .alias(SAIL_MAP_KEY_FIELD_NAME),
```
linhr: The alias should be hardcoded since this is a Spark convention.

docs/guide/catalog/unity-catalog.md
```diff
@@ -0,0 +1,48 @@
...
+
+Unity Catalog can be configured using the following options:
+
+- `type` (required): The string `unity`.
+- `name` (required): The name of the catalog.
+- `uri` (optional): The base URI of the Unity Catalog API. Defaults to `http://localhost:8080/api/2.1/unity-catalog`.
+- `default_catalog` (optional): The default catalog name to use. Defaults to `unity`.
+- `token` (optional): The authentication token to use when connecting to the Unity Catalog API. Defaults to empty (no authentication). Refer to the [Unity Catalog Auth Docs](https://docs.unitycatalog.io/server/auth/) for more information.
+
+## Environment Variables
```
linhr, line 18: Ok for now, but in the future we may need to support the options in the catalog provider. Having multiple Unity Catalog instances would not be possible if we read the options from environment variables.

PR #1006, opened 2025-10-31
fix: ensure absolute paths for cross-engine Iceberg reads

lonless9:


crates/sail-iceberg/src/writer/table_writer.rs
```diff
@@ -269,9 +273,16 @@ impl IcebergTableWriter {
...
                 &full
             );
-            // Use absolute filesystem path for cross-compat (PyIceberg expects an absolute file path)
-            let abs_path = format!("/{}", full);
-            let df = DataFileWriter::new(self.partition_spec_id, abs_path, partition_values)
+            // Local FS needs absolute paths for cross-engine reads; object stores use relative paths
+            let scheme = self.table_url.scheme();
+            let file_path = if scheme.is_empty() || scheme == "file" {
+                // Use the fully-resolved object_store path (which preserves any encoding)
+                format!("/{}", full)
```
linhr: Does this work on Windows?

lonless9: Path like `/C:\path\to\file` won't work on Windows, so it's incorrect here.

linhr (approved):
🚀

PR #1024, opened 2025-11-02
fix: timestamp data type support for to_date function

### LINKED ISSUE #1007: Unsupported data type `Timestamp(Microsecond, None)` for function `to_date`
I was trying lakesail and got this error.

Dataset: [NYC taxi yellow dataset](https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2025-09.parquet).

Code:
```python
from pyspark.sql import SparkSession, DataFrame
from common.read_process import read_parquet_nyc_data, compute_complex_query
from common.timing import timing_decorator

server = SparkConnectServer()
server.start()
addr, port = server.listening_address
print(f'spark {addr:}, {port:}')

df = spark.read.parquet("./data/yellow_tripdata_2025-09.parquet")

# df.printSchema()
# df.selectExpr('date_format(tpep_pickup_datetime, "yyyy-MM-dd")').show(1)
dd = df.selectExpr(
    'tpep_pickup_datetime',
    'to_timestamp(tpep_pickup_datetime) to_timestamp',
    'to_date(to_timestamp(date_format(tpep_pickup_datetime, "yyyy-MM-dd HH:mm:ss"))) to_date_to_timestamp',
)

dd.printSchema()
dd.show(1)
```
Error:
```shell
pyspark.errors.exceptions.connect.AnalysisException: Unsupported data type Timestamp(Microsecond, None) for function to_date
```

I have tried to look at the schema, like using `parquet-cli` (brew):
```
{
    "name" : "tpep_pickup_datetime",
    "type" : [ "null", {
      "type" : "long",
      "logicalType" : "local-timestamp-micros"
    } ],
    "default" : null
  }
```

shehabgamin:
Closes https://github.com/lakehq/sail/issues/1007


crates/sail-plan/src/function/scalar/conversion.rs
```diff
@@ -22,11 +21,14 @@ fn date(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
             args: vec![arg],
         }))
     } else {
-        Ok(expr_fn::to_date(vec![arg]))
+        Ok(expr::Expr::Cast(expr::Cast::new(
```
shehabgamin, line 24: Expressions are typically more performant than UDFs.

crates/sail-plan/src/function/scalar/conversion.rs
```diff
@@ -40,7 +42,10 @@ fn timestamp(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
             args: vec![arg],
         }))
     } else {
-        Ok(expr_fn::to_timestamp_micros(vec![arg]))
+        Ok(expr::Expr::Cast(expr::Cast::new(
```
shehabgamin, line 45: Expressions are typically more performant than UDFs.

crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -173,11 +173,17 @@ fn to_chrono_fmt(format: Expr) -> Expr {
...
-    } else if args.len() == 2 {
-        let (expr, format) = args.two()?;
+fn to_date(input: ScalarFunctionInput) -> PlanResult<Expr> {
+    if input.arguments.len() == 1 {
+        // If format is not supplied, the function is a synonym for cast(expr AS DATE).
+        crate::function::scalar::conversion::cast_to_date(input)
+    } else if input.arguments.len() == 2 {
+        let (expr, format) = input.arguments.two()?;
+        let expr = match expr.get_type(input.function_context.schema) {
+            Ok(DataType::Timestamp(_time_unit, _tz)) => cast(expr, DataType::Utf8),
```
linhr, line 183: In JVM Spark, when the first argument is `timestamp`, the second argument seems ignored. But here, an error will be returned if the format is invalid or is not the same as the default timestamp format.

I feel a more performant implementation would just convert the timestamp to local timestamp and then extract the date from it, without involving formatting. This could be future work. (Since we have implemented most datetime functions, there might be some function in our codebase already to handle local timestamp. But I haven't looked deeper into this.)

linhr (approved):
Thanks for the fix!

crates/sail-plan/src/function/scalar/datetime.rs
nis12ram: Yes @linhr, you are right. We can convert to local timestamp directly using `expr_fn::to_local_time` method and then cast it to `DataType::Date32` without using format.

I’ve opened a PR for this https://github.com/lakehq/sail/pull/2025https://github.com/lakehq/sail/pull/2025

PR #1032, opened 2025-11-07
test: set up SQL BDD tests

linhr:
This PR is a proof-of-concept that support writing SQL tests using [Gherkin](https://cucumber.io/docs/gherkin/reference) syntax. The `pytest-bdd` library is used to run the tests within pytest and utilize the powerful fixture system.

The idea is similar to sqllogictest (SLT) but SLT is not easily customizable unless we implement our own parser and test runner to extent it with our own logic for test setup, result comparison, etc.

Compared with writing Python tests or Python doctests, the `.feature` file does put some restriction on the expressiveness of test assertions, since each step needs to have a corresponding function defined in Python. But the restriction would make the test more readable, compared with tests written in Python/Java/Rust code.

There are plenty of opportunities after this PR:
1. Port function gold tests and doctests to Gherkin format. The `.feature` file can also contain the description for the function, and we can tag the tests to be included in the documentation. This is a clear path for generating documentation for all SQL functions.
2. Write most lakehouse tests in Gherkin format. (Tests for the DataFrame API and tests involving complex assertions can remain in its current form.) (1) We could port tests from other repositories using AI tools. I feel AI could generate Gherkin tests more reliably than Python tests, and it's easier for us to review the results. The correctness of the tests can be verified by running the tests against JVM Spark. (2) Some tests can be tagged for use in the documentation.

### DIFF
--- a/python/pysail/tests/spark/conftest.py
+++ b/python/pysail/tests/spark/conftest.py
@@ -229,4 +229,4 @@ def query_error(error, query, spark):
     that will be used to partially match the actual error message.
     """
     with pytest.raises(Exception, match=error):
-        spark.sql(query)
+        _ = spark.sql(query).collect()


pyproject.toml
```diff
@@ -146,8 +152,6 @@ dependencies = [
     "pytest-xdist>=3.7,<4",
     "pytest-timeout>=2.4,<3",
     "pytest-reportlog>=0.4,<0.5",
-    "pyiceberg[sql-sqlite,pyiceberg-core]==0.10.0",
-    "pydantic>=2.11,<2.12",
```
linhr, line 150: This is not needed for the `test-spark` environment.

python/pysail/tests/spark/conftest.py
```diff
@@ -91,3 +94,139 @@ def pytest_collection_modifyitems(session, config, items):  # noqa: ARG001
...
+        """The string representation of the path."""
+        return str(self.path)
+
+    @property
+    def sql(self):
+        """The corresponding SQL string literal for the path."""
+        return f"'{escape_sql_string_literal(str(self.path))}'"
+
+
+@given(parsers.parse("variable {name} for temporary directory {directory}"), target_fixture="variables")
```
linhr, line 133: We could add more `@then` steps later to define the expected content in the directory after lakehouse data is written. This could be a more standardized approach than assertions in Python.

linhr:
I was able to run the tests against JVM Spark using the following command:

```bash
env \
  SPARK_REMOTE="local" \
  PYSPARK_SUBMIT_ARGS='--packages io.delta:delta-spark_2.13:4.0.0 --conf "spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension" --conf "spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog" pyspark-shell' \
  hatch run pytest -k "test_delta_features" -vv
```

See also:
* <https://docs.delta.io/quick-start/#pyspark-shell>
* <https://docs.lakesail.com/sail/latest/development/build/python.html#testing-with-jvm-spark>

lonless9 (approved):
Great to see this idea come true! 🚀

Issue #501, opened 2025-05-27
[EPIC] Full Coverage of Ibis Tests


shehabgamin:
Ibis tests are not enabled in CI yet but can be run locally.

<https://docs.lakesail.com/sail/latest/development/spark-tests/test-ibis.html>

This issue tracks the remaining work needed to fully support Ibis with the Spark backend.

davidlghellin:
deprecated msg, its before update bump ibis

davidlghellin:
deprecated msg, its before update bum ibis

davidlghellin:
related https://github.com/ibis-project/ibis/issues/11859

davidlghellin:
https://github.com/ibis-project/ibis/commit/e9c35e6a3574855daf4e02f4217727df0d54872d

PR #1033, opened 2025-11-08
feat: add CI ibis test

davidlghellin:
part https://github.com/lakehq/sail/issues/501

### DIFF
--- a/.github/workflows/ibis-tests.yml
+++ b/.github/workflows/ibis-tests.yml
@@ -51,6 +51,12 @@ jobs:
           repository: ibis-project/testing-data
           path: opt/ibis-testing-data
 
+      - name: Download Python Package
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package
+          path: target/wheels
+
       - name: Build native extension
         run: |
           hatch run install-pysail
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -44,10 +44,3 @@ jobs:
           workflow_run_id: ${{ github.event.workflow_run.id }}
           artifact_name: gold-data-report
           report_name: gold-data
-
-      - uses: ./.github/actions/pull-request-report
-        if: github.event.workflow_run.name == 'Ibis Tests' && github.event.workflow_run.conclusion == 'success'
-        with:
-          workflow_run_id: ${{ github.event.workflow_run.id }}
-          artifact_name: ibis-report
-          report_name: ibis


davidlghellin:
need to write in a comment in PR
image

davidlghellin:
these is the output format:
```
## 🧪 Ibis Tests Summary
= 262 failed, 1290 passed, 154 skipped, 37678 deselected, 171 xfailed, 37 xpassed, 702 warnings, 66 errors in 252.02s (0:04:12) =
```

I think when merged, we write in PR

davidlghellin:
try install maturin hatch (not uses pip) [72d9082](https://github.com/lakehq/sail/pull/1033/commits/72d90829a10846c09013be421f4a3e86fad43ff5)

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+        with:
+          pattern: "\\[(ibis )?tests?\\]"
+          ignore_case: true
+
+      - uses: actions/github-script@v7
+        id: label
+        with:
+          script: |
+            const labels = context.payload.pull_request?.labels?.map(label => label.name) || [];
+            core.setOutput('has_label', labels.includes('ibis tests') ? 'true' : 'false');
```
linhr: The label should be `run ibis tests`. I'm going to create the label after this PR is merged.

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable
+
+      - name: Setup Protobuf Compiler
+        uses: arduino/setup-protoc@v3
+
+      - name: Clone IBIS
+        run: |
+          git clone https://github.com/ibis-project/testing-data.git opt/ibis-testing-data
```
linhr: Let's use `actions/checkout@v4` with the `path` parameter for the clone location.

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Install Hatch and Maturin
+        run: |
+          pip install hatch maturin
```
linhr: Let's use `pypa/hatch@install`.

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Build native extension
+        run: |
+          hatch run maturin develop
+          hatch run maturin build
```
linhr: We don't need to build the package. Let's follow `spark-tests.yml` to download the Python package built in the previous step and use `hatch run install-pysail`.

.github/workflows/report.yml
```diff
@@ -4,7 +4,7 @@ on:
   # When this workflow is triggered, the workflow definition comes from the `main` branch.
   # If a pull request modifies this file, it will have an effect only after the pull request is merged.
   workflow_run:
-    workflows: ["Build", "Gold Data"]
+    workflows: ["Build", "Gold Data", "Ibis Tests"]
```
linhr: This is not needed.

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+      # Set the Hatch environment for `hatch run` commands.
+      HATCH_ENV: "test-ibis"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: "corretto"
+          java-version: "17"
```
linhr: This is not needed.

.github/workflows/ibis-tests.yml
```diff
@@ -0,0 +1,122 @@
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install stable
+
+      - name: Setup Protobuf Compiler
+        uses: arduino/setup-protoc@v3
```
linhr: This won't be needed if we don't build the package.

.github/workflows/ibis-tests.yml
davidlghellin: Thanks for the suggestion! I was doing it the other way and didn't know about this option, which is why I thought I needed those dependencies. I appreciate the clarification.

davidlghellin:
Sorry about that — I wasn’t aware of the different ways to install Sail and had only been using the compile-from-source method. Thanks @linhr

PR #1042, opened 2025-11-11
fix: iceberg path handling on Windows platform

lonless9:
The previous time travel for iceberg implementation used `ListingTableUrl`, which is wrong and causing some iceberg test failure.

Besides, this PR also introduces some improvements for Iceberg:

1.  UUID-Based Snapshot ID Generation
2.  Monotonic Timestamp Generation (which needs some design consideration later)

### DIFF
--- a/crates/sail-iceberg/src/physical_plan/commit/bootstrap.rs
+++ b/crates/sail-iceberg/src/physical_plan/commit/bootstrap.rs
@@ -107,12 +107,13 @@ pub async fn bootstrap_new_table(
         .ok_or_else(|| DataFusionError::Plan("No snapshot in bootstrap commit".to_string()))?;
 
     // Build minimal TableMetadata V2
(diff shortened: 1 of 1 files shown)


crates/sail-iceberg/src/table_format.rs
```diff
@@ -1,184 +1,56 @@
...
-        options: TableIcebergOptions,
-    ) -> Result<Arc<dyn TableProvider>> {
-        let (schema, snapshot, partition_specs) =
-            load_table_metadata_with_options(ctx, &table_url, options).await?;
-
-        let provider =
-            IcebergTableProvider::new(table_url.to_string(), schema, snapshot, partition_specs)?;
-        Ok(Arc::new(provider))
+/// Parse a table URL from a path string
+pub async fn parse_table_url(ctx: &dyn Session, paths: Vec<String>) -> Result<Url> {
```
linhr, line 27: This seems a duplicated function.

(No action is needed for now. This can happen as we work on cleanup in the near future.)

PR #1054, opened 2025-11-13
chore: apache license header

shehabgamin:
I just spoke with the Linux Foundation because folks from Delta Lake (Databricks) raised a concern that we had not attributed credit properly. They provided a screenshot as evidence, which is attached. **If you look at line 30 in the screenshot, you can see that we clearly do attribute credit.**

image

Before the Linux Foundation showed me the screenshot, I explained that we have always attributed credit with exact links in our code files and that this has been our practice since the very first day of Sail. They then noticed that the screenshot itself clearly included the attribution. The LF representative felt bad about the situation and said it seemed like Delta Lake was acting in bad faith, but he was only relaying the message he had been given.

I asked what more we could possibly do and told the LF representative that if there was anything else needed, we would absolutely do it. I also told the LF representative that the accusation really personally hurt because we have consistently gone above and beyond in attributing credit in our code, and I have rarely seen other projects maintain this level of diligence. I also mentioned our contributions to `delta-kernel-rs` and the fact that since `DataFusion` is tested on Sail every month, we end up testing `delta-rs` as well because `delta-rs` depends on `DataFusion`. We do all this work for the broader community without ever advertising it.

The LF representative agreed that we clearly attribute credit and apologized for wasting my time. He said that the only additional step we could take would be to include the Apache License header at the top of each file, but also noted that this is a moot point since Sail is already an open-source project licensed under Apache 2.0.

Moving forward, let’s continue attributing credit using the `// [Credit]:` format, as we have always done since the very beginning. In addition, for every file in `sail-catalog-iceberg`, `sail-catalog-unity`, `sail-delta-lake`, `sail-iceberg`, and any related locations, let’s also include the Apache License header at the top.

### EDIT:
Adding this here as well for easier reading.
On a related note, @lonless9 contributed a PR in Delta-rs to upgrade https://github.com/delta-io/delta-rs/pull/3933 earlier today...

### EDIT 2:
This PR introduced incorrect license headers in several places. The issue is resolved by https://github.com/lakehq/sail/pull/1055, which fixes the incorrect headers.

### DIFF
--- a/crates/sail-catalog-iceberg/Cargo.toml
+++ b/crates/sail-catalog-iceberg/Cargo.toml
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
 [package]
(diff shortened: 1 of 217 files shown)


shehabgamin:
On a related note, @lonless9 contributed a PR in Delta-rs to upgrade https://github.com/delta-io/delta-rs/pull/3933 earlier today...

PR #1071, opened 2025-11-20
feat: add `percentile` aggregate function

davidlghellin:
Part of https://github.com/lakehq/sail/issues/309

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.percentile.html?highlight=percentile


davidlghellin:
closes #1072

crates/sail-execution/src/codec.rs
```diff
@@ -1709,6 +1710,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             || node.inner().as_any().is::<MaxByFunction>()
             || node.inner().as_any().is::<MinByFunction>()
             || node.inner().as_any().is::<ModeFunction>()
+            || node.inner().as_any().is::<PercentileFunction>()
```
davidlghellin, line 1844: need `PercentileDiscFunction`

crates/sail-execution/src/codec.rs
```diff
@@ -1709,6 +1711,8 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
...
+            || node.inner().as_any().is::<PercentileDiscFunction>()
```
linhr: We need to handle these two aggregate functions in `try_decode_udaf` as well.

crates/sail-function/src/aggregate/percentile.rs
```diff
@@ -0,0 +1,1291 @@
...
+            arrow::datatypes::Field::new("percentile", DataType::Float64, false).into(),
+            arrow::datatypes::Field::new("data_type_id", DataType::UInt8, false).into(),
+        ])
+    }
+
+    fn accumulator(&self, acc_args: AccumulatorArgs) -> Result<Box<dyn Accumulator>> {
+        let data_type = &acc_args.exprs[0].data_type(acc_args.schema)?;
+
+        // Extract the percentile value from the second argument if present
+        let percentile = 0.5f64;
```
linhr: The second argument must be a literal so we can use `LiteralEvaluator` to get the percentile here. The accumulator can then just ignore the second column in the input batch data.

Also `percentile` accepts an array of percentiles. We can add a TODO for this.

crates/sail-spark-connect/tests/gold_data/function/agg.json
```diff
@@ -2027,7 +2027,7 @@
         }
       },
       "output": {
-        "failure": "not implemented: unknown aggregate function: percentile_cont"
+        "success": "ok"
```
linhr, line 2414: The gold tests right now do not validate whether the results are correct or not. The tests are marked as `ok` if the SQL statements complete without errors. I'm actually not sure if we support the `WITHIN GROUP` syntax properly. Could you have a look?

crates/sail-function/src/aggregate/percentile.rs
```diff
@@ -0,0 +1,1291 @@
...
+    fn size(&self) -> usize {
+        std::mem::size_of_val(&self.values)
+            + self.values.capacity() * std::mem::size_of::<i64>()
+            + std::mem::size_of::<f64>()
+            + std::mem::size_of::<DataType>()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
linhr: It's Ok to skip Rust unit tests as long as the Python tests cover similar situations.

python/pysail/tests/spark/aggregate/test_agg_percentile.txt
linhr, line 1: Nice tests!

crates/sail-function/src/aggregate/percentile.rs
davidlghellin: I've been looking into using the `LiteralEvaluator` expression directly, but it's defined in the physical plan crate, and importing it would create a circular dependency. Because of that, I tried evaluating the percentile argument using a dummy RecordBatch. It took me some time to get that part working, and for now it's just a temporary workaround. I hope to improve this solution soon and replace it with something cleaner.

crates/sail-function/src/aggregate/percentile.rs
davidlghellin: These tests were only checking the part after the percentile value was already passed through, which is why the results looked strange. I didn’t realize that the percentile was still hard-coded to 0.5

crates/sail-function/src/aggregate/percentile.rs
linhr: Got it. We can move `LiteralEvaluator` to `sail-common-datafusion` now that it's needed for functions.

linhr: I've moved `LiteralEvaluator` as part of #1138.

linhr:
I'll take a look at this shortly!

PR #1073, opened 2025-11-21
chore: docker test for Iceberg

lonless9:


lonless9:
In favor of https://github.com/lakehq/sail/pull/1631

PR #1074, opened 2025-11-21
feat: add `uniform` function

davidlghellin:
## Summary

Implements `uniform(min, max[, seed])` producing values **bit-exact with Spark JVM** for any given seed.

Closes the last missing function of `function/math.json` gold tests.

## Implementation

Uses the existing [`SparkXorShiftRandom`](https://github.com/lakehq/sail/blob/main/crates/sail-function/src/scalar/math/xorshift.rs) — a port of Apache Spark's [`XORShiftRandom`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/random/XORShiftRandom.scala) with MurmurHash3-hashed seed — already used by `random()` / `rand()`.

Scaling formulas mirror [Spark's `Uniform` expression](https://github.com/apache/spark/blob/34f3d4a46097f171868633254000c9f9727535a6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala#L208):

- **Integer**: `min + floor(nextDouble() * (max - min))`
- **Float / Double**: `min + nextDouble() * (max - min)` (computed in f64, cast to target type)
- **Decimal**: double-space interpolation, cast back to scaled i128

The RNG is built **once per `invoke_with_args` call** and advances across rows, matching Spark's "one RNG per partition, N values" semantics (instead of the per-row reseed that would repeat the same value).

## Behavior parity with Spark

| Aspect | Status |
|---|---|
| Signature (`uniform(min, max[, seed])`, arity 2-3) | ✅ |
| Type promotion (Int8/16/32/64, Float32/64, Decimal, mixed) | ✅ |
| NULL bound → NULL result (Float64 type) | ✅ |
| Foldability check (rejects column references) | ✅ |
| Seed type validation (only INT / BIGINT) | ✅ |
| Equal bounds short-circuit | ✅ |
| Bit-exact values per seed | ✅ |
| Multi-row seed sequences | ✅ |


crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,644 @@
...
+
+impl SparkUniform {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::user_defined(Volatility::Volatile),
+        }
+    }
+
+    fn calculate_output_type(t_min: &DataType, t_max: &DataType) -> DataType {
+        if t_min.is_integer() && t_max.is_integer() {
```
shehabgamin, line 115: We shouldn't coerce from `DataType::Int64 | DataType::UInt32 | DataType::UInt64` to `DataType::Int32`.

It also looks like `SMALLINT`, `INT`, and `BIGINT` are all supported as `t_min` and `t_max` values, which means those could be possible return types I believe:
https://docs.databricks.com/aws/en/sql/language-manual/functions/uniform#arguments

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,644 @@
...
+            return DataType::Int32;
+        }
+
+        match (t_min, t_max) {
```
shehabgamin: We can also support `DataType::Decimal32` and  `DataType::Decimal64`.

Also if `DataType::Decimal256` has precision <= `DECIMAL128_MAX_PRECISION` and scale <= `DECIMAL128_MAX_SCALE`, that can be supported as well.

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,644 @@
...
+                    "max must be a scalar value, not an array",
+                ))
+            }
+        };
+
+        // Extract seed if present
+        let seed: Option<u64> = if args.len() == 3 {
+            match &args[2] {
+                ColumnarValue::Scalar(scalar) => match scalar {
+                    ScalarValue::Int64(Some(value)) => Some(*value as u64),
```
shehabgamin: Seed can be negative:
```
>>> spark.sql("SELECT uniform(5, 105, 3)").show()
+------------------+
|uniform(5, 105, 3)|
+------------------+
|                30|
+------------------+

>>> spark.sql("SELECT uniform(5, 105, -3)").show()
+-------------------+
|uniform(5, 105, -3)|
+-------------------+
|                 81|
+-------------------+
```

shehabgamin (changes requested):
Exciting! Just have a couple of small comments

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,1200 @@
...
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::user_defined(Volatility::Volatile),
+        }
+    }
+
+    /// Helper to extract precision and scale from any decimal type
+    fn extract_decimal_info(dt: &DataType) -> Option<(u8, i8)> {
+        match dt {
+            DataType::Decimal128(p, s) => Some((*p, *s)),
```
shehabgamin:
```suggestion
            DataType::Decimal32(p, s) | DataType::Decimal64(p, s) | DataType::Decimal128(p, s) => Some((*p, *s)),
```

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,1200 @@
...
+        } else {
+            // Same precision → use max scale
+            (p1, s1.max(s2))
+        }
+    }
+
+    fn calculate_output_type(t_min: &DataType, t_max: &DataType) -> DataType {
+        if t_min.is_integer() && t_max.is_integer() {
+            // Check for 64-bit types (signed or unsigned)
+            let is_64: bool = matches!(t_min, DataType::Int64 | DataType::UInt64)
```
shehabgamin: Should we throw an error for `UInt64` since it can't safely be casted to `Int64`?

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,1200 @@
...
+            (p1, s1)
+        } else if p2 > p1 {
+            (p2, s2)
+        } else {
+            // Same precision → use max scale
+            (p1, s1.max(s2))
+        }
+    }
+
+    fn calculate_output_type(t_min: &DataType, t_max: &DataType) -> DataType {
```
shehabgamin, line 99: The return type can be:
A `DataType::Int16`, `DataType::Int32`, `DataType::Int64`, `DataType::Float32`, `DataType::Float64`, or `DataType::Decimal128`.

This function only seems to support:
`DataType::Int32`, `DataType::Int64`, `DataType::Float64`, and `DataType::Decimal128`.

crates/sail-function/src/scalar/math/spark_uniform.rs
davidlghellin: > Should we throw an error for `UInt64` since it can't safely be casted to `Int64`?

Seeing this, would it be better to remove it? So it doesn’t cause confusion or lead to misunderstandings.

https://github.com/lakehq/sail/blob/b1715057a9e5a4649903393ce8cfb0bf9ee0a8a6/docs/guide/dataframe/data-types/compatibility.md?plain=1#L62
image

shehabgamin:
> (+1)       26 DocTestFailure
> (-1)        0 UnsupportedOperationException: function: uniform

@davidlghellin When expanding the `Error Counts` ([here](https://github.com/lakehq/sail/pull/1074#issuecomment-3563977857)), we find an additional `DocTestFailure` for the newly supported function. Any chance you've looked into it?

davidlghellin:
> > (+1)       26 DocTestFailure
> > (-1)        0 UnsupportedOperationException: function: uniform
> 
> @davidlghellin When expanding the `Error Counts` ([here](https://github.com/lakehq/sail/pull/1074#issuecomment-3563977857)), we find an additional `DocTestFailure` for the newly supported function. Any chance you've looked into it?

Thanks for pointing it out, I’ll take a look and follow up in the PR.

davidlghellin:
related https://github.com/lakehq/sail/issues/1266

davidlghellin:
https://github.com/apache/datafusion/pull/19908

crates/sail-function/src/scalar/math/spark_uniform.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        internal_err!("return_field_from_args should be used instead")
+    }
+
+    fn return_field_from_args(&self, args: ReturnFieldArgs) -> Result<FieldRef> {
+        let t_min = args.arg_fields[0].data_type();
+        let t_max = args.arg_fields[1].data_type();
+        let return_type = Self::calculate_output_type(t_min, t_max);
```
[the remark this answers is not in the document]
davidlghellin: [1530cd8](https://github.com/lakehq/sail/pull/1074/commits/1530cd8f3d2949080ee969008e90f6ae9634bd65)[1530cd8](https://github.com/lakehq/sail/pull/1074/commits/1530cd8f3d2949080ee969008e90f6ae9634bd65)

python/pysail/tests/spark/function/features/math/uniform.feature
```diff
@@ -0,0 +1,694 @@
...
+      When query
+        """
+        SELECT uniform(10, 20) AS result
+        """
+      Then query schema
+        """
+        root
+         |-- result: integer (nullable = false)
+        """
+
```
davidlghellin, line 127: Maybe we need to remove these schema tests, but I remember they gave me trouble, which is why I added them in the first place. Maybe they're no longer necessary now.

shehabgamin (approved):
@davidlghellin Nice!!

Issue #1066, opened 2025-11-18
Feature request: make it easier for users to evaluate if sail already supports required PySpark functions


keen85:
I'd love to use `sail` for faster local _development_ of PySpark code - and on the long run maybe even execute production workloads on `sail`.
However, currently `sail` does not have 100% feature parity with PySpark.

So, I need to check if at the PySpark functions that we use are already supported by sail - and I need to repeat this after every `sail` release.

What I do is to search our code repo for `import pyspark.sql.function` occurrences and identify the function names that we use.
Then I need to cross reference this with the `sail` documentation: https://docs.lakesail.com/sail/main/guide/functions/scalar.html

I think it would be very helpful if `sail` came with a "compatibility checker" that automates this and highlights gaps.

shehabgamin:
@keen85 Agreed, this would be super useful. I'll brainstorm a couple of ways to create a script for this.

Lemme know if you have any ideas as well!

keen85:
@shehabgamin I already have a Python script for analyzing which PySpark functions are used in `*.py` and `*.ipynb` files for a given root directory and counts how many times these occur.

I could create a PR to show what I got so far.

But I'm unsure how to handle the "look up if supported" part. I could use the doc `*.md` files like [this one](https://github.com/lakehq/sail/blob/main/docs/guide/functions/scalar.md?plain=1).
Are these generated or maintained manually?

shehabgamin:
> @shehabgamin I already have a Python script for analyzing which PySpark functions are used in `*.py` and `*.ipynb` files for a given root directory and counts how many times these occur.
> 
> I could create a PR to show what I got so far.
> 
> But I'm unsure how to handle the "look up if supported" part. I could use the doc `*.md` files like [this one](https://github.com/lakehq/sail/blob/main/docs/guide/functions/scalar.md?plain=1).
> Are these generated or maintained manually?


@keen85 Yeah, that would be a great first start!

The docs are manually maintained for now, we haven't come up with a good story around how to generate them yet.

@linhr may have some additional thoughts here, I know he's been thinking about docs generation.

linhr:
> @linhr may have some additional thoughts here, I know he's been thinking about docs generation.

We'll later consolidate the function docs with test cases so passing tests will indicate supported functions. I'm still figuring out the details of the idea.

keen85:
@shehabgamin, I created a PR containing my rudimentary scanner for PySpark function calls: https://github.com/lakehq/sail/pull/1075

[...]

PR #1075, opened 2025-11-21
feat: add PySpark function scanner & sail coverage

keen85:
addresses #1066 

After installing `sail` via `pip install pysail`, the CLI command `python -m pysail.examples.spark.compatibility_check <directory>` becomes available. It scans all `*.py` and `*.ipynb` files in `<directory>` for PySpark function usage and cross-references the detected functions with the `sail` documentation. This shows whether a PySpark project is compatible with `sail` as a drop-in replacement.


keen85:
Hi @shehabgamin & @linhr,

I'd appreciate some feedback on my current implementation approach:

1.  **Scan:** I use the built-in Python [`ast`](https://docs.python.org/3/library/ast.html) module to scan code for function calls.
2.  **Verify:** I use the [`jedi`](https://jedi.readthedocs.io) package (additional dependency) to confirm if the function stems from a relevant PySpark module.
3.  **Count:** Found functions are stored in a [`Counter object`](https://docs.python.org/3/library/collections.html#collections.Counter).
4.  **Validate:** I parse the Markdown files from the `sail` documentation (e.g. https://github.com/lakehq/sail/blob/main/docs/guide/functions/scalar.md) to derive the support status.
5.  **Report:** The code usage is cross-referenced with the documentation. The final output tells the user which functions were found, how often, and if `sail` claims to support them.

**Current Limitation:**
The scanner is only as accurate as the documentation. If a function is supported but not documented (e.g. `pyspark.sql.functions.col`, `pyspark.sql.functions.lit`), it displays as "unknown."

**Request for Feedback:**
I am also looking for general advice on the module/method structure I chose. I am not a seasoned Python Software Engineer, so I am very open to overall best-practice suggestions there.

shehabgamin:
> Hi @shehabgamin & @linhr,
> 
> I'd appreciate some feedback on my current implementation approach:
> 
>     1. **Scan:** I use the built-in Python [`ast`](https://docs.python.org/3/library/ast.html) module to scan code for function calls.
> 
>     2. **Verify:** I use the [`jedi`](https://jedi.readthedocs.io) package (additional dependency) to confirm if the function stems from a relevant PySpark module.
> 
>     3. **Count:** Found functions are stored in a [`Counter object`](https://docs.python.org/3/library/collections.html#collections.Counter).
> 
>     4. **Validate:** I parse the Markdown files from the `sail` documentation (e.g. https://github.com/lakehq/sail/blob/main/docs/guide/functions/scalar.md) to derive the support status.
> 
>     5. **Report:** The code usage is cross-referenced with the documentation. The final output tells the user which functions were found, how often, and if `sail` claims to support them.
> 
> 
> **Current Limitation:** The scanner is only as accurate as the documentation. If a function is supported but not documented (e.g. `pyspark.sql.functions.col`, `pyspark.sql.functions.lit`), it displays as "unknown."
> 
> **Request for Feedback:** I am also looking for general advice on the module/method structure I chose. I am not a seasoned Python Software Engineer, so I am very open to overall best-practice suggestions there.

@keen85 Exciting!! I'll take a look today/tomorrow 🚀

linhr:
This is useful! Thanks @keen85 for having this!

I feel the scanner entrypoint might best fit the `pysail/examples/spark` directory. We have a TPC-H script there already and the user can invoke it via `python -m pysail.examples.spark.tpch`. The function scanner seems another valuable addition as we gradually build a repertoire of handy utilities around Sail. (For the package-level entrypoint, we need to be careful about breaking changes, so it would be harder for making adjustments to the scanner in the long term.)

The utilities and tests seems in good locations, though I prefer `utils` over `util` for the directory name (for consistency with the entire codebase). And I wonder if the functions can be prefixed with `_` so that it's not part of Sail's public interface. This is again for better maintainability.

One unfortunate blocker for this PR is that this assumes the Markdown files are available. This is not true for released packages, so the end user would not be able to run the scanner when they install Sail. One solution is to have function support metadata as JSON files that are packaged with the release, and the same JSON files can be used to render Markdown files. In this way we also won't need to parse Markdown docs. I'll let @shehabgamin to review the changes more thoroughly and decide the best approach forward.

Besides my notes above, I feel this PR is great for Sail and I'd love to see a path for it to be merged!

keen85:
> One unfortunate blocker for this PR is that this assumes the Markdown files are available. This is not true for released packages, so the end user would not be able to run the scanner when they install Sail.

Good point! I did not think of that...
I could also imagine not to scan the *.md files from within the repo, but rather scraping the actual HTML websites.
However, this would require an internet connection during execution.

I'd like the thought of having some JSON as single point of truth that can also be used for doc generation much more.

shehabgamin:
> > One unfortunate blocker for this PR is that this assumes the Markdown files are available. This is not true for released packages, so the end user would not be able to run the scanner when they install Sail.
> 
> Good point! I did not think of that... I could also imagine not to scan the *.md files from within the repo, but rather scraping the actual HTML websites. However, this would require an internet connection during execution.
> 
> I'd like the thought of having some JSON as single point of truth that can also be used for doc generation much more.

Perhaps for a v1 we can fetch the `*.md` files for the version of Sail that the user is using? For example for Sail `0.4.2`, it would be `https://github.com/lakehq/sail/tree/v0.4.2/docs/guide/functions`

linhr:
I feel it might be easier to just create the JSON files for this PR? Then we can separately follow up to see how the JSON files can be used to generate the Markdown files (which requires writing a custom Vue component but should be easy). In this way we also don't need any extra Markdown/HTML parsing work.

keen85:
@linhr, how to move forward with this?

1. I could create a JSON and add it to the repo.
2. another option would be not to scan the *local* md files but rather the *remote* md files from GitHub repo (e.g. https://raw.githubusercontent.com/lakehq/sail/refs/heads/main/docs/guide/functions/scalar.md). 

I implemented 2. yesterday and it works fine for now.

If you prefer 1., should it be one JSON containing _all_ coverage information (where the category would be another attribute per item), or should it be rather multiple JSON files, maintaining the categories per file as in the current documentation: https://docs.lakesail.com/sail/main/guide/functions/

linhr:
Thanks for validating the ideas!

I feel let's move forward with option 1 since we know it's future-proof. You can create one JSON file per category in the `python/pysail/data/functions` directory. In the installed package, the files will be available at `Path(pysail.__file__).parent / "data" / "functions"`.

linhr:
For formatting and linting Python code, you can run `hatch fmt` in the project root directory. We use `# noqa: XXX` comments to suppress lint errors for particular lines if not following the rule improves readability. (Make sure to use the latest Hatch version 1.16.1 since there are recent breaking changes to the lint rules.)

keen85:
> I feel let's move forward with option 1 since we know it's future-proof. You can create one JSON file per category in the `python/pysail/data/functions` directory. In the installed package, the files will be available at `Path(pysail.__file__).parent / "data" / "functions"`.

@linhr, taking https://docs.lakesail.com/sail/main/guide/functions/scalar.html as example, there are several ways to model this information as JSON:
1. One large nested JSON/dict:
```json
{
    "pyspark.sql.functions": {
        "array": {
            "support_status": "supported",
            "cluster": "Array"
        },
        "array_append": {
            "support_status": "supported",
            "cluster": "Array"
        },
        ...
        "add_years": {
            "support_status": "supported",
            "cluster": "Datetime"
        },
        "add_months": {
            "support_status": "supported",
            "cluster": "Datetime"
        }
        ...
    }
}
````
2. Array of (flat) JSON/dicts
```json
[
    {
        "module": "pyspark.sql.functions",
        "function": "array",
        "support_status": "supported",
        "cluster": "Array"
    },
    {
        "module": "pyspark.sql.functions",
        "function": "array_append",
        "support_status": "supported",
        "cluster": "Array"
    },
    ...
    {
        "module": "pyspark.sql.functions",
        "function": "array_compact",
        "support_status": "supported",
        "cluster": "Datetime"
    },
    {
        "module": "pyspark.sql.functions",
        "function": "add_months",
        "support_status": "supported",
        "cluster": "Datetime"
    },
    ...
]
````

Do you have any preference here? Taking into account that maybe at some point the md files should be generated from this JSON.

**Comment by linhr** [comment]
Thanks for sharing the ideas!

Both options make sense to me, though I feel option 2 is a cleaner data model which makes it easier for serde in different languages or libraries. (In option 1, the object key is free-form function name, which is sometimes slightly inconvenient to work with.)

nitpick: I feel `support_status` could simply be `status`. And we could leave out `cluster` if we have one JSON file per category. (We could have `scalar/<category>.json`, `aggregate.json` etc. When rendering the scalar function page, multiple JSON files could be loaded at once.)

File: python/pysail/data/compatibility/functions/scalar/bitwise.json
Comment (keen85) [inline python/pysail/data/compatibility/functions/scalar/bitwise.json:1]: these are rather operators

**Comment by keen85** [comment]
> For formatting and linting Python code, you can run `hatch fmt` in the project root directory. We use `# noqa: XXX` comments to suppress lint errors for particular lines if not following the rule improves readability. (Make sure to use the latest Hatch version 1.16.1 since there are recent breaking changes to the lint rules.)

@linhr local formatting does not work properly in my (Windows) environment;
I needed to add
```toml
[tool.ruff]
line-length = 120
```
to `pyproject.toml` and also fiddle with `.vscode\settings.json`.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ test = [
     "pillow>=10.3.0",
     "pyiceberg[sql-sqlite,pyiceberg-core]==0.10.0",
     "pydantic>=2.11,<2.12",
+    "jedi>=0.19.2",
```
linhr: Let's have an upper bound on all occurrence of the `jedi` the version to avoid breaking changes for future package versions.

```
jedi>=0.19.2,<0.20
```

And could you help with the `pillow` package (for all occurrence in this file) as well? I just realize it's not bounded either.

```
pillow>=10.3.0,<11
```

(A few bounds in the `test-ibis` environment are missing but that's fine since we copy the dependency specification from the upstream project.)

python/pysail/data/compatibility/functions/scalar.json
```diff
@@ -0,0 +1,2022 @@
+[
```
linhr: I feel this file should be removed since we have one JSON file per category for scalar functions.

python/pysail/data/compatibility/sparksession.json
```diff
@@ -0,0 +1,17 @@
...
+  {
+    "module": "pyspark.sql.DataFrame",
+    "function": "read",
+    "status": "supported"
+  },
+  {
+    "module": "pyspark.sql.DataFrame",
+    "function": "readStream",
+    "status": "in progress"
+  }
```
linhr, line 16: The module should be `pyspark.sql.session.SparkSession` for these two functions.

python/pysail/tests/utils/test_pyspark_function_scanner.py
```diff
@@ -0,0 +1,195 @@
...
+    code = cleandoc(
+        """
+        from pyspark.sql.types import StructType
+        import json
+
+        json_str = "..."
+
+        scheme = StructType.fromJson(json.loads(json_str))
+        scheme.simpleString()
+        scheme.jsonValue()
```
linhr: nitpick: This should be named `schema` instead of `scheme`.

python/pysail/tests/utils/test_pyspark_function_scanner.py
```diff
@@ -0,0 +1,195 @@
...
+    code = cleandoc(
+        """
+        import pyspark.sql.functions as F
+        import pyspark.sql.window as W
+        F.col("a")
+        F.when(True, 1)
+        W.Window.partitionBy("x")
+        """
+    )
+    for i in range(15):
```
linhr: nitpick: A few iterations (e.g. `range(3)`) might be enough already.

The entire test suite runs on CI frequently, so we may want to control the overall test time as we have more and more tests (especially tests involving IO) over time.

python/pysail/utils/pyspark_function_scanner.py
```diff
@@ -0,0 +1,200 @@
...
+from typing import TYPE_CHECKING
+
+import jedi
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+from pysail.utils.logging_config import setup_logging
+
+setup_logging()
```
linhr: nitpick: The logging setup is a side-effect during module import and we may want to just remove it. Usually utility modules are not supposed to alter global state. It's best to have the setup only inside the entrypoint `main()` function.

python/pysail/tests/utils/test_pyspark_function_scanner.py
```diff
@@ -0,0 +1,195 @@
...
+    assert _scan_file(path) == Counter(
+        {
+            # ("pyspark.sql.types.StructType", "fromJson"): 1,  # is not recognized yet...
+            ("pyspark.sql.types.StructType", "simpleString"): 1,
+            ("pyspark.sql.types.StructType", "jsonValue"): 1,
+        }
+    )
+
+
+def test_infer_column(tmp_path):
```
linhr, line 108: In the CI this test somehow fails for PySpark 3.5. I'm not sure about the actual reason though. Feel free to look deeper into this, or you can skip it conditionally based on the PySpark version via `@pytest.mark.skipif`.

linhr:
> I needed to add
> 
> ```toml
> [tool.ruff]
> line-length = 120
> ```
> 
> to `pyproject.toml` and also fiddle with `.vscode\settings.json`.

I feel the explicit `line-length` configuration in `pyproject.toml` is nice to have so feel free to add that to this PR!

For `.vscode\settings.json`, do you think your local change is suitable for inclusion in this PR as well?

python/pysail/tests/utils/test_pyspark_function_scanner.py
linhr: You can run `hatch run test.spark-3.5.5:pytest -k "test_infer_column"` to reproduce the test failure locally.

PR #1077, opened 2025-11-22
feat: add `div` to interval

davidlghellin:
solve test gold_data failed


crates/sail-execution/src/codec.rs
```diff
@@ -1489,6 +1490,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             || node_inner.is::<SparkFloor>()
             || node_inner.is::<SparkFromCSV>()
             || node_inner.is::<SparkHex>()
+            || node_inner.is::<SparkIntervalDiv>()
```
linhr, line 1494: We need to add the function to `try_decode_udf` as well.

linhr (approved):
This looks great! This is ready to merge once the minor codec issue is fixed.

PR #1084, opened 2025-11-25
chore: update to delta-kernel-rs 0.18.1 arrow 57.1.0

lonless9:


Cargo.toml
linhr: We might want to update the version here as well.

linhr (approved):
🚀

Cargo.toml
lonless9: Nice catch!

PR #1087, opened 2025-11-27
chore: remove OS native SSL dependency

linhr:
I noticed this when I tried out cross-compilation for Linux on macOS host. It seems that `reqwest` uses native SSL while all other dependencies uses `rustls`. The former would be challenging for cross-compilation and adds a few `.so` libraries in Linux wheels. So we need to consolidate this.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -39,6 +39,8 @@ jobs:
           python-version: "3.11"
 
       - uses: pypa/hatch@install
+        with:
+          version: 1.15.1
 
       - uses: ./.github/actions/setup-rust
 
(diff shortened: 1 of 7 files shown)


crates/sail-cli/src/runner.rs
```diff
@@ -79,6 +80,15 @@ enum SparkCommand {
...
 pub fn main(args: Vec<String>) -> Result<(), Box<dyn std::error::Error>> {
+    if rustls::crypto::aws_lc_rs::default_provider()
+        .install_default()
+        .is_err()
+    {
+        Err(CommonError::InternalError(
+            "failed to install crypto provider".to_string(),
+        ))?;
+    }
+
```
linhr, line 91: This is moved here so the initialization can happen when Sail is used as a binary or a Python library.

shehabgamin (approved):

PR #1090, opened 2025-11-27
chore: revert the pin of hatch version and resolve ruff warnings

lonless9:
This is a follow-up for #1088 
We need to notice that recently `hatch` released its 1.16.0 version (https://github.com/pypa/hatch/pull/2106) and directly upgrade `ruff` from 0.6.8 to 0.13.2 (https://github.com/pypa/hatch/pull/1890)
Thus we need to take care of this kind of breaking change from `hatch`.

### DIFF
--- a/crates/sail-cli/src/python/spark_mcp_server.py
+++ b/crates/sail-cli/src/python/spark_mcp_server.py
@@ -9,6 +9,8 @@ import uvicorn.server
 from mcp.server.fastmcp import Context, FastMCP
 from pyspark.sql import SparkSession
 
+logger = logging.getLogger(__name__)
+
 
 def _is_temp_view(table):
     return table.catalog is None and not table.namespace and table.tableType == "TEMPORARY" and table.isTemporary
@@ -65,10 +67,10 @@ def configure_fastmcp_log_level():
 def create_spark_mcp_server(host: str, port: int, spark_remote: str):
     @asynccontextmanager
     async def lifespan(server: FastMCP) -> AsyncIterator[SparkSession]:  # noqa: ARG001
-        logging.info("Creating Spark session")
+        logger.info("Creating Spark session")
         spark = SparkSession.builder.remote(spark_remote).getOrCreate()
         yield spark
-        logging.info("Stopping Spark session")
+        logger.info("Stopping Spark session")
         spark.stop()
 
     mcp = FastMCP("Sail MCP server for Spark SQL", lifespan=lifespan, host=host, port=port)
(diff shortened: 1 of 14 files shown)


crates/sail-cli/src/python/spark_mcp_server.py
```diff
@@ -9,6 +10,15 @@
 from mcp.server.fastmcp import Context, FastMCP
 from pyspark.sql import SparkSession
 
+try:
+    import boto3
+except ImportError as _BOTO3_IMPORT_ERROR:  # pragma: no cover - optional dependency
+    boto3 = None
+else:
+    _BOTO3_IMPORT_ERROR = None
```
linhr: I feel having `import` inside functions is more readable and manageable than the global flag. We could simply add a `noqa` comment.

I came across notes somewhere saying that many lints are very strict and mostly advisory, so it's fine to scatter `noqa` in a few places if not following the lint improves the code for particular use cases.

python/pysail/examples/spark/tpch.py
```diff
@@ -1,9 +1,20 @@
...
 
+try:
+    import readline
+except ImportError:  # pragma: no cover - optional dependency on non-Unix systems
+    readline = None
+
+try:
+    from rlcompleter import Completer
+except ImportError:  # pragma: no cover - optional dependency on non-Unix systems
+    Completer = None
```
linhr: Same here regarding local `import`.

python/pysail/tests/spark/utils.py
```diff
@@ -70,6 +72,8 @@ def __eq__(self, other):
     def __repr__(self):
         return f"AnyOf({', '.join(repr(v) for v in self.values)})"
 
+    __hash__ = None
+
```
linhr: Should we implement these functions?

scripts/spark-tests/plugins/spark.py
```diff
@@ -12,22 +12,50 @@
...
 
 def _is_spark_testing():
     return os.environ.get("SPARK_TESTING") == "1"
 
 
+def _require_dependency(obj, description: str):
+    if obj is None:
+        msg = f"{description} requires PySpark test dependencies. Please install pyspark and pandas."
+        raise RuntimeError(msg)
+    return obj
```
linhr: Same here regarding local `import` for this file. I feel `noqa` is more straightforward.

crates/sail-cli/src/python/spark_mcp_server.py
lonless9: Done! If we encounter a similar problem again, we can consider disabling this linter rule.

python/pysail/tests/spark/utils.py
lonless9: Done!

PR #1093, opened 2025-11-28
feat: basic merge operation for Delta Lake

lonless9:
part of #171 

$$ Op(T) = \underbrace{\text{Write}(\Pi_{New}(\sigma_{Keep}(T \bowtie S)))}_{\text{Add Actions}} \cup \underbrace{\text{Remove}(\Pi_{Path}(\sigma_{Touch}(T \bowtie S)))}_{\text{Remove Actions}} $$

These should be addressed in the following PRs:
- Cardinality check
- Optimization for fast append paths
- Metric

### DIFF
--- a/python/pysail/tests/spark/analyst/__snapshots__/test_features/test_explain_codegen_shows_codegen_notice_and_physical_plan.plan
+++ b/python/pysail/tests/spark/analyst/__snapshots__/test_features/test_explain_codegen_shows_codegen_notice_and_physical_plan.plan
@@ -24,7 +24,7 @@ Projection: t.#2 AS #4, sum(t.#3) AS #5
 analyzed_logical_plan:
 SAME TEXT AS ABOVE
 
-logical_plan after ExpandMergeRule:
+logical_plan after expand_merge:
 SAME TEXT AS ABOVE
 
 logical_plan after eliminate_nested_union:
@@ -97,7 +97,7 @@ Projection: t.#2 AS #4, sum(t.#3) AS #5
       Projection: column1 AS #2, column2 AS #3
         Values: (Int32(1), Int32(2)), (Int32(1), Int32(3))
 
-logical_plan after ExpandMergeRule:
+logical_plan after expand_merge:
 SAME TEXT AS ABOVE
 
 logical_plan after eliminate_nested_union:


python/pysail/tests/spark/delta/__snapshots__/test_delta_features/test_explain_codegen_includes_plan_steps_and_merge_rewrite_artifacts.plan
```diff
@@ -0,0 +1,1183 @@
...
+                              FilterExec: __sail_merge_source_row_present@10 IS NOT NULL AND id@0 = __sail_src_id@6 AND category@1 IS NOT NULL AND (category@1 = vip OR category@1 = stale) OR __sail_merge_source_row_present@10 IS NULL AND category@1 = stale
+                                CoalesceBatchesExec: target_batch_size=8192
+                                  CoalesceBatchesExec: target_batch_size=8192
+                                    CoalesceBatchesExec: target_batch_size=8192
+                                      HashJoinExec: mode=CollectLeft, join_type=Left, on=[(id@0, __sail_src_id@0)], filter=category@0 IS NOT NULL
+                                        CoalescePartitionsExec
+                                          ProjectionExec: expr=[id@0 as id, category@1 as category, amount@2 as amount, note@3 as note, __sail_file_path@4 as __sail_file_path, true as __sail_merge_target_row_present]
+                                            CoalesceBatchesExec: target_batch_size=8192
+                                              CoalesceBatchesExec: target_batch_size=8192
+                                                CoalesceBatchesExec: target_batch_size=8192
```
linhr, line 1172: We may want to see why there are duplicated `CoalesceBatchesExec`. This could be future work.

It's so nice that we have good `EXPLAIN` output for logical/physical plans. It's now much easier to investigate DML planning issues!

crates/sail-plan-lakehouse/src/optimizer.rs
```diff
@@ -0,0 +1,291 @@
...
+                        node: Arc::new(write_node),
+                    })));
+                }
+            }
+            Ok(Transformed::no(plan))
+        })
+    }
+
+    fn name(&self) -> &str {
+        "ExpandMergeRule"
```
linhr: To follow the convention of DataFusion's built-in rules, we can name this `expand_merge` so the `EXPLAIN` output is more consistent.

crates/sail-delta-lake/src/physical_plan/planner/op_merge.rs
```diff
@@ -0,0 +1,129 @@
...
+        // Ensure a single partition so we can de-duplicate paths and avoid re-loading
+        // the snapshot multiple times.
+        let touched_plan = Arc::new(CoalescePartitionsExec::new(touched_plan));
+
+        // Convert Path stream -> Add(JSON) stream on the Worker.
+        let lookup_plan = Arc::new(DeltaFileLookupExec::new(
+            touched_plan,
+            table_url.clone(),
+            version,
+        ));
```
linhr: This could be achieved by `required_input_distribution` for `DeltaFileLookupExec`. Is the corresponding optimizer not in effect for some reason? We can have a TODO to look into it.

crates/sail-logical-plan/src/merge.rs
```diff
@@ -0,0 +1,1025 @@
...
+    }
+}
+
+impl PartialOrd for MergeIntoWriteNode {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let a = self as *const Self as usize;
+        let b = other as *const Self as usize;
+        a.partial_cmp(&b)
+    }
+}
```
linhr, line 326: Let's add a TODO to implement these without pointer operations. We've been using helper structs to skip the `schema` field so `PartialOrd` can be derived.

I was also considering using proc macro to make it easier. There are a few options but I haven't decided on the choice though:
* <https://github.com/mcarton/rust-derivative>
* <https://github.com/magiclen/educe>

crates/sail-logical-plan/src/merge.rs
```diff
@@ -0,0 +1,1025 @@
...
+
+fn all_placeholder_schema(schema: &DFSchemaRef, path_column: &str) -> bool {
+    let names: Vec<&str> = schema.fields().iter().map(|f| f.name().as_str()).collect();
+    let non_path: Vec<&str> = names
+        .iter()
+        .copied()
+        .filter(|name| *name != path_column)
+        .collect();
+    !non_path.is_empty() && non_path.iter().all(|name| name.starts_with('#'))
```
linhr, line 1030: Let's add a TODO here. The plan resolver might need to provide some utilities so it's easier to work with "resolved" opaque field names.

The need to work with the original schema in this file indicates that there might be limitations in the current plan resolver design. So the merge operation would become a good example for future improvements on the plan resolver.

python/pysail/tests/spark/delta/__snapshots__/test_delta_features/test_explain_codegen_includes_plan_steps_and_merge_rewrite_artifacts.plan
lonless9: Yeah we can pay attention to this part later. This may also be related to the implementation mechanism of `explain` itself.

crates/sail-plan-lakehouse/src/optimizer.rs
lonless9: Done!

crates/sail-delta-lake/src/physical_plan/planner/op_merge.rs
lonless9: Yeah there's redundancy here, I remove it and it works fine now.

crates/sail-logical-plan/src/merge.rs
lonless9: Done!

crates/sail-logical-plan/src/merge.rs
lonless9: Done!

keen85:
will this allow implementing [`DataFrame.mergeInto`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.mergeInto.html#pyspark.sql.DataFrame.mergeInto)?

linhr:
> will this allow implementing `DataFrame.mergeInto`?

Yeah the same code path will be used to support this DataFrame operation in the near future!

PR #1097, opened 2025-11-29
feat: support OpenTelemetry exporter and distributed tracing

linhr:
1. Support OpenTelemetry exporter configuration.
2. Add a proof of concept for distributed tracing, using worker registration as the example. It adds support for tracing in our actor model. Context propagation between the driver and the worker is done via environment variables. It also shows how to carefully keep span context among async function calls so that span dependencies are modeled correctly in the trace. As a next step, we can refine span naming to make it more readable, and extend tracing setup to the core query execution logic.


crates/sail-python/src/spark/server.rs
```diff
@@ -136,7 +135,6 @@ impl SparkConnectServer {
             _ = rx => { }
         }
         info!("Shutting down the Spark Connect server...");
-        fastrace::flush();
```
linhr, line 139: We can't call `shutdown_telemetry()` here since it affects the global telemetry setup. I'm thinking about a breaking change in our next minor release where `init_telemetry()` is removed from the `SparkConnectServer` class and called implicitly during module import. And `shutdown_telemetry()` can be registered as a hook via the Python `atexit` module.

crates/sail-telemetry/src/telemetry.rs
```diff
@@ -1,86 +1,203 @@
...
-            }
-            let level = record.level();
-            let target = record.target();
-            let style = buf.default_level_style(level);
-            let timestamp = buf.timestamp();
-            let args = record.args();
-            if let Some(span_context) = SpanContext::current_local_parent() {
-                let trace_id = span_context.trace_id.0;
-                let span_id = span_context.span_id.0;
-                writeln!(buf, "[{timestamp} {style}{level}{style:#} {target} trace: {trace_id} span: {span_id}] {args}")
```
linhr, line 68: We no longer print trace and span ID in the env logger. I feel the tracing information would be more meaningful in Grafana etc. and it's OK to leave it out in the console output to make the debug log easier to read.

PR #1099, opened 2025-12-01
refactor: invert dependency for data source registration

lonless9:


crates/sail-session/src/formats.rs
```diff
@@ -0,0 +1,43 @@
...
+use sail_data_source::formats::text::TextTableFormat;
+use sail_delta_lake::DeltaTableFormat;
+use sail_iceberg::IcebergTableFormat;
+
+pub fn create_table_format_registry() -> Result<Arc<TableFormatRegistry>> {
+    let registry = Arc::new(TableFormatRegistry::new());
+    register_builtin_formats(&registry)?;
+    register_external_formats(&registry)?;
+    Ok(registry)
+}
```
lonless9, line 23:
```suggestion

static TABLE_FORMAT_REGISTRY: OnceLock<Arc<TableFormatRegistry>> = OnceLock::new();

pub fn create_table_format_registry() -> Result<Arc<TableFormatRegistry>> {
    if let Some(registry) = TABLE_FORMAT_REGISTRY.get() {
        return Ok(registry.clone());
    }

    let registry = Arc::new(TableFormatRegistry::new());
    register_builtin_formats(&registry)?;
    register_external_formats(&registry)?;

    if TABLE_FORMAT_REGISTRY.set(registry.clone()).is_err() {
        return TABLE_FORMAT_REGISTRY
            .get()
            .cloned()
            .ok_or_else(|| plan_datafusion_err!("failed to load table format registry"));
    }

    Ok(registry)
}
```

lonless9: Do we need to use static?

crates/sail-spark-connect/src/session_manager.rs
```diff
@@ -104,6 +105,7 @@ impl SessionManagerActor {
             // We do not use the DataFusion catalog and schema since we manage catalogs ourselves.
             .with_create_default_catalog_and_schema(false)
             .with_information_schema(false)
+            .with_extension(create_table_format_registry()?)
```
lonless9, line 108: We added the registration logic for the table format here. I guess that using only the default session in the `sail-execution` session creation location won't have unexpected effects, since `sail-execution` doesn't consume the catalog or format.

crates/sail-common-datafusion/src/extension.rs
```diff
@@ -20,3 +21,11 @@ impl SessionExtensionAccessor for SessionContext {
             .ok_or_else(|| internal_datafusion_err!("session extension not found: {}", T::name()))
     }
 }
+
+impl SessionExtensionAccessor for SessionState {
```
linhr, line 25: Nice!

crates/sail-spark-connect/src/session_manager.rs
linhr: Glad that you pointed out this! Yeah I was aware that we're using the default session creation logic for workers in `sail-execution` and I've been thinking about consolidating it. But yes even if the registry is not available for workers it won't have unexpected effects.

crates/sail-session/src/formats.rs
linhr: I feel per-session registry is fine since the registry creation overhead is low. (Conceptually, global static registry works equally well for this case though.)

Issue #1067, opened 2025-11-19
New function in Datafusion 51 to migrate Sail


davidlghellin:
We use the PR structure https://github.com/lakehq/sail/issues/907 (some functions not migrated in the last PR)

In Datafusion Proyect
needs https://github.com/lakehq/sail/pull/1031
```
❯ git checkout 50.0.0
❯ cd datafusion/spark
for f in **/mod.rs(.N); do
  # empty vec
  if rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[\s*\S' "$f" >/dev/null; then
    echo "---- $f ----"
    rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[([\s\S]*?\S[\s\S]*?)\]' -or '$1' "$f"
    echo
  fi
done > datafusion-50
```
```
❯ git checkout 51.0.0
for f in **/mod.rs(.N); do
  # empty vec
  if rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[\s*\S' "$f" >/dev/null; then
    echo "---- $f ----"
    rg -nUP 'pub fn functions\(\)[\s\S]*?vec!\[([\s\S]*?\S[\s\S]*?)\]' -or '$1' "$f"
    echo
  fi
done > datafusion-51
```

```
❯ diff datafusion-50 datafusion-51
1a2
> 33:avg()
4c5
< 32:array()
---
> 39:array(), shuffle()
10c11,17
< 39:bit_get(), bit_count()
---
> 66:        bit_get(),
> 67:        bit_count(),
> 68:        bitwise_not(),
> 69:        shiftleft(),
> 70:        shiftright(),
> 71:        shiftrightunsigned(),
> 72:    
22c29,35
< 59:date_add(), date_sub(), last_day(), next_day()
---
> 74:        date_add(),
> 75:        date_sub(),
> 76:        last_day(),
> 77:        make_dt_interval(),
> 78:        make_interval(),
> 79:        next_day(),
> 80:    
33a47
> 45:map_from_arrays(), map_from_entries()
50c64,73
< 64:ascii(), char(), ilike(), like(), luhn_check()
---
> 93:        ascii(),
> 94:        char(),
> 95:        concat(),
> 96:        elt(),
> 97:        ilike(),
> 98:        length(),
> 99:        like(),
> 100:        luhn_check(),
> 101:        format_string(),
> 102:    
57c80
< 32:parse_url()
---
> 43:parse_url(), try_parse_url()
```

[...]

PR #1105, opened 2025-12-03
feat: refactor functions datafusion 51

davidlghellin:
part #1067

### DIFF
--- a/crates/sail-plan/src/function/scalar/datetime.rs
+++ b/crates/sail-plan/src/function/scalar/datetime.rs
@@ -347,96 +347,6 @@ fn make_ym_interval(args: Vec<Expr>) -> PlanResult<Expr> {
     Ok(ScalarUDF::from(SparkMakeYmInterval::new()).call(vec![years, months]))
 }
 
-fn make_interval(args: Vec<Expr>) -> PlanResult<Expr> {
-    let (years, months, weeks, days, hours, mins, secs) = match args.len() {
-        7 => {
-            let (y, mo, w, d, h, m, s) = args.seven()?;
-            (y, mo, w, d, h, m, s)
-        }
-        6 => {
-            let (y, mo, w, d, h, m) = args.six()?;
-            (y, mo, w, d, h, m, lit(0_f64))
-        }
-        5 => {
-            let (y, mo, w, d, h) = args.five()?;
-            (y, mo, w, d, h, lit(0_i32), lit(0_f64))
-        }
-        4 => {
-            let (y, mo, w, d) = args.four()?;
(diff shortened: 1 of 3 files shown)


davidlghellin:
Hi
I had forgotten about this PR

crates/sail-plan/src/function/scalar/string.rs
```diff
@@ -216,7 +220,7 @@ pub(super) fn list_built_in_string_functions() -> Vec<(&'static str, ScalarFunct
         ("concat_ws", F::var_arg(concat_ws)),
         ("contains", F::custom(contains)),
         ("decode", F::udf(SparkDecode::new())),
-        ("elt", F::udf(SparkElt::new())),
```
shehabgamin, line 219: I think we can leave this line unchanged? Is the newly added `elt` function above needed?

crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -347,6 +347,96 @@ fn make_ym_interval(args: Vec<Expr>) -> PlanResult<Expr> {
     Ok(ScalarUDF::from(SparkMakeYmInterval::new()).call(vec![years, months]))
 }
 
+fn make_interval(args: Vec<Expr>) -> PlanResult<Expr> {
```
shehabgamin: Given that the signature already handles variable length args, is this function needed?
https://github.com/apache/datafusion/blob/main/datafusion/spark/src/function/datetime/make_interval.rs#L58

crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -347,6 +347,96 @@ fn make_ym_interval(args: Vec<Expr>) -> PlanResult<Expr> {
...
+            return Err(PlanError::invalid(
+                "make_dt_interval expects between 0 and 4 arguments",
+            ));
+        }
+    };
+    Ok(ScalarUDF::from(SparkMakeInterval::new())
+        .call(vec![years, months, weeks, days, hours, mins, secs]))
+}
+
+fn make_dt_interval(args: Vec<Expr>) -> PlanResult<Expr> {
```
shehabgamin: Given that the signature already handles variable length args, is this function needed?
https://github.com/apache/datafusion/blob/main/datafusion/spark/src/function/datetime/make_dt_interval.rs#L61

crates/sail-plan/src/function/scalar/string.rs
davidlghellin: Makes sense — I added it just to match the vararg pattern, but it’s not necessary here. I’ll keep the original SparkElt registration and drop the extra var_arg one.

PR #1110, opened 2025-12-05
feat: explain statement rework

lonless9:
This PR optimizes the output of `explain` and introduces snapshot testing.
After modifying the relevant content, you can use the following command to update the snapshot file.
`hatch run pytest --snapshot-update`


pyproject.toml
```diff
@@ -36,6 +36,7 @@ test = [
     "duckdb>=1.0,<2",
     "pytest>=8.4,<9",
     "pytest-bdd>=8.1,<9",
+    "syrupy>=5.0,<6",
```
linhr, line 39: Is there a reason to use a different `syrupy` version (`>=5.0,<6`) here? All the other environments use `>=4.6,<5`.

python/pysail/tests/spark/analyst/__init__.py
```diff
@@ -0,0 +1 @@
+# Mark analyst tests as a package for linting tools.
```
linhr: nitpick: We could remove this comment since this is a well-known pattern for the existence of other `__init__.py` as well.

python/pysail/tests/spark/analyst/__plan_snapshots__/test_features/test_explain_codegen_shows_codegen_notice_and_physical_plan.raw
```diff
@@ -0,0 +1,683 @@
...
+initial_logical_plan:
+Projection: t.#2 AS #4, sum(t.#3) AS #5
+  Aggregate: groupBy=[[t.#2]], aggr=[[sum(t.#3)]]
+    SubqueryAlias: t
+      Projection: #0 AS #2, #1 AS #3
+        Projection: column1 AS #0, column2 AS #1
+          Values: (Int32(1), Int32(2)), (Int32(1), Int32(3))
+
+logical_plan after resolve_grouping_function:
+Projection: t.#2 AS #4, sum(t.#3) AS #5
```
linhr: DataFusion shows "the same as above" if the logical plan does not change for certain optimizers. We could consider supporting this to make the output easier to read. This could be a follow-up PR though.

python/pysail/tests/spark/conftest.py
```diff
@@ -221,6 +260,27 @@ def query_result(datatable, ordered, query, spark):
         assert sorted(rows) == sorted(r)
 
 
+@then("query plan equals")
+def query_plan_equals(docstring, query, spark, snapshot: SnapshotAssertion):
```
linhr: Is this unused currently? I wonder if this is still needed since all the tests use `query_plan_matches_snapshot`.

And it seems there are `__plan_snapshots__` and `__snapshots__` directories with identical content. Are these duplicates? (And is the directory name controlled by `file_extension = "plan"`?)

python/pysail/tests/spark/conftest.py
```diff
@@ -221,6 +260,27 @@ def query_result(datatable, ordered, query, spark):
...
+    # Persist a normalized snapshot for easier updates and diffing.
+    snapshot.use_extension(PlanSnapshotExtension).assert_match(plan)
+
+
+@then("query plan matches snapshot")
+def query_plan_matches_snapshot(query, spark, snapshot: SnapshotAssertion):
+    """Executes the SQL query and only asserts against the stored snapshot."""
+
+    plan = _collect_plan(query, spark)
+    snapshot.use_extension(PlanSnapshotExtension).assert_match(plan)
```
linhr: nitpick: According to the Syrupy documentation, we could replace `assert_match` with `==` which is more idiomatic in pytest.

linhr (approved):
This is exciting!! 🎉

python/pysail/tests/spark/conftest.py
linhr: And are we able to use the `extension_class` parameter? I saw this in the documentation:

> `snapshot.use_extension` is retained for compatibility. However, it is limited to only overriding the default extension class.

I don't have experience with Syrupy so I may not have the full picture of the setup though.

python/pysail/tests/spark/conftest.py
lonless9: Yes, I originally used a different naming scheme and asserting without syrupy, but it seems to have some conflicts in the CI environment, which appears to be due to me using the mismatch wrong version of syrupy.

python/pysail/tests/spark/analyst/__plan_snapshots__/test_features/test_explain_codegen_shows_codegen_notice_and_physical_plan.raw
lonless9: I expect to accomplish this in this PR; I don't think it will be too difficult.

pyproject.toml
lonless9: This should be the reason I encountered the problem with CI. Now I can figure out how to correctly set the path and file extension.

linhr (approved):

PR #1115, opened 2025-12-09
refactor: use arrow take_record_batch

gstvg:
Currently, this just simplifies the code, but after [arrow-rs#8879](https://www.github.com/apache/arrow-rs/issues/8879) lands, this should improve performance due to reduced check_bounds/better vectorization without using unsafe here

### DIFF
--- a/crates/sail-iceberg/src/operations/write/partition.rs
+++ b/crates/sail-iceberg/src/operations/write/partition.rs
@@ -165,12 +165,7 @@ pub fn split_record_batch_by_partition(
     let mut out: Vec<PartitionBatchResult> = Vec::with_capacity(groups.len());
     for (dir, grp) in groups.into_iter() {
         let indices = UInt32Array::from(grp.indices);
-        let mut cols = Vec::with_capacity(batch.num_columns());
-        for col in batch.columns() {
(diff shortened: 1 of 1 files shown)


crates/sail-iceberg/src/operations/write/partition.rs
```diff
@@ -165,12 +165,7 @@ pub fn split_record_batch_by_partition(
...
     for (dir, grp) in groups.into_iter() {
         let indices = UInt32Array::from(grp.indices);
-        let mut cols = Vec::with_capacity(batch.num_columns());
-        for col in batch.columns() {
-            let taken = compute::take(col.as_ref(), &indices, None).map_err(|e| e.to_string())?;
-            cols.push(taken);
-        }
-        let rb = RecordBatch::try_new(batch.schema(), cols).map_err(|e| e.to_string())?;
+        let rb = compute::take_record_batch(batch, &indices).map_err(|e| e.to_string())?;
         out.push(PartitionBatchResult {
```
gstvg, line 169: [Link to take_record_batch docs](https://docs.rs/arrow/latest/arrow/compute/fn.take_record_batch.html)

PR #1116, opened 2025-12-10
feat: report execution metrics

linhr:
This PR implements a metric registry for use in the entire application. Similar to how we implement data source options, a few YAML configuration files are used to generate Rust code, and the YAML files can later be used to generate documentation automatically.

This PR also provide basic integration to bridge DataFusion metrics with OpenTelemetry. More work will come in future PRs.


crates/sail-telemetry/src/execution/physical_plan.rs
```diff
@@ -202,3 +224,32 @@ impl ExecutionPlan for TracingExec {
...
+
+impl Stream for MetricEmitterStream {
+    type Item = Result<RecordBatch>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.project();
+        let poll = this.inner.poll_next(cx);
+        if poll.is_ready() {
+            if let Some(metrics) = this.plan.metrics() {
+                this.emitter.emit(&metrics, this.registry)
```
linhr, line 246: Emitting metrics on every batch may incur too much performance overhead. In a follow-up PR I'm considering reporting metrics at a fixed interval.

crates/sail-telemetry/src/execution/metrics/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::metrics::{MetricAttribute, MetricRegistry};
+
+pub enum MetricEmitted {
+    #[expect(unused)]
+    Yes,
+    No,
+}
+
+pub trait MetricEmitterExtension: Send {
+    fn try_emit(&self, metric: &Metric, registry: &MetricRegistry) -> MetricEmitted;
```
linhr, line 14: This will be implemented for each `ExecutionPlan` in a follow-up PR.

crates/sail-telemetry/src/execution/physical_plan.rs
lonless9: Besides fixed interval, I think we could also filter based on the current partition or report when the stream ends.

crates/sail-telemetry/src/execution/metrics/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+            match self.extension.try_emit(metric, registry) {
+                MetricEmitted::Yes => continue,
+                MetricEmitted::No => {}
+            }
+            match metric.value() {
+                MetricValue::OutputRows(count) => registry
+                    .execution_output_row_count
+                    .recorder(count)
+                    .with_optional_attribute(MetricAttribute::PARTITION, metric.partition())
+                    .emit(),
```
lonless9, line 51: I wonder if there's any other query-related information that can be carried besides the partition.

crates/sail-telemetry/src/metrics/data/registry.yaml
```diff
@@ -0,0 +1,52 @@
+- name: execution.output_size
+  description: The total output size.
+  unit: byte
+  type: Gauge
+  value_type: u64
+  attributes:
+    - partition
```
lonless9, line 7: I don't think there's a problem about Time/Gauge/Counter semantically, but I'm not sure if external systems will do reset and calculate the rate based on the Gauge and Counter types.

crates/sail-telemetry/src/metrics/data/attributes.yaml
```diff
@@ -0,0 +1,2 @@
+- name: partition
+  description: The execution partition.
```
lonless9, line 2: Can we add information related to the query?

crates/sail-telemetry/src/execution/metrics/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+                    .recorder(gauge)
+                    .with_optional_attribute(MetricAttribute::PARTITION, metric.partition())
+                    .emit(),
+                MetricValue::Count { .. }
+                | MetricValue::Gauge { .. }
+                | MetricValue::Time { .. }
+                | MetricValue::PruningMetrics { .. }
+                | MetricValue::Ratio { .. }
+                | MetricValue::Custom { .. } => {}
+                MetricValue::StartTimestamp(_) | MetricValue::EndTimestamp(_) => {}
```
lonless9, line 88: Is here TODO needed?

crates/sail-telemetry/src/telemetry.rs
```diff
@@ -190,18 +203,29 @@ fn init_datafusion_telemetry() -> TelemetryResult<()> {
...
             if let Some(provider) = &state.meter_provider {
                 let _ = provider.shutdown();
             }
             if let Some(provider) = &state.logger_provider {
                 let _ = provider.shutdown();
             }
+            *status = TelemetryStatus::Finalized;
         }
     }
 }
```
lonless9, line 217: Will this limit re-initialization? Is there a way to allow re-initialization from Finalized? Or could we simply set it as Uninitialized?

lonless9 (approved):
Exciting!

crates/sail-telemetry/src/execution/physical_plan.rs
linhr: Nice suggestions! I'll take care of it in the follow-up PR.

crates/sail-telemetry/src/execution/metrics/mod.rs
linhr: Yeah I'm going to add job/task/operator information in the follow-up PR.

crates/sail-telemetry/src/metrics/data/attributes.yaml
linhr: Yes this will come in the follow-up PR!

crates/sail-telemetry/src/execution/metrics/mod.rs
linhr: The timestamps are not metrics so they are ignored. And I plan to add some debug information for the other unknown metrics. I'm polishing this part in the follow-up PR.

crates/sail-telemetry/src/telemetry.rs
linhr: It's unclear to me whether fastrace and the OpenTelemetry SDK can reset their internal states, so I introduced this `Finalized` status to be safe. We can revisit this once we're more confident about the internals of these libraries.

crates/sail-telemetry/src/metrics/data/registry.yaml
linhr: Yeah this is a very good point. It's unfortunate that DataFusion does not provide callbacks for metric updates so we cannot report increments for its counters, but can only model the DataFusion counter as a gauge by reading the counter's current value. This may have some implications how the metrics can be analyzed externally.

lonless9 (approved):

PR #1119, opened 2025-12-11
feat: more attributes and definitions for execution metrics

linhr:
This is a follow-up of #1116.

1. Define more attributes for metrics.
2. Define metrics for various execution plans and implement metric emitters. Add unit tests to ensure that we're aware of metric set changes in DataFusion in the future.

Some of the metrics for join operations are not emitted yet due to upstream blockers being addressed in this PR: <https://github.com/apache/datafusion/pull/19283>.

### DIFF
--- a/crates/sail-telemetry/src/execution/metrics/default.rs
+++ b/crates/sail-telemetry/src/execution/metrics/default.rs
@@ -92,3 +92,32 @@ impl MetricEmitter for DefaultMetricEmitter {
         MetricHandled::Yes
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use datafusion::arrow::datatypes::{DataType, Field, Schema};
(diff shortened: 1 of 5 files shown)


crates/sail-execution/src/worker/actor/handler.rs
```diff
@@ -331,7 +332,14 @@ impl WorkerActor {
...
             DisplayableExecutionPlan::new(plan.as_ref()).indent(true)
         );
-        let plan = trace_execution_plan(plan)?;
+        let options = TracingExecOptions {
+            metric_registry: global_metric_registry(),
+            job_id: None, // TODO: propagate job ID
+            task_id: Some(task_id.into()),
+            task_attempt: Some(attempt),
+            operator_id: None,
+        };
```
lonless9, line 341: Yeah we can subsequently integrate the job ID transmission across the driver, worker and tracing layers.

PR #1127, opened 2025-12-13
feat: document compatibility check and update support status

keen85:
This is a follow-up of https://github.com/lakehq/sail/pull/1075.
It adds documentation for the compatibility check feature and updates which PySpark functions are supported currently.
Also the meta information on which PySpark functionalities are supported by Sail were extensively refined.

### DIFF
--- a/python/pysail/data/compatibility/functions/scalar/array.json
+++ b/python/pysail/data/compatibility/functions/scalar/array.json
@@ -23,7 +23,7 @@
     "module": "pyspark.sql.functions",
     "function": "array_contains_all",
     "status": "supported",
-    "note": "not part of Apache Spark SQL or PySpark; sail exclusive"
+    "note": "Sail exclusive; not available in PySpark or Spark SQL"
   },
   {
     "module": "pyspark.sql.functions",


python/pysail/data/compatibility/dataframe.json
```diff
@@ -1,102 +1,393 @@
...
   {
     "module": "pyspark.sql.DataFrame",
-    "function": "show",
-    "status": "supported"
+    "function": "approxQuantile",
+    "status": "not supported"
+  },
+  {
+    "module": "pyspark.sql.DataFrame",
+    "function": "cache",
```
keen85, line 19: @shehabgamin what is the status here?
I know, that `df.cache()` does not fail in Sail; but does it actually do something?

shehabgamin: `df.cache()` is planned but not supported atm. Strange that it does not fail.

keen85: @shehabgamin, thanks for clarification.
In this case I actually appreciate, that it does not fail, because this would block Sail adoption.

There are other operations (e.g. [`checkpoint`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.checkpoint.html), [`localCheckpoint`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.localCheckpoint.html), [`broadcast`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.broadcast.html#)) where simply ignoring the operation at runtime until it is implemented would be acceptable in my eyes. It would allow more users to adopt Sail - without changing their code.

But this should be some **general decision**, how to handle this and maybe to log some warning at runtime but not raising an exception.

docs/guide/functions/scalar.md
keen85, line 1: @shehabgamin, while cross referencing Sail doc with [PySpark doc](https://spark.apache.org/docs/latest/api/python/_sources/reference/pyspark.sql/functions.rst.txt), I found some functions that don't exist in PySpark: so I'd remove those.

Probably this also means that some rust files need to be adjusted right?

shehabgamin: @keen85 The PySpark doc is not complete and doesn't list all available functions. You can add these back in.

python/pysail/data/compatibility/dataframe.json
shehabgamin: @keen85 Good idea! We can make this controlled by a config value to determine whether to fail or emit a warning. Besides `cache`, `checkpoint`, `localCheckpoint`, and `broadcast`, are there any other operations?

docs/guide/functions/scalar.md
linhr: The missing functions are still available in SQL so we need to keep them. Maybe in the JSON file we can remove its `module` field?

python/pysail/data/compatibility/functions/scalar/sort.json
```diff
@@ -0,0 +1,32 @@
+[
```
linhr, line 1: This seems a duplicate of `column.json`.

linhr (approved):
Thanks for having this! I can see that a lot of time has been devoted to the accuracy of these JSON files, so I really appreciate the contribution!!

Let me have a follow-up PR to use these JSON files to generate documentations so that the data here becomes the single source of truth moving forward.

There are a few "not planned" functions that we can actually support. I guess there are no corresponding GitHub issues either because the functions were not discovered in our Spark test suites, or were introduced in recent Spark versions. We'll update the JSON files once we create issues for them.

shehabgamin (approved):

python/pysail/data/compatibility/functions/scalar/sort.json
keen85: @linhr, I thought so too first. But there is a subtle difference:
- `df.sort(F.col("id").asc())`
  - `asc` is a [function of a Column object](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.asc.html).
- `df.sort(F.asc(F.col("id")))`
  - `asc` is a [_standalone_ function where a Column object is passed in as an argument](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.asc.html).

linhr: Ahh I see. This makes perfect sense. Thanks for catching such subtlety!

PR #1132, opened 2025-12-16
feat: add snapshot validation for query plan benchmarking

lonless9:


lonless9:
Done in https://github.com/lakehq/sail/pull/1509

PR #1133, opened 2025-12-16
fix: ensure delta log json files include operation details and set up snapshot tests for Delta Lake

lonless9:
This is a follow-up of #1093

### DIFF
--- a/crates/sail-delta-lake/src/physical_plan/writer_exec.rs
+++ b/crates/sail-delta-lake/src/physical_plan/writer_exec.rs
@@ -81,7 +81,7 @@ pub struct DeltaWriterExec {
     table_exists: bool,
     sink_schema: SchemaRef,
     condition: Option<Arc<dyn PhysicalExpr>>,
-    /// Optional override for commit operation metadata (used by MERGE/DELETE/UPDATE planners).
+    /// Optional override for commit operation metadata.
     operation_override: Option<DeltaOperation>,
     cache: PlanProperties,
 }


python/pysail/tests/spark/delta/__snapshots__/test_delta_features/test_explain_codegen_includes_plan_steps_and_delta_rewrite_artifacts_for_delete.plan
```diff
@@ -0,0 +1,342 @@
...
+                      CooperativeExec
+                        RepartitionExec: partitioning=RoundRobinBatch(<partitions>), input_partitions=<partitions>
+                          CooperativeExec
+                            DeltaFindFilesExec(table_path=file://<tmp>/delete_explain_codegen/)
+      DeltaRemoveActionsExec
+        CoalescePartitionsExec
+          CooperativeExec
+            RepartitionExec: partitioning=RoundRobinBatch(<partitions>), input_partitions=<partitions>
+              CooperativeExec
+                DeltaFindFilesExec(table_path=file://<tmp>/delete_explain_codegen/)
```
linhr, line 342: From the plan it's now obvious that the entire table is rewritten. I guess there is an opportunity for optimization so that we only rewrite files that contain rows matching the predicate?

linhr (approved):
This is great! 🚀

python/pysail/tests/spark/delta/__snapshots__/test_delta_features/test_explain_codegen_includes_plan_steps_and_delta_rewrite_artifacts_for_delete.plan
lonless9: I will track this issue in the following PR!

PR #1135, opened 2025-12-17
refactor: split driver logic

linhr:
This PR splits the control plane logic in `DriverActor` into components such as `WorkerPool` and `JobScheduler`. We no longer maintain a single `DriverState` instance since this makes the actor handler logic harder to extend.

Besides the refactoring, this PR makes the following changes to the cluster protocol:
* The task ID is now only unique within a job. We no longer maintain a global counter for task IDs.
* The worker location (host and port) is sent in a separate field when running tasks. Previously the worker location is encoded in the physical plan (done by the codec for `ShuffleReadExec` and `ShuffleWriteExec`).


shehabgamin (approved):
Well done!

Issue #520, opened 2025-06-06
Support Arrow Flight SQL


linhr:
We would like to support Arrow Flight SQL as an alternative protocol to Spark Connect. This would allow integration with SQL clients (e.g. BI tools) outside of PySpark. The user would be able to start the server via something like `sail flight server` in the command line.

Using the [Arrow Flight SQL JDBC Driver](https://arrow.apache.org/docs/java/flight_sql_jdbc_driver.html), Java applications can connect to Sail via `jdbc:arrow-flight-sql://`. This is similar to how the [Spark Thrift Server](https://www.russellspitzer.com/2017/05/19/Spark-Sql-Thriftserver/) allows connection via `jdbc:hive2://`.

(As far as I can tell, the Spark Thrift Server uses the JDBC protocol, and do not use Apache Thrift.)

Ballista used to support Arrow Flight SQL (~1,000 LoC) but the functionality has been removed (<https://github.com/apache/datafusion-ballista/pull/1228>). I expect the effort would be similar in Sail. The design should be straightforward and not too much knowledge of Sail internals is required.

DeliveranceTechSolutions:
I've worked quite a bit with raw arrow tables so I'd like to take a crack at this if possible?

I mainly use Go but I've built small things in rust.

shehabgamin:
> I've worked quite a bit with raw arrow tables so I'd like to take a crack at this if possible?
> 
> I mainly use Go but I've built small things in rust.

@DeliveranceTechSolutions Thank you for your interest in this issue! Your expertise with Arrow tables sounds like a great fit.

Please feel free to take a crack at this issue. If you have any questions about the codebase or need any help getting started, do not hesitate to ask. We are happy to help along the way!

shehabgamin:
@DeliveranceTechSolutions How are things going? Let us know if you need help with anything!

DeliveranceTechSolutions:
@shehabgamin hey Shehab, some things came up on a project I'm on, I've been looking over crates and I have the arrow flight part implemented; however, I'm looking into how it'll interface with you guys.  Rust hard lol

shehabgamin:
@DeliveranceTechSolutions Rust is indeed hard! You'll get the hang of it quickly.

Feel free to open up a draft PR if you'd like. @linhr, @lonless9, and I are happy to give pointers and lend a hand.

Also, if communicating over Slack would be easier/better for you, feel free to join our Community Slack:
https://lakesail.com/slack

[...]

PR #1138, opened 2025-12-18
refactor: session manager

linhr:
1. Move `SessionManager` to the `sail-session` crate and remove its dependency on concepts specific to Spark. This would be helpful when we implement Arrow Flight server (#520).
2. Consolidate session creation logic into `SessionFactory` in the `sail-session` crate. This would make it easier to reason about how the session is created for local/driver/worker use cases. It would also make it cleaner to implement session memory/disk configuration in a future PR.
3. Move `JobRunner`, `PlanFormatter`, and `CatalogDisplay` to session extensions. The similar idea can later be used for SQL parser and plan resolver. This means **all these components are now (or soon) accessible at various query planning stages**. This would make it possible to implement distributed planning for operations such as `tail`. We could also get rid of the unfortunate dependency on sqlparser very soon. (Right now we use sqlparser for Delta Lake expression parsing and schema parsing in the `from_csv` function.)

The number of changed files is surprisingly large in this PR since I have to move a few `struct`s/`enum`s from `sail-catalog` to `sail-common-datafusion` to accommodate `CatalogDisplay`, so the `use` statements in many files have changed.


crates/sail-session/src/session_factory/worker.rs
```diff
@@ -0,0 +1,39 @@
...
+impl<'a> SessionFactory<()> for WorkerSessionFactory<'a> {
+    fn create(&mut self, _info: ()) -> Result<SessionContext> {
+        let runtime = {
+            let registry = DynamicObjectStoreRegistry::new(self.runtime.clone());
+            let builder =
+                RuntimeEnvBuilder::default().with_object_store_registry(Arc::new(registry));
+            Arc::new(builder.build()?)
+        };
+        let config = SessionConfig::default();
+        let state = SessionStateBuilder::new()
```
shehabgamin, line 31: In `crates/sail-session/src/session_factory/server.rs` we have this:
```
    fn create_session_state(&mut self, info: &I) -> Result<SessionState> {
        let config = self.create_session_config(info)?;
        let runtime = self.create_runtime_env(info)?;
        let builder = SessionStateBuilder::new()
            .with_config(config)
            .with_runtime_env(runtime)
            .with_default_features()
            .with_analyzer_rules(default_analyzer_rules())
            .with_optimizer_rules(default_optimizer_rules())
            .with_physical_optimizer_rules(get_physical_optimizers(PhysicalOptimizerOptions {
                enable_join_reorder: self.config.optimizer.enable_join_reorder,
            }))
            .with_query_planner(new_query_planner());
        let builder = self.mutator.mutate_state(builder, info)?;
        Ok(builder.build())
    }
```

Should the construction of `SessionStateBuilder` be consistent across the codebase? If so, should `create_session_state` be a shared utility function that multiple crates can use?

shehabgamin: I checked out this branch and found more usage of `SessionStateBuilder`:
- https://github.com/lakehq/sail/blob/d67c835386a65f5ffb0f891aa935006b7a49cbac/crates/sail-delta-lake/src/physical_plan/commit_exec.rs#L428
- https://github.com/lakehq/sail/blob/d67c835386a65f5ffb0f891aa935006b7a49cbac/crates/sail-delta-lake/src/physical_plan/find_files_exec.rs#L140
- https://github.com/lakehq/sail/blob/d67c835386a65f5ffb0f891aa935006b7a49cbac/crates/sail-delta-lake/src/physical_plan/scan_by_adds_exec.rs#L108
- https://github.com/lakehq/sail/blob/d67c835386a65f5ffb0f891aa935006b7a49cbac/crates/sail-delta-lake/src/table_format.rs#L290

crates/sail-session/src/session_factory/worker.rs
```diff
@@ -0,0 +1,39 @@
...
+
+impl<'a> WorkerSessionFactory<'a> {
+    pub fn new(runtime: &'a RuntimeHandle) -> Self {
+        Self { runtime }
+    }
+}
+
+impl<'a> SessionFactory<()> for WorkerSessionFactory<'a> {
+    fn create(&mut self, _info: ()) -> Result<SessionContext> {
+        let runtime = {
```
shehabgamin, line 24: In `crates/sail-session/src/session_factory/server.rs` we have this:
```
    fn create_runtime_env(&mut self, info: &I) -> Result<Arc<RuntimeEnv>> {
        let registry = DynamicObjectStoreRegistry::new(self.runtime.clone());
        let cache_config = CacheManagerConfig::default()
            .with_files_statistics_cache(self.create_file_statistics_cache())
            .with_list_files_cache(self.create_file_listing_cache())
            .with_file_metadata_cache(Some(self.create_file_metadata_cache()));
        let builder = RuntimeEnvBuilder::default()
            .with_object_store_registry(Arc::new(registry))
            .with_cache_manager(cache_config);
        let builder = self.mutator.mutate_runtime_env(builder, info)?;
        Ok(Arc::new(builder.build()?))
    }
 ```
   
Should the construction of `RuntimeEnvBuilder` be consistent across the codebase?

File: crates/sail-session/src/session_factory/server.rs [hunk 3]
```diff
@@ -0,0 +1,309 @@
...
+        parquet.created_by = concat!("sail version ", env!("CARGO_PKG_VERSION")).into();
+        parquet.enable_page_index = self.config.parquet.enable_page_index;
+        parquet.pruning = self.config.parquet.pruning;
+        parquet.skip_metadata = self.config.parquet.skip_metadata;
+        parquet.metadata_size_hint = self.config.parquet.metadata_size_hint;
+        parquet.pushdown_filters = self.config.parquet.pushdown_filters;
+        parquet.reorder_filters = self.config.parquet.reorder_filters;
+        parquet.schema_force_view_types = self.config.parquet.schema_force_view_types;
+        parquet.binary_as_string = self.config.parquet.binary_as_string;
+        parquet.max_predicate_cache_size = Some(self.config.parquet.max_predicate_cache_size);
```
linhr, line 282: I've incorporated this line from #1141.

crates/sail-session/src/session_factory/worker.rs
linhr: Yeah I plan to consolidate it in a follow-up PR. The purpose of this PR is to move session construction logic to the same place so that such inconsistency problem becomes obvious.

crates/sail-session/src/session_factory/worker.rs
linhr: Nice find!

The goal is that all the other crates should not construct sessions but accept sessions passed as function arguments. I believe the (improper) session creation in Delta Lake is inherited from the upstream and we need to clean up these places.

PR #1139, opened 2025-12-19
fix: correct MergeIntoWriteNode Hash & avoid the side effect of explain statement

lonless9:


crates/sail-logical-plan/src/merge.rs
```diff
@@ -203,14 +204,18 @@ impl UserDefinedLogicalNodeCore for MergeIntoNode {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Educe)]
+#[educe(Eq, Hash, PartialOrd)]
 pub struct MergeIntoWriteNode {
```
linhr, line 190: We need to handle `MergeIntoNode` as well.

crates/sail-plan/src/explain.rs
```diff
@@ -168,6 +168,11 @@ async fn collect_plan_with(
     let initial_logical = plan.clone();
     let mut stringified = vec![initial_logical.to_stringified(PlanType::InitialLogicalPlan)];
 
+    // TODO: The current implementation executes the logical plan to get the DataFrame,
+    // which triggers side effects for commands like CREATE TABLE or INSERT.
+    // This should be avoided in EXPLAIN. A possible solution is to instead rewrite the logical plan
+    // to replace command nodes with dummy nodes (e.g., empty relation with the same schema)
+    // to allow generating the physical plan and showing potential errors without execution.
     let df = execute_logical_plan(ctx, plan).await?;
```
linhr: The comment is nice! Let's comment out the code as well to avoid the side effect.

```rust
// let df = execute_logical_plan(ctx, plan).await?;
```

crates/sail-logical-plan/src/merge.rs
lonless9: Nice catch!

crates/sail-plan/src/explain.rs
lonless9: `WithPreconditionsNode` appears to be an unrecognized logical plan node, and I tend to ignore it for now.

crates/sail-plan/src/explain.rs
```diff
@@ -310,6 +317,24 @@ async fn collect_plan_with(
     })
 }
 
+/// Remove `WithPreconditionsNode` wrappers without executing them.
+/// FIXME: This is a temporary solution to avoid executing this logical plan during EXPLAIN.
```
linhr: Good idea!

We also need to strip `CatalogCommandNode` which is the actual cause of side effect. Some catalog commands are not wrapped by the precondition node.

linhr: We could possibly replace `CatalogCommandNode` with an empty plan but we need to use the same schema.

lonless9: Done!

PR #1140, opened 2025-12-19
chore: various ops should not fail

shehabgamin:
cc @keen85 https://github.com/lakehq/sail/pull/1127#discussion_r2617238425

Tested manually:
```

spark = SparkSession.builder.remote(f"sc://localhost:50051").getOrCreate()

df = spark.createDataFrame([(1, "a"), (2, "b"), (3, "c")],["id", "letter"])
df_cached = df.cache()
df.persist()
df_cached.unpersist()

df.checkpoint(True)
df.checkpoint(False)

df.localCheckpoint(True)
df.localCheckpoint(False)

big = spark.range(0, 100)
small = spark.createDataFrame([(1, "one"), (2, "two"), (3, "three")], ["id", "label"])
joined = big.join(sf.broadcast(small), on="id", how="left")
joined.show()
```

### DIFF
--- a/crates/sail-plan/src/resolver/query/mod.rs
+++ b/crates/sail-plan/src/resolver/query/mod.rs
@@ -1,6 +1,5 @@
 use async_recursion::async_recursion;
 use datafusion_expr::{Expr, LogicalPlan, LogicalPlanBuilder};
-use log::warn;
 use sail_common::spec;
 
 use crate::error::{PlanError, PlanResult};
@@ -216,14 +215,10 @@ impl PlanResolver<'_> {
                     .await?
             }
             QueryNode::CachedLocalRelation { .. } => {
-                // TODO: Implement
-                warn!("CachedLocalRelation operation is not yet supported and is a no-op");
-                self.resolve_query_empty(false)?
+                return Err(PlanError::todo("cached local relation"));
             }
             QueryNode::CachedRemoteRelation { .. } => {
-                // TODO: Implement
-                warn!("CachedRemoteRelation operation is not yet supported and is a no-op");
-                self.resolve_query_empty(false)?
+                return Err(PlanError::todo("cached remote relation"));
             }
             QueryNode::CommonInlineUserDefinedTableFunction(udtf) => {
                 self.resolve_query_common_inline_udtf(udtf, state).await?


crates/sail-plan/src/resolver/query/misc.rs
```diff
@@ -130,12 +131,14 @@ impl PlanResolver<'_> {
...
+        input: spec::QueryPlan,
         _name: String,
         _parameters: Vec<spec::Expr>,
-        _state: &mut PlanResolverState,
+        state: &mut PlanResolverState,
     ) -> PlanResult<LogicalPlan> {
-        Err(PlanError::todo("hint"))
+        // TODO: Implement
+        warn!("Hint operation is not yet supported and is a no-op");
+        self.resolve_query_plan(input, state).await
```
shehabgamin: Not sure if `resolve_query_plan_with_hidden_fields` should be used instead.

crates/sail-plan/src/resolver/query/mod.rs
```diff
@@ -215,10 +216,14 @@ impl PlanResolver<'_> {
...
-                return Err(PlanError::todo("cached local relation"));
+                // TODO: Implement
+                warn!("CachedLocalRelation operation is not yet supported and is a no-op");
+                self.resolve_query_empty(false)?
             }
             QueryNode::CachedRemoteRelation { .. } => {
-                return Err(PlanError::todo("cached remote relation"));
+                // TODO: Implement
+                warn!("CachedRemoteRelation operation is not yet supported and is a no-op");
+                self.resolve_query_empty(false)?
```
linhr: These are fetching "cached relations" instead of "caching relations". IMO these should fail since we don't know how to return the expected data.

linhr (approved):

crates/sail-plan/src/resolver/query/misc.rs
linhr: Yes we should use `resolve_query_plan_with_hidden_fields`.

crates/sail-plan/src/resolver/query/mod.rs
shehabgamin: @linhr  I found that after adding the no-op for `CheckpointCommandResult`, a request is sent to the server for fetching "cached relations" when calling `checkpoint` and `localCheckpoint`. This is because those commands return data:
- https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.localCheckpoint.html
- https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.checkpoint.html

Given this, should we still fail here? I'm open either way.

linhr: I thought the empty result may break things due to schema mismatch, but if this yields better client-side behavior than returning errors, let's keep the change then.

shehabgamin: This is what ends up happening in Sail with these changes:
```
>>> df.localCheckpoint(False)
DataFrame[]
>>> df.localCheckpoint(False).show()
++
++
++

>>> df.localCheckpoint(True)
DataFrame[]
>>> df.localCheckpoint(True).show()
++
++
++

>>> df.checkpoint(True)
DataFrame[]
>>> df.checkpoint(True).show()
++
++
++

>>> df.checkpoint(False)
DataFrame[]
>>> df.checkpoint(False).show()
++
++
++
```

Versus Spark's behavior:
```
>>> df.checkpoint(True)
DataFrame[age: bigint, name: string]
>>> df.checkpoint(True).show()
+---+-----+
|age| name|
+---+-----+
| 14|  Tom|
| 23|Alice|
| 16|  Bob|
+---+-----+

>>> df.checkpoint(False)
DataFrame[age: bigint, name: string]
>>> df.checkpoint(False).show()
+---+-----+
|age| name|
+---+-----+
| 14|  Tom|
| 23|Alice|
| 16|  Bob|
+---+-----+

>>> df.localCheckpoint(True)
DataFrame[age: bigint, name: string]
>>> df.localCheckpoint(True).show()
+---+-----+
|age| name|
+---+-----+
| 14|  Tom|
| 23|Alice|
| 16|  Bob|
+---+-----+

>>> df.localCheckpoint(False)
DataFrame[age: bigint, name: string]
>>> df.localCheckpoint(False).show()
+---+-----+
|age| name|
+---+-----+
| 14|  Tom|
| 23|Alice|
| 16|  Bob|
+---+-----+
```

So it does break things for users that are using the results of those commands.

crates/sail-plan/src/resolver/query/misc.rs
shehabgamin: @linhr Will do, can you explain why though?

linhr: It has been a while so I don't have the details on top of my mind, but I just looked at similar logic such as `resolve_query_subquery_alias` and I feel we can follow the same pattern here.

crates/sail-plan/src/resolver/query/mod.rs
linhr: If that's the case, an explicit error could be better user experience. So probably we're not yet ready to implement the checkpoint command.

PR #1141, opened 2025-12-20
chore: update parquet config options

shehabgamin:
- Add `max_predicate_cache_size` which is an Arrow config that DataFusion exposes.
- Update default value of `metadata_size_hint` to align with DataFusion. This should lead to better read performance.


crates/sail-common/src/config/application.yaml
```diff
@@ -336,6 +336,16 @@
     (Reading) Whether to use available bloom filters when reading Parquet files.
   experimental: true
 
+- key: parquet.max_predicate_cache_size
+  type: number
+  default: "104857600"
```
shehabgamin, line 344: DataFusion defaults to `None` which defaults to the default Arrow Parquet reader setting of `104857600` (`100 * 1024 * 1024`)

PR #1159, opened 2025-12-24
feat: clean up session and SQL usage for Delta Lake

linhr:
This is the first attempt to remove ad-hoc session creation and SQL parsing/unparsing in our Delta Lake integration. There are still a few blockers and I added comments in the code for future work.

See also <https://github.com/lakehq/sail/pull/1138#discussion_r2644660451> for a related discussion for these changes.


crates/sail-common-datafusion/src/datasource.rs
```diff
@@ -131,31 +131,24 @@ pub enum MergeNotMatchedByTargetActionInfo {
...
     pub predicate: Option<String>,
 }
 
 /// Optional override metadata for operation commit logs (currently used by Delta MERGE).
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[derive(Debug, Clone)]
 pub enum OperationOverride {
-    // TODO: extend or rename like `DeltaMerge`, `IcebergMerge` to better distinguish between different implementations.
-    #[serde(rename_all = "camelCase")]
```
linhr, line 154: It seems that `MergePredicateInfo` and `OperationOverride` do not need to support serde since it will be converted to the corresponding structs in the `sail-delta-lake` crate.

crates/sail-delta-lake/src/physical_plan/writer_exec.rs
```diff
@@ -83,7 +82,6 @@ pub struct DeltaWriterExec {
     sink_mode: PhysicalSinkMode,
     table_exists: bool,
     sink_schema: SchemaRef,
-    condition: Option<Arc<dyn PhysicalExpr>>,
```
linhr, line 86: I removed `condition` since it can be inferred from `sink_mode`.

crates/sail-physical-plan/src/merge.rs
linhr, line 145: These fields appear unused. Are they for usage in the future?

python/pysail/tests/spark/delta/__snapshots__/test_delta_features.ambr
```diff
@@ -131,7 +131,7 @@
       'writeTimeMs': '<time_ms>',
     }),
     'operationParameters': dict({
-      'predicate': 'department = Engineering',
+      'predicate': "department = 'Engineering' ",
```
linhr, line 134: We now have predicates that are more likely to be syntactically correct.

python/pysail/tests/spark/delta/__snapshots__/test_delta_features.ambr
```diff
@@ -93,14 +93,14 @@
       'matchedPredicates': list([
         dict({
           'actionType': 'update',
-          'predicate': "(flag = 'archive')",
+          'predicate': "t . flag = 'archive' ",
         }),
       ]),
-      'mergePredicate': '(id = __sail_src_id)',
+      'mergePredicate': 't . id = s . id ',
```
linhr, line 99: The predicates are now user-facing. But this also means that we should not "unparse" it. I added a `FIXME` in `DeltaCommitExec`. As far as I can tell, the "unparsed" physical expressions are only passed to the conflict checker, but the conflict checker does not use it currently. So this does not break things if my analysis is correct.

Note that there are unnecessary whitespaces in the SQL string since the `TreeText` trait does not have sophisticated logic for formatting SQL ASTs yet.

crates/sail-physical-plan/src/merge.rs
lonless9: These fields are likely remnants after adopting the expand_merge strategy. If there are any subsequent merge operations performed with pre_expanded: false, these fields may need to be retained. But in my plan all other formats of merge operations are intended to adopt this rewrite strategy, so I think these fields and the related condition judgments in build_merge_plan can be safely removed. This can be completed in a subsequent PR.

crates/sail-delta-lake/src/physical_plan/commit_exec.rs
```diff
@@ -425,6 +425,13 @@ impl ExecutionPlan for DeltaCommitExec {
             };
             let reference = snapshot.as_ref().map(|s| *s as &dyn TableReference);
 
+            // FIXME: avoid creating a new session state
+            //   This session state is used to parse predicates in the commit info
+            //   back to physical expressions. But the predicate string is not always available
+            //   (e.g., for DML operations constructed from the DataFrame API). Should we
+            //   store the physical expression directly when creating `DeltaCommitExec`?
+            //   The physical expression should be known at planning time so it does not
+            //   need to be sent back from the writer.
```
lonless9, line 434: Yes the current round-tripping is a bit too lengthy. We really should find a way to pass it directly to `DeltaCommitExec` during the planning construction phase, rather than having it travel as a string through the entire data stream.

python/pysail/tests/spark/delta/__snapshots__/test_delta_features.ambr
lonless9: Yes, currently it's only for informational purposes. I will open a PR later to clean up transaction-related issues. In addition to the UDF issue in predicates we previously mentioned, aliases column name might also cause problems in conflict checking.

crates/sail-common-datafusion/src/physical_expr.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl Eq for PhysicalExprWithSource {}
+
+impl Hash for PhysicalExprWithSource {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.expr.as_ref().hash(state);
+        self.source.hash(state);
+    }
+}
```
lonless9, line 34: I'm not sure whether including the source string when implementing `PartialEq` and `Hash` for `PhysicalExprWithSource` or `ExprWithSource` might lead to some unexpected side effects.

lonless9 (approved):
Excellent! With a healthier session management, we can now clearly clean up all related side effects going forward.

crates/sail-common-datafusion/src/physical_expr.rs
linhr: I'm fine either way. We can remove it if needed in follow-up PRs. For now, `ExprWithSource` and `PhysicalExprWithSource` are not part of the actual logical/physical plan, so the optimizer won't be affected.

PR #1161, opened 2025-12-26
feat: arrow-native actions and distributed log scanning for Delta Lake

lonless9:


crates/sail-physical-plan/src/format_tag.rs
```diff
@@ -0,0 +1,138 @@
...
+            FormatTag::DuckLake => write!(f, "DuckLake"),
+            FormatTag::Custom(name) => write!(f, "{}", name),
+        }
+    }
+}
+
+lazy_static::lazy_static! {
+    static ref FORMAT_TAG_REGISTRY: RwLock<HashMap<TypeId, FormatTag>> =
+        RwLock::new(HashMap::new());
+}
```
lonless9, line 33: Should we place this file in `sail_common_datafusion`? Is there a better session level registry logic?

crates/sail-delta-lake/src/physical_plan/scan_by_adds_exec.rs
```diff
@@ -187,71 +127,134 @@ impl ExecutionPlan for DeltaScanByAddsExec {
...
+            let mut input_stream = input.execute(partition, context.clone())?;
+            let mut candidate_adds = Vec::new();
+            let mut is_partition_scan = true;
 
-            while let Some(batch_result) = input_stream.next().await {
-                let batch = batch_result?;
+            while let Some(batch) = input_stream.try_next().await? {
                 if batch.num_rows() == 0 {
                     continue;
                 }
```
lonless9, line 144: I plan to later buffer Adds here and execute them in a streaming manner using `async_stream`, which may require the involvement of a parquet opener or similar infrastructure. Perhaps we should take a look at the implementation of `filestream`, though it is not defined as `pub`.

crates/sail-delta-lake/src/physical_plan/log_scan_exec.rs
```diff
@@ -0,0 +1,341 @@
...
+                if batch.num_rows() == 0 {
+                    continue;
+                }
+                let out = this.to_file_rows(&batch)?;
+                if out.num_rows() == 0 {
+                    continue;
+                }
+                yield out;
+            }
+        };
```
lonless9: I'm curious if this makes sense.

crates/sail-delta-lake/src/datasource/expressions.rs
```diff
@@ -16,35 +16,24 @@
...
-    AggregateUDF, BinaryExpr, Expr, Operator, ScalarUDF, TableProviderFilterPushDown, TableSource,
-};
+use datafusion::logical_expr::{BinaryExpr, Expr, Operator, TableProviderFilterPushDown};
 use datafusion::optimizer::simplify_expressions::ExprSimplifier;
 use datafusion::physical_expr::PhysicalExpr;
 use datafusion::physical_plan::expressions::Column as PhysicalColumn;
-use datafusion::sql::planner::{ContextProvider, SqlToRel};
-use datafusion::sql::sqlparser::dialect::GenericDialect;
-use datafusion::sql::sqlparser::parser::Parser;
-use datafusion::sql::sqlparser::tokenizer::Tokenizer;
```
lonless9, line 41: I guess there's no usage of sqlparser in this crate anymore.

crates/sail-plan/src/resolver/command/write_v1.rs
```diff
@@ -29,6 +29,14 @@ impl PlanResolver<'_> {
...
         } = write;
 
+        let replace_where = options.iter().find_map(|(k, v)| {
+            if k.eq_ignore_ascii_case("replaceWhere") || k.eq_ignore_ascii_case("replace_where") {
+                Some(v.clone())
+            } else {
+                None
+            }
+        });
+
```
lonless9, line 39: The v1 option condition was previously consuming at an incorrect location.

python/pysail/tests/spark/analyst/__snapshots__/test_features/test_explain_analyze_executes_and_returns_physical_plan.plan
```diff
@@ -1,10 +1,10 @@
...
-  ProjectionExec: expr=[#2@0 as #4, sum(t.#3)@1 as #5], metrics=[output_rows=<metric>, elapsed_compute=<metric>, output_bytes=<metric>], statistics=[Rows=Inexact(2), Bytes=Inexact(<bytes>), [(Col[0]:),(Col[1]:)]], schema=[#4:Int32;N, #5:Int64;N]
-    AggregateExec: mode=FinalPartitioned, gby=[#2@0 as #2], aggr=[sum(t.#3)], metrics=[output_rows=<metric>, elapsed_compute=<metric>, output_bytes=<metric>, spill_count=<metric>, spilled_bytes=<metric>, spilled_rows=<metric>, peak_mem_used=<metric>, aggregate_arguments_time=<metric>, aggregation_time=<metric>, emitting_time=<metric>, time_calculating_group_ids=<metric>], statistics=[Rows=Inexact(2), Bytes=Absent, [(Col[0]:),(Col[1]:)]], schema=[#2:Int32;N, sum(t.#3):Int64;N]
+ProjectionExec: expr=[#4@<i> as k, #5@<i> as SUM(v)], metrics=[output_rows=<metric>, elapsed_compute=<metric>, output_bytes=<metric>], statistics=[Rows=Inexact(2), Bytes=Inexact(<bytes>), [(Col[0]:),(Col[1]:)]], schema=[k:Int32;N, SUM(v):Int64;N]
```
linhr: I assume the statistics also refer to the column via `Col[0]` etc. Is the index stable enough? I thought it would be the same as `@<i>` indices (if I understand the notation correctly).

python/pysail/tests/spark/steps/plan.py
```diff
@@ -72,6 +72,54 @@ def normalize_path(path: str) -> str:
         flags=re.IGNORECASE,
     )
 
+    # Normalize unstable expression / column indices (can vary across runs when a session is reused).
```
linhr: Interesting. It's a bit counter-intuitive to me. I thought the column indices would be deterministic even with session reuse.

python/pysail/tests/spark/delta/features/merge.feature
```diff
@@ -509,3 +545,99 @@ Feature: Delta Lake Merge
...
+        """
+        INSERT INTO delta_merge_explain_partition_pushdown
+        SELECT * FROM VALUES
+          (1, 2023, 100),
+          (2, 2024, 200),
+          (3, 2024, 700)
+        """
+      Given statement
+        """
+        CREATE OR REPLACE TEMP VIEW src_merge_explain_partition_pushdown AS
```
linhr, line 611: We need to drop the `TEMP VIEW` in `final statement`.

python/pysail/tests/spark/delta/features/log_scan_parallel.feature
```diff
@@ -0,0 +1,77 @@
...
+        INSERT INTO delta_logscan_parallel SELECT * FROM VALUES (2, 'v2')
+        """
+      Given statement
+        """
+        INSERT INTO delta_logscan_parallel SELECT * FROM VALUES (3, 'v3')
+        """
+
+      Given statement
+        """
+        CREATE OR REPLACE TEMP VIEW src_logscan_parallel AS
```
linhr, line 39: We need to drop the `TEMP VIEW` in `final statement`.

crates/sail-physical-plan/src/format_tag.rs
linhr: In general, the static registry may cause surprising behavior in tests if we are not careful enough to isolate test executions. (It doesn't seem an issue with the current tests though.) We can move it to the session but of course this can be future work.

crates/sail-delta-lake/src/physical_plan/discovery_exec.rs
```diff
@@ -0,0 +1,487 @@
...
+                    }
+
+                    let table_provider =
+                        DeltaTableProvider::try_new(snapshot, log_store, scan_config)
+                            .map_err(|e| DataFusionError::External(Box::new(e)))?;
+
+                    let session_state = SessionStateBuilder::new()
+                        .with_runtime_env(context.runtime_env().clone())
+                        .with_config(context.session_config().clone())
+                        .build();
```
linhr, line 362: Creating session inside `execute()` indicates that there might be something we can do to improve the architecture. I guess we can revisit whether `DeltaTableProvider` is still the right abstraction moving forward.

crates/sail-delta-lake/src/physical_plan/discovery_exec.rs
```diff
@@ -0,0 +1,487 @@
...
+                let size_arr = batch.column_by_name("size_bytes").and_then(|c| {
+                    c.as_any()
+                        .downcast_ref::<datafusion::arrow::array::Int64Array>()
+                });
+                let mod_time_arr = batch.column_by_name("modification_time").and_then(|c| {
+                    c.as_any()
+                        .downcast_ref::<datafusion::arrow::array::Int64Array>()
+                });
+
+                let part_arrays: Vec<(String, Arc<dyn Array>)> = partition_columns
```
linhr, line 244: I guess we have to work with individual columns in `RecordBatch` instead of using `serde_arrow` here due to the dynamic schema caused by partition columns?

It seems we convert `MapArray` to columns in `DeltaLogScanExec` and then here we read it back as a `HashMap`. I'm not sure if this roundtrip is due to any limitations in the code I'm not aware of.

crates/sail-delta-lake/src/physical_plan/action_schema.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AddAction {
+    path: String,
+    partition_values: BTreeMap<String, Option<String>>,
+    size: i64,
+    modification_time: i64,
+    data_change: bool,
+    stats_json: Option<String>,
```
linhr, line 56: Once we stop depending on kernel for data structures, we can do JSON serde + Arrow serde together (via `serde(deserialize_with = "...")`) so that we stored parsed `stats` instead of `stats_json` (which requires the downstream to do extra serde). This will fully isolate serde from application logic.

crates/sail-delta-lake/src/physical_plan/action_schema.rs
```diff
@@ -0,0 +1,394 @@
...
+const COL_PARTITION_VALUES: &str = "partition_values";
+
+static ACTION_FIELDS: LazyLock<
+    std::result::Result<Vec<datafusion::arrow::datatypes::FieldRef>, String>,
+> = LazyLock::new(delta_action_fields_build);
+static ACTION_SCHEMA: LazyLock<std::result::Result<SchemaRef, String>> =
+    LazyLock::new(|| match (*ACTION_FIELDS).as_ref() {
+        Ok(fields) => Ok(Arc::new(Schema::new(fields.clone()))),
+        Err(msg) => Err(msg.clone()),
+    });
```
linhr: I feel `lazy_static!` with `unwrap()` may be simpler. Any error in schema creation should panic the entire program immediately since such constant creation is not supposed to fail.

linhr (approved):
This is exciting! 🚀

Glad to see that we're moving part of the Delta Lake planning logic to physical execution and we've started using `serde_arrow` to simplify metadata processing.

python/pysail/tests/spark/steps/plan.py
lonless9: These issues may have been caused by previous incorrect implementations such as file lookups. After migrating to discovery, these problems should no longer exist. I have removed the placeholderization here.

crates/sail-delta-lake/src/physical_plan/action_schema.rs
lonless9: Yes, I plan to complete this in the subsequent PR, and a lot of redundant logic should be cleaned up.

crates/sail-delta-lake/src/physical_plan/discovery_exec.rs
lonless9: Yeah here the dynamic issue it is.
This is not a limitation; resolving this roundtrip may require enabling `Add` to support columnar/struct, which could involve lots, so I expect to address this issue in a following PR.

crates/sail-delta-lake/src/physical_plan/discovery_exec.rs
lonless9: Yes, the refactoring for the `DeltaTableProvider` should also be on the way.

crates/sail-physical-plan/src/format_tag.rs
lonless9: I will address this issue in a subsequent PR.

python/pysail/tests/spark/delta/features/log_scan_parallel.feature
lonless9: Done!

python/pysail/tests/spark/delta/features/merge.feature
lonless9: Done!

python/pysail/tests/spark/analyst/__snapshots__/test_features/test_explain_analyze_executes_and_returns_physical_plan.plan
lonless9: This could be a previous wrong implementation.

crates/sail-delta-lake/src/physical_plan/action_schema.rs
lonless9: Done!

PR #1163, opened 2025-12-28
fix: resolve streaming write to console sink hanging

santosh-d3vpl3x:
## Summary

Streaming queries writing to console sink experienced multiple issues:
1. **Deadlock**: `ConsoleSinkExec` blocked on unbounded streams
2. **"operation not found"**: PySpark clients failed to reattach after start
3. **Premature inactivity**: `query.isActive` became `False` while still processing

### 1. Console sink deadlock (`ConsoleSinkExec::execute`)
- **Problem**: `for_each().await` waited for entire stream before returning
- **Fix**: Process stream in background `tokio::spawn`, return immediately

### 2. Missing executor registration (`handle_execute_write_stream_operation_start`)
- **Problem**: No reattachable executor registered for the operation
- **Fix**: Create executor with empty stream, add initial output, register it

### 3. Query becoming inactive (`StreamingQuery` lifecycle)
- **Problem**: Returned stream completed immediately, triggering `stopped=true`
- **Fix**: `BackgroundTaskStream` stays pending while task runs, completes only when processing finishes

## Additional Improvements
- **Graceful shutdown**: abort background task when stream is dropped
- **Non-blocking I/O**: stdout writes moved to `spawn_blocking`

## Test Plan
- [x] Verified streaming query to console sink no longer hangs
- [x] Verified `query.isActive` returns `True` while processing
- [x] Verified `query.stop()` properly terminates the query
- [ ] Run existing streaming tests

### DIFF
--- a/crates/sail-data-source/src/formats/console/writer.rs
+++ b/crates/sail-data-source/src/formats/console/writer.rs
@@ -7,22 +7,21 @@ use std::task::{Context, Poll};
 use datafusion::arrow::datatypes::{Schema, SchemaRef};
 use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::arrow::util::pretty::pretty_format_batches;
+use datafusion::common::runtime::SpawnedTask;
 use datafusion::execution::{RecordBatchStream, SendableRecordBatchStream, TaskContext};
 use datafusion::physical_expr::{EquivalenceProperties, Partitioning};
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::{DisplayAs, ExecutionPlan, PlanProperties};
 use datafusion_common::{plan_err, Result};
-use futures::{Stream, StreamExt};
-use futures::FutureExt;
-use tokio::task::JoinHandle;
+use futures::{FutureExt, Stream, StreamExt};
 
 /// A stream that stays active while a background task runs.
 /// - First poll: emits an empty batch to signal the operation started
 /// - Subsequent polls: stays pending until background task completes
-/// - On drop: aborts the background task for graceful shutdown
+/// - On drop: SpawnedTask automatically aborts the background task
 struct BackgroundTaskStream {
     schema: SchemaRef,
-    handle: Option<JoinHandle<()>>,
+    task: Option<SpawnedTask<()>>,
     sent_initial_batch: bool,
 }
 
@@ -38,10 +37,10 @@ impl Stream for BackgroundTaskStream {
         }
 
         // Subsequent polls: wait for background task to complete
-        if let Some(ref mut handle) = self.handle {
-            match handle.poll_unpin(cx) {
+        if let Some(ref mut task) = self.task {
+            match task.poll_unpin(cx) {
                 Poll::Ready(_) => {
-                    self.handle = None;
+                    self.task = None;
                     Poll::Ready(None) // Task done, stream ends
                 }
                 Poll::Pending => Poll::Pending, // Task running, query stays active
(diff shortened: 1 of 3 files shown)


shehabgamin:
@santosh-d3vpl3x Thanks for this PR! It looks like the build is failing, but once you commit the changes after running `cargo +nightly fmt`, the full test suite should continue running.

crates/sail-data-source/src/formats/console/writer.rs
```diff
@@ -80,32 +134,40 @@ impl ExecutionPlan for ConsoleSinkExec {
...
-                        Err(e) => {
-                            format!("error: {e}")
-                        }
-                    };
+        let schema = self.schema();
+
+        // Process the stream in the background and return an empty stream immediately.
+        // This prevents deadlock: the caller can poll the returned stream while we
+        // consume and print batches asynchronously.
+        let handle = tokio::spawn(async move {
```
linhr: `tokio::spawn` should not be used directly in physical execution. DataFusion has its own mechanism to spawn tasks to ensure the tasks are terminated when the stream is dropped.

crates/sail-spark-connect/src/service/plan_executor.rs
```diff
@@ -271,28 +273,47 @@ pub(crate) async fn handle_execute_write_stream_operation_start(
...
+    executor.add_initial_output(out)?;
+
+    let rx = executor.start()?;
+    spark.add_executor(executor)?;
+
     Ok(ExecutePlanResponseStream::new(
         spark.session_id().to_string(),
         operation_id,
-        Box::pin(stream::iter(output)),
+        Box::pin(rx),
```
linhr: I don't think this handler is expected to return executor output stream directly.

linhr (commented):
The console sink is supposed to be blocking since it is designed to work with the Spark structured streaming API only. Since we haven't fully supported streaming queries yet, the console sink was added for development purposes, and it's not intended to be used in normal cases yet.

Therefore, we may not be able to merge this PR in its current state unfortunately. But if there is any use case that needs support, we'd be happy to look into it and see how we can fix it together!

santosh-d3vpl3x:
Thanks for the review! I'll address your code comments:

1. `tokio::spawn` → Replaced with `SpawnedTask` from DataFusion's runtime
2. Executor stream pattern → Reverted to `stream::iter()` matching other handlers, its fine for console

Regarding the general overall comment: I believe the console sink cannot be blocking in `execute()`. Here's why:

In Spark, the console sink is passive - the engine calls `commit()` from a background thread. In
DataFusion, sinks are active - they must consume the stream in `execute()`.

If `execute()` blocks waiting for stream completion:
- For unbounded sources (rate, socket), it blocks forever
- The handler never returns the query ID
- Client can't call `isActive()` or `stop()`

The fix spawns consumption in background and returns a placeholder stream that stays pending until done - same end result as Spark's background thread, just different mechanics.

santosh-d3vpl3x:
Hmm interesting indeed, those sequence of commands did not work for me, thus this PR! Can you tell me more about your local setup, is it local mode or local-cluster?

What I am up to: I wanted to give a shot to python datasources for sail. Most of the first iteration seems to work except for streaming where it hangs. I tried to use streaming without any change directly from main, it hung there too. Probably there is more to this issue, like you said, streaming is still in dev mode, I will focus on getting other sources in better shape first.

EDIT: on a closer reading, @linhr you are right, the implementation in the PR is creating a separate nested tasks inside the existing background task which should already be non-blocking. 

That said, spark does support batch behaviour for console, not quite sure how useful that is in any case.

linhr:
I was using the code from `main` and both `local` and `local-cluster` modes work for me. (The `local-cluster` mode may have issues with stopping the query properly since streaming is in very early development stage and its distributed execution is not well tested.)

If the issue is with Python data source, I guess GIL might be something to consider, especially if the server and the client run in the same Python process. This is just my rough guess.

From the docs it seems there are separate methods for batch and streaming use cases for data source implementations: <https://spark.apache.org/docs/latest/api/python/tutorial/sql/python_data_source.html>. So yeah it might be a good idea to consider the batch aspect of the Python data source API first, as streaming is still under development.

linhr:
> I tried to use streaming without any change directly from main, it hung there too.

This does seem interesting. If there are steps to reproduce, I'd be happy to look into it!

santosh-d3vpl3x:
I have bothered you enough already :) I will try to focus on batch side of the things first and leave streaming for later, there is ample learning curve here so this makes my life a little easier!

PR #1164, opened 2025-12-29
feat: redesign job graph and task scheduling logic

linhr:
This PR aims to deliver a more feature-rich control plane for distributed query execution. We would like to support (1) task retry, (2) task scheduling for blocking and pipelined shuffle, (3) slot sharing, and (4) task placements on the driver.

### Notable Changes

(*I wrote these notes manually without using AI, so it's worth a careful read.*)

1. Remove `TaskId` and `ChannelName` and identify tasks and task streams by *stage*, *partition*, and *channel*. Rename `TaskInstance` to `TaskKey`.
1. Update the `JobGraph` data structure to track stage dependencies.
1. Introduce `TaskAssigner` to reduce the complexity of `WorkerPool`.
1. Schedule tasks by *region* and allow task within a region to share task slots under certain conditions. The tasks within a region are scheduled together with undeterministic order, so we have to update local stream management logic to support fetching streams before they are created.
1. Support task retry and redesign the job output to detect output streams from failed attempts.
1. The driver no longer rewrites the physical plan to use `ShuffleReadExec` and `ShuffleWriteExec` before running task. Instead, a `TaskDefinition` data structure is used to represents logical input/output information and `TaskRunner` is responsible for rewriting the plan to include shuffle readers and writers.
1. Support running tasks and managing task streams on the driver. This is not in use yet but is needed when we later move catalog operations (which requires using the driver session) to physical execution.
1. Add the logic to handle both local and remote streams during job scheduling. The data plane for remote streams is not implemented yet but can later be used to support blocking shuffle.
1. Update metrics attributes for telemetry.

### Testing

I've tested it manually using Python tests. The Delta Lake `MERGE` tests are especially handy since the physical plans are complex enough and cover common physical plan nodes, and they have helped me identify many job graph planner bugs during development.


linhr:
@claude Please review the code in the `sail-execution` crate. There is no need to look at the changes in this PR, but just look at the code at the branch for this PR. Please identify logical bugs and performance issues for distributed job scheduling and execution. Do not create PRs for bug fixes. Do not make any changes to the branch.

linhr:
Responses to Claude's code review:

1. The analysis is incorrect. The race condition won't exist since the actor event loop is single-threaded.
2. This is acceptable for now.
3. This is acceptable for now.
4. Fixed.
5. This is acceptable for now. If the driver actor has stopped, there is no need for further actions.
6. This is acceptable for now.
7. This is acceptable for now.
8. This is acceptable for now.
9. This is by design.
10. Fixed.
11. The analysis is incorrect. The signal is not created yet if the task fails at this point.
12. This is by design.
13. This is acceptable for now but can indeed be optimized via caching.
14. This is acceptable for now.
15. This is by design to avoid borrowing `self`.
16. This is by design.
17. This is by design.
18. The analysis is probably incorrect. I don't think `TaskKeyDisplay` creates formatted strings directly.

crates/sail-execution/src/worker/actor/handler.rs
```diff
@@ -82,7 +64,7 @@ impl WorkerActor {
 
     pub(super) fn handle_start_heartbeat(&mut self, ctx: &mut ActorContext<Self>) -> ActorAction {
         let worker_id = self.options.worker_id;
-        let client = self.driver_client.clone();
+        let client = self.driver_client_set.core.clone();
         let interval = self.options.worker_heartbeat_interval;
         ctx.spawn(async move {
             loop {
```
lonless9, line 70: Do we need to clean up this spawned task when `server.stop` is called?

crates/sail-execution/src/plan/shuffle_read.rs
lonless9, line 1: Is it needed to limit the number of concurrently open streams?
https://github.com/lakehq/sail/pull/1164/files#diff-0c4f854439e08d2707d290effe684def5c6507c18795cd14be483763f86613cdR132-R142

crates/sail-execution/src/worker/peer_tracker/core.rs
lonless9, line 1: If the pod is rebuilt, will the binding between the ID and IP/port still remain secure?
```rust
        for peer in peers {
            self.peers
                .entry(peer.worker_id)
                .or_insert_with(|| Peer::new(peer.host, peer.port));
        }
```

https://github.com/lakehq/sail/pull/1164/files#diff-3964d0d009c064af780377df36a2f5a328a6c7ec3d6da7607614e08f30543bbeR25-R29

crates/sail-execution/src/job_graph/planner.rs
```diff
@@ -173,36 +167,99 @@ fn build_job_graph(
...
+    let schema = plan.schema();
+    let partitioning = plan.output_partitioning().clone();
+    let (plan, inputs) = rewrite_inputs(plan.clone())?;
+    let stage = Stage {
+        inputs,
+        plan,
+        group: String::new(),
+        mode: OutputMode::Pipelined,
+        distribution: OutputDistribution::RoundRobin { channels: 1 },
+        placement: TaskPlacement::Worker,
```
lonless9, line 197: Obviously we will implement `OutputMode::Blocking` to break up potentially large regions. Perhaps we can also implement batch scheduling within a region, I don’t have a clear idea yet.

crates/sail-execution/src/worker_manager/kubernetes.rs
lonless9, line 1: The current stop implementation seems to be a no-op. Would a long-running driver accumulate many completed or error pods?

crates/sail-execution/src/worker/actor/core.rs
```diff
@@ -24,20 +24,20 @@ impl Actor for WorkerActor {
     }
 
     fn new(options: WorkerOptions) -> Self {
-        let driver_client = DriverClient::new(ClientOptions {
+        let driver_client_set = DriverClientSet::new(ClientOptions {
             enable_tls: options.enable_tls,
             host: options.driver_host.clone(),
             port: options.driver_port,
```
lonless9, line 30: I feel we need to have documentation on enabling TLS for production environments.

crates/sail-execution/src/job_runner.rs
```diff
@@ -76,13 +76,16 @@ impl JobRunner for ClusterJobRunner {
...
         plan: Arc<dyn ExecutionPlan>,
     ) -> Result<SendableRecordBatchStream> {
         let (tx, rx) = oneshot::channel();
         self.driver
-            .send(DriverEvent::ExecuteJob { plan, result: tx })
+            .send(DriverEvent::ExecuteJob {
+                plan,
+                context: ctx.task_ctx(),
+                result: tx,
+            })
```
lonless9, line 88: 👍

lonless9 (approved):
This is a highly complex task and also our first milestone achievement of the year.
My focus mainly include resource recycling, unexpected configuration changes, fault tolerance, etc.. But these more boundary conditions should be easily discoverable in the future. 
Such as caching plans, recomputing/reusing or tracking the availability of intermediate results, and adding conditions that require throttling or isolation (for example slow-read consumers) can all obviously serve as points for future improvement.
I feel our architecture is already robust enough to support endeavors, and I can't wait to see this brand-new execution model work its magic!

crates/sail-execution/src/worker/actor/handler.rs
linhr: The tasks are managed by a `JoinSet` so they will be aborted when `ActorContext` is dropped.

crates/sail-execution/src/worker/peer_tracker/core.rs
linhr: The worker will be lost when the pod is terminated, so new pods will always have different worker IDs with different IPs and/or port numbers.

crates/sail-execution/src/job_graph/planner.rs
linhr: A task region by definition will have blocking inputs and outputs. But a task region can be large (the entire job, which is the current case) or small (one partition in one stage, for jobs with only blocking shuffle). And yeah `OutputMode::Blocking` will be implemented soon!

crates/sail-execution/src/worker_manager/kubernetes.rs
linhr: Yeah all the worker pods will still be there so that the logs remain accessible. The terminated pods do not consume physical resources but some metadata is kept in K8s. All the pods will be removed once the driver pod terminates (because we've configured owner reference when starting worker pods).

I think there is an optimization opportunity for cleaning up worker pods more promptly for long-running jobs.

crates/sail-execution/src/worker/actor/core.rs
linhr: Good point! And this will be important for enterprise users.

crates/sail-execution/src/plan/shuffle_read.rs
linhr: Yeah this is an area for optimization. For jobs with large number of partitions, we may even have to do multi-stage merge to reduce the number of streams for a single task. There are many research papers around shuffle algorithms, and with the redesign in this PR (especially `TaskDefinition`), we could support these advanced algorithms more easily.

PR #1169, opened 2025-12-30
refactor: arrow action schema for Iceberg

lonless9:


crates/sail-iceberg/src/physical_plan/action_schema.rs
```diff
@@ -0,0 +1,482 @@
...
+
+pub const COL_ACTION: &str = "action";
+
+static ACTION_FIELDS: LazyLock<std::result::Result<Vec<FieldRef>, String>> =
+    LazyLock::new(iceberg_action_fields_build);
+static ACTION_SCHEMA: LazyLock<std::result::Result<SchemaRef, String>> =
+    LazyLock::new(|| match (*ACTION_FIELDS).as_ref() {
+        Ok(fields) => Ok(Arc::new(Schema::new(fields.clone()))),
+        Err(msg) => Err(msg.clone()),
+    });
```
linhr: Similar to Delta Lake, I feel we should panic if static schema creation fails. This could be changed in a future PR though.

crates/sail-iceberg/src/physical_plan/action_schema.rs
```diff
@@ -0,0 +1,482 @@
...
+                Some(other) => Err(DataFusionError::Internal(format!(
+                    "unsupported non-primitive partition literal in DataFile: {other:?}"
+                ))),
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        Ok(AddFileAction {
+            content: format!("{:?}", df.content),
+            file_path: df.file_path,
+            file_format: format!("{:?}", df.file_format),
```
linhr: Is it intentional to use debug formatting (`{:?}`) for these fields?

crates/sail-iceberg/src/physical_plan/action_schema.rs
```diff
@@ -0,0 +1,482 @@
...
+        })
+    }
+}
+
+pub fn encode_actions(rows: Vec<ActionRow>) -> Result<RecordBatch> {
+    if rows.is_empty() {
+        return Ok(RecordBatch::new_empty(iceberg_action_schema()?));
+    }
+    serde_arrow::to_record_batch(iceberg_action_fields()?, &rows)
+        .map_err(|e| DataFusionError::External(Box::new(e)))
```
linhr, line 322: Yeah I guess you've realized already that `serde_arrow::to_record_batch` does not support empty batch. 😆 I have some utilities (being refactored in #1216) to handle empty inputs. We can add a TODO here to consolidate the logic throughout our codebase.

crates/sail-iceberg/src/physical_plan/action_schema.rs
lonless9: Done!

crates/sail-iceberg/src/physical_plan/action_schema.rs
lonless9: Forgot about this part, temporarily use a function to solve it now. Maybe I'll need to do a complete refactor of serde later.

crates/sail-iceberg/src/physical_plan/action_schema.rs
lonless9: Added!

PR #1171, opened 2025-12-31
chore: refine CI Workflow

lonless9:


.github/actions/setup-rust/action.yml
```diff
@@ -1,5 +1,12 @@
 name: Rust setup
 description: Prepare Rust build environment and build cache
+inputs:
+  cache-key:
+    description: Cache key prefix (e.g., 'setup-rust' or 'rust-lint')
+    default: "setup-rust"
```
linhr: I feel we can make it a required input so that the caching sharing behavior is controlled explicitly. We can have `rust-test`, `rust-lint`, and `rust-build`.

.github/actions/setup-rust/action.yml
```diff
@@ -60,7 +73,9 @@ runs:
       if: ${{ github.event_name == 'push' }}
       with:
         path: ${{ steps.config.outputs.path }}
-        key: setup-rust-${{ steps.config.outputs.label }}-${{ steps.config.outputs.hash }}
+        key: ${{ inputs.cache-key }}-${{ steps.config.outputs.label }}-${{ steps.config.outputs.hash }}
+        restore-keys: |
+          ${{ inputs.cache-key }}-${{ steps.config.outputs.label }}-
```
linhr: We should not have `restore-keys` for `push` events since we'd like the `main` branch build to start with a clean state so that stale dependencies can be purged.

.github/workflows/build.yml
```diff
@@ -46,8 +46,8 @@ jobs:
           workflow_name: Build
           artifact_name: python-package
 
-  build:
-    name: Build
+  rust-unit-tests:
+    name: Rust Unit Tests
```
linhr: For consistency I'm thinking about the names for these jobs:
* `rust-unit-tests` -> `rust-tests`
* `build-wheels` -> `python-build`
* `python-build` -> `python-tests`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,59 @@
+name: Lint
```
linhr: I was actually thinking about breaking down this further into ~`lint-rust.yml` and `lint-python.yml`~ `rust-lint.yml` and `python-lint.yml` so we can configure different trigger conditions, and the failure would not interfere with each other (otherwise we would need `if: always()`).

And for `push` events we should always run this without `paths` filters in case there are missing/incorrect globs in `paths` as the directly layout involves, so that we get a chance to detect and fix such bugs.

.github/workflows/build.yml
linhr: And maybe `python-build.yml` should be renamed to `python-tests.yml` now that it's dedicated to testing.

.github/actions/setup-rust/action.yml
lonless9: Done!

.github/actions/setup-rust/action.yml
lonless9: Done!

.github/workflows/build.yml

.github/workflows/lint.yml
lonless9: Done!

.github/workflows/gold-data-script-validation.yml
```diff
@@ -35,6 +35,9 @@ jobs:
           java-version: "17"
 
       - uses: ./.github/actions/setup-rust
+        with:
+          cache-key: rust-build
```
linhr: Haha glad that we make the cache key explicit. I believe this needs to be `rust-tests` due to the different profile being used.

.github/actions/setup-rust/action.yml
```diff
@@ -1,5 +1,12 @@
 name: Rust setup
 description: Prepare Rust build environment and build cache
+inputs:
+  cache-key:
+    description: Cache key prefix (e.g., 'setup-rust' or 'rust-lint')
```
linhr:
```
(e.g., 'rust-lint')
```

linhr (approved):

PR #1172, opened 2026-01-05
docs: add Windows LLVM linker requirement

pomykalakyle:
Added warning to the Windows rust compilation to make the user use llvm.


shehabgamin (approved):
LGTM! Will merge after build finishes.

Thanks for the contribution!!

PR #1179, opened 2026-01-06
feat: add Delta Lake support and include-load-time option to TPC-H benchmark

pomykalakyle:
Modified the benchmarks a bit to include file loading. Wanted to compare with loading - sail is approx 40x faster on my local machine.

### DIFF
--- a/docs/development/recipes/benchmark.md
+++ b/docs/development/recipes/benchmark.md
@@ -28,7 +28,7 @@ Please follow the README in that repository to generate the TPC-H data.
 
 - You can use `--console` instead of `--query [N]` to start a PySpark console and explore the TPC-H data interactively.
 - You can use `--url` to specify the Spark URL. Use `local` for Spark local mode (JVM) and use `sc://localhost:50051` for Spark Connect (the default).
-- You can use `--format` to specify the data format: `parquet` (default) or `delta`.
+- You can use `--format` to specify the data format: `parquet` (default), `delta`, or `iceberg`.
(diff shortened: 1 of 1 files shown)


docs/development/recipes/benchmark.md
```diff
@@ -28,8 +28,24 @@ Please follow the README in that repository to generate the TPC-H data.
 
 - You can use `--console` instead of `--query [N]` to start a PySpark console and explore the TPC-H data interactively.
 - You can use `--url` to specify the Spark URL. Use `local` for Spark local mode (JVM) and use `sc://localhost:50051` for Spark Connect (the default).
+- You can use `--format` to specify the data format: `parquet` (default) or `delta`.
```
shehabgamin: Should we add Iceberg too?

pomykalakyle: Yea it's probably worth doing that too since its getting traction.

pomykalakyle: > Should we add Iceberg too?

added

results.txt
```diff
@@ -0,0 +1,745 @@
+-- SQLBench-H query 1 derived from TPC-H query 1 under the terms of the TPC Fair Use Policy.
```
linhr: This file needs to be removed.

pomykalakyle: Removed!

docs/development/recipes/benchmark.md
shehabgamin:
```suggestion
- You can use `--format` to specify the data format: `parquet` (default), `delta`, or `iceberg`.
```

PR #1180, opened 2026-01-06
refactor: extract command handler and add documentation

pomykalakyle:
### Summary

Improves code organization in the Spark Connect server by extracting command handling into a dedicated function and adds documentation comments throughout the execution path.

#### Refactoring
- Extracted command type dispatch from `execute_plan` into a new `handle_command` utility function
- Standardized unimplemented command errors to use `SparkError::todo()` instead of raw `Status::unimplemented`

#### Documentation
Added doc comments to:
- `JobRunner::execute` trait method
- `ClusterJobRunner::execute` implementation
- `SparkConnectServer.start()`, `run_blocking()`, and `run()` methods
- `serve` gRPC entrypoint function
- `handle_execute_sql_command` handler
- `Relation` to `Plan` conversion
- `sail-spark-connect` build script

### Files Changed
- `crates/sail-common-datafusion/src/session/job.rs`
- `crates/sail-execution/src/runner.rs`
- `crates/sail-python/src/spark/server.rs`
- `crates/sail-spark-connect/build.rs`
- `crates/sail-spark-connect/src/entrypoint.rs`
- `crates/sail-spark-connect/src/proto/plan.rs`
- `crates/sail-spark-connect/src/server.rs`
- `crates/sail-spark-connect/src/service/plan_executor.rs`

### DIFF
--- a/crates/sail-spark-connect/src/server.rs
+++ b/crates/sail-spark-connect/src/server.rs
@@ -148,8 +148,7 @@ impl SparkConnectService for SparkConnectServer {
                 command_type: command,
             }) => {
                 let command = command.required("command")?;
-                handle_command(&ctx, command, metadata)
-                    .await?
+                handle_command(&ctx, command, metadata).await?
             }
             plan::OpType::CompressedOperation(_) => {
                 return Err(Status::unimplemented("compressed operation plan"));


linhr:
I wonder if we could reduce the amount of documentation especially for internal functions. I found many cases when the documentation and comments are repetition of the code itself. Since the internals are evolving, it would make it slightly inconvenient if we later have to keep the code and function arguments in sync with the comments. It also requires more review efforts to ensure the comments are accurate.

This is just a thought on how to potentially make the collaboration more productive. But I really appreciate your interest in Sail and the time you've devoted to the exploration!

pomykalakyle:
> I wonder if we could reduce the amount of documentation especially for internal functions. I found many cases when the documentation and comments are repetition of the code itself. Since the internals are evolving, it would make it slightly inconvenient if we later have to keep the code and function arguments in sync with the comments. It also requires more review efforts to ensure the comments are accurate.
> 
> This is just a thought on how to potentially make the collaboration more productive. But I really appreciate your interest in Sail and the time you've devoted to the exploration!

Gotcha, we don’t need these if they aren’t helpful. 
I assumed internal documentation was something desired since it’s that way for most codebases. I also like having documentation so that if I forget what a function does if I hover over it in my IDE I can easily see what it’s meant for.

linhr:
> I assumed internal documentation was something desired since it’s that way for most codebases.

Agreed! Yeah I'm not against documentation for internal functions in general. I feel the comments would be the most valuable if it complements the code to keep track of high-level ideas, hidden assumptions, background knowledge etc. I feel it's fine (and maybe desired) to leave out information that can be inferred immediately from function signature (since the names and types themselves convey the idea already).

crates/sail-spark-connect/src/server.rs
```diff
@@ -47,18 +48,100 @@ fn is_reattachable(
...
+    ///
+    /// * `request` - The request containing the plan to execute.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Status::unimplemented` if the plan contains a `CompressedOperation`.
+    ///
+    /// # Returns
+    ///
+    /// A stream of responses containing the results of the plan execution.
```
linhr: We typically don't have docs for trait methods since this method already has documentation inherited from trait definition (visible in IDE).

crates/sail-spark-connect/src/server.rs
```diff
@@ -87,81 +170,9 @@ impl SparkConnectService for SparkConnectServer {
...
-                    CommandType::PipelineCommand(_) => {
-                        return Err(Status::unimplemented("pipeline command"));
-                    }
-                    CommandType::Extension(_) => {
-                        return Err(Status::unimplemented("command extension"));
-                    }
-                }
+                handle_command(&ctx, command, metadata)
+                    .await
+                    .map_err(Status::from)?
```
linhr: `.map_err(Status::from)` is not needed due to `?`.

crates/sail-spark-connect/src/server.rs
pomykalakyle: good point - removing

crates/sail-spark-connect/src/server.rs
pomykalakyle: agreed - fixing

pomykalakyle:
> > I assumed internal documentation was something desired since it’s that way for most codebases.
> 
> Agreed! Yeah I'm not against documentation for internal functions in general. I feel the comments would be the most valuable if it complements the code to keep track of high-level ideas, hidden assumptions, background knowledge etc. I feel it's fine (and maybe desired) to leave out information that can be inferred immediately from function signature (since the names and types themselves convey the idea already).

Thanks for the feedback! Yea let's go that way, I'll remove the function signature related stuff that's hard to maintain and focus on the high level concepts.

PR #1183, opened 2026-01-07
feat: set up bdd testing & implement partition transforms for Iceberg

lonless9:


lonless9:
@copilot+claude-sonnet-4.6 Our testing foundation has undergone significant changes, so the current iceberg tests need to be refactored to adapt to the existing structure. Now merge the main branch, appropriately refactor the relevant iceberg tests/foundation, and remove the existing iceberg snapshots to regenerate them.

python/pysail/tests/spark/iceberg/features/partition_transforms.feature
```diff
@@ -0,0 +1,342 @@
...
+
+    Scenario: Bucket transform creates consistent partitions
+      Given statement template
+        """
+        CREATE TABLE bucket_test (
+          id INT,
+          user_id BIGINT,
+          action STRING
+        )
+        USING iceberg
```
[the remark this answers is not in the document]
lonless9: Done!

python/pysail/tests/spark/iceberg/features/partition_transforms.feature
```diff
@@ -0,0 +1,342 @@
...
+
+    Scenario: Truncate on string column groups similar values
+      Given statement template
+        """
+        CREATE TABLE truncate_test (
+          id INT,
+          code STRING,
+          value DOUBLE
+        )
+        USING iceberg
```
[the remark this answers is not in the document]
lonless9: Done!

crates/sail-sql-analyzer/src/statement.rs
```diff
@@ -1153,9 +1153,10 @@ fn from_ast_table_definition(
         .into_iter()
         .flatten()
         .map(|x| match x {
-            PartitionColumn::Typed(c) => c.name.value.into(),
-            PartitionColumn::Name(x) => x.value.into(),
+            PartitionByItem::ColumnDefinition(column) => column.name.value,
+            PartitionByItem::Expression(expr) => expr.text().trim().to_string(),
```
linhr, line 1157: I wonder if we should update the spec so that partition_by accepts expressions instead of Vec<Identifier>. Using expr.text() for the spec and parsing it again later seems less clean, and we cannot distinguish the following cases in the spec:

```
PARTITIONED BY `years(ts)` -- an identifier for the column
PARTITIONED BY years(ts)
```

crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -32,6 +32,10 @@ fn integer_part(expr: Expr, part: &str) -> Expr {
     )
 }
 
+fn years_transform(arg: Expr) -> Expr {
```
linhr: For naming consistency, just call this `years`.

And it seems the existing `year` function does not have `cast`. Is the difference on purpose?

crates/sail-iceberg/src/utils/partition_transform.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+    raw.split('.')
+        .map(strip_quotes)
+        .map(str::trim)
+        .filter(|p| !p.is_empty())
+        .collect::<Vec<_>>()
+        .join(".")
+}
+
+pub fn parse_partition_field_expr(raw: &str) -> Result<PartitionFieldSpec, String> {
```
linhr, line 88: We won't need this parsing logic if we store the partition expression in the spec and resolve the expression directly into `PartitionFieldSpec`.

crates/sail-iceberg/src/utils/mod.rs
```diff
@@ -49,6 +50,33 @@ pub fn join_table_uri(table_uri: &str, rel: &str, mode: &WritePathMode) -> Strin
...
+    // object_store::path::Path requires slash-delimited paths.
+    let p = p.replace('\\', "/");
+    let path_no_leading = p.strip_prefix('/').unwrap_or(&p);
+    if is_file {
+        object_store::path::Path::parse(path_no_leading)
+            .map_err(|e| DataFusionError::External(Box::new(e)))
+    } else {
+        object_store::path::Path::parse(path_no_leading)
+            .map_err(|e| DataFusionError::External(Box::new(e)))
+    }
```
linhr: The logic seems identical for `if` and `else`.

python/pysail/tests/spark/iceberg/features/partition_transforms.feature
```diff
@@ -0,0 +1,308 @@
...
+          (2, DATE '2023-12-20', 'purchase'),
+          (3, DATE '2024-01-10', 'login'),
+          (4, DATE '2024-06-25', 'view')
+        """
+      Then iceberg metadata matches snapshot
+      Then iceberg snapshot operation is append
+      Then file tree in location matches
+        """
+        📂 data
+          📂 event_date_year=2023
```
linhr, line 118: For my own knowledge, is the partition transform directory name (`event_date_year=xxx` for `years(event_date)`) part of the Iceberg specification? Or is it specific to our implementation?

crates/sail-plan/src/function/scalar/datetime.rs
lonless9: Done! Here is indeed an unnecessary implementation.

crates/sail-iceberg/src/utils/mod.rs
lonless9: Done!

python/pysail/tests/spark/iceberg/features/partition_transforms.feature
lonless9: The absolute path in the metadata can determine the file location. This seems to be a style issue, and there is also the object store style. I will investigate later which one we should support. https://github.com/apache/iceberg/blob/68e555b94f4706a2af41dcb561c84007230c0bc1/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L213-L228

crates/sail-sql-analyzer/src/statement.rs
lonless9: Changing the spec may involve many files and include catalog changes, so let's handle it in the follow-up PR.

crates/sail-iceberg/src/utils/partition_transform.rs
lonless9: https://github.com/lakehq/sail/pull/1183#discussion_r2925057596

PR #1184, opened 2026-01-09
docs: add documentation to execution module and rename TaskSchedule to ScheduledTask

pomykalakyle:
## Summary

Adds documentation comments to key components in the execution module and renames `TaskSchedule` to `ScheduledTask` for clarity. A TaskSchedule sounds like a list of Tasks in a Vec or something else, not something more atomic. 

### Documentation
- `DriverEvent` enum and `ExecuteJob` variant
- `JobRunner` trait
- `ClusterJobRunner` struct
- `ActorHandle` struct
- `handle_execute_job`, `scale_up_workers`, `schedule_tasks` methods in driver actor
- `get_table` method in Unity Catalog provider

### Refactor
- Renamed `TaskSchedule` → `ScheduledTask` for better readability

### DIFF
--- a/crates/sail-execution/src/driver/event.rs
+++ b/crates/sail-execution/src/driver/event.rs
@@ -14,6 +14,7 @@ use crate::driver::gen;
 use crate::error::ExecutionResult;
 use crate::id::{JobId, TaskInstance, WorkerId};
 
+/// Events that the driver can recieve and react to.
 pub enum DriverEvent {
     ServerReady {
         /// The local port that the driver server listens on.
@@ -45,6 +46,7 @@ pub enum DriverEvent {
         worker_id: WorkerId,
         instant: Instant,
     },
+    /// Exactly what you think - executes a plan. Contains a sender so we can send back the result too.
     ExecuteJob {
         plan: Arc<dyn ExecutionPlan>,
         result: oneshot::Sender<ExecutionResult<SendableRecordBatchStream>>,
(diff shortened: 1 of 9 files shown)


crates/sail-execution/src/driver/event.rs
```diff
@@ -45,6 +46,7 @@ pub enum DriverEvent {
         worker_id: WorkerId,
         instant: Instant,
     },
+    /// Exactly what you think - executes a plan. Contains a sender so we can send back the result too.
```
linhr, line 49: I wonder if terms like "Exactly what you think - " can be skipped in the comments. 😆 We'd like the documentation to read a bit formal and concise.

linhr (commented):
Thanks for the PR. Could you remove the changes to the following files? We're working on heavy rewrites on distributed job scheduling in #1164 so the changes will no longer be relevant soon. This would help avoid merge conflicts for the ongoing work.

* `crates/sail-execution/src/driver/actor/handler.rs`
* `crates/sail-execution/src/driver/job_scheduler/core.rs`
* `crates/sail-execution/src/driver/job_scheduler/mod.rs`
* `crates/sail-execution/src/driver/worker_pool/core.rs`

crates/sail-execution/src/driver/event.rs
pomykalakyle: I think I'll just close the pr if there is such a big refactor coming in.

Issue #749, opened 2025-08-12
SparkRuntimeException: Csv error: incorrect number of fields for line 1, expected 53 got 10


djouallah:
i read a mal formated csv, where the column numbers is variable, the solution is to provide a schema and just ignore etra columns, but it is not working with sail, and I am getting this 

`SparkRuntimeException: Csv error: incorrect number of fields for line 1, expected 53 got 10`

technically, it is an upstream issue, but it is kind of a blocker for me,as we can't control the csv source https://github.com/apache/datafusion/issues/12772

linhr:
Thanks for reporting this!

We're working on a text format right now, which could provide a workaround since you can then use SQL functions or Python UDFs to process the lines. But I can see that it would be ideal if this can be supported inside the CSV source.

djouallah:
I trolled polars, duckdb and daft till they fix it, I think i will expect the same here :)  and after that I will ask for returning the filename :)

djouallah:
@linhr can we add this, it is already implemented in datafusion  

truncated_rows=True

https://github.com/apache/datafusion/pull/17553

djouallah:
@linhr in case you want to validate, it is a very dirty csv 

https://github.com/djouallah/Fabric_Notebooks_Demo/blob/main/ETL/Light_ETL_Python_Notebook.ipynb

djouallah:
holly Cow !!! it seems to be working !!!! give me filename 🥇

djouallah:
<img width="1043" height="480" alt="Image" src="https://github.com/user-attachments/assets/2d0d78da-fbeb-44af-9630-1a2598cff844" />

shehabgamin:
[...]

PR #1185, opened 2026-01-09
feat: allow truncated rows in CSV reader

linhr:
Part of #749.


djouallah:
any option to test this, maybe a night release or something, it is very important feature to me :)

PR #1187, opened 2026-01-10
fix: ignore nulls in collect_set collet_list

davidlghellin:
example bug in sail
```python
spark.sql("""
SELECT COLLECT_LIST(DISTINCT x) as result
FROM (VALUES (1), (2), (NULL), (2), (NULL), (1)) AS t(x)
""").show(truncate=False)
+------------+
|result      |
+------------+
|[NULL, 2, 1]|
+------------+
```
image

--------

example in spark
```python
spark.sql("""
SELECT COLLECT_LIST(DISTINCT x) as result
FROM (VALUES (1), (2), (NULL), (2), (NULL), (1)) AS t(x)
""").show(truncate=False)
+------+                                                                        
|result|
+------+
|[1, 2]|
+------+
```
image

### DIFF
--- a/crates/sail-plan/src/function/aggregate.rs
+++ b/crates/sail-plan/src/function/aggregate.rs
@@ -262,22 +262,26 @@ fn collect_set(input: AggFunctionInput) -> PlanResult<expr::Expr> {
 
     // WORKAROUND: DataFusion's array_agg doesn't properly handle null_treatment when distinct=true
     // So we need to add an explicit filter for NULLs
-    let (filter, null_treatment) = if ignore_nulls == Some(true) {
-        let arg = input.arguments.clone().one()?;
+    let (args, filter, null_treatment) = if ignore_nulls == Some(true) {
+        let arg = input.arguments.one()?;
         let null_filter = arg.clone().is_not_null();
         let combined_filter = match input.filter {
             Some(existing) => Some(Box::new(existing.as_ref().clone().and(null_filter))),
             None => Some(Box::new(null_filter)),
         };
-        (combined_filter, None) // Don't use null_treatment when we have explicit filter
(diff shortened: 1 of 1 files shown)


crates/sail-plan/src/function/aggregate.rs
```diff
@@ -257,27 +257,69 @@ fn count_if(input: AggFunctionInput) -> PlanResult<expr::Expr> {
...
+    let ignore_nulls = input.ignore_nulls.or(Some(true));
+
+    // WORKAROUND: DataFusion's array_agg doesn't properly handle null_treatment when distinct=true
+    // So we need to add an explicit filter for NULLs
+    let (filter, null_treatment) = if ignore_nulls == Some(true) {
+        let arg = input
+            .arguments
+            .first()
+            .cloned()
+            .unwrap_or_else(|| lit(ScalarValue::Null));
```
linhr: I wonder if we should have `let arg = input.arguments.one()?;` if we only allow one argument. This ensures the error is reported if the input is invalid.

crates/sail-plan/src/function/aggregate.rs
```diff
@@ -257,27 +257,69 @@ fn count_if(input: AggFunctionInput) -> PlanResult<expr::Expr> {
...
+    let ignore_nulls = input.ignore_nulls.or(Some(true));
+
+    // WORKAROUND: DataFusion's array_agg doesn't properly handle null_treatment when distinct=true
+    // So we need to add an explicit filter for NULLs when both distinct and ignore_nulls are true
+    let (filter, null_treatment) = if input.distinct && ignore_nulls == Some(true) {
+        let arg = input
+            .arguments
+            .first()
+            .cloned()
+            .unwrap_or_else(|| lit(ScalarValue::Null));
```
linhr: Same here. Should we ensure only one argument?

PR #1188, opened 2026-01-11
feat: implement percentile_cont aggregate function

pomykalakyle:
- Add percentile_cont_expr to handle WITHIN GROUP syntax
- Extract column from ORDER BY and combine with percentile argument
- Rename AggFunctionInput to ResolvedAggFuncComps with improved docs
- Add docstrings to AggFunctionBuilder methods
- Add test for percentile_cont
- Update aggregate.json status to supported
- Add .cargo/ to gitignore for local config

Part of #309

### DIFF
--- a/crates/sail-plan/src/function/aggregate.rs
+++ b/crates/sail-plan/src/function/aggregate.rs
@@ -155,8 +155,8 @@ fn mode(input: AggFunctionInput) -> PlanResult<expr::Expr> {
 /// in order_by and the percentile in arguments. This function combines them.
 fn percentile_cont_expr(input: AggFunctionInput) -> PlanResult<expr::Expr> {
     // Extract the single column expression from ORDER BY (error if multiple)
-    let sort = input.order_by.one()?;
-    let column = sort.expr.clone();
+    let sort = input.order_by.clone().one()?;
+    let column = sort.expr;
 
     // Get the percentile value from arguments
     let percentile = input.arguments.one()?;


linhr:
Would you mind reverting the rename to `AggFunctionInput`? I feel the original name is fine so I prefer to keep it to reduce merge conflicts with other PRs that the community is working on.

BTW I updated the PR description to avoid mentioning "Closes #XXX" otherwise the issue will be auto-closed once this PR is merged.

crates/sail-plan/src/function/aggregate.rs
```diff
@@ -148,7 +148,39 @@ fn mode(input: AggFunctionInput) -> PlanResult<expr::Expr> {
...
+/// Builds a percentile_cont aggregate expression from WITHIN GROUP syntax.
+///
+/// DataFusion's percentile_cont expects args = [column, percentile], but Spark's
+/// SQL syntax `percentile_cont(0.5) WITHIN GROUP (ORDER BY col)` puts the column
+/// in order_by and the percentile in arguments. This function combines them.
+fn percentile_cont_expr(input: ResolvedAggFuncComps) -> PlanResult<expr::Expr> {
+    // Extract the column expression from ORDER BY
+    let column = input
+        .order_by
+        .first()
```
linhr: If DataFusion does not support multiple `ORDER BY` columns, we should return an error if `input.order_by` contains more than one items.

crates/sail-plan/src/function/aggregate.rs
```diff
@@ -360,6 +392,9 @@ fn approx_count_distinct(input: AggFunctionInput) -> PlanResult<expr::Expr> {
     ))
 }
 
+/// Creates a list of built-in aggregate functions.
+/// This is used to create a hashmap that the resolver uses to look up
+/// aggregate functions by name.
```
linhr, line 393: nitpick: Notes about the usage details will get outdated if we later use this function in a different way. There is no mechanism yet to remember updating the documentation here, especially if the future changes are made by another developer.

linhr (commented):
I left a few minor comments but I feel it looks good in general!

crates/sail-plan/src/function/aggregate.rs
pomykalakyle: Gotcha - any chance you have some guidelines on how you want comments to look like in the future? Just to make sure I match the style you want.

crates/sail-plan/src/function/aggregate.rs
pomykalakyle: Thanks for catching! Should be in the update.

pomykalakyle:
> Would you mind reverting the rename to `AggFunctionInput`? I feel the original name is fine so I prefer to keep it to reduce merge conflicts with other PRs that the community is working on.
> 
> BTW I updated the PR description to avoid mentioning "Closes #XXX" otherwise the issue will be auto-closed once this PR is merged.

Done.

crates/sail-plan/src/function/aggregate.rs
linhr: For this particular case I'm fine keeping it (and hence the "nitpick" in my comment). I brought this up mostly as a general discussion point. We're in an early stage of improving the documentation so we don't have comprehensive guidelines yet. But as we accumulate cases like this, we may collectively build a checklist that everyone can follow. And maybe the style we develop over time can be used to direct AI-assisted code review to ensure consistency of the entire codebase.

BTW the docs you add is good and valuable. I was just thinking from a different angle regarding how the practice can be scalable and future-proof. I feel this is important as the project and the community grow. If you have ideas (especially how docs quality assurance can be automated via AI tools), I'd love to discuss in GitHub/Slack!

linhr (approved):

linhr:
Thanks for the contribution!

crates/sail-plan/src/function/aggregate.rs
pomykalakyle: Do you think it's worth adding a list of tentative preferences for comments in the docs somewhere? 

I'm thinking that we don't need any formal rules for the moment but it might be nice to have it so that we can list out things like for now we don't want to add param documentation for functions because of the reasons you mentioned before. Just so we can accumulate thoughts on it until we want to get a bit more formal.

linhr: I love the idea! Let me draft something in the dev guide and we can enrich it together over time. I guess this can also readily become context for AI tools if we point the agents to the Markdown files as rules to follow during review etc.

PR #1190, opened 2026-01-11
fix: return inf/nan for division by zero in all ANSI modes

davidlghellin:
Testing ibis tests

### DIFF
--- a/crates/sail-plan/src/function/scalar/math.rs
+++ b/crates/sail-plan/src/function/scalar/math.rs
@@ -1,12 +1,10 @@
 use std::sync::Arc;
 
 use datafusion::arrow::datatypes::{DataType, IntervalUnit, TimeUnit};
-use datafusion::arrow::error::ArrowError;
 use datafusion::functions::expr_fn;
 use datafusion_common::ScalarValue;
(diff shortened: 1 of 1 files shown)


davidlghellin:
Closing this PR - ANSI mode conflict

After investigation, the issue is that fixing the Ibis test_divide_by_zero tests conflicts with existing PySpark tests:

The Problem:

Ibis tests expect: Division by zero returns inf/nan (pandas/numpy behavior)
PySpark tests expect: Division by zero throws ArithmeticException (Spark ANSI mode behavior)
What this PR attempted:

Remove compile-time division by zero checks to let DataFusion handle it at runtime (returns inf/nan)
Respect ANSI mode: throw error when ON, return inf/nan when OFF
Change default ANSI mode from true to false
Why it doesn't work:
Changing the default ANSI mode to false breaks PySpark tests that expect errors by default. Keeping it as true means Ibis tests fail.

Possible solutions to explore:

Keep ANSI mode default as true (Spark compatible) and configure Ibis backend to use ANSI=false
Add a separate configuration for division-by-zero behavior independent of ANSI mode
Accept that Ibis test_divide_by_zero tests remain marked as notyet for the pyspark backend
Needs further discussion on the right approach.

linhr:
Thanks for the investigation! I've created an issue to track the follow-up work: #1196.

PR #1193, opened 2026-01-11
feat: support dynamic expressions as map keys in element access

davidlghellin:
uses https://github.com/lakehq/sail/issues/501 to detect errors

This PR fixes map element access (`map[key]`) to support dynamic expressions as keys, not just literals. Previously, Sail required map keys to be compile-time literals, which was more restrictive than Spark's behavior and caused 64 Ibis tests to fail.

## Problem

Before this change:
- `map[literal_key]` ✅ worked
- `map[expression_key]` ❌ failed with "extraction must be a literal"

### DIFF
--- a/crates/sail-plan/src/resolver/expression/misc.rs
+++ b/crates/sail-plan/src/resolver/expression/misc.rs
@@ -112,22 +112,15 @@ impl PlanResolver<'_> {
 
         // For Maps, we support non-literal expressions as keys
         if matches!(data_type, DataType::Map(_, _)) {
-            // Check if extraction is a literal to generate appropriate column name
-            let result_name = if let spec::Expr::Literal(ref lit) = extraction {
-                let literal_value = self.resolve_literal(lit.clone(), state)?;
-                let service = self.ctx.extension::<PlanService>()?;
-                let literal_name = service
-                    .plan_formatter()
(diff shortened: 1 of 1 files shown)


crates/sail-plan/src/resolver/expression/misc.rs
```diff
@@ -106,6 +106,34 @@ impl PlanResolver<'_> {
...
+                let literal_name = service
+                    .plan_formatter()
+                    .literal_to_string(&literal_value, &self.config.session_timezone)?;
+                format!("{}[{}]", name.one()?, literal_name)
+            } else {
+                format!("{}[<expr>]", name.one()?)
+            };
+
+            let extraction_expr = self
+                .resolve_named_expression(extraction, schema, state)
```
linhr: Could we use the resolved name instead of `<expr>` in the result name?

linhr (commented):
Nice! This is a valuable bug fix with high impact. I have one minor comment but otherwise it looks good!

crates/sail-plan/src/resolver/expression/misc.rs
linhr: Actually I just realized that we don't need to handle literals separately. `resolve_named_expression` internally calls `literal_to_string` for literal expressions. So the logic here can actually be largely simplified.

PR #1194, opened 2026-01-11
fix: handle NULL values in map_concat function

davidlghellin:
uses https://github.com/lakehq/sail/issues/501 to detect errors

Fix `map_concat` function to properly handle NULL values by returning NULL when any input map is NULL, matching Spark's behavior.


linhr (approved):
Thanks for this nice fix!

PR #1204, opened 2026-01-12
fix: complete date sequence support in spark_sequence function

davidlghellin:
- **Allow 2-argument date sequences**: `sequence(date, date)` now works without explicit step, defaulting to `INTERVAL '1' DAY` (ascending) or `INTERVAL '-1' DAY` (descending)
image

### DIFF
--- a/python/pysail/tests/spark/function/test_features.py
+++ b/python/pysail/tests/spark/function/test_features.py
@@ -1,12 +1,5 @@
 """BDD tests for DataFrame operations using pytest-bdd."""
 
-from __future__ import annotations
-
-import pytest
 from pytest_bdd import scenarios
-
-from pysail.tests.spark.utils import is_jvm_spark
-
-pytestmark = pytest.mark.skipif(is_jvm_spark(), reason="Sail only")
 
 scenarios("features")


crates/sail-function/src/scalar/array/spark_sequence.rs
```diff
@@ -228,45 +228,101 @@ fn gen_sequence_timestamp(args: &[ArrayRef]) -> Result<ArrayRef> {
...
+                // Note: interval can be represented as days OR as nanoseconds (86400000000000 ns = 1 day)
+                let nanos_per_day = 86_400_000_000_000i64;
+                if months == 0 && days == 0 && nanoseconds.abs() < nanos_per_day {
                     return exec_err!(
                         "Spark `sequence` function cannot generate date range less than 1 day."
                     );
                 }
 
-                let negative = months < 0 || days < 0;
+                let negative = months < 0 || days < 0 || nanoseconds < 0;
```
linhr, line 296: I'm not sure if the way we detect negative steps is general enough. Could `1 month -100 days` be a valid interval?

crates/sail-function/src/scalar/array/spark_sequence.rs
```diff
@@ -228,45 +228,101 @@ fn gen_sequence_timestamp(args: &[ArrayRef]) -> Result<ArrayRef> {
...
+                // Validate step direction matches sequence direction
+                // For error message, show the non-zero component
+                let step_value = if months != 0 {
+                    months
+                } else if days != 0 {
+                    days
+                } else {
+                    // Convert nanoseconds to days for display
+                    (nanoseconds / nanos_per_day) as i32
+                };
```
linhr, line 307: I feel this step value for display may be confusing if both the months and the days are non-zero.

PR #1216, opened 2026-01-15
feat: system catalog

linhr:
This PR introduces a built-in catalog named `system` so that the user can query internal states (sessions, workers, jobs, stages, and tasks) for both running and deleted sessions.

### Catalog

1. Add a YAML file to define system databases and tables. Add a build script to generate corresponding data structures, such as (1) table row struct compatible with `serde_arrow`, and (2) database, table, and column metadata for the `CatalogProvider` implementation.
1. Add a `CatalogProvider` implementation to access databases and tables for the system catalog.
1. Add logical and physical nodes for reading system tables. **The table supports projection pruning, filter pushdown, and limit pushdown.**

### State Observation

1. Add actor events to observe internal states. This is invoked by `SystemTableExec` for "reading" from the system table.
  1. Update `SessionManagerActor` to observe states of all sessions.
  1. Update `DriverActor` to observe states of a single session.
1. Support taking "snapshots" when the driver stops so that the states of deleted sessions can still be observed via `JobRunnerHistory`.
1. Introduce `Predicate` and `ArrowPredicateEvaluator` to support filtering internal data structures via physical expressions. Introduce the `PredicateExt` trait to assist collecting information when iterating internal data structures.
1. Introduce the `StateObservable<O: Observer>` trait to observe states via Tokio oneshot channels.

### Session Management

1. Update `SessionManagerActor` to keep deleted sessions. We now return an error on deleted session. (This is a behavior change. Previously, when the user sends request with a deleted session ID, the server would silently recreate a session with the same ID. This would result in obscure errors such as "temporary view not found" since the recreated session does not have the resources created before.)
1. Use only session ID for session lookup. The user ID is no longer considered part of the session lookup key.

## Miscellaneous Changes

1. Move `catalog` and `database` from `TableKind` enum variants to the `TableStatus` struct, so that temporary views can have catalog and database information. (All system tables are actually temporary views.)
1. Move `planner.rs` physical planner from the `sail-plan` crate to the `sail-session` crate.
1. Add `ArrowSerializer` as a utility for `serde_arrow`.
1. Remove the `SessionKey` trait and various `K: SessionKey` generic parameters from the session manager. (We now explicitly handle session ID and user ID since we need to expose such information in system tables.)

## Testing

The changes are verified by unit tests, BDD tests, and manual testing.


crates/sail-session/src/session_manager/actor/handler.rs
```diff
@@ -61,18 +88,19 @@ impl<K: SessionKey> SessionManagerActor<K> {
...
+        let session = self.sessions.get_mut(&session_id);
+        if let Some(session) = session {
+            if let ServerSessionState::Active { context } = &mut session.state {
+                if let Ok(tracker) = context.extension::<ActivityTracker>() {
+                    if tracker.active_at().is_ok_and(|x| x <= instant) {
+                        info!("removing idle session {session_id}");
+                        Self::delete_session(ctx, session_id, context);
+                        session.deleted_at = Some(Utc::now());
+                        session.state = ServerSessionState::Deleting;
                     }
```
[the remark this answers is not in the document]
linhr: It's intended to keep deleted sessions so that they can be queried in the system table.

crates/sail-common-datafusion/src/system/predicate.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.fetch == 0 {
+            return None;
+        }
+        for item in self.input.by_ref() {
+            match (self.predicate)((self.key)(&item)) {
+                Ok(true) => {
+                    self.fetch -= 1;
+                    return Some(Ok((self.value)(item)));
```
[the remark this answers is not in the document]
linhr: The analysis is incorrect. The `for` loop is only entered when `self.fetch > 0` and `self.fetch` is decremented at most once inside the loop due to the early `return`.

crates/sail-catalog-system/src/predicate.rs
```diff
@@ -0,0 +1,238 @@
...
+            predicate,
+            schema,
+            phantom: PhantomData,
+        })
+    }
+
+    pub fn evaluate(&self, value: &T) -> Result<bool> {
+        let array = value.to_scalar()?.to_array()?;
+        let batch = RecordBatch::try_new(self.schema.clone(), vec![array])?;
+        let result = self.predicate.evaluate(&batch)?.into_array(1)?;
```
lonless9, line 138: Is here constructing a single-row RecordBatch per value?

crates/sail-execution/src/driver/actor/core.rs
```diff
@@ -126,14 +127,18 @@ impl Actor for DriverActor {
                 schema,
                 result,
             } => self.handle_fetch_remote_stream(ctx, uri, key, schema, result),
-            DriverEvent::Shutdown => ActorAction::Stop,
+            DriverEvent::ObserveState { observer } => self.handle_observe_state(ctx, observer),
+            DriverEvent::Shutdown { history } => self.handle_shutdown(ctx, history),
```
lonless9, line 131: It seems our current driver scheduling is FCFS. I feel these tasks have different priorities, costs and latency requirements, so we may need some hot/cold orchestration or offloading strategies. For example we can introduce a `historyActor` to share the overhead, open dedicated lanes and so on. By the way, I don't know if we have a logic like ttl or limit to restrict the amount of historical data in driver memory.

lonless9: Additionally, I think it is needed to check whether workers should automatically release resources when the driver disconnects, and whether some hacky UDFs might monopolize resources, preventing workers from reporting heartbeats.

python/pysail/tests/spark/catalog/__snapshots__/test_features/test_filter_and_limit_pushdown_for_system_tables.1.plan
```diff
@@ -0,0 +1,6 @@
+== Physical Plan ==
+ProjectionExec: expr=[#5@0 as session_id, #6@1 as user_id, #7@2 as status, #8@3 as created_at, #9@4 as deleted_at]
+  ProjectionExec: expr=[session_id@0 as #5, user_id@1 as #6, status@2 as #7, created_at@3 as #8, deleted_at@4 as #9]
+    GlobalLimitExec: skip=0, fetch=1
+      CooperativeExec
+        SystemTableExec: table=sessions, projection=Some([0, 1, 2, 3, 4]), filters=[BinaryExpr { left: Column { name: "session_id", index: 0 }, op: NotEq, right: Literal { value: Utf8(""), field: Field { name: "lit", data_type: Utf8 } }, fail_on_overflow: false }, BinaryExpr { left: Column { name: "session_id", index: 0 }, op: NotEq, right: Literal { value: Utf8("x"), field: Field { name: "lit", data_type: Utf8 } }, fail_on_overflow: false }], fetch=Some(1)
```
lonless9, line 6: 👍

crates/sail-catalog-system/src/predicate.rs
linhr: Yes it is.

I've considered batching but it would add complexity for the implementation, so I decided to defer it to the future when the performance becomes a concern here.

crates/sail-execution/src/driver/actor/core.rs
linhr: > It seems our current driver scheduling is FCFS. I feel these tasks have different priorities, costs and latency requirements, so we may need some hot/cold orchestration or offloading strategies. For example we can introduce a `historyActor` to share the overhead, open dedicated lanes and so on. By the way, I don't know if we have a logic like ttl or limit to restrict the amount of historical data in driver memory.

Yeah right now each actor has a single event queue and the handler is supposed to be lightweight. Observing internal state as part of event handling could indeed be heavy if there are a large number of jobs. If we want to collect states by spawning a task, we will have to use lock to protect the state and we're essentially moving away from the actor model that assumes single-threaded access to internal states. In that case, we probably want to support TTL to clean up completed jobs as you mentioned. This would fit the actor model nicely. Having prioritized event queues for different kinds of events are feasible as well, though it's a bit complex to design so we need to ensure the benefit is measurable.

> Additionally, I think it is needed to check whether workers should automatically release resources when the driver disconnects

Yeah I think it would nice to stop the worker when certain number of heartbeat fails. Let me figure it out in a follow-up PR.

> whether some hacky UDFs might monopolize resources, preventing workers from reporting heartbeats

If the user code such as UDFs consumes all CPU resources which prevents the control plane from running, this is probably a very bad situation and the driver would consider the worker as "lost" after missing heartbeats. This detection is there in the driver already.

Issue #219, opened 2024-09-29
Spark SQL Functions Coverage and Parity - JSON


shehabgamin:
- [ ] `from_json`
- [x] `get_json_object` (https://github.com/lakehq/sail/pull/213)
- [x] `json_array_length` (https://github.com/lakehq/sail/pull/213)
- [x] `json_object_keys`
- [ ] `json_tuple`
- [x] `schema_of_json`
- [x] `to_json`

kentkr:
@shehabgamin I'll take from_json!

shehabgamin:
> @shehabgamin I'll take from_json!

Wonderful!

kentkr:
@shehabgamin sadly I wasn't able to finish `from_json` within the month I gave myself. Hopefully I have more free time, but it won't be before the end of the year

davidlghellin:
> [@shehabgamin](https://github.com/shehabgamin) sadly I wasn't able to finish `from_json` within the month I gave myself. Hopefully I have more free time, but it won't be before the end of the year

Hi @kentkr , maybe this https://github.com/apache/datafusion-comet/blob/main/native/spark-expr/src/json_funcs/from_json.rs help

anarefolio:
Hi, I'd like to work on implementing json_tuple. It appears to be marked as F::unknown in json.rs and planned in the compatibility data. Is it still open for contribution?

linhr:
> Hi, I'd like to work on implementing json_tuple. It appears to be marked as F::unknown in json.rs and planned in the compatibility data. Is it still open for contribution?

Thanks for your interest! `json_tuple` is implemented in DataFusion so all we need is a few lines of code change to bring it to Sail. But I feel this is a perfect task for first contribution.

PR #1218, opened 2026-01-15
feat: add `to_json` function

davidlghellin:
part #219 
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_json.html

To run the tests (with my config):

## test Spark
```sh
sdk use java 17.0.9-tem && SPARK_REMOTE="local" hatch run pytest python/pysail/tests/spark/function/test_features.py -v
```

## test Sail
```sh
env RUST_LOG='sail=debug' cargo run -p sail-cli -- spark server
```

```sh
export SPARK_REMOTE="sc://localhost:50051" && hatch run pytest python/pysail/tests/spark/function/test_features.py -v
```

### DIFF
--- a/crates/sail-function/src/scalar/json/to_json.rs
+++ b/crates/sail-function/src/scalar/json/to_json.rs
@@ -161,11 +161,12 @@ fn to_json_inner(args: &[ArrayRef]) -> Result<ArrayRef> {
     }
 
     let options = if let Some(opts_array) = args.get(1) {
-        if let Some(map_array) = opt_downcast_arg!(opts_array, MapArray) {
-            ToJsonOptions::from_map(map_array)
-        } else {
-            ToJsonOptions::default()
-        }
+        let map_array = opt_downcast_arg!(opts_array, MapArray).ok_or_else(|| {
+            datafusion_common::DataFusionError::Plan(
+                "[INVALID_OPTIONS.NON_MAP_FUNCTION] Invalid options: Must use the `map()` function for options.".to_string(),
+            )
+        })?;
+        ToJsonOptions::from_map(map_array)
     } else {
         ToJsonOptions::default()
     };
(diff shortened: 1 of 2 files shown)


python/pysail/tests/spark/function/features/to_json.feature
```diff
@@ -0,0 +1,119 @@
+Feature: to_json function converts complex types to JSON strings
```
linhr, line 1: Yeah it's a nice move for using BDD tests for SQL functions!

crates/sail-function/src/scalar/json/to_json.rs
```diff
@@ -0,0 +1,490 @@
...
+fn to_json_inner(args: &[ArrayRef]) -> Result<ArrayRef> {
+    if args.is_empty() || args.len() > 2 {
+        return Err(datafusion_common::DataFusionError::Plan(
+            "to_json requires 1 or 2 arguments".to_string(),
+        ));
+    }
+
+    let options = if let Some(opts_array) = args.get(1) {
+        if let Some(map_array) = opt_downcast_arg!(opts_array, MapArray) {
+            ToJsonOptions::from_map(map_array)
```
linhr: We should return an error if the second argument is not `MapArray` instead of falling back to the default options.

crates/sail-function/src/scalar/json/to_json.rs
```diff
@@ -0,0 +1,490 @@
...
+        fractional_part,
+        width = scale as usize
+    )
+}
+
+fn decimal_to_json_number(value: i128, scale: i8) -> Value {
+    let decimal_str = format_decimal(value, scale);
+    // Parse the formatted decimal string as a JSON number
+    // This preserves the exact decimal representation
+    serde_json::from_str(&decimal_str).unwrap_or(Value::String(decimal_str))
```
linhr, line 458: I suspect this is not very reliable. Let's add a TODO here.

PR #1221, opened 2026-01-16
fix: try_cast returns NULL for invalid date/timestamp values

### LINKED ISSUE #1192: [bug] `try_cast()` fails for invalid dates and timestamps
[pyspark.sql.Column.try_cast()](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.try_cast.html) is a  special version of `pyspark.sql.Column.cast()` that performs the same operation, but returns a `NULL` value instead of raising an error if the invoke method throws exception.

I ran into the issue, that it does not appear to work reliably for data types `DateType()`, `TimestampType()`, `TimestampNTZType()` (maybe there are even more):

# Reproducing the issue
```python
df = (
    spark
    .createDataFrame(schema="id STRING, some_date STRING", data=[("a", "2025-99-99")])
    .select(
        F.col("id").try_cast("integer"),
        F.col("some_date").try_cast("date"),  # same for try_cast("timestamp"), try_cast("timestamp_ntz")
    )
)
df.printSchema()
df.show()
```

```python
root
 |-- id: integer (nullable = true)
 |-- some_date: date (nullable = true)

---------------------------------------------------------------------------
AnalysisException                         Traceback (most recent call last)
Cell In[20], line 10
      1 df = (
      2     spark
      3     .createDataFrame(schema="id STRING, some_date STRING", data=[("a", "2025-99-99")])
   (...)      7     )
      8 )
      9 df.printSchema()
---> 10 df.show()

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\dataframe.py:1153, in DataFrame.show(self, n, truncate, vertical)
   1152 def show(self, n: int = 20, truncate: Union[bool, int] = True, vertical: bool = False) -> None:
-> 1153     print(self._show_string(n, truncate, vertical))

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\dataframe.py:906, in DataFrame._show_string(self, n, truncate, vertical)
    889     except ValueError:
    890         raise PySparkTypeError(
    891             errorClass="NOT_BOOL",
    892             messageParameters={
   (...)    895             },
    896         )
    898 table, _ = DataFrame(
    899     plan.ShowString(
    900         child=self._plan,
    901         num_rows=n,
    902         truncate=_truncate,
    903         vertical=vertical,
    904     ),
    905     session=self._session,
--> 906 )._to_table()
    907 return table[0][0].as_py()

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\dataframe.py:1837, in DataFrame._to_table(self)
   1835 def _to_table(self) -> Tuple["pa.Table", Optional[StructType]]:
   1836     query = self._plan.to_proto(self._session.client)
-> 1837     table, schema, self._execution_info = self._session.client.to_table(
   1838         query, self._plan.observations
   1839     )
   1840     assert table is not None
   1841     return (table, schema)

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\client\core.py:981, in SparkConnectClient.to_table(self, plan, observations)
    979 req = self._execute_plan_request_with_metadata()
    980 req.plan.CopyFrom(plan)
--> 981 table, schema, metrics, observed_metrics, _ = self._execute_and_fetch(req, observations)
    983 # Create a query execution object.
    984 ei = ExecutionInfo(metrics, observed_metrics)

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\client\core.py:1697, in SparkConnectClient._execute_and_fetch(self, req, observations, self_destruct)
   1694 properties: Dict[str, Any] = {}
   1696 with Progress(handlers=self._progress_handlers, operation_id=req.operation_id) as progress:
-> 1697     for response in self._execute_and_fetch_as_iterator(
   1698         req, observations, progress=progress
   1699     ):
   1700         if isinstance(response, StructType):
   1701             schema = response

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\client\core.py:1674, in SparkConnectClient._execute_and_fetch_as_iterator(self, req, observations, progress)
   1672     raise kb
   1673 except Exception as error:
-> 1674     self._handle_error(error)

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\client\core.py:1982, in SparkConnectClient._handle_error(self, error)
   1980     self.thread_local.inside_error_handling = True
   1981     if isinstance(error, grpc.RpcError):
-> 1982         self._handle_rpc_error(error)
   1983     raise error
   1984 finally:

File c:\...\.venv\Lib\site-packages\pyspark\sql\connect\client\core.py:2066, in SparkConnectClient._handle_rpc_error(self, rpc_error)
   2057                 logger.info(
   2058                     "Disabling plan compression for the session due to "
   2059                     "CONNECT_INVALID_PLAN.CANNOT_PARSE error."
   2060                 )
   2061                 self._plan_compression_threshold, self._plan_compression_algorithm = (
   2062                     -1,
   2063                     "NONE",
   2064                 )
-> 2066             raise convert_exception(
   2067                 info,
   2068                 status.message,
   2069                 self._fetch_enriched_error(info),
   2070                 self._display_server_stack_trace(),
   2071                 status_code,
   2072             ) from None
   2074     raise SparkConnectGrpcException(
   2075         message=status.message,
   2076         grpc_status_code=status_code,
   2077     ) from None
   2078 else:

AnalysisException: invalid argument: DateValue { year: 2025, month: 99, day: 99 }
```

pomykalakyle:
Fixes #1192

- Add is_try support to SparkDate and SparkTimestamp UDFs
- When is_try=true, return NULL instead of throwing error on parse failure
- Update cast.rs to use try variants for string->date/timestamp conversions
- Add test for try_cast with invalid date/timestamp values

### DIFF
--- a/crates/sail-execution/src/codec.rs
+++ b/crates/sail-execution/src/codec.rs
@@ -1345,9 +1345,12 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
                 let udf = TimestampNow::new(Arc::from(timezone), time_unit);
                 return Ok(Arc::new(ScalarUDF::from(udf)));
             }
-            UdfKind::SparkTimestamp(gen::SparkTimestampUdf { timezone }) => {
-                let udf = SparkTimestamp::try_new(timezone.map(Arc::from), false)?;
+            UdfKind::SparkTimestamp(gen::SparkTimestampUdf { timezone, is_try }) => {
+                let udf = SparkTimestamp::try_new(timezone.map(Arc::from), is_try)?;
                 return Ok(Arc::new(ScalarUDF::from(udf)));
+            }
+            UdfKind::SparkDate(gen::SparkDateUdf { is_try }) => {
+                return Ok(Arc::new(ScalarUDF::from(SparkDate::new(is_try))));
             }
         };
         match name {
@@ -1448,7 +1451,6 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_elt" | "elt" => Ok(Arc::new(ScalarUDF::from(SparkElt::new()))),
             "spark_decode" | "decode" => Ok(Arc::new(ScalarUDF::from(SparkDecode::new()))),
             "spark_bin" | "bin" => Ok(Arc::new(ScalarUDF::from(SparkBin::new()))),
-            "spark_date" => Ok(Arc::new(ScalarUDF::from(SparkDate::new(false)))),
             "spark_year_month_interval" => {
                 Ok(Arc::new(ScalarUDF::from(SparkYearMonthInterval::new())))
             }
@@ -1529,7 +1531,6 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             || node_inner.is::<SparkConcat>()
             || node_inner.is::<SparkConv>()
             || node_inner.is::<SparkCrc32>()
-            || node_inner.is::<SparkDate>()
             || node_inner.is::<SparkDayTimeInterval>()
             || node_inner.is::<SparkDecode>()
             || node_inner.is::<SparkElt>()
@@ -1654,7 +1655,11 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             })
         } else if let Some(func) = node.inner().as_any().downcast_ref::<SparkTimestamp>() {
             let timezone = func.timezone().map(|x| x.to_string());
-            UdfKind::SparkTimestamp(gen::SparkTimestampUdf { timezone })
+            let is_try = func.is_try();
+            UdfKind::SparkTimestamp(gen::SparkTimestampUdf { timezone, is_try })
+        } else if let Some(func) = node.inner().as_any().downcast_ref::<SparkDate>() {
+            let is_try = func.is_try();
+            UdfKind::SparkDate(gen::SparkDateUdf { is_try })
         } else {
             return Ok(());
         };
--- a/crates/sail-function/src/scalar/datetime/spark_date.rs
+++ b/crates/sail-function/src/scalar/datetime/spark_date.rs
@@ -17,12 +17,6 @@ pub struct SparkDate {
     is_try: bool,
 }
 
-impl Default for SparkDate {
-    fn default() -> Self {
-        Self::new(false)
-    }
-}
-
 impl SparkDate {
     /// Creates a SparkDate.
     ///
@@ -38,6 +32,10 @@ impl SparkDate {
             ),
             is_try,
         }
+    }
+
+    pub fn is_try(&self) -> bool {
+        self.is_try
     }
 
     fn string_to_date32(value: &str, is_try: bool) -> Result<Option<i32>> {
--- a/crates/sail-function/src/scalar/datetime/spark_timestamp.rs
+++ b/crates/sail-function/src/scalar/datetime/spark_timestamp.rs
@@ -100,6 +100,10 @@ impl SparkTimestamp {
     pub fn timezone(&self) -> Option<&str> {
         self.timezone.as_deref()
     }
+
+    pub fn is_try(&self) -> bool {
+        self.is_try
+    }
 }
 
 impl ScalarUDFImpl for SparkTimestamp {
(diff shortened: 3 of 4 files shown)


crates/sail-function/src/scalar/datetime/spark_date.rs
```diff
@@ -23,21 +24,34 @@ impl Default for SparkDate {
...
         Self {
             signature: Signature::coercible(
                 vec![Coercion::new_exact(TypeSignatureClass::Native(
                     logical_string(),
                 ))],
                 Volatility::Immutable,
             ),
+            is_try,
         }
     }
```
linhr: I wonder if we should keep a single `new()` method and make `is_try` a mandatory parameter. This conveys the behavior of the UDF explicitly in all cases. (It's Ok to have breaking changes in the internal interface.)

Also you would need to update `codec.rs`. (Sorry again for this manual reminder as we don't have test suites for that yet.) You need to update `.proto` in this case.

(In fact, if we have `pub fn new(is_try: bool) -> Self`, the missing codec would be caught by the compiler in this case.)

crates/sail-function/src/scalar/datetime/spark_timestamp.rs
```diff
@@ -70,6 +99,7 @@ impl SparkTimestamp {
                 ))],
                 Volatility::Immutable,
             ),
+            is_try,
         })
     }
```
linhr: Same here, I feel we could have a single `new` method. And you need the `codec.rs` and `.proto` changes as well.

python/pysail/tests/spark/test_column.py
```diff
@@ -77,3 +77,27 @@ def test_get_item_nested_map(spark):
...
+
+def test_try_cast_invalid_date(spark):
+    """Test that try_cast returns NULL for invalid date/timestamp values.
+
+    See: https://github.com/lakehq/sail/issues/1192
+    """
+    # try_cast was added in Spark 4.0
+    col = F.col("x")
+    if not callable(getattr(col, "try_cast", None)):
+        pytest.skip("try_cast not available in this Spark version")
```
linhr, line 90: Nice compatibility guard!

crates/sail-function/src/scalar/datetime/spark_date.rs
pomykalakyle: I actually prefer that approach a lot and was what I first thought about! But I wanted to isolate any changes to the codebase so this wouldn't impact other places.

linhr: Got it. Glad to see that our thoughts are aligned here!

crates/sail-execution/src/codec.rs
```diff
@@ -1448,7 +1448,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_elt" | "elt" => Ok(Arc::new(ScalarUDF::from(SparkElt::new()))),
             "spark_decode" | "decode" => Ok(Arc::new(ScalarUDF::from(SparkDecode::new()))),
             "spark_bin" | "bin" => Ok(Arc::new(ScalarUDF::from(SparkBin::new()))),
-            "spark_date" => Ok(Arc::new(ScalarUDF::from(SparkDate::new()))),
+            "spark_date" => Ok(Arc::new(ScalarUDF::from(SparkDate::new(false)))),
```
linhr: You'll need to add `is_try` for `SparkTimestampUdf` in `crates/sail-execution/proto/sail/plan/physical.proto` and similarly define a proto message `SparkDateUdf`. (`SparkDate` didn't take arguments so we don't have a message defined for it yet.)

You could test the changes locally.

(Terminal 1)
```bash
hatch run SAIL_MODE=local-cluster scripts/spark-tests/run-server.sh
```
(Terminal 2)
```bash
env SPARK_CONNECT_MODE_ENABLED=1 SPARK_REMOTE="sc://localhost:50051" hatch run pytest -k "test_try_cast_invalid_date"
```

pomykalakyle: Working on that now haha - just was pushing a bit too early (I have a shortcut for it I press)

crates/sail-function/src/scalar/datetime/spark_date.rs
```diff
@@ -14,30 +14,38 @@ use sail_sql_analyzer::parser::parse_date;
...
     signature: Signature,
+    is_try: bool,
 }
 
 impl Default for SparkDate {
     fn default() -> Self {
-        Self::new()
+        Self::new(false)
     }
 }
```
linhr: We could just remove `Default`.

PR #1222, opened 2026-01-16
feat: distributed log replay for Delta Lake

lonless9:
part of #171 

This PR implements:
- Streaming Log Replay
- Clustered Writer Architecture
- Improvements about metadata and plan tree.


crates/sail-delta-lake/src/physical_plan/log_replay_exec.rs
```diff
@@ -0,0 +1,701 @@
...
+        );
+        vec![Distribution::HashPartitioned(vec![expr])]
+    }
+
+    fn required_input_ordering(&self) -> Vec<Option<OrderingRequirements>> {
+        // The streaming replay logic relies on all rows for the same `COL_REPLAY_PATH`
+        // being adjacent within each partition, so we require a local ordering by
+        // (COL_REPLAY_PATH ASC, COL_LOG_VERSION DESC).
+        // TODO: Add COL_LOG_IS_REMOVE ASC as a tie-breaker so Add beats Remove within
+        // the same path/version (needed for DV updates: Remove(old dv) + Add(new dv)).
```
lonless9, line 332: This is related to follow-up work about operations like Deletion Vectors and Metadata Updates.

crates/sail-delta-lake/src/physical_plan/log_replay_exec.rs
```diff
@@ -0,0 +1,701 @@
...
+            // Input is expected to be sorted by (replay_path ASC, log_version DESC), so the first
+            // decision we make for a given path must be final (older actions cannot override it).
+            if self.current_removed || self.current_row.is_some() {
+                continue;
+            }
+
+            if !is_remove.is_null(row) && is_remove.value(row) {
+                self.current_removed = true;
+                continue;
+            }
```
lonless9, line 282: Here implements the "Newest-First" resolution. Once we've seen any action for a specific path (either an active Add stored in current_row or a Remove marking it current_removed), we explicitly ignore all subsequent rows for that path in the stream.
This confirms why the upstream SortExec order is needed for correctness—we are relying on the assumption that we will never see a "newer" action after an "older" one for the same path.

crates/sail-delta-lake/src/datasource/mod.rs
```diff
@@ -32,6 +32,8 @@ use crate::kernel::snapshot::LogDataHandler;
 use crate::kernel::{DeltaResult, DeltaTableError};
 use crate::table::DeltaTableState;
 pub const PATH_COLUMN: &str = "__sail_file_path";
+pub const COMMIT_VERSION_COLUMN: &str = "_commit_version";
+pub const COMMIT_TIMESTAMP_COLUMN: &str = "_commit_timestamp";
```
lonless9, line 36: These virtual columns could serve as the infrastructure for time travel, and perhaps some internal tables could be exposed as well.

crates/sail-delta-lake/src/operations/write/writer.rs
```diff
@@ -142,101 +147,128 @@ impl DeltaWriter {
...
+    async fn switch_partition_if_needed(
+        &mut self,
+        partition_key: String,
+        partition_values: IndexMap<String, Scalar>,
+    ) -> Result<(), DeltaTableError> {
+        if self.current_partition_key.as_deref() == Some(partition_key.as_str())
+            && self.current_writer.is_some()
+        {
+            return Ok(());
+        }
```
lonless9, line 219: The previous Mechanism (`HashMap<PartitionKey, PartitionWriter>`) used unordered writing, keeping an active writer open for each partition simultaneously, causing linear growth in memory and file handles. 

New Mechanism (`Option<PartitionWriter>`) requires upstream data to be clustered by partition key, maintains one active writer at a time, and closes the previous writer before opening a new one on partition switch.

crates/sail-delta-lake/src/physical_plan/planner/op_write.rs
```diff
@@ -45,18 +47,105 @@ pub async fn build_write_plan(
...
+    input: Arc<dyn ExecutionPlan>,
+    sort_order: Option<LexRequirement>,
 ) -> Result<Arc<dyn ExecutionPlan>> {
     let input_schema = input.schema();
-    build_standard_write_layers(ctx, input, &sink_mode, sort_order, input_schema)
+
+    let target_partitions = ctx.session().config().target_partitions().max(1);
+    let plan = create_projection(input, ctx.partition_columns().to_vec())?;
+    let plan = create_repartition(plan, ctx.partition_columns().to_vec(), target_partitions)?;
+    let plan = create_sort(plan, ctx.partition_columns().to_vec(), sort_order)?;
```
lonless9, line 79: This is where the writer data is organized.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
```diff
@@ -48,6 +103,9 @@ pub fn build_standard_write_layers(
         None,
     )?);
 
+    // DeltaCommitExec is single-partition; gather writer partitions first.
+    let writer: Arc<dyn ExecutionPlan> = Arc::new(CoalescePartitionsExec::new(writer));
```
lonless9, line 107: Currently, there are some places where `CoalescePartitionsExec` has been manually added to commit nodes, which is worth reconsidering in conjunction with optimizer behavior. Maybe we don't need manually do and optimizer will take over.

python/pysail/tests/spark/delta/__snapshots__/test_delta_features.yaml
```diff
@@ -59,11 +59,11 @@ data:
...
-    numAddedFiles: 1
-    numFiles: 1
-    numOutputBytes: 1093
-    numOutputFiles: 1
+    numAddedBytes: 2122
+    numAddedFiles: 2
+    numFiles: 2
+    numOutputBytes: 2122
+    numOutputFiles: 2
     numOutputRows: 4
```
lonless9, line 67: This is caused by the concurrent writes. We could adopt different strategies for the partitioned and non-partitioned, although I'm not sure if it's worth it.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
```diff
@@ -12,28 +12,83 @@
...
+    /// Whether to include `stats_json` in the replay output (as a Utf8 column).
+    pub include_stats_json: bool,
+    /// Optional inclusive log version range for commit JSON files.
+    pub commit_version_range: Option<(i64, i64)>,
+    /// Optional metadata-stage filter applied after log replay.
+    pub log_filter: Option<LogReplayFilter>,
+    /// Optional predicate pushed down to checkpoint parquet scan.
+    pub parquet_predicate: Option<Arc<dyn PhysicalExpr>>,
+}
+
```
lonless9, line 63: These options may be helpful for subsequent optimizations.

crates/sail-delta-lake/src/physical_plan/planner/log_scan.rs
```diff
@@ -16,9 +18,22 @@ use object_store::{ObjectMeta, ObjectStore};
...
+pub struct LogScanOptions {
+    /// Optional projection of top-level log columns (e.g. ["add", "remove", "metaData"]).
+    ///
+    /// When set, the scan will only read these columns plus any required partition columns.
+    pub projection: Option<Vec<String>>,
+    /// Optional inclusive log version range for commit JSON files.
+    pub commit_version_range: Option<(i64, i64)>,
+    /// Optional pushdown predicate for checkpoint parquet scans.
+    pub parquet_predicate: Option<Arc<dyn PhysicalExpr>>,
+}
```
lonless9, line 35: These options may be helpful for subsequent optimizations.

python/pysail/tests/spark/delta/__snapshots__/test_delta_features/test_explain_plan_for_conditional_overwrite_replace_where_category__a.plan
```diff
@@ -3,21 +3,24 @@ DeltaCommitExec(table_path=file://<tmp>/delta_overwrite/)
...
+                    RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=1
```
lonless9, line 25: Currently, options take effect on paths like this, which can be reflected in the datasource.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
```diff
@@ -104,3 +162,272 @@ pub fn align_schemas_for_union(
...
+    let remove_path = guard_with(
+        remove_is_not_null.clone(),
+        get_field_expr(remove_col_expr.clone(), "path"),
+    );
+
+    let replay_path = simplify(Expr::ScalarFunction(ScalarFunction::new_udf(
+        datafusion::functions::core::coalesce(),
+        vec![add_path, remove_path.clone()],
+    )))?;
+
```
lonless9, line 261: Here an `Expr` is constructed and then simplify is called, I'm not sure if this is a straightforward implementation.

crates/sail-delta-lake/src/physical_plan/log_replay_exec.rs
linhr: I just realized that this sounds similar to an aggregation such as "max by". I wonder if we can even further optimize it so that the sort is not needed (since sorting is more expensive than a single-pass iteration). We just need to implement an aggregator (or even express the replay logic using existing logical expressions) and then use `AggregateExec`.

This is of course something to explore in the future. (And I just have a rough idea so I may not be correct here.)

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
linhr: I think `required_input_distribution` and the optimizer would take care of this.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
```diff
@@ -12,28 +12,83 @@
...
-    create_projection, create_repartition, create_sort, DeltaCommitExec, DeltaWriterExec,
+    create_projection, create_repartition, create_sort, DeltaCommitExec, DeltaLogReplayExec,
+    DeltaPhysicalExprAdapterFactory, DeltaWriterExec, COL_LOG_IS_REMOVE, COL_LOG_VERSION,
+    COL_REPLAY_PATH,
 };
 
+/// Options that control what the log replay pipeline materializes as payload columns.
+///
+/// This is intentionally kept small: it is primarily used to avoid scanning/transporting
+/// `stats_json` unless downstream pruning (data skipping) actually needs it.
```
linhr, line 51: As we move more logic to logical plans, I wonder if the optimizer can decide automatically whether this column is needed when doing projection pruning.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
linhr: This sounds another place where logical planning may help. 😆

crates/sail-delta-lake/src/physical_plan/writer_exec.rs
```diff
@@ -190,7 +194,58 @@ impl ExecutionPlan for DeltaWriterExec {
...
 
     fn required_input_distribution(&self) -> Vec<Distribution> {
-        vec![Distribution::SinglePartition]
+        if self.partition_columns.is_empty() {
+            // Upstream repartitioning controls file counts and small-file behavior.
+            return vec![Distribution::UnspecifiedDistribution];
+        }
+
+        // For partitioned tables, require grouping by the partition key so that each task can
+        // write its partitions correctly without opening many writers concurrently.
```
linhr, line 203: I feel this may cause imbalance if different partitions have different amounts of data.

I guess the "bucket by" columns should be used for hash partitioning, but since we don't have proper support for "bucket by" yet, this is future work.

crates/sail-delta-lake/src/operations/write/writer.rs
linhr: I feel the previous mechanism has its benefit if the cardinality of partition keys is small, as it allows us to avoid extra sorting or clustering of the input data, so we may want to revisit this in the future.

In my other comment, I mentioned the distinction between "partition by" and "bucket by". We may consider this altogether in the writer design.

crates/sail-delta-lake/src/physical_plan/log_replay_exec.rs
lonless9: I feel the same way this relational algebra need to be pushed up into the logical plan!

crates/sail-delta-lake/src/operations/write/writer.rs
lonless9: The redesign for the writer will be truly exciting!

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
lonless9: I will try to address the issues here in subsequent PRs. This is likely due to inconsistent handling of some properties during my implementation.

crates/sail-delta-lake/src/physical_plan/planner/utils.rs
lonless9: Yeah I think so! After switching from physical expressions to logical expressions, I observed simplification benefits in the expressions on the execution plan side.

PR #1224, opened 2026-01-16
feat: add `json_tuple` function

davidlghellin:
part #219 
## test Spark
```sh
sdk use java 17.0.9-tem && SPARK_REMOTE="local" hatch run pytest python/pysail/tests/spark/function/test_features.py -v
```

## test Sail
```sh
env RUST_LOG='sail=debug' cargo run -p sail-cli -- spark server
```

```sh
export SPARK_REMOTE="sc://localhost:50051" && hatch run pytest python/pysail/tests/spark/function/test_features.py -v
```

### DIFF
--- a/python/pysail/tests/spark/function/features/json_tuple.feature
+++ b/python/pysail/tests/spark/function/features/json_tuple.feature
@@ -118,7 +118,7 @@ Feature: json_tuple() extracts multiple values from JSON strings
 
   Rule: Star expansion
 
-    Scenario: json_tuple with star expansion
+    Scenario: json_tuple with star expansion over multiple keys
       When query
       """
       SELECT * FROM (SELECT json_tuple('{"name":"Alice","age":"30"}', 'name', 'age'))
@@ -126,3 +126,12 @@ Feature: json_tuple() extracts multiple values from JSON strings
       Then query result ordered
       | c0    | c1 |
       | Alice | 30 |
+
+    Scenario: json_tuple with star expansion over single key
+      When query
+      """
+      SELECT * FROM (SELECT json_tuple('{"x":100}', 'x'))
+      """
+      Then query result ordered
+      | c0  |
+      | 100 |
--- a/crates/sail-plan/src/resolver/query/project.rs
+++ b/crates/sail-plan/src/resolver/query/project.rs
@@ -155,16 +155,20 @@ impl PlanResolver<'_> {
                 Expr::ScalarFunction(ScalarFunction { func, args }) => {
                     if func.inner().as_any().is::<MultiExpr>() {
                         // The metadata from the original expression are ignored.
-                        // For wildcard-derived MultiExpr (multiple names matching args count),
-                        // pair original names with args. For function-derived MultiExpr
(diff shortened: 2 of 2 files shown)


python/pysail/tests/spark/function/test_features.py
```diff
@@ -1,5 +1,7 @@
 """BDD tests for DataFrame operations using pytest-bdd."""
 
+from __future__ import annotations
+
```
linhr: This is not needed.

crates/sail-plan/src/function/scalar/json.rs
```diff
@@ -38,6 +38,24 @@ fn json_object_keys(json_data: expr::Expr) -> expr::Expr {
...
+        return Err(crate::error::PlanError::invalid(
+            "json_tuple requires at least 2 arguments: json string and at least one key",
+        ));
+    }
+
+    // Return the SparkJsonTuple call directly.
+    // The JsonTupleRewriter will transform this into field access expressions
+    // wrapped in MultiExpr during the projection rewrite phase.
+    let func = ScalarUDF::from(SparkJsonTuple::new());
+    Ok(func.call(arguments))
```
linhr: I wonder if we can move the `MultiExpr` logic here so that we don't need to introduce another rewriter. 

The issue with rewriters is that we need to remember to add them in multiple places, so it's a little bit hard to maintain. (For example, in the codebase, you will see that `ExplodeRewriter` is actually needed in many places.)

The rewriter is needed if the expression also expand "rows" so the logical plan needs adjustment accordingly. For `json_tuple`, since it only expands "columns", a rewriter is not strictly necessary.

python/pysail/tests/spark/function/features/json_tuple.feature
```diff
@@ -0,0 +1,146 @@
+Feature: json_tuple() extracts multiple values from JSON strings
+
+  Rule: Basic value extraction
+
+    Scenario: json_tuple extracts simple string values
+      When query
+      """
+      SELECT c0, c1 FROM (SELECT json_tuple('{"name":"Alice","age":"30"}', 'name', 'age') AS (c0, c1))
```
linhr, line 8: I think `SELECT ... FROM (SELECT ... AS ...)` can be simplified to `SELECT ... AS ...` (i.e. no subquery) for all the tests here.

python/pysail/tests/spark/function/features/json_tuple.feature
```diff
@@ -0,0 +1,146 @@
...
+      """
+      SELECT c0 FROM (SELECT json_tuple('{"a":"hello\\"world"}', 'a') AS (c0))
+      """
+      Then query result ordered
+      | c0           |
+      | hello"world  |
+
+  Rule: Direct usage without explicit alias
+
+    Scenario: json_tuple direct select without subquery
```
linhr: We can keep one "with subquery" test while the remaining tests are all "without subquery".

davidlghellin:
https://github.com/apache/datafusion/pull/20412

davidlghellin:
Closing as this has been merged into DataFusion

PR #1225, opened 2026-01-16
feat: improve `schema_of_json` function

davidlghellin:
part #219 
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.schema_of_json.html


python/pysail/tests/spark/function/features/schema_of_json.feature
```diff
@@ -0,0 +1,109 @@
...
+  Scenario: schema_of_json with null values
+    When query
+    """
+    SELECT schema_of_json('{"name":"Alice","age":null}') AS result
+    """
+    Then query result ordered
+    | result                                  |
+    | STRUCT<age: STRING, name: STRING>       |
+
+  Scenario: schema_of_json with mixed types in array
```
linhr: This test doesn't seem to contain mixed types in an array.

python/pysail/tests/spark/function/features/schema_of_json.feature
```diff
@@ -0,0 +1,109 @@
+Feature: schema_of_json() returns the schema of a JSON string as DDL
```
linhr, line 2: Nice tests!

Could you also add a test for an input of multiple rows?

linhr (commented):
This looks great!

I have some minor comments around the tests, but I think it's almost ready to be merged.

python/pysail/tests/spark/function/features/schema_of_json.feature
davidlghellin: ~~I add this test~~
```python
  Scenario: schema_of_json rejects non-foldable column input
    When query
    """
    SELECT schema_of_json(json_col) AS result
    FROM VALUES ('{"name":"Alice"}') AS t(json_col)
    """
    Then query error NON_FOLDABLE_INPUT
```
~~and I change the logic for the result/error as the same~~
 [89d254d](https://github.com/lakehq/sail/pull/1225/commits/89d254de767f20e38cb8bbc547d12ca2f911ca10)

davidlghellin: My bad, I misunderstood — I've added a test with a JSON array containing multiple objects with null, normal and unusual values `([{"a": null}, {"a": 1}, {"a": "weird"}])` which covers the multi-record schema merging case.

crates/sail-function/src/scalar/json/schema_of_json.rs
```diff
@@ -535,6 +584,249 @@ fn is_unquoted_name_char(c: char) -> bool {
...
+
+/// Spark trims leading/trailing whitespace and ISO control characters.
+fn is_whitespace_or_iso_control(b: u8) -> bool {
+    b <= b' ' || b == 0x7f
+}
+
+/// A faithful port of Spark's `SparkDateTimeUtils.parseTimestampString`. Splits
+/// the input into timestamp segments using Spark's exact delimiter rules, or
+/// returns `None` if the string is not shaped like a timestamp.
+fn parse_timestamp_string(s: &str) -> Option<ParsedTimestamp> {
```
linhr, line 615: This may become easier after #2099.

linhr (approved):
Thanks for the follow-up for the improvements! This looks good to me!

PR #1226, opened 2026-01-17
feat: initial Arrow Flight integration in Sail

davidlghellin:
part #520
Thanks @DeliveranceTechSolutions 

Basic Flight SQL server (sail-flight crate)

Command-line interface with configurable host/port
```sh
cargo run -p sail-cli -- flight server --port 32010
```

### DIFF
--- a/python/pysail/tests/flight/test_flight.py
+++ b/python/pysail/tests/flight/test_flight.py
@@ -6,6 +6,7 @@ from adbc_driver_flightsql import dbapi
 
 from pysail.flight import FlightSqlServer
 
+
 @pytest.fixture(scope="module")
 def flight_uri():
     if "SAIL_FLIGHT_URI" in os.environ:


davidlghellin:
Configure dbeaver

```sh
wget https://repo1.maven.org/maven2/org/apache/arrow/flight-sql-jdbc-driver/18.3.0/flight-sql-jdbc-driver-18.3.0.jar
```

image

Create new
image
Add library
image


image
image


Example working
image

Not implemented yet
image
image

davidlghellin:
https://github.com/lakehq/sail/pull/1226/commits/006b26aa8d076485f08d74cac440b846233e3887
```sql
CREATE TABLE test
USING csv
OPTIONS (header 'true', delimiter ';')
LOCATION '/tmp/test.csv';

SELECT * FROM test;
```
image

davidlghellin:
I played with grafana (report from claude) to saw a basic report
image

```sh
export SAIL_TELEMETRY__EXPORT_METRICS=true
export SAIL_TELEMETRY__EXPORT_TRACES=true
export SAIL_TELEMETRY__OTLP_ENDPOINT=http://localhost:4317
cargo run --bin sail-flight -- server --port 32010
```

crates/sail-flight/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+
+# Serialization
+serde = { workspace = true }
+
+# Config
+figment = { workspace = true }
+
+# Cache
+lru = "0.12"
+
```
lonless9: Declaring version numbers within sub crates is not ideal. Additionally, I recall there was a discussion about this when Datafusion introduced the caching mechanism. (https://github.com/apache/datafusion/pull/17031#discussion_r2252442774)
That means we might end up relying on three crates simultaneously: lru-rs, moka-rs, and foyer?

davidlghellin: Thanks for the context — that makes sense.

The version pinning was purely experimental and not intended as a final design choice. It will be removed in the next commit to avoid unnecessary noise, and we can revisit cache dependency management later in the appropriate cache-related issue if needed.

lonless9: Yeah, I'm just providing some context. I think `lru-rs` is a lightweight enough library, and if we have some handy cache implementations, we can find optimization opportunities in more places. However cache issues are always tricky, we need to have an overall consideration.

.github/workflows/flight-tests.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Wait for Flight SQL Server to Start
+        run: |
+          scripts/flight-tests/wait-for-server.sh
+
+      - name: Run Flight SQL Tests
+        timeout-minutes: 5
+        env:
+          SAIL_FLIGHT_URI: "grpc://127.0.0.1:32010"
+          TEST_RUN_GIT_COMMIT: ${{ github.sha }}
+          TEST_RUN_GIT_REF: ${{ github.ref }}
```
linhr: This can be removed since we are not generating the test report.

crates/sail-flight/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+        disable_cache_stats: bool,
+
+        /// Maximum rows to return per query (0 = unlimited, default: 0)
+        #[arg(long, default_value = "0")]
+        max_rows: usize,
+    },
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
linhr: We need to integrate the Flight server in the `sail-cli` crate so that we don't need to build the binary separately in `flight-tests.yml`.

crates/sail-flight/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+                return Err(e);
+            }
+        }
+    }
+
+    shutdown_telemetry();
+    Ok(())
+}
+
+async fn run_flight_server(
```
linhr: We should use `ServerBuilder` from the `sail-server` crate to create the gRPC server.

python/pysail/tests/flight/test_flight_features.py
```diff
@@ -0,0 +1,15 @@
...
+
+This module uses pytest-bdd to run Gherkin feature files that test
+the Sail Arrow Flight SQL server functionality.
+"""
+
+import pytest
+from pytest_bdd import scenarios
+
+# Load all scenarios from the features directory
+scenarios("features/")
```
linhr: I feel BDD tests for Flight SQL are not necessary. The tests will focus on connection management and request/response handling, which is easier to write in Python tests. It seems like you already have Python tests for this.

crates/sail-flight/src/cache.rs
```diff
@@ -0,0 +1,335 @@
...
+pub struct CacheStats {
+    pub hits: u64,
+    pub misses: u64,
+    pub evictions: u64,
+    pub current_size_bytes: usize,
+    pub current_entries: usize,
+}
+
+/// LRU cache for prepared statements
+pub struct PreparedStatementCache {
```
linhr: I feel we can support prepared statements in a follow-up PR, as caching may require some design. For this PR, we can focus on getting the basic project setup in a good state and see if we can integrate the session manager.

crates/sail-flight/src/session.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Key features:
+/// - Uses Sail's custom analyzers and optimizers (shared with Spark Connect)
+/// - Registers PlanService extension for plan formatting (same as Spark Connect)
+/// - Registers CatalogManager for table creation and querying
+/// - Registers TableFormatRegistry for file format support (parquet, csv, etc.)
+/// - Uses custom query planner for extension nodes
+///
+/// # Differences with sail-spark-connect
+/// - **NO SessionManager**: Flight SQL uses simple per-request sessions
+/// - **NO ActorSystem**: No need for multi-session orchestration
```
linhr: This assumption is incorrect. It seems that Arrow Flight SQL supports session management, and we need to support multi-session using `SessionManager`.

.github/workflows/flight-tests.yml
```diff
@@ -0,0 +1,155 @@
+name: Flight SQL Tests
+
+on:
+  workflow_call:
+    secrets:
+      CODECOV_TOKEN:
+        required: false
+
+jobs:
+  setup:
```
linhr: I'm making changes to testing CI in #1300. You may want to take a look and adjust `flight-tests.yml` accordingly. We no longer have the `setup` step in the test workflows.

.github/workflows/flight-tests.yml
```diff
@@ -0,0 +1,131 @@
...
+
+      - name: Install grcov
+        uses: taiki-e/install-action@grcov
+
+      - name: Install protobuf compiler
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+          protoc --version
+
```
linhr: Remove.

python/pysail/tests/flight/test_flight.py
```diff
@@ -0,0 +1,353 @@
...
+        logger.info("Server not running at %s, starting new instance", uri)
+    except OSError as e:
+        logger.warning("Unexpected error checking server: %s", e)
+    else:
+        yield uri
+        return
+
+    # Start the server process
+    workspace_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(__file__)))))
+    cargo_path = shutil.which("cargo")
```
linhr: We should not use `cargo` in Python tests.

.github/workflows/python-tests.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           fi
 
       - name: Run Tests
-        run: hatch run pytest --pyargs pysail
+        run: hatch run pytest --pyargs pysail.tests --ignore-glob='**/flight/**'
```
linhr: We cannot filter tests like this since we need the `pytest` command to run without extra arguments in the release workflow and user installation.

Maybe we can add a Python API to create the Arrow Flight server programmatically so we can launch the server in Python, similar to how we manage `SparkConnectServer`? In this way we won't need `flight-tests.yml` at all.

crates/sail-cli/src/runner.rs
```diff
@@ -17,10 +18,43 @@ struct Cli {
...
+            default_value_t = 32010,
+            help = "The port number that the server listens on"
+        )]
+        port: u16,
+        #[arg(
+            long,
+            default_value_t = 0,
+            help = "Maximum rows to return per query (0 = unlimited)"
+        )]
+        max_rows: usize,
```
linhr: I feel it's better to have `max_rows` as a Sail configuration instead of a CLI argument.

crates/sail-flight/src/service.rs
```diff
@@ -0,0 +1,1351 @@
...
+            .await
+            .map_err(|e| Status::internal(format!("Session error: {}", e)))
+    }
+
+    // =========================================================================
+    // Metrics helpers
+    // =========================================================================
+
+    /// Get query type for metrics labeling
+    fn get_query_type(sql: &str) -> &'static str {
```
linhr: I feel this is not a good way to infer the query type. We should parse the SQL and infer the query type from the AST.

crates/sail-flight/src/session.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Uses the same SessionManager infrastructure as `sail-spark-connect`:
+/// - Actor-based session management for concurrent requests
+/// - Session factory for creating new sessions on demand
+/// - Automatic session cleanup and lifecycle management
+///
+/// # Shared components with sail-spark-connect
+/// - `sail_session::SessionManager`: Multi-session orchestration
+/// - `sail_session::ServerSessionFactory`: Session creation and configuration
+/// - `sail_plan`: PlanResolver, PlanConfig, execute_logical_plan
+/// - `sail_session`: Optimizer rules, analyzer rules, table format registry
```
linhr: Let's remove lengthy AI-generated documentation.

crates/sail-flight/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+name = "sail-flight"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+description = "Arrow Flight SQL server for Sail"
+
+[lib]
+name = "sail_flight"
+path = "src/lib.rs"
+
```
linhr, line 8: Let's follow the style from existing crates.

pyproject.toml
```diff
@@ -95,6 +95,9 @@ dependencies = [
     "jedi>=0.19.2,<0.20",
     "jsonpath-ng>=1.6,<2",
     "PyYAML>=6.0,<7",
+    "pyarrow>=19.0,<20",
```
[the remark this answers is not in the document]
linhr: This env is for dev purpose so we want to be restrictive on the versions.

crates/sail-flight/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// This module defines the configuration options for the Flight SQL server:
+/// - Server binding (host/port)
+/// - Query execution limits
+use std::net::SocketAddr;
+
+use serde::{Deserialize, Serialize};
+
+/// Main server configuration
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
```
linhr: I don't think we need serde support here. Also in our codebase, it's more conventional to name this `Options` instead of `Config`.

davidlghellin:
```sh
hatch run maturin develop
```

```python
"""Quick Flight SQL smoke test."""
from pysail.flight import FlightSqlServer
from adbc_driver_flightsql import dbapi

server = FlightSqlServer(ip="127.0.0.1", port=0)
server.start(background=True)

addr = server.listening_address
print(f"Server on {addr[0]}:{addr[1]}")

conn = dbapi.connect(f"grpc://{addr[0]}:{addr[1]}")
cur = conn.cursor()

cur.execute("SELECT 1 + 2 AS result")
print(cur.fetchall())

cur.execute("SELECT upper('hello') AS greeting")
print(cur.fetchall())

cur.close()
conn.close()
server.stop()
print("OK")
```

```sh
hatch run python /tmp/test_flight.py
```
image

.github/workflows/python-tests.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           fi
 
       - name: Run Tests
-        run: hatch run pytest --pyargs pysail
+        run: hatch run pytest --pyargs pysail.tests
```
linhr: Revert.

pyproject.toml
linhr, line 33: Let's update the test dependencies here.

python/pysail/libsail_python.dylib.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
linhr: Remove this file.

crates/sail-cli/src/spark/server.rs
```diff
@@ -120,6 +120,33 @@ where
     })
 }
 
-pub fn run_spark_connect_server(ip: IpAddr, port: u16) -> Result<(), Box<dyn std::error::Error>> {
-    with_spark_connect_server((ip, port), shutdown(), |_| async { Ok(()) })
+pub fn run_spark_connect_server(
+    ip: IpAddr,
+    port: u16,
+    flight_port: Option<u16>,
```
linhr: Let's revert the changes to the Spark server. We'd like Flight server and Spark Connect server to be separate.

crates/sail-flight/src/error.rs
```diff
@@ -0,0 +1,36 @@
+use std::fmt;
+
+use datafusion::error::DataFusionError;
+
+/// Flight SQL error types
+#[derive(Debug)]
+pub enum FlightError {
```
linhr, line 5: Let's derive `thiserror::Error` instead of implementing this manually.

And the `Internal` variant should be called `DataFusion` instead.

python/pysail/tests/flight/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+These tests require a running Sail Flight SQL server and are excluded from
+the default test suite. To run these tests, use:
+
+    pytest python/pysail/tests/spark/server-flight/
+
+Or set the SAIL_FLIGHT_URI environment variable and run:
+
+    SAIL_FLIGHT_URI=grpc://localhost:32010 pytest python/pysail/tests/spark/server-flight/
+"""
```
linhr: Remove this comment. This is probably outdated.

python/pysail/tests/flight/conftest.py
```diff
@@ -0,0 +1,14 @@
+"""Configuration for Arrow Flight SQL tests."""
+
+
+def pytest_configure(config):
+    """Register custom markers for Flight SQL tests."""
+    config.addinivalue_line(
+        "markers",
+        "flight_sql: mark test as requiring a running Flight SQL server",
+    )
```
linhr: Remove this marker. We run the Flight tests by default.

python/pysail/tests/flight/test_flight.py
```diff
@@ -0,0 +1,242 @@
...
+    with pytest.raises(
+        (adbc_driver_manager.ProgrammingError, adbc_driver_manager.InternalError),
+        match="nonexistent_table",
+    ):
+        cur.execute("SELECT * FROM nonexistent_table_12345")
+
+    cur.close()
+
+
+def test_flight_not_implemented_function(flight_connection):
```
linhr: I feel this test should be removed since the function may be implemented in the future.

crates/sail-flight/src/session.rs
```diff
@@ -0,0 +1,90 @@
...
+        let config = config.clone();
+        let runtime = runtime.clone();
+        let system = system.clone();
+        Box::new(move || {
+            create_flight_session_factory(config.clone(), runtime.clone(), system.clone())
+        })
+    };
+    let options = SessionManagerOptions::new(runtime.clone(), system, factory)
+        .with_session_timeout(std::time::Duration::from_secs(
+            config.spark.session_timeout_secs,
```
linhr: We need to define `config.flight.session_timeout_secs`.

crates/sail-flight/src/service.rs
```diff
@@ -0,0 +1,1311 @@
...
+    // ========================================================================
+    // Unimplemented Flight SQL Operations (Future Work)
+    // ========================================================================
+    //
+    // The following methods are part of the Arrow Flight SQL specification
+    // but are not yet implemented. They return Status::unimplemented() to
+    // indicate to clients that these operations are not supported.
+    //
+    // See ARCHITECTURE.md for implementation roadmap.
+
```
linhr: Remove this comment.

crates/sail-flight/src/service.rs
```diff
@@ -0,0 +1,1311 @@
...
+            _ => Self::Other,
+        }
+    }
+}
+
+pub struct SailFlightSqlService {
+    session_manager: SessionManager,
+    config: Arc<PlanConfig>,
+    /// Maximum rows to return per query (0 = unlimited)
+    max_rows: usize,
```
linhr: Consider having `SailFlightSqlOptions` and use it as the argument to `fn new` even though we only have `max_rows` for now.

PR #1228, opened 2026-01-18
fix: onelake catalog type mapping and schema lookup

djouallah:
## Summary
Apologies - should have tested properly before merging #1217. Was too excited when a simple SELECT worked.

Fixes:
- Prefer `type_name` over `type_text` (OneLake API returns `type_text: null`)
- Use fully qualified schema name for `get_database()` (fixes USE SCHEMA 400 error)
- Parse `decimal(precision,scale)` format from `type_name` field

## Test plan
Tested with TPC-H SF1 on Fabric Lakehouse: USE SCHEMA, SELECT, aggregations, 3-table joins with decimal arithmetic.

### DIFF
--- a/crates/sail-catalog-onelake/src/provider.rs
+++ b/crates/sail-catalog-onelake/src/provider.rs
@@ -104,7 +104,6 @@ struct TableInfo {
 #[derive(Debug, Deserialize)]
 struct ColumnInfo {
     name: Option<String>,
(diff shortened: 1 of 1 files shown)


linhr (approved):
This looks good! Thanks for the fix!

djouallah:
image

Issue #483, opened 2025-05-24
PySpark DataFrame Doctest: randomSplit


shehabgamin:
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.randomSplit`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.randomSplit.html

SparkApplicationMaster:
The step-by-step plan to resolve randomSplit in spark:
1) Spark-connect creates a Sample plan for each interval with deterministic_order=True:
https://github.com/apache/spark/blob/c951611e8403953f102a756df5aab9b3951ab08d/python/pyspark/sql/connect/dataframe.py#L1071

2) If no MapType columns in DF, spark applies SortWithinPartitions with all orderable columns
if there are no orderable columns, it applies DF.cache() instead:
https://github.com/apache/spark/blob/c951611e8403953f102a756df5aab9b3951ab08d/sql/connect/server/src/main/scala/org/apache/spark/sql/connect/planner/SparkConnectPlanner.scala#L377

3) The sample is without replacement, so physical SampleExec calls randomSampleWithRange:
https://github.com/apache/spark/blob/c951611e8403953f102a756df5aab9b3951ab08d/sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala#L336

4) The BernoulliCellSampler is created with seed salted by partition_number calls the partition sampling
https://github.com/apache/spark/blob/c951611e8403953f102a756df5aab9b3951ab08d/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L603
BernoulliCellSampler implements RandomSampler trait:  
https://github.com/apache/spark/blob/322b3d06d3f86c23bbe14d3514c1ffd731c216bb/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L41

5) BernoulliCellSampler uses XORShiftRandom to generate flag whether the row is needed or not:
https://github.com/apache/spark/blob/322b3d06d3f86c23bbe14d3514c1ffd731c216bb/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L112
https://github.com/apache/spark/blob/322b3d06d3f86c23bbe14d3514c1ffd731c216bb/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L120

[...]

PR #1235, opened 2026-01-21
feat: implement deterministic_order sorting for randomSplit

pomykalakyle:
Meant to solve #483

Note that the doctest fails itself when run. This makes it so that when we run the doctest with sail it matches the output that comes from spark. 

- Add sail-spark-random crate with Spark-compatible XORShiftRandom
  - Implements MurmurHash3 seed hashing matching Scala's implementation
  - XORShift algorithm with shifts 21, 35, 4 matching Spark
  - Unit tests verify output matches Spark's XORShiftRandom (values in test came from running scala directly)

- Update Random UDF to use SparkXorShiftRandom instead of ChaCha8Rng
  - Ensures sampling operations produce same results as Spark

- Fix SortWithinPartitionsNode for DataFusion optimizer compatibility (this was causing test to fail becaus it errored out)
  - expressions() now returns actual sort expressions (was returning empty vec)
  - with_exprs_and_inputs() properly rebuilds with transformed expressions

- Implement deterministic_order logic in sample.rs (required to run the test consistently)
  - Sorts by all orderable, non-hidden columns ascending with nulls first
  - Uses resolve_sort_orders for proper column name resolution

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6090,6 +6090,7 @@ dependencies = [
  "sail-common-datafusion",
  "sail-data-source",
  "serde",
+ "serde_arrow",
  "serde_json",
  "url",
  "uuid",
--- a/crates/sail-iceberg/Cargo.toml
+++ b/crates/sail-iceberg/Cargo.toml
@@ -40,6 +40,7 @@ log = { workspace = true }
 # itertools = { workspace = true }
 # percent-encoding = { workspace = true }
 once_cell = { workspace = true }
+serde_arrow = { workspace = true }
 ordered-float = { workspace = true }
 apache-avro = { workspace = true }
 # num-bigint = { workspace = true }
--- a/crates/sail-iceberg/src/physical_plan/action_schema.rs
+++ b/crates/sail-iceberg/src/physical_plan/action_schema.rs
@@ -1,0 +1,478 @@
+use std::collections::BTreeMap;
+use std::sync::{Arc, LazyLock};
+
+use datafusion::arrow::datatypes::{DataType, Field, FieldRef, Schema, SchemaRef};
+use datafusion::arrow::record_batch::RecordBatch;
+use datafusion_common::{DataFusionError, Result};
+use serde::{Deserialize, Serialize};
+
+use crate::spec::types::values::{Literal, PrimitiveLiteral};
+use crate::spec::{
+    DataContentType, DataFile, DataFileFormat, Operation, PartitionSpec, Schema as IcebergSchema,
+    TableRequirement,
+};
+
+pub const COL_ACTION: &str = "action";
+
+static ACTION_FIELDS: LazyLock<Vec<FieldRef>> = LazyLock::new(|| {
+    #[expect(
+        clippy::unwrap_used,
+        reason = "ACTION_FIELDS is a process-global constant."
+    )]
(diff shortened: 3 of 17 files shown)


pomykalakyle:
It looks like there are some CI issues with hatch. Still making the pr public to get a code review in the meanwhile.

linhr:
> It looks like there are some CI issues with hatch.

These seem intermittent installation errors we've seen before. I triggered a rerun.

crates/sail-logical-plan/src/sort.rs
```diff
@@ -63,10 +64,21 @@ impl UserDefinedLogicalNodeCore for SortWithinPartitionsNode {
     }
 
     fn with_exprs_and_inputs(&self, exprs: Vec<Expr>, inputs: Vec<LogicalPlan>) -> Result<Self> {
-        exprs.zero()?;
+        // Rebuild sort_expr using new expressions but keeping same asc/nulls_first
```
linhr, line 76: We may need to return `Err` if the length of `exprs` and `self.sort_expr` do not match.

linhr:
> I triggered a rerun.

Reran still failed. So there may be some issues with the upstream. Let me try later today or tomorrow.

crates/sail-logical-plan/src/sort.rs
```diff
@@ -45,7 +45,8 @@ impl UserDefinedLogicalNodeCore for SortWithinPartitionsNode {
     }
 
     fn expressions(&self) -> Vec<Expr> {
-        vec![]
+        // Return the inner expression from each Sort so the optimizer can transform them
+        self.sort_expr.iter().map(|s| s.expr.clone()).collect()
```
linhr, line 49: Thanks for fixing the issue with `SortWithinPartitionsNode`! This is indeed an important one we overlooked when implementing this logical node a long time ago.

crates/sail-spark-random/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Spark-compatible random number generation.
```
linhr: Usually the utilities can be put directly inside `sail-function`. But I'm fine if we decide to have a dedicated crate in case we will expand it later with more functionalities.

I wonder if we can make this crate a bit more general though. Maybe we name it `sail-random` so that all sampling utilities and randomized algorithms (not necessarily specific to Spark) can be put here in the future?

pomykalakyle: Now that I think about it, I agree it should go in sail function.

crates/sail-logical-plan/src/sort.rs
pomykalakyle: done

pomykalakyle:
interesting - these failures look unrelated to my changes... will check in a sec

pomykalakyle:
> interesting - these failures look unrelated to my changes... will check in a sec

https://github.com/lakehq/sail/pull/1237

yea it's unrelated - seems like we are getting more ci/cd problems @linhr

linhr:
I'm looking into it now.

It seems the intermittent Hatch installation error is gone but we got new errors. It's possibly due to the new Pandas 3.0.0 version released today.

<https://pypi.org/project/pandas/#history>

pomykalakyle:
Thank you @linhr !

Issue #742, opened 2025-08-11
create table as select don't seems to be supported


djouallah:
something like this will be really nice

```
spark.sql("""
CREATE TABLE xx
USING DELTA
LOCATION 'abfss://xxxx@onelake.dfs.fabric.microsoft.com/rrrr.Lakehouse/Tables/dbo/xxx'
AS SELECT 43
""")
```

currently you need first to create a table with a schema then do insert

linhr:
Yeah this is not supported but it's in our plan! We recently revamped the internals of catalog and write operations, so this feature likely won't be complex for us to implement. Stay tuned!

mwc360:
FYI that this isn't clear in the docs: https://docs.lakesail.com/sail/latest/guide/sql/features.html

Also, other obvious commands like MERGE aren't listed.

linhr:
> FYI that this isn't clear in the docs: https://docs.lakesail.com/sail/latest/guide/sql/features.html

> Also, other obvious commands like MERGE aren't listed.

Nice catch! I've updated the docs for the unsupported SQL features in #788.

PR #1236, opened 2026-01-21
feat: support create as select

fafacao86:
fix:https://github.com/lakehq/sail/issues/742
I think we can easily support this using the resolve_write_with_builder. Feel free to correct me.

### DIFF
--- a/crates/sail-plan/src/resolver/command/catalog/table.rs
+++ b/crates/sail-plan/src/resolver/command/catalog/table.rs
@@ -5,6 +5,7 @@ use sail_common::spec;
 use sail_common_datafusion::catalog::{
     CatalogTableBucketBy, CatalogTableConstraint, CatalogTableSort,
 };
+use sail_common_datafusion::rename::logical_plan::rename_logical_plan;
 use sail_common_datafusion::utils::items::ItemTaker;
 use uuid::Uuid;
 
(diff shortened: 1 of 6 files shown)


crates/sail-plan/src/resolver/command/catalog/table.rs
```diff
@@ -76,12 +77,32 @@ impl PlanResolver<'_> {
 
     pub(in super::super) async fn resolve_catalog_create_table_as_select(
         &self,
-        _table: spec::ObjectName,
-        _definition: spec::TableDefinition,
-        _query: spec::QueryPlan,
-        _state: &mut PlanResolverState,
+        table: spec::ObjectName,
+        definition: spec::TableDefinition,
```
linhr, line 81: We need to handle all the fields in table definition. This is what was done similarly in `resolve_catalog_create_table`:

```rust
        let spec::TableDefinition {
            columns,
            comment,
            constraints,
            location,
            file_format,
            row_format,
            partition_by,
            sort_by,
            bucket_by,
            cluster_by,
            if_not_exists,
            replace,
            options,
            properties,
        } = definition;
```

Then you resolve the information and set corresponding fields after `WritePlanBuilder::new()` via various `with_` methods. You can refer to `resolve_command_write_to` for an example. For unsupported fields, we need to return `Err`.

The tests so far are working since they uses the default `parquet` format. You could try a different format (e.g. `csv`) and partitioned tables and see if the data is written as expected.

python/pysail/tests/spark/test_write_table.py
```diff
@@ -229,3 +229,262 @@ def test_write_with_partition_columns(spark, tmp_path):
...
+        assert actual.iloc[i]["is_active"] == expected["is_active"]
+        assert str(actual.iloc[i]["birth_date"]) == expected["birth_date"]
+        assert actual.iloc[i]["extra_str"] == expected["extra_str"]
+        assert actual.iloc[i]["extra_int"] == expected["extra_int"]
+        assert abs(float(actual.iloc[i]["extra_double"]) - expected["extra_double"]) < allowed_error
+        assert actual.iloc[i]["extra_bool"] == expected["extra_bool"]
+
+    # Cleanup
+    spark.sql("DROP TABLE source_table")
+    spark.sql("DROP TABLE ctas_table")
```
linhr: The cleanup is usually part of `try: ... finally: ...` so that the statements are always run even there are failures. And we can have `DROP TABLE IF EXISTS ...` in case the table creation itself fails.

But if we use BDD testing, we have a step named `final statement` for such cleanup statements. You can refer to the codebase for examples.

python/pysail/tests/spark/test_write_table.py
```diff
@@ -229,3 +229,262 @@ def test_write_with_partition_columns(spark, tmp_path):
     assert spark.read.parquet(location).select("a", "b").collect() == [
         ("42", "foo"),
     ]
+
+
+def test_create_table_as_select(spark, tmp_path):
```
linhr: Nice tests!

We are migrating SQL tests to BDD styles. You could consider writing the tests as `.feature` files. You can refer to the codebase for examples. This is not a blocker for this PR, so feel free to decide whether you'd like to tackle this now!

linhr (commented):
Thanks for helping with this! Yeah I think the general direction of using `WritePlanBuilder` is correct.

I added a comment regarding more complete handling for `spec::TableDefinition`. Once this is addressed, I feel the PR can be merged.

crates/sail-plan/src/resolver/command/catalog/table.rs
fafacao86: Got it. I've tried. It's easy to set like partition_by, sort_by, bucket_by etc. But I find it hard to write test for these, I couldn't find a proper way to write assert check in tests. I saw SHOW TBLPROPERTIES, DESCRIBE, these kind of commands are in construction.

fafacao86: So in this PR, I might just handle file_format, if_not_exists, options? For other fields just throw todo Error. If columns are present, throw invalid 'Schema may not be specified in a Create Table As Select (CTAS) statement.' Error. Or I can change this PR status to draft util we support proper check commands.

python/pysail/tests/spark/test_write_table.py
fafacao86: I tried BDD, it's cool. AI is indeed good at wrting these feature files.

crates/sail-plan/src/resolver/command/catalog/table.rs
linhr: > I might just handle file_format, if_not_exists, options? For other fields just throw todo Error. If columns are present, throw invalid 'Schema may not be specified in a Create Table As Select (CTAS) statement.' Error.

Yeah this sounds good to me! We can merge this PR with TODOs for fields that are not yet ready to support.

I feel `partition_by` and `location` might be useful, but that probably requires some changes to `WritePlanBuilder` so this can be out of scope for this PR. (Once we're able to write the table to a particular location, we can use the `file tree` step to verify directory layout in BDD tests.)

We could also accept `columns` and verify if it's consistent with the input query schema, but this can also be future work.

fafacao86: Got it. But for `columns` I tried Spark, it does not allow specifying columns while doing CTAS, it will throw 'Schema may not be specified in a Create Table As Select (CTAS) statement.' error message. I think we need to be sync with Spark's semantics here.

linhr: > But for columns I tried Spark, it does not allow specifying columns while doing CTAS, it will throw 'Schema may not be specified in a Create Table As Select (CTAS) statement.' error message. I think we need to be sync with Spark's semantics here.

I see! That makes sense and yeah we can follow the same semantics here.

python/pysail/tests/spark/steps/sql.py
```diff
@@ -56,7 +56,7 @@ def statement(template, docstring, spark, variables):
     spark.sql(s)
 
 
-@given(parsers.re("statement(?P<template>( template)?) with error {error}"))
```
fafacao86, line 59: Couldn't find any other feature files which used this function, as a reference. And it doesn't match my error message, so I changed this.

crates/sail-plan/src/resolver/command/catalog/table.rs
fafacao86: > I feel `partition_by` and `location` might be useful, but that probably requires some changes to `WritePlanBuilder` so this can be out of scope for this PR. (Once we're able to write the table to a particular location, we can use the `file tree` step to verify directory layout in BDD tests.)

Location can be set via options, I think.

python/pysail/tests/spark/write_table/features/ctas_error_handling.feature
```diff
@@ -0,0 +1,290 @@
...
+      | 1  | Alice |
+      | 2  | Bob   |
+    Given final statement
+      """
+      DROP TABLE IF EXISTS source_table_exists
+      """
+    Given final statement
+      """
+      DROP TABLE IF EXISTS ctas_exists_table
+      """
```
linhr: It's probably a bit counter-intuitive, but `final statement` needs to be defined *before* `statement` so that the cleanup is executed when the statement fails. Otherwise the error in `statement` will fail the test without registering final statements.

I create an issue to validate this automatically: #1246.

python/pysail/tests/spark/write_table/features/ctas_error_handling.feature
```diff
@@ -0,0 +1,290 @@
...
+      (1, 'Alice'),
+      (2, 'Bob')
+      """
+    Given statement template with error COMMENT in CREATE TABLE AS SELECT statement
+      """
+      CREATE TABLE ctas_error_comment
+      USING PARQUET
+      LOCATION {{ location.sql }}
+      COMMENT 'This is a comment'
+      AS SELECT id, name FROM source_table_error_comment
```
linhr: nitpick: I feel some of the tests can be simplified by `SELECT` from `VALUES` directly instead of `SELECT` from a source table. This makes the test shorter and faster to execute (less IO since the source does not need to be materialized on disk).

crates/sail-plan/src/resolver/command/catalog/table.rs
```diff
@@ -76,12 +77,102 @@ impl PlanResolver<'_> {
...
+        query: spec::QueryPlan,
+        state: &mut PlanResolverState,
     ) -> PlanResult<LogicalPlan> {
-        Err(PlanError::todo("CREATE TABLE ... AS SELECT ..."))
+        use super::super::write::{WriteMode, WritePlanBuilder, WriteTableAction, WriteTarget};
+        let spec::TableDefinition {
+            columns,
+            comment,
+            constraints: _,
```
linhr: Should we return an error if `constraints` is non-empty as well?

python/pysail/tests/spark/steps/sql.py
linhr: Nice catch!

linhr:
@fafacao86 Is there a reason for closing this PR? I feel this PR is almost ready for merge. If there is anything I can help with, let me know!

fafacao86:
> @fafacao86 Is there a reason for closing this PR? I feel this PR is almost ready for merge. If there is anything I can help with, let me know!

Oops, I mistakenly deleted my remote branch. I'll amend the code as suggested later, and let's get this merge.

fafacao86:
@linhr Code amended.

PR #1239, opened 2026-01-22
chore: improve env mangement for BDD testing

lonless9:


lonless9:
System catalog should help better manage test scenarios, which is clearly a better approach.

PR #1242, opened 2026-01-23
fix: ensure task completion upon closed sink

linhr:
This pull request fixes the `test_rate_with_filtering` tests that hangs in cluster mode previously. The issue is that we need to stop the execution for streaming nodes when all the stream receivers have closed.

### DIFF
--- a/crates/sail-execution/src/driver/output.rs
+++ b/crates/sail-execution/src/driver/output.rs
@@ -101,6 +101,8 @@ pub fn build_job_output(
         // via prior task status updates.
         // Therefore, it is safe to clean up the job. The job can reach a terminal state and all
         // its streams can be removed.
+        // When `tx` is dropped at the end of this async block, the job output stream built around
+        // `rx` will also end, signaling to the consumer that the job has completed.
         // This is how we ensure the data plane event (output stream termination) is consistent
         // with the control plane event (job termination).
         let _ = handle.send(DriverEvent::CleanUpJob { job_id }).await;
--- a/crates/sail-execution/src/plan/shuffle_write.rs
+++ b/crates/sail-execution/src/plan/shuffle_write.rs
@@ -183,12 +183,14 @@ async fn shuffle_write(
             partitions[p] = Some(batch);
             Ok(())
         })?;
-        let mut active = false;
+        let mut active = 0;
         for p in 0..partitions.len() {
             let Some(sink) = partition_sinks[p].as_mut() else {
                 continue;
             };
-            active = true;
+            // We should update the number of active sinks here,
+            // even if the current batch does not have data for this partition.
+            active += 1;
             if let Some(batch) = partitions[p].take() {
                 match sink.write(Ok(batch)).await {
                     TaskStreamSinkState::Ok => {}
@@ -197,14 +199,19 @@ async fn shuffle_write(
                     }
                     TaskStreamSinkState::Closed => {
                         partition_sinks[p] = None;
+                        // This sink is closed when writing this batch,
+                        // so we should not consider it active anymore.
+                        active -= 1;
                     }
                 }
             }
         }
-        if !active {
+        if active == 0 {
             break;
         }
     }
+    // TODO: Ensure the sinks are cleaned up properly when an error causes an early return
+    //   of this function. We need to consider this for sinks that handle remote data.
     let futures = partition_sinks
         .into_iter()
         .filter_map(|s| s.map(|x| x.close()));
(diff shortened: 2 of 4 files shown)


crates/sail-execution/src/driver/output.rs
```diff
@@ -94,6 +94,15 @@ pub fn build_job_output(
...
         }
+        // When we reach here, the job output stream has ended, which means the job output manager
+        // has been dropped (otherwise the job output stream would still be active waiting for
+        // new streams to be added). Since the job output manager is dropped, we know that the job
+        // is no longer in the running state, which means all tasks have reached a terminal state
+        // via prior task status updates.
+        // Therefore, it is safe to clean up the job. The job can reach a terminal state and all
+        // its streams can be removed.
+        // This is how we ensure the data plane event (output stream termination) is consistent
+        // with the control plane event (job termination).
```
linhr, line 107: This comment is unrelated to this PR, but I feel it's nice to have here.

crates/sail-execution/src/stream_manager/local.rs
```diff
@@ -70,19 +70,31 @@ struct MemoryStreamReplicaSender {
...
-                    *sender = None;
+                match s.send(batch.clone()).await {
+                    Ok(()) => {
+                        sent += 1;
+                    }
+                    Err(_) => {
+                        // This can happen under normal operation when the receiver no longer needs
+                        // more data (e.g., after a LIMIT operator has received enough rows).
+                        warn!("memory stream replica receiver has been dropped");
+                        *sender = None;
```
lonless9, line 85: If this is the expected behavior, after confirming that the relevant logic is correct, we need to change the log here to debug level. We can add a `todo` for now.

crates/sail-execution/src/plan/shuffle_write.rs
```diff
@@ -166,25 +166,48 @@ async fn shuffle_write(
...
             .map(|location| writer.open(location, schema.clone()));
-        try_join_all(futures).await?
+        try_join_all(futures)
+            .await?
+            .into_iter()
+            .map(Some)
+            .collect::<Vec<_>>()
     };
     while let Some(batch) = stream.next().await {
         let batch = batch?;
```
lonless9, line 180: If an error occurs here, will the sink be dropped without calling close?

crates/sail-execution/src/plan/shuffle_write.rs
```diff
@@ -166,25 +166,48 @@ async fn shuffle_write(
...
+                    TaskStreamSinkState::Closed => {
+                        partition_sinks[p] = None;
+                    }
+                }
             }
         }
+        if !active {
+            break;
+        }
     }
```
lonless9, line 212: Using loop may reduce one final batch fetch at the end.

crates/sail-execution/src/driver/output.rs
```diff
@@ -94,6 +94,15 @@ pub fn build_job_output(
                 break;
             }
         }
+        // When we reach here, the job output stream has ended, which means the job output manager
+        // has been dropped (otherwise the job output stream would still be active waiting for
```
lonless9, line 98: I think it would be helpful to add comments for when the consumer disconnects.

crates/sail-execution/src/stream_manager/local.rs
linhr: Good point! Let me address this in the current PR.

crates/sail-execution/src/plan/shuffle_write.rs
linhr: Nice catch! Let me think about the right behavior we should have on error.

lonless9: I forgot to mention that the iteration over `partition_sinks` and the usage of `try_join_all` below might also need to be taken into account. Perhaps we need to collect errors or perform some conversion to warnings.

crates/sail-execution/src/plan/shuffle_write.rs
linhr: Really nice observation! So when the result matches `TaskStreamSinkState::Closed` I have `partition_sinks[p] = None;` but the `active` flag is not updated, so it takes another fetch to discover the fact that all the sinks are really closed.

Let me change `active` to a counter so it can be updated again after setting the sink to `None`. This way we avoid an extra fetch and also avoid an extra `partition_sinks.iter()` to examine the sinks again after the `for` loop.

crates/sail-execution/src/plan/shuffle_write.rs
linhr: I realized that we may want a separate `abort()` method instead of `close()` if the write encounters any error. Since we don't have non-trivial cleanup logic for memory sinks to inspire the design for this yet, I decided to defer this to the future when we support remote sinks. I added a TODO in the code for this.

But you made a good point: all the errors in this function should be handled consistently regarding sink cleanup.

Issue #473, opened 2025-05-24
PySpark DataFrame Doctest: colRegex


shehabgamin:
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.colRegex`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.colRegex.html

PR #1243, opened 2026-01-23
feat: add `colRegex`

davidlghellin:
part #473

### DIFF
--- a/python/pysail/tests/spark/function/test_col_regex.py
+++ b/python/pysail/tests/spark/function/test_col_regex.py
@@ -1,7 +1,5 @@
 """Tests for colRegex column selection functionality."""
 
-import pytest
-
 
 class TestColRegex:
     """Tests for DataFrame.colRegex() method."""


crates/sail-plan/src/resolver/expression/misc.rs
```diff
@@ -91,12 +91,73 @@ impl PlanResolver<'_> {
...
-        _plan_id: Option<i64>,
-        _schema: &DFSchemaRef,
-        _state: &mut PlanResolverState,
+        col_name: String,
+        plan_id: Option<i64>,
+        schema: &DFSchemaRef,
+        state: &mut PlanResolverState,
     ) -> PlanResult<NamedExpr> {
-        Err(PlanError::todo("unresolved regex"))
+        use fancy_regex::Regex;
```
linhr: I wonder if we should just use `regex::Regex` to avoid the extra dependency. I feel `colRegex` is typically used for simple regex filtering so we don't need to support advanced features such as lookaround and backtracking.

python/pysail/tests/spark/function/features/col_regex.feature
```diff
@@ -0,0 +1,70 @@
+Feature: Column selection using colRegex in PySpark
```
linhr: It's nice to see the exploration of using BDD tests for DataFrame APIs. But I wonder if we should consider using doctests for this situation instead, since it's challenging to make the steps general enough for all the future scenarios related to DataFrame APIs (e.g., regarding data type handling for the input/output etc.)

So I feel we can keep it for this PR (so that we keep this example in the commit history for future reference), but we may want to rewrite the tests as Python tests or doctests.

For BDD testing, so far we found it works best for SQL queries since the setup can be highly generalized.

python/pysail/tests/spark/function/features/col_regex.feature
```diff
@@ -0,0 +1,70 @@
...
+  As a developer
+  I want to select columns from a DataFrame using a regular expression
+  So that I can work with dynamic schemas flexibly
```
linhr, line 4: It would also be great to test the error case when no column matches.

python/pysail/tests/spark/function/features/col_regex.feature
linhr: For choosing between BDD tests and doctests, I feel we can ask these questions: (1) Can BDD tests be written without adding new steps? (2) For the steps added, are they applicable for the current tests only, or can they be reused for future tests?

python/pysail/tests/spark/function/features/col_regex.feature
davidlghellin: added

python/pysail/tests/spark/function/features/col_regex.feature
davidlghellin: I've updated the tests based on your feedback:

- Migrated the specific colRegex tests to regular Python tests (test_col_regex.py)
- Kept a generic step for DataFrame creation that can be reused in future BDD tests
- Added a simple BDD example showing basic usage of colRegex in col_regex.feature

This way we have both: comprehensive Python tests for colRegex and a reference example for how to use BDD steps with DataFrame APIs.

crates/sail-plan/src/resolver/expression/misc.rs
davidlghellin: Switched to regex::Regex. Updated tests to use standard regex patterns since we no longer support advanced features like possessive quantifiers.

python/pysail/tests/spark/function/test_col_regex.py
```diff
@@ -0,0 +1,100 @@
+"""Tests for colRegex column selection functionality."""
+
+import pytest
+
+
+@pytest.fixture
+def spark_session(spark):
+    """Use the shared spark fixture."""
+    return spark
```
linhr: I don't think we need this.

crates/sail-plan/src/resolver/expression/misc.rs
linhr: @davidlghellin I don't see the original tests using any advanced regex features such as possessive quantifiers. Let me know if I'm wrong.

python/pysail/tests/spark/function/features/col_regex.feature
linhr: Ultimately we would like to go one way, since have both BDD tests and Python tests for DataFrame APIs may confuse future contributors.

But there is no need to make further changes in this PR. As I mentioned, the code could be valuable reference in the commit history.

crates/sail-plan/src/resolver/expression/misc.rs
davidlghellin: https://github.com/lakehq/sail/pull/1243/changes/fbf6c9595650821be93738a2aa731d51819b695a#diff-a6374027e5f592667bdff49a9c2477e0a7017c0003ac953b7e509fc048843ddcL12

linhr: I see. Thanks for pointing out!

PR #1244, opened 2026-01-24
chore: update deps, fix lint errors, and remove usage of unwrap and expect

shehabgamin:


crates/sail-delta-lake/src/operations/write/stats.rs
```diff
@@ -261,8 +261,8 @@ impl StatsScalar {
         match (stats, logical_type) {
             (Statistics::Boolean(v), _) => Ok(Self::Boolean(get_stat!(v))),
             (Statistics::Int32(v), Some(LogicalType::Date)) => {
-                #[allow(clippy::unwrap_used)]
-                let epoch_start = chrono::NaiveDate::from_ymd_opt(1970, 1, 1).unwrap();
+                let epoch_start = chrono::NaiveDate::from_ymd_opt(1970, 1, 1)
+                    .ok_or_else(|| DeltaTableError::generic("Failed to create epoch date"))?;
```
linhr: Creating date from this constant should never fail.

crates/sail-function/src/scalar/math/spark_unhex.rs
```diff
@@ -72,33 +72,25 @@ fn hex_int64(num: i64) -> String {
...
-            #[allow(clippy::unwrap_used)]
-            write!(&mut s, "{b:02x}").unwrap();
+            write!(&mut s, "{b:02x}")?;
         }
     } else {
         for b in data.as_ref() {
             // Writing to a string never errors, so we can unwrap here.
-            #[allow(clippy::unwrap_used)]
-            write!(&mut s, "{b:02X}").unwrap();
+            write!(&mut s, "{b:02X}")?;
```
linhr: The comment should be removed:

```rust
// Writing to a string never errors, so we can unwrap here.
```

crates/sail-iceberg/src/spec/types/values.rs
```diff
@@ -353,35 +353,33 @@ impl Literal {
 
         use crate::spec::types::{PrimitiveType, Type};
 
-        fn days_to_date_str(days: i32) -> String {
-            #[allow(clippy::unwrap_used)]
-            let epoch = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap();
+        fn days_to_date_str(days: i32) -> Result<String, String> {
+            let epoch =
+                NaiveDate::from_ymd_opt(1970, 1, 1).ok_or_else(|| "Bad epoch".to_string())?;
```
linhr: Creating date from this constant should never fail.

crates/sail-iceberg/src/spec/types/values.rs
```diff
@@ -353,35 +353,33 @@ impl Literal {
...
-        fn micros_to_time_str(us: i64) -> String {
+        fn micros_to_time_str(us: i64) -> Result<String, String> {
             let secs = us.div_euclid(1_000_000);
-            let rem = (us.rem_euclid(1_000_000)) as u32;
-            #[allow(clippy::unwrap_used)]
+            let rem = us.rem_euclid(1_000_000) as u32;
             let t = NaiveTime::from_num_seconds_from_midnight_opt(secs as u32, rem * 1000)
-                .unwrap_or(NaiveTime::from_hms_opt(0, 0, 0).unwrap());
-            t.format("%H:%M:%S%.f").to_string()
+                .unwrap_or(NaiveTime::from_hms_opt(0, 0, 0).ok_or_else(|| "Bad time".to_string())?);
```
linhr: Creating time from this constant should never fail.

crates/sail-iceberg/src/spec/types/values.rs
```diff
@@ -353,35 +353,33 @@ impl Literal {
...
             let secs = us.div_euclid(1_000_000);
-            let rem = (us.rem_euclid(1_000_000)) as u32;
-            #[allow(clippy::unwrap_used)]
+            let rem = us.rem_euclid(1_000_000) as u32;
             let base = NaiveDate::from_ymd_opt(1970, 1, 1)
-                .unwrap()
+                .ok_or_else(|| "Bad epoch".to_string())?
                 .and_hms_nano_opt(0, 0, 0, 0)
-                .unwrap();
+                .ok_or_else(|| "Bad epoch".to_string())?;
```
linhr: Creating date and time from these constants should never fail.

crates/sail-physical-optimizer/src/lib.rs
```diff
@@ -32,21 +32,25 @@ pub struct PhysicalOptimizerOptions {
...
         .iter()
         .find(|rule| rule.name() == "LimitPushPastWindows")
         .cloned()
-        .unwrap()
+        .ok_or_else(|| {
+            datafusion::common::DataFusionError::Internal(
+                "LimitPushPastWindows optimizer rule not found".to_string(),
+            )
+        })
 }
```
linhr, line 46: This is a situation where I feel `unwrap()` would make it easier to detect breaking changes upon DataFusion upgrade. A missing optimizer is catastrophic and should fail the entire program immediately. So I feel deferring the error to session creation is less ideal.

shehabgamin: @linhr Maybe `expect()` instead?

crates/sail-iceberg/src/spec/types/values.rs
shehabgamin: Right, I think it's better not to use `unwrap` here, regardless. Otherwise, other contributors might think it's acceptable to use `unwrap` throughout the codebase if they're just following our coding style. Thoughts?

crates/sail-iceberg/src/spec/types/values.rs
shehabgamin: https://github.com/lakehq/sail/pull/1244#discussion_r2723771696

crates/sail-iceberg/src/spec/types/values.rs
shehabgamin: https://github.com/lakehq/sail/pull/1244#discussion_r2723771696

crates/sail-delta-lake/src/operations/write/stats.rs
shehabgamin: https://github.com/lakehq/sail/pull/1244#discussion_r2723771696

crates/sail-physical-optimizer/src/lib.rs
linhr: Yeah `expect()` sounds a better way than `unwrap()` to convey the assumption in the code.

crates/sail-iceberg/src/spec/types/values.rs
linhr: I feel in our review we've never accepted code when `unwrap` is used for fallible code. But you made a good point for setting code style examples for others to follow.

We should make it clear for contributors that `#[expect(clippy::unwrap_used)]` should always come with a comment explaining the assumption.

For this particular situation for constant initialization, maybe `.expect()` achieves a good balance.

PR #1247, opened 2026-01-26
fix: job cleanup

linhr:
It was previously discovered (<https://github.com/lakehq/sail/pull/1163#issuecomment-3696050624>) that infinite streaming jobs cannot be stopped properly in cluster mode. This PR fixes this issue and improves job cleanup logic in general.

### DIFF
--- a/crates/sail-execution/src/driver/job_scheduler/core.rs
+++ b/crates/sail-execution/src/driver/job_scheduler/core.rs
@@ -550,10 +550,11 @@ impl JobScheduler {
                 // For running jobs, the job output is dropped here.
                 // Internally, the job output manages the receiving end of the output stream.
                 // So once the stream receiver is no longer available, all the running tasks
-                // will ultimately find that the sink is closed (even if the task produces
-                // infinite stream) and the task will stop running.
-                // Once the tasks are stopped, the worker gRPC server will have no active clients
-                // subscribing to local streams, so the server can proceed with shutdown.
+                // will ultimately find that the sink (owned by the shuffle write node) is closed.
+                // The task will stop running even if the task produces an infinite stream.
+                // Once the tasks are stopped, the inputs (gRPC streaming responses owned by
+                // shuffle read nodes) are dropped. So the worker gRPC server will have no active
+                // clients subscribing to local streams, and the server can proceed with shutdown.
                 job.state = JobState::Canceled;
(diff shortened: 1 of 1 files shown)


crates/sail-execution/src/driver/output.rs
```diff
@@ -85,11 +85,18 @@ pub fn build_job_output(
...
     let handle = ctx.handle().clone();
     ctx.spawn(async move {
-        let mut stream = stream;
-        while let Some(batch) = stream.next().await {
+        let mut stream = JobOutputStream::new(receiver);
+        loop {
+            let next = tokio::select! {
+                biased;
+                x = stream.next() => x,
+                _ = tx.closed() => break,
```
linhr, line 95: This is the key of this fix.

crates/sail-execution/src/driver/job_scheduler/core.rs
```diff
@@ -544,6 +544,37 @@ impl JobScheduler {
...
     }
 
+    pub fn stop(&mut self) {
+        for (_, job) in self.jobs.iter_mut() {
+            if matches!(job.state, JobState::Running { .. } | JobState::Draining) {
+                // For running jobs, the job output is dropped here.
+                // Internally, the job output manages the receiving end of the output stream.
+                // So once the stream receiver is no longer available, all the running tasks
+                // will ultimately find that the sink (owned by the shuffle write node) is closed.
+                // The task will stop running even if the task produces an infinite stream.
```
lonless9, line 554: With the fix #1242 we can ensure this. 👍

crates/sail-execution/src/driver/output.rs
lonless9: I think as the complexity of our execution crate increases, we need to consider designing appropriate unit tests/integration tests to verify scenarios such as race conditions or client disconnections mid-process.

linhr: > I think as the complexity of our execution crate increases, we need to consider designing appropriate unit tests/integration tests to verify scenarios such as race conditions or client disconnections mid-process.

Absolutely! The tests around concurrent behavior are important to have. I feel we could also consider tools such as this one to discover concurrency bugs: <https://github.com/tokio-rs/loom>. This is going to be an interesting but challenging task.

PR #1256, opened 2026-01-27
fix: return Null for skewness/kurtosis with fewer than 2 values

davidlghellin:


crates/sail-function/src/aggregate/skewness.rs
```diff
@@ -161,14 +161,17 @@ impl SkewnessAccumulator {
...
     }
 
+    /// Returns NULL when fewer than 2 values have been accumulated,
+    /// matching Spark's behavior for skewness and kurtosis.
+    /// With n=1, m2 (variance) is zero, causing division by zero in the
+    /// skewness formula `sqrt(n) * m3 / m2^(3/2)`, which produces NaN.
+    /// Spark returns NULL for n < 2 and computes from n >= 2 onward.
     pub fn null_or_value(&self, value: f64) -> Result<ScalarValue> {
-        Ok(ScalarValue::Float64(
-            if (self.n() == 0.0) || (self.avg() == 0.0) {
```
davidlghellin, line 166: mean(-10, 10) => null

linhr (approved):
Nice fix! Thanks!

PR #1260, opened 2026-01-27
fix: `count_if` window function missing filter condition

### LINKED ISSUE #1259: [bug] count_if window function counts all rows instead of filtering by condition
`count_if` used as a window function counts all rows instead of only counting rows where the condition is true.

## Reproduction

```python
spark.sql("""SELECT value,
        count_if(value > 2) OVER (ORDER BY value ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS result
FROM VALUES (1), (2), (3), (4), (5) AS t(value)
ORDER BY value""").show()
```
pyspark
Image

sail:

Image

davidlghellin:
closes #1259

### DIFF
--- a/crates/sail-plan/src/function/window.rs
+++ b/crates/sail-plan/src/function/window.rs
@@ -274,7 +274,7 @@ fn count_if(input: WinFunctionInput) -> PlanResult<expr::Expr> {
             Ok(expr::Expr::WindowFunction(Box::new(expr::WindowFunction {
                 fun: WindowFunctionDefinition::AggregateUDF(count::count_udaf()),
                 params: WindowFunctionParams {
-                    args: arguments,
+                    args: vec![lit(0)],
                     partition_by,
                     order_by,
                     window_frame,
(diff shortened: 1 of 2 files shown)


python/pysail/tests/spark/function/features/count_if.feature
```diff
@@ -0,0 +1,100 @@
...
+
+    Scenario: count_if with string condition
+      When query
+      """
+      SELECT count_if(color = 'red') AS result
+      FROM VALUES ('red'), ('blue'), ('red'), ('green') AS t(color)
+      """
+      Then query result
+      | result |
+      | 2      |
```
davidlghellin: clean

crates/sail-plan/src/function/window.rs
```diff
@@ -266,18 +266,24 @@ fn count_if(input: WinFunctionInput) -> PlanResult<expr::Expr> {
...
-        }))),
+        1 => {
+            let filter = arguments
+                .first()
+                .ok_or_else(|| PlanError::invalid("`count_if` requires 1 argument"))?
+                .clone();
+            Ok(expr::Expr::WindowFunction(Box::new(expr::WindowFunction {
+                fun: WindowFunctionDefinition::AggregateUDF(count::count_udaf()),
+                params: WindowFunctionParams {
+                    args: arguments,
```
linhr: I wonder if we can simply pass `lit(0)` as a single argument here to avoid unnecessary expression evaluation.

linhr (approved):

PR #1263, opened 2026-01-28
feat: add soundex and randstr string functions

pomykalakyle:
## Summary

Implements the `soundex` and `randstr` scalar functions for Spark SQL compatibility.

- `soundex`: Phonetic algorithm that encodes strings by sound (e.g., "Robert" -> "R163")
- `randstr`: Generates random alphanumeric strings with optional seed for deterministic output

Part of #508

### DIFF
--- a/crates/sail-function/src/scalar/string/randstr.rs
+++ b/crates/sail-function/src/scalar/string/randstr.rs
@@ -145,7 +145,7 @@ impl ScalarUDFImpl for Randstr {
 fn generate_random_string(rng: &mut SparkXorShiftRandom, length: usize) -> String {
     let mut result = String::with_capacity(length);
     for _ in 0..length {
-        let v = (rng.next_int().abs() % 62) as u8;
+        let v = ((rng.next_int() as u32) % 62) as u8;
         let ch = if v < 10 {
             b'0' + v
         } else if v < 36 {
--- a/crates/sail-function/src/scalar/string/soundex.rs
+++ b/crates/sail-function/src/scalar/string/soundex.rs
@@ -62,16 +62,24 @@ impl ScalarUDFImpl for Soundex {
     }
 }
 
-/// Maps a character to its Soundex digit code (1-6), or None for vowels/H/W/Y.
-fn soundex_code(c: char) -> Option<char> {
+/// Represents character categories in Soundex algorithm.
+enum SoundexChar {
+    Code(char),
+    Separator,
+    Ignored,
+}
+
+/// Classifies a character for Soundex processing.
+fn classify_char(c: char) -> SoundexChar {
     match c.to_ascii_uppercase() {
-        'B' | 'F' | 'P' | 'V' => Some('1'),
-        'C' | 'G' | 'J' | 'K' | 'Q' | 'S' | 'X' | 'Z' => Some('2'),
-        'D' | 'T' => Some('3'),
-        'L' => Some('4'),
-        'M' | 'N' => Some('5'),
-        'R' => Some('6'),
-        _ => None,
+        'B' | 'F' | 'P' | 'V' => SoundexChar::Code('1'),
+        'C' | 'G' | 'J' | 'K' | 'Q' | 'S' | 'X' | 'Z' => SoundexChar::Code('2'),
+        'D' | 'T' => SoundexChar::Code('3'),
+        'L' => SoundexChar::Code('4'),
+        'M' | 'N' => SoundexChar::Code('5'),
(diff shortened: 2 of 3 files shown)


crates/sail-function/src/scalar/string/randstr.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        Ok(result)
+    }
+}
+
+/// Generates a random alphanumeric string (0-9, a-z, A-Z) of the specified length.
+fn generate_random_string(rng: &mut SparkXorShiftRandom, length: usize) -> String {
+    let mut result = String::with_capacity(length);
+    for _ in 0..length {
+        let v = (rng.next_int().abs() % 62) as u8;
```
lonless9: I think to avoid overflow we can use `as u32 % 62` here.

crates/sail-function/src/scalar/string/soundex.rs
```diff
@@ -0,0 +1,123 @@
...
+        'L' => Some('4'),
+        'M' | 'N' => Some('5'),
+        'R' => Some('6'),
+        _ => None,
+    }
+}
+
+/// Computes the 4-character Soundex code for a string.
+fn compute_soundex(s: &str) -> String {
+    let mut chars = s.chars().filter(|c| c.is_ascii_alphabetic());
```
lonless9: I think we can distinguish three categories: Digit(code), Separator (vowels/Y), and Ignored (H/W, possibly punctuation), or handle vowels explicitly in the loop by resetting last_code when a separator is encountered.

python/pysail/tests/spark/test_soundex.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+    df.createOrReplaceTempView("test_soundex")
+
+    actual = spark.sql("SELECT soundex(name) AS code FROM test_soundex").toPandas()
+
+    assert actual["code"].iloc[0] == "R163"
+    assert actual["code"].iloc[1] == "R163"
+    assert actual["code"].iloc[2] == "S530"
+    assert actual["code"].iloc[3] == ""
+    assert pd.isna(actual["code"].iloc[4])
```
lonless9: We can add these cases:
`Tymczak` -> `T522`
`Ashcraft` -> `A261`

lonless9: Migrating these tests to BDD style could be handled as follow-up work and does not need to be addressed in this PR.

crates/sail-function/src/scalar/string/randstr.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        match seed {
+            Some(seed_val) => {
+                let mut rng = SparkXorShiftRandom::new(seed_val);
+                let values: Vec<String> = (0..number_rows)
+                    .map(|_| generate_random_string(&mut rng, length))
+                    .collect();
+                let array = StringArray::from(values);
+                Ok(ColumnarValue::Array(Arc::new(array)))
+            }
```
lonless9, line 98: This comment does not need to be addressed in this PR. 
The current merge cardinality check uses `UUID`, which is a heavy implementation. Ideally, `MonotonicallyIncreasingID` should be used. Similar to the issue here, state maintenance should not be `batch-aware` but `partition-aware`. These two issues belong to the same category and should be resolved through the physical execution plan. I attempted to fix this issue in this branch (https://github.com/lakehq/sail/tree/feat-monotonic) but forgot what the blocker was at the time. I will resolve it when I pick it up again (#1264 ).

linhr: Wow really nice catch! I believe all our random functions suffer from the same problem when the RNG is seeded on every batch instead of on every partition.

For example, I constructed this SQL query and we can see the result rows are all the same. (The SQL is kind of strange due to how physical optimizers coalesce and repartition batches internally, and `OR r != r` is added to avoid filter pushdown.)

```sql
SELECT * FROM (SELECT id, randn(42) AS r FROM range(1048576)) WHERE id % 32768 IN (0, 1024, 2048, 3072) OR r != r
```

(@pomykalakyle You can ignore the note regarding merge cardinality check UUID in case you haven't looked at that part of the codebase. The UUID issue is deeply inside the internals in our DML statement implementation. But the phenomenon is similar: batch-aware vs partition-aware function state, as @lonless9 has mentioned.)

linhr: I created an issue to track the work: #1266.

pomykalakyle:
@lonless9 I think the previous commit helps out with that?

PR #1264, opened 2026-01-28
feat: support monotonically_increasing_id

lonless9:
part of #507 
related #1266

### DIFF
--- a/python/pysail/tests/spark/test_monotonic_id.py
+++ b/python/pysail/tests/spark/test_monotonic_id.py
@@ -1,3 +1,6 @@
+import pytest
+
+
 def test_monotonically_increasing_id_smoke(spark):
     rows = spark.sql("SELECT monotonically_increasing_id() AS id").collect()
     assert len(rows) == 1
@@ -35,6 +38,7 @@ def test_monotonically_increasing_id_same_when_called_twice_in_select(spark):
         assert r["id1"] == r["id2"]
 
 
+@pytest.mark.skip(reason="Temporarily skipped until we have a fix for the issue")
 def test_monotonically_increasing_id_in_aggregate_projection(spark):
     rows = spark.sql(
         """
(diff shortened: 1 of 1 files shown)


crates/sail-session/src/session_factory/server.rs
```diff
@@ -195,7 +195,8 @@ impl ServerSessionFactory {
     }
 
     fn apply_execution_config(&mut self, config: &mut SessionConfig) {
-        let execution = &mut config.options_mut().execution;
+        let options = config.options_mut();
+        let execution = &mut options.execution;
```
linhr: I guess this is not needed?

crates/sail-plan/src/resolver/query/project.rs
```diff
@@ -35,6 +36,7 @@ impl PlanResolver<'_> {
         let (input, expr) = self.rewrite_wildcard(input, expr, state)?;
         let (input, expr) = self.rewrite_projection::<ExplodeRewriter>(input, expr, state)?;
         let (input, expr) = self.rewrite_projection::<WindowRewriter>(input, expr, state)?;
+        let (input, expr) = self.rewrite_projection::<MonotonicIdRewriter>(input, expr, state)?;
```
linhr: We might need to apply the same rewrites for aggregation, lateral view, etc.

You can search for all the places where the `ExplodeRewriter` is used and see if we need the `MonotonicIdRewriter` there as well. It might also be good to add a few unit tests for aggregation etc.

crates/sail-logical-plan/src/monotonic_id.rs
```diff
@@ -0,0 +1,107 @@
...
+impl MonotonicIdNode {
+    pub fn try_new(input: Arc<LogicalPlan>, column_name: String) -> Result<Self> {
+        let mut fields = input.schema().as_ref().as_arrow().fields().to_vec();
+        fields.push(Arc::new(Field::new(
+            column_name.clone(),
+            DataType::Int64,
+            false,
+        )));
+        let arrow_schema = ArrowSchema::new(fields);
+        let schema = Arc::new(DFSchema::from_unqualified_fields(
```
linhr: This loses qualifiers of existing fields. We need to construct `DFSchema` in a different way.

crates/sail-logical-plan/src/monotonic_id.rs
```diff
@@ -0,0 +1,107 @@
...
+            column_name: &node.column_name,
+        }
+    }
+}
+
+impl PartialOrd for MonotonicIdNode {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        MonotonicIdNodeOrd::from(self).partial_cmp(&other.into())
+    }
+}
```
linhr: We can now use `educe` to derive these.

crates/sail-physical-plan/src/monotonic_id.rs
```diff
@@ -0,0 +1,224 @@
...
+                "MonotonicIdExec expects Int64 field for {}, got {:?}",
+                column_name,
+                field.data_type()
+            );
+        }
+
+        let properties = PlanProperties::new(
+            EquivalenceProperties::new(schema.clone()),
+            input.output_partitioning().clone(),
+            EmissionType::Both,
```
linhr: I wonder if we should use the emission type of the input.

crates/sail-physical-plan/src/monotonic_id.rs
```diff
@@ -0,0 +1,224 @@
...
+            Poll::Ready(Some(Ok(batch))) => {
+                let mut cols = batch.columns().to_vec();
+                let id_col = match self.make_ids(batch.num_rows()) {
+                    Ok(c) => c,
+                    Err(e) => return Poll::Ready(Some(Err(e))),
+                };
+                if self.col_idx >= cols.len() {
+                    cols.push(id_col);
+                } else {
+                    cols[self.col_idx] = id_col;
```
linhr: I'm not sure if this is correct. It seems to override an existing column rather than inserting an additional column.

crates/sail-session/src/session_factory/server.rs
lonless9: This should be the part I forgot to revert in my previous debugging. 😆 
Done!

crates/sail-plan/src/resolver/query/project.rs
lonless9: Done!

crates/sail-logical-plan/src/monotonic_id.rs
lonless9: Done!

crates/sail-logical-plan/src/monotonic_id.rs
lonless9: Done!

crates/sail-physical-plan/src/monotonic_id.rs
lonless9: Done!

crates/sail-session/src/session_factory/server.rs
lonless9: Besides I previously encountered some strange snapshot instability issues, which might have been caused by column ordering. This phenomenon disappeared after I merged the main branch, possibly related to the recent refactoring of Delta Lake. After I added the implementation of the execution plan trait for statistics, the distributed snapshot also stabilized.

crates/sail-plan/src/resolver/query/aggregate.rs
```diff
@@ -105,6 +107,42 @@ impl PlanResolver<'_> {
         state: &mut PlanResolverState,
     ) -> PlanResult<LogicalPlan> {
         let grouping = self.resolve_grouping_positions(grouping, &projections)?;
+        let mut monotonic_rewriter = MonotonicIdRewriter::new_from_plan(input, state);
```
linhr: I am not sure if we need the `MonotonicIdRewriter` this early to handle projections, `GROUP BY`, and `HAVING` expressions explicitly.

I feel the rewriter can be called after the expression rebase logic in this function (where expressions involved in aggregation are changed to column references for `HAVING` etc.). Could you try if we can just have the one-line change similar to how it is done for `ExplodeRewriter`?

lonless9: I made the following changes in the new commit (if I understand correctly). There might still retain the non-degraded `monotonically_increasing_id()` in the Aggregate.

linhr: Thanks for trying this out! I realized we actually found a bug in aggregation where the `explode` function is not supported either. For example, the following query gives an error:

```python
spark.sql("SELECT id, max(explode(sequence(id))) FROM range(0, 10) GROUP BY id ORDER BY id").show()
```

```
explode should be rewritten during logical plan analysis
```

(Note that JVM Spark does not support such queries though (`UNSUPPORTED_GENERATOR.NESTED_IN_EXPRESSIONS`). But I feel this is a well-defined query that Sail can support.)

So it's the same issue with `monotonically_increasing_id`. Let me see if there is a way to solve this in general.

linhr: @lonless9 Let's keep the one-line solution (for simplicity, even if we know it doesn't work) and mark the test as skipped, so that we can merge this PR.

Your previous solution doesn't work in all cases either. For example, I tried the following and it doesn't work.

```sql
SELECT max(id), monotonically_increasing_id()
FROM range(10)
GROUP BY monotonically_increasing_id()
ORDER BY monotonically_increasing_id()
```

My hope is to figure out an elegant solution so that it works for all rewriters. So it requires some investigation. I created an issue to track the work: #1361.

lonless9: Make sense to me! Let me do the modifications about the test case later.

PR #1273, opened 2026-01-29
feat: support `Dataframe.mergeInto`

lonless9:
Currently only working on Delta Lake. 
pyspark[connect]>=4.0.0
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.mergeInto.html
Document update required. 
part of #171

### DIFF
--- a/crates/sail-spark-connect/src/service/plan_executor.rs
+++ b/crates/sail-spark-connect/src/service/plan_executor.rs
@@ -30,8 +30,7 @@ use crate::spark::connect::{
     CommonInlineUserDefinedFunction, CommonInlineUserDefinedTableFunction,
     CreateDataFrameViewCommand, ExecutePlanResponse, GetResourcesCommand, LocalRelation,
     MergeIntoTableCommand, Relation, SqlCommand, StreamingQueryCommand,
-    StreamingQueryCommandResult,
-    StreamingQueryListenerBusCommand, StreamingQueryManagerCommand,
+    StreamingQueryCommandResult, StreamingQueryListenerBusCommand, StreamingQueryManagerCommand,
     StreamingQueryManagerCommandResult, WriteOperation, WriteOperationV2,
     WriteStreamOperationStart, WriteStreamOperationStartResult,
 };


crates/sail-plan/src/resolver/command/merge.rs
```diff
@@ -18,6 +18,19 @@ use crate::error::{PlanError, PlanResult};
...
+// - if it exists only in source -> treat as source
+// - if it exists in both -> default to target
+//
+// FIXME: We currently disambiguate `plan_id=None` attributes only for MERGE.
+// Similar ambiguity can arise in other multi-input operations (e.g. joins) when using
+// unqualified column names from ExpressionString / col("..."). Consider a shared resolver utility.
+// These are hardcoded values and need to pay attention if other implementations got conflicting values.
+
+const MERGE_TARGET_DEFAULT_PLAN_ID: i64 = i64::MIN + 4242;
+const MERGE_SOURCE_DEFAULT_PLAN_ID: i64 = i64::MIN + 4243;
```
lonless9, line 32: Note here.

crates/sail-plan/src/resolver/command/merge.rs
```diff
@@ -18,6 +18,19 @@ use crate::error::{PlanError, PlanResult};
 use crate::resolver::state::PlanResolverState;
 use crate::resolver::PlanResolver;
 
+// When we receive an unqualified attribute with `plan_id=None`, we need:
+// - if it exists only in target -> treat as target
+// - if it exists only in source -> treat as source
+// - if it exists in both -> default to target
+//
+// FIXME: We currently disambiguate `plan_id=None` attributes only for MERGE.
+// Similar ambiguity can arise in other multi-input operations (e.g. joins) when using
```
linhr, line 27: I guess ambiguity may just be treated as an error in join. So resolving the ambiguity (with target as the default) could be a unique situation for `MERGE`.

It would be interesting to see if we can resolve the ambiguity without assigning static plan ID. I don't have good ideas for this yet. So this is future work.

keen85:
this is fantastic!

PR #1278, opened 2026-01-30
feat: support distributed build-probe join

lonless9:


crates/sail-common/src/config/application.yaml
```diff
@@ -676,6 +676,14 @@
     Whether to enable cost-based join reorder in the query optimizer.
   experimental: true
 
+- key: optimizer.enable_distributed_collect_left
+  type: boolean
+  default: "true"
```
lonless9, line 681: Temporarily set to true for testing convenience, should be adjusted to false later.

PR #1279, opened 2026-01-30
feat: dispatch Glue table creation based on format (Iceberg vs Hive)

pomykalakyle:
## Summary

- Add format-based dispatching for Glue table creation:
  - `format: "iceberg"` → uses OpenTableFormatInput API with support for partition transforms (day, month, year, hour, bucket, truncate)
  - Other formats (parquet, csv, etc.) → uses TableInput API (Hive-style, identity partitions only)
- Split table creation logic into dedicated modules (`iceberg.rs`, `hive.rs`)
- Make Iceberg format version configurable via `format-version` property (defaults to v2)

### DIFF
--- a/crates/sail-catalog-glue/src/data_type.rs
+++ b/crates/sail-catalog-glue/src/data_type.rs
@@ -87,7 +87,10 @@ pub fn arrow_to_iceberg_type(data_type: &DataType) -> CatalogResult<Document> {
         }
         DataType::Float16 | DataType::Float32 => Ok(Document::String("float".to_string())),
         DataType::Float64 => Ok(Document::String("double".to_string())),
-        DataType::Decimal128(precision, scale) | DataType::Decimal256(precision, scale) => {
+        DataType::Decimal32(precision, scale)
+        | DataType::Decimal64(precision, scale)
+        | DataType::Decimal128(precision, scale)
+        | DataType::Decimal256(precision, scale) => {
             Ok(Document::String(format!("decimal({precision},{scale})")))
         }
         DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => {
@@ -194,11 +197,9 @@ pub fn arrow_to_iceberg_type(data_type: &DataType) -> CatalogResult<Document> {
             "Union types are not supported by Iceberg".to_string(),
         )),
         DataType::Dictionary(_, value_type) => arrow_to_iceberg_type(value_type),
-        DataType::RunEndEncoded(_, _) | DataType::Decimal32(_, _) | DataType::Decimal64(_, _) => {
-            Err(CatalogError::NotSupported(format!(
-                "Data type {data_type:?} is not supported by Iceberg"
-            )))
-        }
+        DataType::RunEndEncoded(_, _) => Err(CatalogError::NotSupported(format!(
+            "Data type {data_type:?} is not supported by Iceberg"
+        ))),
     }
 }
 


crates/sail-catalog-glue/src/data_type.rs
```diff
@@ -73,6 +74,134 @@ pub fn arrow_to_glue_type(data_type: &DataType) -> CatalogResult<String> {
...
+            }
+            Err(CatalogError::InvalidArgument(
+                "Map type must have key and value fields".to_string(),
+            ))
+        }
+        DataType::Union(_, _) => Err(CatalogError::NotSupported(
+            "Union types are not supported by Iceberg".to_string(),
+        )),
+        DataType::Dictionary(_, value_type) => arrow_to_iceberg_type(value_type),
+        DataType::RunEndEncoded(_, _) | DataType::Decimal32(_, _) | DataType::Decimal64(_, _) => {
```
shehabgamin: I think we should be able to support `DataType::Decimal32` and `DataType::Decimal64`:
```
        DataType::Decimal32(precision, scale) | DataType::Decimal64(precision, scale) |DataType::Decimal128(precision, scale) | DataType::Decimal256(precision, scale) => {
            Ok(Document::String(format!("decimal({precision},{scale})")))
        }
```

crates/sail-catalog-glue/src/provider.rs
```diff
@@ -435,18 +456,18 @@ impl GlueCatalogProvider {
...
-                "partition transforms are not supported by Glue catalog".to_string(),
-            ));
-        }
+
+        let location = location.ok_or_else(|| {
+            CatalogError::InvalidArgument("Location is required for Iceberg tables".to_string())
+        })?;
+
+        // Note: comment is not used for Iceberg tables through the OpenTableFormatInput API
+        let _ = comment;
```
shehabgamin: Nitpick. This line isn't needed. Can do `comment: _` above like you did with format.

crates/sail-catalog-glue/tests/table_tests.rs
```diff
@@ -540,3 +540,82 @@ async fn test_drop_table() {
...
+
+    match &created_table.kind {
+        sail_common_datafusion::catalog::TableKind::Table {
+            columns,
+            comment,
+            location,
+            ..
+        } => {
+            assert_eq!(columns.len(), 4);
+            assert_eq!(
```
shehabgamin: The below comment was added in `crates/sail-catalog-glue/src/provider.rs`, but this test seems to be passing:
```
// Note: comment is not used for Iceberg tables through the OpenTableFormatInput API
```

crates/sail-catalog-glue/src/provider.rs
```diff
@@ -634,34 +653,50 @@ impl CatalogProvider for GlueCatalogProvider {
...
+            iceberg_table_builder = iceberg_table_builder.set_properties(Some(props));
+        }
+
+        let create_iceberg_table = iceberg_table_builder.build().map_err(|e| {
+            CatalogError::InvalidArgument(format!("Failed to build CreateIcebergTableInput: {e}"))
+        })?;
+
+        let iceberg_input = IcebergInput::builder()
+            .metadata_operation(MetadataOperation::Create)
+            .version("2")
```
shehabgamin: Should this be a const?

crates/sail-catalog-glue/src/provider.rs
```diff
@@ -634,34 +653,50 @@ impl CatalogProvider for GlueCatalogProvider {
...
             columns,
-            comment,
             location,
-            format,
             partition_by,
             if_not_exists,
             properties,
         } = Self::validate_create_table_options(options)?;
 
-        let format_info = GlueStorageFormat::from_format(&format)?;
```
shehabgamin, line 645: There is no more need for `format`? If so, should we maybe add it to `properties` so the information is not lost?

Also, should `comment` be added to `properties` as well?

shehabgamin (commented):
Just have a couple comments/questions.

Also, `OpenTableFormatInput` seems to support only Iceberg, but Glue as a whole supports a variety of file and table formats ([see here](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-format.html)). Is switching to `OpenTableFormatInput` the right move?

pomykalakyle:
> Just have a couple comments/questions.
> 
> Also, `OpenTableFormatInput` seems to support only Iceberg, but Glue as a whole supports a variety of file and table formats ([see here](https://docs.aws.amazon.com/glue/latest/dg/aws-glue-programming-etl-format.html)). Is switching to `OpenTableFormatInput` the right move?

@shehabgamin  This is actually a really good comment, might need to rethink this whole PR. I'm putting it back in draft for now. 

 The original point of this was to make it so that the new struct for partitions can be used across more providers to unify things but making users use the Iceberg system might not be the best way to go about this since there are others too.... need to think about this.

pomykalakyle:
@shehabgamin It's fixed - thanks for catching that for me!

shehabgamin:
> @shehabgamin It's fixed - thanks for catching that for me!

@pomykalakyle Forsure! Let me know your thoughts on the other PR comments.

pomykalakyle:
@shehabgamin They've all been addressed.

crates/sail-catalog-glue/src/provider.rs
```diff
@@ -630,57 +462,12 @@ impl CatalogProvider for GlueCatalogProvider {
...
-                    Err(CatalogError::External(format!(
-                        "Failed to create table: {service_err}"
-                    )))
-                }
-            }
+        if format_lower == "iceberg" {
+            iceberg::create_iceberg_table(self, client, database, table, options).await
+        } else {
+            hive::create_hive_table(self, client, database, table, options).await
         }
```
linhr, line 471: Nice approach!

Issue #1069, opened 2025-11-20
Follow-up tasks for DataFusion 51 upgrade


linhr:
This is a follow-up for #1031.

1. Look into failed test cases.
2. Adjust default `metadata_size_hint` configuration value.

linhr:
Related issue upstream: <https://github.com/apache/datafusion/issues/18527>

PR #1280, opened 2026-01-30
fix: preserve plan properties in stage inputs

linhr:
Part of #1069.

We need to preserve equivalence properties when building distributed query plans. Otherwise, some validation may fail for window aggregation.

This PR also preserves the boundness property for shuffle input so that certain streaming queries can run in cluster mode. These queries were previously known to fail due to boundness validation.


lonless9 (approved):
Great fix!

PR #1288, opened 2026-02-01
feat: catalog SQL commands (SHOW CATALOGS, USE CATALOG) and field serialization fixes

davidlghellin:
part #501 

Fix PySpark catalog API field names to use camelCase serialization, matching the expected format by PySpark clients.


python/pysail/tests/spark/function/features/catalog.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Catalog operations return correct field names
```
linhr: Let's rewrite these tests to Python tests and avoid defining new BDD steps. Since we are testing PySpark APIs, this would follow the same design reasoning that we have discussed before: <https://github.com/lakehq/sail/pull/1243#discussion_r2723813373>.

PR #1302, opened 2026-02-03
feat!: update Python module initialization logic

linhr:
This PR contains a few breaking changes to the PySail package.

1. Drop Python 3.9 support and fix a few lint errors accordingly.
2. Remove the `SparkConnectServer.init_telemetry()` method from the Python API. OpenTelemetry is now initialized when the `pysail.spark` module is imported. Telemetry shutdown is performed using a Python `atexit` hook.
3. The Sail configuration is loaded globally and cannot be changed. We emit a warning if we detect environment variable changes when creating a new `SparkConnectServer`.

### DIFF
--- a/crates/sail-python-udf/src/python/spark.py
+++ b/crates/sail-python-udf/src/python/spark.py
@@ -740,7 +740,7 @@ class PySparkArrowTableUdf:
             if passthrough is None:
                 passthrough = last  # noqa: PLW2901
             if passthrough is not None:
-                for v, name in zip(passthrough, self._input_names, strict=True):
+                for v, name in zip(passthrough, self._input_names, strict=False):
                     df[name] = [v] * len(out)
             else:
                 for name in self._input_names:


crates/sail-python-udf/src/python/spark.py
```diff
@@ -308,7 +308,7 @@ def to_pyspark(self, array: pa.Array) -> Sequence[Any]:
                 result.append(None)
             else:
                 (start, end) = (offsets[i], offsets[i + 1])
-                result.append(dict(zip(keys[start:end], values[start:end])))
+                result.append(dict(zip(keys[start:end], values[start:end], strict=False)))
```
lonless9: Is it better to set this to true or false?

crates/sail-python/src/globals.rs
```diff
@@ -0,0 +1,147 @@
...
+        // the snapshot we capture here is a good enough approximation to warn users
+        // about changes to the environment variables later.
+        let environment = EnvironmentSnapshot::initialize();
+
+        let config = AppConfig::load().map_err(|e| {
+            PyErr::new::<PyRuntimeError, _>(format!("failed to load configuration: {e}"))
+        })?;
+
+        let runtime = RuntimeManager::try_new(&config.runtime).map_err(|e| {
+            PyErr::new::<PyRuntimeError, _>(format!("failed to create the runtime: {e}"))
```
lonless9, line 111: What will happen with fork/spawn and potential pytest-xdist? Do we need to consider the following scenarios?
```python
import pysail.spark  # init
from multiprocessing import Pool

def job(x):
  # sail logic
  ...
if __name__ == "__main__":
  with Pool(8) as p: p.map(job, range(100))
```

crates/sail-telemetry/src/telemetry.rs
lonless9, line 194: What happens if multiple pyo3 projects compete for the logger within the same Python process?

crates/sail-python-udf/src/python/spark.py
linhr: Nice catch! Those lints were fixed by Hatch automatically, but yeah let me manually review them and change the places where `strict=True` should be used instead.

crates/sail-python/src/globals.rs
linhr: If we use `multiprocessing`, the process is forked, so each one will load the native module individually. Therefore, I think there will be no conflict.

For multi-threading, since the native library will be imported only once, there should be no issue with initialization either.

crates/sail-telemetry/src/telemetry.rs
linhr: I assume internally the logger uses a lock to ensure concurrent access. If we have no issue with that within Rust, it should also work in PyO3 and Python.

And I think native modules in Python are fully isolated, so they actually use different loggers even if the Rust code is the same.

But yeah it may not be possible to lock stdout or stderr globally. So the text written to those file descriptors from different threads (that run code from different libraries) may be interleaved.

linhr:
It seems `strict=True` causes UDTF test failures in PySpark 3.5. Let me take a look.

PR #1303, opened 2026-02-03
feat: implement ANSI mode support for Spark 4.x compatibility

davidlghellin:
- Division/modulo by zero returns NULL when ANSI=false, throws error when ANSI=true
- Array out of bounds returns NULL when ANSI=false, throws error when ANSI=true  
- CAST invalid values returns NULL when ANSI=false, throws error when ANSI=true
- unix_timestamp invalid format returns NULL when ANSI=false, throws error when ANSI=true
- Add pytest-bdd test scenarios for all ANSI mode behaviors

### DIFF
--- a/crates/sail-execution/proto/sail/plan/physical.proto
+++ b/crates/sail-execution/proto/sail/plan/physical.proto
@@ -128,6 +128,7 @@ message ExplodeUdf {
 
 message SparkUnixTimestampUdf {
   string timezone = 1;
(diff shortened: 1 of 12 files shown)


davidlghellin:
Closing — superseded by smaller focused ANSI PRs. The only remaining bit (element_at
NULL on out-of-bounds with ANSI=false) I'll do in a separate PR.

Issue #1196, opened 2026-01-12
Investigate proper divide-by-zero behavior for compatibility with both PySpark tests and Ibis tests


linhr:
This is the follow-up work for #1190.

PR #1304, opened 2026-02-03
fix: handle decimal division by zero for ANSI mode

davidlghellin:
## Ibis test compatibility note
The Ibis tests `test_divide_by_zero[pyspark-*]` fail with `ArithmeticException` instead of the expected `AssertionError`. This is correct behavior for Spark 4.x where `spark.sql.ansi.enabled=true` by default.

https://github.com/ibis-project/ibis/blob/dbdd13191b9cf1b6ff04a7f648cc2ed65eabc557/ibis/backends/tests/test_numeric.py#L1273
```python
@pytest.mark.notyet(["mysql", "pyspark"], raises=AssertionError)
```
related #501 #1196 #1190

need https://github.com/lakehq/sail/pull/1337

### DIFF
--- a/crates/sail-plan/src/resolver/tree/window.rs
+++ b/crates/sail-plan/src/resolver/tree/window.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::mem;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


davidlghellin:
https://github.com/ibis-project/ibis/issues/11802

PR #1319, opened 2026-02-05
feat!: lazy config loading

linhr:
#1302 introduces a breaking change that ignores Sail environment variable changes once the PySail module is imported. This was meant to make the server behavior predictable and consistent with telemetry initialization that is done globally. However, I now realize that such a way of configuring the Sail server may result in confusing user behavior, even if we have emitted a warning when we ignore environment variable changes.

For example, in benchmark scripts, we have this common pattern:

```python
import os

# This is ignored after #1302.
os.environ["SAIL_OPTIMIZER__ENABLE_JOIN_REORDER"] = "true"

server = SparkConnectServer()
```

Since such scripts run automatically, we have no way to observe the warning and know that the server was initialized improperly.

This PR reverts the breaking change around configuration. Each `SparkConnectServer` instance now loads the configuration again from the environment variables (the same way before #1302). We only ignore the few configuration options related to telemetry and the runtime stack.

This should also revert the ~3% coverage decrease since #1302 (although we've already been thinking about fixing it).

### DIFF
--- a/crates/sail-python/src/spark/server.rs
+++ b/crates/sail-python/src/spark/server.rs
@@ -56,13 +56,15 @@ impl SparkConnectServer {
     #[new]
     #[pyo3(signature = (ip, port, /))]
     fn new(py: Python<'_>, ip: &str, port: u16) -> PyResult<Self> {
+        let config = AppConfig::load().map_err(|e| {
+            PyErr::new::<PyRuntimeError, _>(format!("failed to load the application config: {e}"))
+        })?;
         let globals = GlobalState::instance(py)?;
-        let config = globals.config.clone();
         let runtime = globals.runtime.handle();
         Ok(Self {
             ip: ip.to_string(),
             port,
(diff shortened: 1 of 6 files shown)


crates/sail-python/src/spark/server.rs
```diff
@@ -56,13 +56,15 @@ impl SparkConnectServer {
     #[new]
     #[pyo3(signature = (ip, port, /))]
     fn new(py: Python<'_>, ip: &str, port: u16) -> PyResult<Self> {
+        let config = AppConfig::load().map_err(|e| {
+            PyErr::new::<PyRuntimeError, _>(format!("failed to load the application config: {e}"))
+        })?;
         let globals = GlobalState::instance(py)?;
```
linhr, line 62: The global state (telemetry and Tokio runtime) is now lazily initialized when the server is created for the first time.

PR #1325, opened 2026-02-06
feat: add Geometry and Geography geospatial types (Spark 4.1)

james-willis:
## Summary

Adds comprehensive support for Spark 4.1 Geometry and Geography geospatial data types to Sail, implementing the full type system integration from SQL parsing through Arrow metadata to Spark Connect protocol.

### Core Type System
- Add `Geometry { srid }` and `Geography { srid, algorithm }` variants to `spec::DataType`
- Implement `EdgeInterpolationAlgorithm` enum (Spherical for Geography, Planar as Sail-internal for Geometry distinction)
- Define `GEOSPATIAL_*` constants for default/mixed SRID handling

### SQL Parser Integration
- Support `GEOMETRY(srid)` and `GEOGRAPHY(srid)` syntax with required parentheses (Spark 4.1 standard)
- Support `GEOMETRY(ANY)` and `GEOGRAPHY(ANY)` for mixed SRID (-1)
- **SRID validation** matching Spark 4.1's CRS mappers

### Arrow Integration
- Store geo types as Binary with `geoarrow.wkb` extension metadata
- Use `ARROW:extension:name` and `ARROW:extension:metadata` per Apache Arrow standard
- Support round-trip conversion: spec → Arrow Field → proto
- Create `GeoArrowMetadata` struct for unified metadata parsing
- Filter extension metadata from Spark client responses (no data leakage)

### Spark Connect Protocol
- Implement proto ↔ spec conversion for `Geometry` and `Geography` proto structs
- Default Geography algorithm to Spherical (only algorithm in Spark 4.1)

### JSON Serialization
- Support parameterized syntax: `geometry(4326)`, `geometry(ANY)`
- Support `geography(4326, spherical)`, `geography(ANY, spherical)`
- Follow `char(n)`/`decimal(p,s)` pattern with regex-based serde

## Spark 4.1 Compliance

| Feature | Spark 4.1 | Sail Implementation |
|---------|-----------|---------------------|
| SQL syntax | `GEOMETRY(srid)` required | ✅ Parentheses required |
| Default SRID | 4326 | ✅ 4326 |
| Mixed SRID | -1 / `ANY` keyword | ✅ Supported |
| CRS format | `OGC:CRS84` for 4326, `EPSG:3857` for 3857 | ✅ Proper mapping via `srid_to_crs()` |
| Geometry SRID validation | 0, 3857, 4326 valid | ✅ Validated with error on invalid |
| Geography SRID validation | Only 4326 valid | ✅ Validated with error on invalid |
| Edge algorithm | Spherical only | ✅ Default Spherical |
| JSON format | `geometry(4326)` | ✅ Parameterized |
| Per-row SRID storage | SRID stored per geometry in WKB | ⚠️ Column-level only (see below) |

### Per-row SRID Storage

**Current behavior:** SRID is stored at the column level (in field metadata), not per-row.

**Spark 4.1 behavior:** When SRID=-1 (mixed), each row can have a different SRID stored in the WKB data.

**Plan:** This requires architectural changes:
1. Extend WKB encoding to include SRID prefix per geometry
2. Add SRID extraction during query execution
3. Implement SRID coercion rules for operations on mixed-SRID columns

Since this PR doesn't have any constructors, I felt we could tackle this limitation once its possible to actually create geometries that contain their own CRS.

## Test Coverage

- Proto ↔ spec conversion tests
- Arrow metadata round-trip tests  
- JSON serialization tests
- Gold data tests for SQL parsing
- CRS format mapping tests
- **SRID validation tests** (valid and invalid cases)

## Files Changed

- `sail-common/src/spec/data_type.rs` - Core types and constants
- `sail-common/src/spec/literal.rs` - Null literal handling
- `sail-plan/src/resolver/data_type.rs` - Arrow field resolution with metadata
- `sail-spark-connect/src/proto/data_type.rs` - Proto conversions
- `sail-spark-connect/src/proto/data_type_arrow.rs` - Arrow ↔ proto with geoarrow parsing
- `sail-spark-connect/src/proto/data_type_json.rs` - JSON serde with SRID validation
- `sail-sql-parser/src/ast/data_type.rs` - Parser AST
- `sail-sql-analyzer/src/data_type.rs` - AST → spec conversion with SRID validation


crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -248,6 +290,52 @@ impl PlanResolver<'_> {
...
+                // Add geoarrow extension type metadata for WKB-encoded geometries
+                // Geometry uses Cartesian (planar) coordinate system
+                //
+                // ARROW:extension:* keys follow the Apache Arrow extension type standard:
+                // - ARROW:extension:name: the extension type identifier (e.g., "geoarrow.wkb")
+                // - ARROW:extension:metadata: extension-specific metadata (JSON format)
+                //
+                // These keys are automatically filtered from Spark client responses and do not
+                // leak user-facing metadata, which allows us to use Arrow's standard namespace
+                // without conflict. See data_type_arrow.rs line 106 for the filtering logic.
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -248,6 +290,52 @@ impl PlanResolver<'_> {
...
+                // Add geoarrow extension type metadata for WKB-encoded geographies
+                // Geography uses spherical (geodetic) coordinate system
+                //
+                // ARROW:extension:* keys follow the Apache Arrow extension type standard:
+                // - ARROW:extension:name: the extension type identifier (e.g., "geoarrow.wkb")
+                // - ARROW:extension:metadata: extension-specific metadata (JSON format)
+                //
+                // These keys are automatically filtered from Spark client responses and do not
+                // leak user-facing metadata, which allows us to use Arrow's standard namespace
+                // without conflict. See data_type_arrow.rs line 106 for the filtering logic.
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-common/src/spec/data_type.rs
```diff
@@ -21,6 +21,58 @@ pub const SAIL_MAP_KEY_FIELD_NAME: &str = "key";
...
+#[num_enum(error_type(name = CommonError, constructor = EdgeInterpolationAlgorithm::invalid))]
+#[repr(i32)]
+pub enum EdgeInterpolationAlgorithm {
+    /// Planar edge interpolation (Cartesian coordinates).
+    /// This is a Sail-internal extension; Spark 4.1 only supports Spherical.
+    /// Used in Arrow metadata to distinguish Geometry (planar) from Geography (spherical).
+    Planar,
+    /// Spherical edge interpolation (geodetic coordinates on a sphere).
+    /// This is the only algorithm supported by Spark 4.1 for Geography types.
+    Spherical,
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-spark-connect/src/proto/data_type_json.rs
```diff
@@ -514,6 +607,37 @@ fn from_spark_json_data_type(data_type: JsonDataType) -> SparkResult<sc::DataTyp
...
+                )));
+            }
+            sc::DataType {
+                kind: Some(dt::Kind::Geometry(dt::Geometry {
+                    srid,
+                    type_variation_reference: 0,
+                })),
+            }
+        }
+        JsonDataType::Geography { srid, algorithm: _ } => {
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-common/src/spec/literal.rs
```diff
@@ -365,5 +365,7 @@ pub fn data_type_to_null_literal(data_type: spec::DataType) -> CommonResult<Lite
         spec::DataType::ConfiguredUtf8 { .. } => Ok(Literal::Utf8 { value: None }),
         spec::DataType::ConfiguredBinary => Ok(Literal::Binary { value: None }),
         spec::DataType::UserDefined { sql_type, .. } => data_type_to_null_literal(*sql_type),
+        spec::DataType::Geometry { .. } => Err(error(&data_type)),
+        spec::DataType::Geography { .. } => Err(error(&data_type)),
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-spark-connect/src/proto/data_type.rs
```diff
@@ -256,6 +266,10 @@ impl TryFrom<DataType> for spec::DataType {
 
 #[cfg(test)]
 mod tests {
+    // Tests are allowed to use panic for assertions
+    #![allow(clippy::panic)]
```
linhr: I wonder if the `panic!` in the tests could simply be replaced by `assert_eq!` on the expected values.

crates/sail-spark-connect/src/proto/data_type.rs
```diff
@@ -244,8 +244,18 @@ impl TryFrom<DataType> for spec::DataType {
...
+                Ok(spec::DataType::Geometry { srid })
+            }
+            Kind::Geography(geography) => {
+                let srid = geography.srid;
+                // Geography always defaults to Spherical algorithm in Spark 4.1
+                Ok(spec::DataType::Geography {
+                    srid,
+                    algorithm: spec::EdgeInterpolationAlgorithm::Spherical,
+                })
+            }
```
linhr, line 263: Let's use something like `let Geometry { srid, type_variation_reference: _ } = *geometry;` to explicitly list all the fields. In this way the compiler will remind us when new unsupported fields are added in the future.

crates/sail-spark-connect/src/proto/data_type_json.rs
```diff
@@ -995,6 +1125,24 @@ mod tests {
...
+            ),
+            (
+                r#""geography(4326, spherical)""#,
+                sc::DataType {
+                    kind: Some(dt::Kind::Geography(dt::Geography {
+                        srid: 4326,
+                        type_variation_reference: 0,
+                    })),
+                },
+            ),
```
linhr, line 1155: Can we add tests around `ANY`?

crates/sail-spark-connect/tests/gold_data/data_type.json
```diff
@@ -848,6 +848,86 @@
       "output": {
         "success": "null"
       }
+    },
+    {
```
linhr: The test cases in the gold data are extracted automatically from the Spark codebase. Any manually added test cases will be removed when these files are generated later.

So we will have to remove changes to this file unfortunately. I feel the unit tests in Rust have covered this PR pretty well.

crates/sail-sql-analyzer/src/data_type.rs
```diff
@@ -258,5 +258,37 @@ pub fn from_ast_data_type(sql_type: DataType) -> SqlResult<spec::DataType> {
...
+                sail_sql_parser::ast::data_type::GeographySrid::Any(_) => -1,
+            };
+            // Validate SRID against Spark 4.1's GeographicSpatialReferenceSystemMapper
+            // Valid SRIDs for Geography: 4326 (WGS84), -1 (mixed)
+            // Note: 3857 (Web Mercator) is NOT valid for Geography as it's a Cartesian projection
+            if !matches!(srid_value, 4326 | -1) {
+                return Err(SqlError::invalid(format!(
+                    "SRID {srid_value} is not valid for GEOGRAPHY. Valid values: 4326 or ANY"
+                )));
+            }
```
linhr: The spec is considered "unverified" so we should just pass SRID to spec without error handling here. (Also, because the spec may be constructed from sources other than SQL in the future, it's good to have centralized error handling.) If we need error handling, we need to move that to the plan resolver.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -5,11 +5,73 @@ use datafusion::arrow::datatypes as adt;
...
+    /// Parse GeoArrow metadata from JSON string.
+    /// Defaults to planar edges and SRID 4326 if parsing fails.
+    ///
+    /// Supports Spark 4.1 CRS formats:
+    /// - "OGC:CRS84" → SRID 4326 (WGS84)
+    /// - "EPSG:{srid}" → direct SRID mapping
+    /// - "SRID:0" → SRID 0 (unspecified)
+    /// - "SRID:ANY" → SRID -1 (mixed)
+    fn from_json(metadata: &str) -> Self {
+        let parsed = serde_json::from_str::<serde_json::Value>(metadata).ok();
```
linhr: I'm wondering if we should return `Result` instead of returning some default value for invalid input.

Also, it might be easier to have an auxiliary struct with `Deserialize` derived, instead of working with `Value` manually.

```rust
#[derive(Deserialize)]
struct RawGeoArrowMetadata { // or name this something better
    edges: String,
    crs: Option<String>,
}
```

crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -248,6 +290,52 @@ impl PlanResolver<'_> {
...
+                // Add geoarrow extension type metadata for WKB-encoded geometries
+                // Geometry uses Cartesian (planar) coordinate system
+                //
+                // ARROW:extension:* keys follow the Apache Arrow extension type standard:
+                // - ARROW:extension:name: the extension type identifier (e.g., "geoarrow.wkb")
+                // - ARROW:extension:metadata: extension-specific metadata (JSON format)
+                //
+                // These keys are automatically filtered from Spark client responses and do not
+                // leak user-facing metadata, which allows us to use Arrow's standard namespace
+                // without conflict. See data_type_arrow.rs line 120 for the filtering logic.
```
linhr: nitpick: Referencing line numbers in the comment is fragile if the code changes later. Maybe we should just remove the line number reference in the comment. (Same below for `Geography`.)

linhr (commented):
This is very exciting! I have a few minor comments, but this overall looks really solid!

crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -248,6 +290,52 @@ impl PlanResolver<'_> {
...
+                metadata.insert(
+                    "ARROW:extension:name".to_string(),
+                    "geoarrow.wkb".to_string(),
+                );
+                // Use mixed CRS format for SRID -1, otherwise map SRID to Spark 4.1 CRS format
+                let crs = srid_to_crs(*srid);
+                let ext_metadata = json!({"crs": crs, "edges": "planar"}).to_string();
```
paleolimbot:
```suggestion
                let ext_metadata = json!({"crs": crs}).to_string();
```

crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -248,6 +290,52 @@ impl PlanResolver<'_> {
...
+                // without conflict. See data_type_arrow.rs line 120 for the filtering logic.
+                metadata.insert(
+                    "ARROW:extension:name".to_string(),
+                    "geoarrow.wkb".to_string(),
+                );
+                // Map SRID to Spark 4.1 CRS format
+                let crs = srid_to_crs(*srid);
+                let edges = match algorithm {
+                    spec::EdgeInterpolationAlgorithm::Spherical => "spherical",
+                    spec::EdgeInterpolationAlgorithm::Planar => "planar",
```
paleolimbot: I am not sure a planar Geography makes sense (at least in GeoArrow this would get resolved as Geometry instead of Geography...there was talk in Spark of making these concepts independent and I'm not sure where it landed).

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -5,11 +5,73 @@ use datafusion::arrow::datatypes as adt;
...
+        let edges = parsed
+            .as_ref()
+            .and_then(|v| v.get("edges").and_then(|e| e.as_str()))
+            .unwrap_or("planar")
+            .to_string();
+
+        let srid = parsed
+            .and_then(|v| {
+                v.get("crs").and_then(|crs| crs.as_str()).and_then(|s| {
+                    // Handle Spark 4.1 CRS string formats
```
paleolimbot: If this will get used to ingest Arrow data from elsewhere (e.g., GeoPandas, SedonaDB, DuckDB), an important case to handle is PROJJSON input, where you could pull the "id" member to get the authority/code:

https://github.com/apache/sedona-db/blob/d17d7f08dd65fe42d9bf64126ae2a383666c72f6/rust/sedona-schema/src/crs.rs#L343-L359

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -5,11 +5,73 @@ use datafusion::arrow::datatypes as adt;
...
     type Error = SparkError;
 
     fn try_from(field: adt::Field) -> SparkResult<sdt::StructField> {
         let is_udt = field.metadata().keys().any(|k| k.starts_with("udt."));
+        let is_geoarrow = field
+            .metadata()
+            .get("ARROW:extension:name")
+            .map(|v| v == "geoarrow.wkb")
+            .unwrap_or(false);
+
```
paleolimbot: I think that an arrow-rs Field has a `get_extension_name()` convenience function for this

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -172,3 +259,160 @@ impl TryFrom<adt::DataType> for DataType {
...
+        // Test mixed SRID (-1) support
+        let metadata: HashMap<String, String> = [
+            (
+                "ARROW:extension:name".to_string(),
+                "geoarrow.wkb".to_string(),
+            ),
+            (
+                "ARROW:extension:metadata".to_string(),
+                r#"{"crs":"SRID:ANY","edges":"planar"}"#.to_string(),
+            ),
```
paleolimbot: In GeoArrow land this wouldn't be interpreted elsewhere (but you could use it internally if this metadata never leaves Sail)

crates/sail-spark-connect/src/proto/data_type_json.rs
```diff
@@ -514,6 +607,43 @@ fn from_spark_json_data_type(data_type: JsonDataType) -> SparkResult<sc::DataTyp
         JsonDataType::Variant => sc::DataType {
             kind: Some(dt::Kind::Variant(dt::Variant::default())),
         },
+        JsonDataType::Geometry { srid } => {
+            // Validate SRID against Spark 4.1's CartesianSpatialReferenceSystemMapper
+            // Valid SRIDs for Geometry: 0 (unspecified), 3857 (Web Mercator), 4326 (WGS84), -1 (mixed)
+            if !matches!(srid, 0 | 3857 | 4326 | -1) {
+                return Err(SparkError::invalid(format!(
+                    "SRID {srid} is not valid for GEOMETRY. Valid values: 0, 3857, 4326, or ANY"
+                )));
```
paleolimbot: It may be worth consolidating the list of these (I think there is already a function that returns an authority/code in another file)

crates/sail-sql-analyzer/src/data_type.rs
```diff
@@ -258,5 +258,37 @@ pub fn from_ast_data_type(sql_type: DataType) -> SqlResult<spec::DataType> {
...
             })
         }
+        DataType::Geometry(_, _, srid, _) => {
+            let srid_value = match srid {
+                sail_sql_parser::ast::data_type::GeometrySrid::Srid(lit) => lit.value as i32,
+                sail_sql_parser::ast::data_type::GeometrySrid::Any(_) => -1,
+            };
+            // Validate SRID against Spark 4.1's CartesianSpatialReferenceSystemMapper
+            // Valid SRIDs for Geometry: 0 (unspecified), 3857 (Web Mercator), 4326 (WGS84), -1 (mixed)
+            if !matches!(srid_value, 0 | 3857 | 4326 | -1) {
```
paleolimbot: This list should also probably be consolidated.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
james-willis: In the mixed CRS case does geoarrow want us to simply not include it here?

crates/sail-spark-connect/src/proto/data_type_json.rs
james-willis: done

crates/sail-spark-connect/src/proto/data_type_arrow.rs
james-willis: read the geoarrow docs, yes we should elide it here

crates/sail-plan/src/resolver/data_type.rs
james-willis: > The edges key must be omitted to indicate planar/linear edges

from the geoarrow spec

crates/sail-plan/src/resolver/data_type.rs
james-willis: Yeah it seems that spark implements only a Spherical in their enum so we can go that route and avoid these ambiguities.

https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/EdgeInterpolationAlgorithm.html

crates/sail-spark-connect/src/proto/data_type.rs
james-willis: done

crates/sail-spark-connect/tests/gold_data/data_type.json
james-willis: done

crates/sail-plan/src/resolver/data_type.rs
james-willis: done

crates/sail-sql-analyzer/src/data_type.rs
james-willis: done. My inclination is that its ok to silently drop the CRS for geometry. I'm not sure what CRSes we can allow for geography. Im not sure how spark behaves in either of these cases though.

Its difficult to verify in the plan resolver because we lose information when we cannot map a CRS to a SRID.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
james-willis: done

crates/sail-sql-analyzer/src/data_type.rs
james-willis: Dewey's idea to use a string internal representation might resolve this problem of verification in the plan resolver.

crates/sail-plan/src/resolver/data_type.rs
james-willis: done

crates/sail-plan/src/resolver/data_type.rs
james-willis: done. removed Planar from the enum

crates/sail-spark-connect/src/proto/data_type_arrow.rs
james-willis: done

crates/sail-spark-connect/src/proto/data_type_arrow.rs
james-willis: done

crates/sail-spark-connect/src/proto/data_type_json.rs
james-willis: done

crates/sail-sql-analyzer/src/data_type.rs
james-willis: done

crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -6,12 +6,57 @@ use sail_common::spec;
...
+/// - org.apache.spark.sql.types.GeographyType#crs
+/// - org.apache.spark.sql.catalyst.util.CartesianSpatialReferenceSystemMapper
+/// - org.apache.spark.sql.catalyst.util.GeographicSpatialReferenceSystemMapper
+/// Returns `None` for SRID -1 (mixed) since there is no single CRS for the column.
+fn srid_to_crs(srid: i32) -> Option<String> {
+    match srid {
+        4326 => Some("OGC:CRS84".to_string()),
+        3857 => Some("EPSG:3857".to_string()),
+        0 => Some("SRID:0".to_string()),
+        -1 => None,
```
james-willis: Note to self: double check that were handling both unknown and mixed correctly and in accordance with the geoarrow spec

crates/sail-spark-connect/src/proto/data_type.rs
james-willis: I think I've implemented this comment correctly. Leaving open because I'm a bit unsure.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -1,15 +1,115 @@
...
+    edges: Option<String>,
+    /// CRS can be a string ("OGC:CRS84", "EPSG:3857") or a PROJJSON object.
+    crs: Option<serde_json::Value>,
+}
+
+/// GeoArrow extension metadata extracted from Arrow field metadata.
+#[derive(Debug, Clone)]
+struct GeoArrowMetadata {
+    /// Edge interpolation: "planar" for Geometry, "spherical" for Geography
+    edges: String,
```
james-willis: It might be better to have this be an Option<String> and when its None we know its a Geometry, otherwise geography

crates/sail-spark-connect/src/proto/data_type.rs
linhr: Yeah this is what I meant!

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -1,15 +1,128 @@
...
+use serde::Deserialize;
 
 use crate::error::{SparkError, SparkResult};
 use crate::spark::connect::{data_type as sdt, DataType};
 
+/// Raw GeoArrow extension metadata deserialized from JSON.
+#[derive(Debug, Deserialize)]
+struct RawGeoArrowMetadata {
+    edges: Option<String>,
+    /// CRS can be a string ("OGC:CRS84", "EPSG:3857") or a PROJJSON object.
```
linhr, line 13: You could have `crs` as an enum to take care of heterogeneous types. The `serde` crate is powerful enough so that we usually won't have to work with `Value` directly.

It may take some time to figure out the `serde` attributes to use for this particular case, so this could be future work. <https://serde.rs/attributes.html>

PR #1327, opened 2026-02-06
fix: handle empty array concatenation in SparkConcat

davidlghellin:
## Summary
- Fix empty array concatenation: `concat(array(), array(1,2,3))` now works
  - Arrays with `List(Null)` type are cast to target type before concatenation
- Remove `array_concat` alias (doesn't exist in Spark, only `concat()`)
- Fix Python test import order to avoid PySail module initialization error

## Test plan
- Validated against Spark JVM with `SPARK_REMOTE="local"`
- Added feature tests for empty array concatenation

```sh
sdk use java 17.0.9-tem && export SPARK_REMOTE="local" && hatch run pytest python/pysail/tests/spark/function/test_features.py -vv
```
error before import test
image

### DIFF
--- a/crates/sail-execution/src/codec.rs
+++ b/crates/sail-execution/src/codec.rs
@@ -1493,7 +1493,9 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_array" | "spark_make_array" | "array" => {
                 Ok(Arc::new(ScalarUDF::from(SparkArray::new())))
             }
-            "spark_concat" | "concat" => Ok(Arc::new(ScalarUDF::from(SparkConcat::new()))),
+            "spark_concat" | "concat" | "array_concat" => {
+                Ok(Arc::new(ScalarUDF::from(SparkConcat::new())))
+            }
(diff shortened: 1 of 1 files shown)


davidlghellin:
> Nice fix!
> 
> I feel we should keep `array_concat` even if it's not supported by Spark.

That makes sense. In that case, I think we should skip or disable those tests when running in “local Spark” mode, since the feature isn’t supported by Spark. I’ll adjust the test setup so they’re not executed when running with Spark.

python/pysail/tests/spark/conftest.py
```diff
@@ -112,8 +119,13 @@ def local_timezone(request):
 
 def pytest_collection_modifyitems(session, config, items):  # noqa: ARG001
     if is_jvm_spark():
+        skip_sail_only = pytest.mark.skip(reason="Sail-only feature, not supported by Spark")
         for item in items:
             if isinstance(item, DoctestItem):
                 for example in item.dtest.examples:
                     if example.options.get(SAIL_ONLY):
                         example.options[doctest.SKIP] = True
+            # Skip pytest-bdd scenarios with @sail-only tag
```
linhr, line 128: Nice!

PR #1329, opened 2026-02-06
feat: implement Spark-compatible `concat_ws` with array support

davidlghellin:
Add SparkConcatWs UDF that properly handles array arguments by
flattening them and joining their elements with the separator,
matching Spark's behavior. Previously, passing arrays to concat_ws
would result in the array's string representation (e.g., "[a, b, c]")
instead of joining elements (e.g., "a,b,c").


davidlghellin:
need https://github.com/lakehq/sail/pull/1327 before

PR #1331, opened 2026-02-06
fix: allow duplicate CTE names with shadowing behavior

davidlghellin:
When Ibis generates SQL with nested expressions, it flattens CTEs into
a single WITH clause, which can result in duplicate CTE names like:
`WITH "x" AS (...), "alias" AS (...), "x" AS (...)`

Previously Sail rejected this with "CTE query name specified more than once".
Now we deduplicate CTEs keeping the last occurrence (shadowing), matching
Spark's behavior.

part #501


linhr (commented):
Actually could you add a BDD test for the shadowing behavior?

PR #1332, opened 2026-02-06
feat: implement `TABLESAMPLE` clause for SQL queries

davidlghellin:
part #501

### DIFF
--- a/crates/sail-plan/src/resolver/query/read.rs
+++ b/crates/sail-plan/src/resolver/query/read.rs
@@ -113,7 +113,7 @@ impl PlanResolver<'_> {
     }
 
     /// Apply TABLESAMPLE clause to a LogicalPlan
-    async fn apply_table_sample(
+    pub(super) async fn apply_table_sample(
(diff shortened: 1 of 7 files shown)


crates/sail-plan/src/resolver/query/read.rs
```diff
@@ -100,7 +103,105 @@ impl PlanResolver<'_> {
...
+            let mut r = rng();
+            r.random::<i64>()
+        });
+
+        // TABLESAMPLE is without replacement
+        Self::apply_sample_to_plan(plan, lower_bound, upper_bound, false, seed, state)
+    }
+
+    /// Evaluate a sample expression to get a float value
+    fn evaluate_sample_expr_to_f64(&self, expr: &spec::Expr) -> PlanResult<f64> {
```
linhr: I feel we should resolve the spec using an empty schema and then use `LiteralEvaluator` to get the float value. This ensures that we can support constant expressions other than literals.

PR #1335, opened 2026-02-08
fix: add missing coalesce on CollectLeft hash join build side

davidlghellin:
part #501 

related https://github.com/apache/datafusion/issues/12955


lonless9:
It seems there have been some changes to the snapshot.
You can use `hatch run pytest -k delta --snapshot-update` to update the corresponding snapshots and then see what exactly happened!

davidlghellin:
#1483

PR #1337, opened 2026-02-08
test: toggle ANSI mode off for Ibis

davidlghellin:
This PR adds a per-test ANSI mode toggle for Ibis tests. Some Ibis tests (e.g. `test_divide_by_zero`) expect non-ANSI behavior (NULL or Infinity for division by zero)

### DIFF
--- a/scripts/spark-tests/plugins/ibis.py
+++ b/scripts/spark-tests/plugins/ibis.py
@@ -106,7 +106,7 @@ def add_ibis_test_markers(items: list[pytest.Item]):
 
 
 def _needs_ansi_disabled(item: pytest.Item) -> bool:
-    return any(all(k in item.name for k in test.keywords) for test in ANSI_DISABLED_IBIS_TESTS)
+    return any(all(any(k in kw for kw in item.keywords) for k in test.keywords) for test in ANSI_DISABLED_IBIS_TESTS)
 
 
 @pytest.fixture(autouse=True)


davidlghellin:
maybe here its better not disable `test_divide_by_zero`

scripts/spark-tests/plugins/ibis.py
```diff
@@ -94,6 +105,37 @@ def add_ibis_test_markers(items: list[pytest.Item]):
                 item.add_marker(pytest.mark.skip(reason=test.reason))
 
 
+def _needs_ansi_disabled(item: pytest.Item) -> bool:
+    return any(all(k in item.name for k in test.keywords) for test in ANSI_DISABLED_IBIS_TESTS)
```
linhr: I think we should have `k in item.keywords for k in test.keywords`.

linhr (commented):
Ibis only support Spark 3.5 for now. Once they support Spark 4 and take care of ANSI handling, would we be able to remove this patch?

The PR itself looks good though. I just have one small comment and we can merge it.

davidlghellin:
> Ibis only support Spark 3.5 for now. Once they support Spark 4 and take care of ANSI handling, would we be able to remove this patch?

Yes, once Ibis upgrades to Spark 4 and enables ANSI mode by default (`spark.sql.ansi.enabled=true`), this patch can be removed.

PR #1347, opened 2026-02-10
feat: format_string

wudihero2:
#508 
- use udf instead of binary
- pytest to test format_string


linhr (approved):
This looks good! And thanks for all the tests!

PR #1350, opened 2026-02-10
feat: implement additional Ibis-compatible features and enable 45 previously xpassed tests

davidlghellin:
to test:

- Add a CI patch script that removes `xfail` markers from 45 Ibis tests that Sail consistently passes
- These tests are marked as expected failures in Ibis for the PySpark backend, but Sail passes them due to additional capabilities beyond standard PySpark

When running the Ibis test suite, pytest treats unexpected passes (xpass) as warnings by default, which makes CI results noisy and hides real regressions. By converting these 45 xpass results into regular passes, we get a cleaner signal: **1417 → 1462 passes** and zero xpass noise.

  | Category | Count | Why Sail passes |
  |---|---|---|
  | Remote UDFs | 21 | Sail supports UDF execution over Spark Connect (Arrow-based); PySpark marks these as unsupported |
  | JSON unwrap | 8 | PySpark fails due to environment/classpath issues; Sail handles JSON natively via DataFusion |
  | File I/O | 8 | `read_csv`/`read_parquet`/`read_json` with glob patterns and `.gz` compression — not available in PySpark's Spark Connect mode |
  | Decimal256 | 2 | Sail supports precision > 38 digits via Arrow's Decimal256 type |
  | Big timestamps | 2 | Out-of-range timestamps (before 1677 / after 2262) handled correctly |
  | String quantiles | 2 | `median`/`quantile` operations on string columns supported |
  | Other | 2 | Uncorrelated subquery support, `first`/`last` with `where` + `include_null` |

The `patch-ibis-xpass.sh` script runs in CI after `pip install` and before test execution. It locates `ibis/backends/conftest.py` and injects an early return in `pytest_runtest_call` that skips the xfail-to-strict conversion for the listed test IDs. The patch is idempotent (safe to run multiple times).

related #501


davidlghellin:
example 
run no changes - changes
image

PR #1352, opened 2026-02-11
refactor: consolide logical predicate passing

lonless9:
- `file_write/file_delete` no longer pre-creates physical expressions. 
In delta `op_write/op_delete`, execute `create_physical_expr(...)` based on the target table schema. 
During the write execution phase, only `source` is retained, no longer relying on carrying logical condition objects

- Streamline unused/legacy fields in `MergeInfo`;
Retain fields required for pre-expanded paths and operation override metadata

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,5 +100,6 @@ docs/.vitepress/.temp
 
 # Claude Code settings
 .claude/
+.worktrees/
 
 CLAUDE.md
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6099,6 +6099,7 @@ dependencies = [
  "pyo3",
  "quote",
  "regex",
+ "sail-common",
  "sail-common-datafusion",
  "serde",
  "serde_yaml",
--- a/crates/sail-cli/src/python/spark_mcp_server.py
+++ b/crates/sail-cli/src/python/spark_mcp_server.py
@@ -116,7 +116,7 @@ def create_spark_mcp_server(host: str, port: int, spark_remote: str):
             Please concatenate the key prefix with the directory name (without any separator) to get the full path.
         """
         try:
-            import boto3  # noqa: PLC0415
+            import boto3
         except ImportError as e:
             message = "Please install boto3 to use AWS functionalities in the MCP server."
             raise RuntimeError(message) from e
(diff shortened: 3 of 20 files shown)


crates/sail-common-datafusion/src/datasource.rs
```diff
@@ -32,7 +31,12 @@ pub enum PhysicalSinkMode {
     IgnoreIfExists,
     Append,
     Overwrite,
-    OverwriteIf { condition: PhysicalExprWithSource },
+    OverwriteIf {
+        /// Driver-side logical predicate. Omitted on remote workers.
+        condition: Option<Box<LogicalPredicateInfo>>,
+        /// SQL source string used by commit metadata.
+        source: Option<String>,
+    },
```
lonless9, line 39: I feeel predicates could be divided into those for the planner phase and those for the execution phase, and converting them into physical predicates in the certain format planner might be better.

crates/sail-common-datafusion/src/logical_expr.rs
```diff
@@ -14,3 +14,6 @@ impl ExprWithSource {
         Self { expr, source }
     }
 }
+
+/// Canonical logical predicate payload used across planning layers.
+pub type LogicalPredicateInfo = ExprWithSource;
```
linhr: Is there a reason to have the type alias for logical predicates?

crates/sail-delta-lake/src/physical_plan/planner/op_write.rs
```diff
@@ -187,15 +206,16 @@ async fn build_overwrite_if_plan(
         } else {
             Some(ctx.partition_columns().to_vec())
         },
-        predicate: condition.source.clone(),
+        predicate: predicate_source.clone(),
     });
     let writer = Arc::new(DeltaWriterExec::new(
```
linhr, line 211: I feel we will ultimately get rid of `PhysicalSinkMode`. (This is just a note for the future.)

linhr (approved):
This looks nice!

I feel we can have more design discussions around how we can simplify logical/physical sink modes, and have better decoupling of the predicate during query planning and execution. This PR is a nice progress toward the clean architecture!

crates/sail-common-datafusion/src/logical_expr.rs
lonless9: Indeed this is a redundant alias 👍

PR #1354, opened 2026-02-11
fix: connect doctest session to Sail server instead of local Spark JVM

davidlghellin:
I think its a error
```
hatch run test-spark.spark-4.1.1:bash scripts/spark-tests/run-tests.sh --doctest-modules --pyargs pyspark.sql.functions.builtin -k "kurtosis" --tb=short -q
```

before
image

after
image

Issue #1314, opened 2026-02-05
Support multi-column IN subqueries


pomykalakyle:
## Summary

When using `struct(...).isin(subquery)` in PySpark, the `in_subquery_values` contains multiple expressions. Currently this returns an unsupported error.

## Example

```python
df.where(struct(df.age, df.name).isin(spark.range(6).select("id", lit("Bob")))).show()
```

SQL equivalent:
```sql
WHERE (age, name) IN (SELECT id, 'Bob' FROM range(6))
```

## Current Behavior

Returns error: `"multi-column IN subquery is not yet supported"`

## Why

DataFusion's `InSubquery` only accepts a single expression:

```rust
pub struct InSubquery {
    pub expr: Box<Expr>,  // single expression, not Vec<Expr>
    pub subquery: Subquery,
    pub negated: bool,
}
```

keen85:
@pomykalakyle, just wondering: couldn't this be rewritten (by the optimizer) to a [`SEMI JOIN`](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-join.html#semi-join)?
```python
df.where(struct(df.age, df.name).isin(spark.range(6).select("id", lit("Bob")))).show()
```
to
```python
df.join(other=spark.range(6).select(F.col("id").alias("age"), lit("Bob").alias("name")), on=["age", "name"], how="semi").show()
```

linhr:
> couldn't this be rewritten (by the optimizer) to a [SEMI JOIN](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-join.html#semi-join)?

Yes, subquery expressions are usually implemented as joins after plan rewrite. I believe this issue is more about how we support the query regarding the user interface.

PR #1356, opened 2026-02-11
fix: support single-column and multi-column IN subquery expressions

davidlghellin:
after create this #1354 I try one test 
related #1314

and solved `doctest-column` test

### DIFF
--- a/crates/sail-plan/src/resolver/expression/subquery.rs
+++ b/crates/sail-plan/src/resolver/expression/subquery.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use datafusion_common::tree_node::{Transformed, TransformedResult, TreeNode};
 use datafusion_common::DFSchemaRef;
 use datafusion_expr::logical_plan::{Filter, Projection};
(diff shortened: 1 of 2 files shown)


crates/sail-plan/src/resolver/expression/subquery.rs
```diff
@@ -66,6 +83,58 @@ impl PlanResolver<'_> {
...
+            let expr = self.resolve_expression(value, schema, state).await?;
+            // Convert column references to OuterReferenceColumn so DataFusion's
+            // decorrelation pass recognizes them as correlated references.
+            let outer_ref = match &expr {
+                Expr::Column(col) => {
+                    let (_, field) = schema.qualified_field_from_column(col)?;
+                    Expr::OuterReferenceColumn(field.clone(), col.clone())
+                }
+                _ => expr,
+            };
```
linhr: This needed to handle columns recursively.

Could you also add a test case for this? I guess `(a + 1, b - 1) IN (...)` might be supported by Spark. I haven't validated this though.

PR #1357, opened 2026-02-11
fix: resolve ON condition columns from both sides in semi/anti joins

davidlghellin:
part #501


linhr (commented):
Nice fix! Once the Copilot comments are addressed, this can be merged.

PR #1373, opened 2026-02-15
chore: prepare v0.5.1

linhr:


python/pysail/tests/spark/glue/conftest.py
```diff
@@ -17,6 +17,9 @@
 if TYPE_CHECKING:
     from collections.abc import Generator
 
+# Skipping the tests at the module level.
+pytest.skip("not working", allow_module_level=True)
```
linhr, line 21: I'm skipping the tests at the module level even though the tests are already skipped individually. It seems the fixtures for Glue catalog interfere with subsequent tests and cause a few failures and errors.

* <https://github.com/lakehq/sail/actions/runs/22028636648>
* <https://github.com/lakehq/sail/actions/runs/22031496628>

But it's unclear to me why this cannot be reproduced in local environments or the manual multi-platform build workflow. More investigation is needed.

PR #1378, opened 2026-02-15
feat: python datasource write mvp

santosh-d3vpl3x:
Add write support for Python data sources in Sail, enabling data to be written through Python implementations.

**Note: Spark Connect `save_type` can be intentionally unset**

In Spark Connect / PySpark 4.1, `WriteOperation.save_type` may be omitted for `df.write.save()` when no path/table destination is provided (for example JDBC/noop-style writes).  
This patch handles that by treating missing `save_type` as an explicit sink destination (not `Path("")`), which matches Spark behavior.

References:
- [SPARK-42733](https://issues.apache.org/jira/browse/SPARK-42733)
- [Spark PR #40356](https://github.com/apache/spark/pull/40356)

### DIFF
--- a/crates/sail-data-source/src/formats/python/arrow_utils.rs
+++ b/crates/sail-data-source/src/formats/python/arrow_utils.rs
@@ -600,13 +600,10 @@ use super::error::{import_cloudpickle, py_err};
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Once;
-
     use super::*;
 
-    fn init_python_once() {
-        static INIT: Once = Once::new();
-        INIT.call_once(Python::initialize);
+    fn init_python() {
+        Python::initialize();
     }
 
     #[test]
@@ -703,7 +700,7 @@ mod tests {
     #[test]
     #[expect(clippy::unwrap_used)]
     fn test_rust_record_batch_to_py() {
-        init_python_once();
+        init_python();
         Python::attach(|py| {
             // Create a simple RecordBatch
             let schema = Arc::new(Schema::new(vec![
@@ -738,7 +735,7 @@ mod tests {
     #[test]
     #[expect(clippy::unwrap_used)]
     fn test_record_batch_to_py_rows() {
-        init_python_once();
+        init_python();
         Python::attach(|py| {
             // Create a simple RecordBatch
             let schema = Arc::new(Schema::new(vec![
@@ -786,7 +783,7 @@ mod tests {
     #[test]
     #[expect(clippy::unwrap_used)]
     fn test_record_batch_to_py_rows_with_nulls() {
-        init_python_once();
+        init_python();
         Python::attach(|py| {
             // Create a RecordBatch with nulls
             let schema = Arc::new(Schema::new(vec![
--- a/crates/sail-data-source/src/formats/python/stream.rs
+++ b/crates/sail-data-source/src/formats/python/stream.rs
@@ -283,7 +283,7 @@ impl PythonDataSourceStream {
 
             // Record total GIL hold time
             metrics.gil_hold_ns.fetch_add(
-                gil_acquired.elapsed().as_nanos().min(u64::MAX as u128) as u64,
+                gil_acquired.elapsed().as_nanos().min(MAX_NANOS_U64) as u64,
                 Ordering::Relaxed,
             );
 
--- a/crates/sail-data-source/src/formats/python/write_exec.rs
+++ b/crates/sail-data-source/src/formats/python/write_exec.rs
@@ -1,55 +1,68 @@
-//! Execution plan for Python data source writes.
+//! Distributed write execution plan for Python data sources.
 //!
-//! This execution plan writes data to a Python datasource using the two-phase
-//! commit protocol: write() → commit() or abort().
-//!
-//! # Architecture
-//!
-//! The write exec is a sink node with one child (the input data plan). It:
-//! 1. Executes the input plan to get RecordBatch streams from all partitions
-//! 2. Feeds batches to Python writer.write() per partition
-//! 3. Collects WriterCommitMessages from all partitions
-//! 4. Calls writer.commit() or writer.abort() based on success/failure
-//!
-//! Only partition 0 orchestrates the entire write operation. Other partitions
-//! return empty streams immediately.
+//! This execution plan runs `writer.write(iterator)` per input partition and
+//! emits one result row per partition. A separate
+//! `PythonDataSourceWriteCommitExec` node consumes these rows and performs
+//! `writer.commit(messages)` or `writer.abort(messages)`.
 //!
 //! # Stateless Writer Boundary
 //!
-//! Each phase (write, commit, abort) deserializes a **fresh** Python writer
-//! instance from the pickled bytes. Writers must not rely on in-memory state
-//! across the write → commit/abort boundary. Commit messages are the only
-//! channel for passing state from write() to commit()/abort(). This matches
-//! PySpark's distributed execution semantics.
-//!
+//! Each phase (write, commit, abort) deserializes a fresh Python writer from
+//! pickled bytes. Writers must not rely on in-memory state across the
+//! write -> commit/abort boundary.
 use std::any::Any;
 use std::fmt;
 use std::sync::Arc;
 
-use arrow::array::RecordBatch;
-use arrow_schema::{Schema, SchemaRef};
+use arrow::array::{BinaryArray, RecordBatch, StringArray, UInt64Array};
+use arrow_schema::{DataType, Field, Schema, SchemaRef};
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
 use datafusion::physical_expr::{EquivalenceProperties, Partitioning};
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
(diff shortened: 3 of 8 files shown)


crates/sail-data-source/src/formats/python/stream.rs
```diff
@@ -277,9 +282,10 @@ impl PythonDataSourceStream {
             }
 
             // Record total GIL hold time
-            metrics
-                .gil_hold_ns
-                .fetch_add(gil_acquired.elapsed().as_nanos() as u64, Ordering::Relaxed);
+            metrics.gil_hold_ns.fetch_add(
+                gil_acquired.elapsed().as_nanos().min(u64::MAX as u128) as u64,
```
linhr: We can use `MAX_NANOS_U64` here as well.

crates/sail-execution/proto/sail/plan/physical.proto
```diff
@@ -592,3 +593,16 @@ message PythonDataSourceExecNode {
...
+message PythonDataSourceWriteExecNode {
+  // Pickled Python DataSourceWriter instance obtained from
+  // DataSource.writer(schema, overwrite), ready for write(), commit(),
+  // and abort() calls (stateless boundary matches write_exec.rs).
+  bytes pickled_writer = 1;
+  // Schema of the data being written
+  bytes schema = 2;
+  // Whether writer is DataSourceArrowWriter (true) or DataSourceWriter (false)
+  bool is_arrow = 3;
+  // Input plan (data to write)
```
linhr: For clarity, we may say "Input plan that generates data to write".

crates/sail-data-source/src/formats/python/arrow_utils.rs
```diff
@@ -233,31 +600,51 @@ use super::error::{import_cloudpickle, py_err};
...
 #[cfg(test)]
 mod tests {
+    use std::sync::Once;
+
     use super::*;
 
+    fn init_python_once() {
+        static INIT: Once = Once::new();
+        INIT.call_once(Python::initialize);
+    }
```
linhr, line 607: You could directly call `Python::initialize()` within each test. This function is backed by a static `Once` already, so it's safe to be called multiple times. `Python::initialize()` is a no-op if the Python interpreter is already initialized.

crates/sail-data-source/src/formats/python/executor.rs
```diff
@@ -89,13 +167,68 @@ pub trait PythonExecutor: Send + Sync + std::fmt::Debug {
...
+    /// Falls back to defaults if config loading fails.
+    pub fn from_app_config() -> Self {
+        Self::try_from_app_config().unwrap_or_else(|_| Self::new())
+    }
+
+    /// Create a new in-process executor from the application config.
+    ///
+    /// Returns an error if config loading fails.
+    pub fn try_from_app_config() -> Result<Self, sail_common::error::CommonError> {
+        let config = sail_common::config::AppConfig::load()?;
```
linhr, line 207: Let me think about how we can pass the `AppConfig` from outside instead of loading it here. This is fine for this PR.

crates/sail-data-source/src/formats/python/write_exec.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// # Arguments
+    ///
+    /// * `input` - Input execution plan providing data to write
+    /// * `pickled_writer` - Pickled Python DataSourceWriter instance
+    /// * `schema` - Schema of the data being written
+    /// * `is_arrow` - Whether writer is Arrow-based (DataSourceArrowWriter)
+    pub fn new(
+        input: Arc<dyn ExecutionPlan>,
+        pickled_writer: Vec<u8>,
+        schema: SchemaRef,
```
linhr: Why do we need to pass the schema separately? Could we use `input.schema()` directly?

crates/sail-data-source/src/formats/python/write_exec.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    /// Get the input plan.
+    pub fn input(&self) -> &Arc<dyn ExecutionPlan> {
+        &self.input
+    }
+
+    /// Get the schema of the data being written.
+    pub fn schema(&self) -> &SchemaRef {
+        &self.schema
+    }
```
linhr: I feel this should be removed since we may accidentally call this function instead of the trait method `ExecutionPlan::schema()` which returns the empty schema for this write node. We can use `self.input().schema()` to access the input schema.

crates/sail-data-source/src/formats/python/write_exec.rs
```diff
@@ -0,0 +1,397 @@
...
+            // Load executor with config from application.yaml (python.data_source_write_channel_capacity,
+            // python.data_source_slow_write_warn_ms). Falls back to defaults if config loading fails.
+            let executor = Arc::new(InProcessExecutor::from_app_config());
+            let num_partitions = input.properties().partitioning.partition_count();
+
+            // Phase 1: Execute writes in parallel for all partitions
+            let mut write_futures = Vec::new();
+
+            for p in 0..num_partitions {
+                let input_stream = input.execute(p, context.clone())?;
```
linhr: For the write to be distributed, we need to split the execution node into two nodes: `PythonDataSourceWriteExec` for write and `PythonDataSourceWriteCommitExec` for commit or abort. 

For the commit node, we need to implement `ExecutionPlan::required_input_distribution` so that the physical optimizer automatically injects `CoalescePartitionsExec` and the commit node receives a single partition containing all the commit messages.

This could be future work after the MVP.

python/pysail/tests/spark/test_python_datasource.py
```diff
@@ -829,3 +832,585 @@ def pushFilters(self, filters):  # noqa: N802
         # Should see both GreaterThan and LessThan filters in the list
         assert "GreaterThan" in pushed_filters_str or "LessThan" in pushed_filters_str
         assert "value" in pushed_filters_str or "id" in pushed_filters_str
+
+
+# ============================================================================
+# Write Support Tests
```
linhr, line 838: The tests rely on class variables to track the methods being called, which doesn't work if the server is running in one terminal while the pytest process is running in a separate terminal (which we sometimes do for development purposes).

But this is just a note. No need to make changes in this PR. (We could use the file system to track the write/commit/abort operations, but this is more work so it could be out of scope for this PR.)

crates/sail-data-source/src/formats/python/stream.rs
santosh-d3vpl3x: Updated to use MAX_NANOS_U64 here as well for consistency with the other nanos->u64 conversion path.

crates/sail-execution/proto/sail/plan/physical.proto
santosh-d3vpl3x: Updated wording to "Input plan that generates data to write" for clarity.

crates/sail-data-source/src/formats/python/executor.rs
santosh-d3vpl3x: Keeping this as-is in this PR. We discussed the config-injection refactor and decided to skip it for now to keep scope focused.

crates/sail-data-source/src/formats/python/arrow_utils.rs
santosh-d3vpl3x: Updated tests to call Python::initialize() directly (removed the extra Once wrapper).

crates/sail-data-source/src/formats/python/write_exec.rs
santosh-d3vpl3x: Addressed: write execution now uses input.schema() directly (the separate write-schema field was removed from PythonDataSourceWriteExec).

crates/sail-data-source/src/formats/python/write_exec.rs
santosh-d3vpl3x: Addressed: removed the custom schema() accessor on PythonDataSourceWriteExec to avoid confusion with ExecutionPlan::schema().

python/pysail/tests/spark/test_python_datasource.py
santosh-d3vpl3x: Acknowledged. No change in this PR (out of scope for MVP); we can revisit with filesystem-backed signaling if we decide to support split-terminal workflows in tests.

crates/sail-data-source/src/formats/python/write_exec.rs
santosh-d3vpl3x: Implemented in this PR: split write into two physical nodes. PythonDataSourceWriteExec now runs distributed partition-local writes and emits per-partition write results; PythonDataSourceWriteCommitExec requires SinglePartition input and performs commit/abort orchestration. Plan/proto/codec were updated accordingly.

crates/sail-execution/src/codec.rs
```diff
@@ -920,13 +921,29 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             }) => {
                 let schema = Arc::new(self.try_decode_schema(&schema)?);
                 let input = self.try_decode_plan(&input, ctx)?;
+                if schema.as_ref() != input.schema().as_ref() {
+                    return plan_err!(
+                        "PythonDataSourceWriteExec schema mismatch: encoded schema does not match input schema"
+                    );
+                }
```
linhr, line 928: We actually don't need to encode input schema in the codec and then we don't need this validation.

linhr (approved):
Nice changes!

I have one minor cleanup suggestion, but this can be done in a follow-up PR.

PR #1388, opened 2026-02-18
fix: preserve start and end fields for day-time interval type

davidlghellin:


crates/sail-plan/src/resolver/data_type.rs
```diff
@@ -129,14 +129,17 @@ impl PlanResolver<'_> {
...
-                //  This may affect the display string in the `data_type_to_simple_string` function.
-                Ok(adt::DataType::Interval(Self::resolve_interval_unit(
+            } => match interval_unit {
+                // Spark's DayTimeInterval has microsecond precision.
+                // Arrow's IntervalUnit::DayTime has millisecond precision.
+                // Use Duration to preserve microsecond precision.
+                spec::IntervalUnit::DayTime => {
+                    Ok(adt::DataType::Duration(adt::TimeUnit::Microsecond))
+                }
+                _ => Ok(adt::DataType::Interval(Self::resolve_interval_unit(
```
linhr: I feel the `resolve_interval_unit` function should be removed to avoid confusion. The logic there needs to move here.

Also, could you update the docs for `IntervalUnit::DayTime` and `DataType::Interval` in `crates/sail-common/src/spec/data_type.rs`? We used to follow the Arrow convention for interval units, but it seems this PR changes this to follow the Spark convention instead.

linhr (approved):
Great! Thanks for having this!

PR #1390, opened 2026-02-19
fix: return null for CAST(date AS numeric) when ANSI mode is disabled

davidlghellin:
Return null for CAST(date AS numeric) to match Spark's legacy mode behavior.                                                                                                                                                                                                                                             

In Spark's legacy cast mode, casting a date to any numeric type (INT, BIGINT, SMALLINT, TINYINT, FLOAT, DOUBLE, DECIMAL) returns null rather than raising an error. 

Related https://github.com/apache/datafusion-comet/pull/3544


linhr (approved):
Nice fix!

PR #1398, opened 2026-02-21
feat: implement st_geomfromwkb, st_geogfromwkb, and st_asbinary functions

james-willis:
## Overview 

These are three of the five new ST functions exposed by Spark 4.1. They only change the column type and validate the wkb structure. This is partly because Sail uses WKB as the internal geometry representation.

Is there a good way to do kind of E2E testing, ie query -> result (and in this case schema)?

## Implementation

**st_geomfromwkb** - Converts WKB binary to Geometry(0) with geoarrow.wkb metadata (SRID:0)
**st_geogfromwkb** - Converts WKB binary to Geography(4326) with geoarrow.wkb metadata (OGC:CRS84)
**st_asbinary** - Strips type metadata from WKB and returns pure binary

WKB Validation: A custom WKB reader was implemented to mirror Spark's validation behavior, which runs at execution time for st_geomfromwkb and st_geogfromwkb:

* Byte order validation
* Geometry type validation (Point, LineString, etc.)
* Dimension support (2D, Z, M, ZM)
* Coordinate validation (non-finite check)
* LineString requires ≥2 points
* Ring requires ≥4 points and must be closed
* Multi* collections validate nested geometry types
* Geography bounds validation (longitude [-180,180], latitude [-90,90])

I chose to implement custom WKB validation instead of using a library because it gives us control over validation details now and in the future if spark changes their rules. Spark implements all their own logic, so it's not as simple as porting JTS logic to GEOS or similar.

### DIFF
--- a/crates/sail-function/src/scalar/geo/st_asbinary.rs
+++ b/crates/sail-function/src/scalar/geo/st_asbinary.rs
@@ -1,0 +1,76 @@
+use std::any::Any;
+use std::sync::Arc;
+
+use datafusion::arrow::datatypes::{DataType, Field};
+use datafusion_common::{exec_err, Result, ScalarValue};
+use datafusion_expr::{
+    ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
+};
+
+/// ST_AsBinary - Convert Geometry/Geography to WKB (Binary)
+///
+/// Input: Binary containing WKB (with geoarrow metadata)
+/// Output: Binary (WKB without metadata)
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct StAsBinary {
+    signature: Signature,
+}
+
+impl Default for StAsBinary {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl StAsBinary {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::exact(vec![DataType::Binary], Volatility::Immutable),
+        }
+    }
+}
+
+impl ScalarUDFImpl for StAsBinary {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn name(&self) -> &str {
+        "st_asbinary"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Binary)
+    }
+
+    fn return_field_from_args(&self, _args: ReturnFieldArgs) -> Result<Arc<Field>> {
+        // st_asbinary strips metadata, so return a plain Binary field
+        Ok(Arc::new(Field::new(self.name(), DataType::Binary, true)))
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let args = args.args;
+
+        if args.len() != 1 {
+            return exec_err!(
+                "st_asbinary requires exactly 1 argument, got {}",
+                args.len()
+            );
+        }
+
+        match &args[0] {
+            ColumnarValue::Scalar(ScalarValue::Binary(Some(b))) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(Some(b.to_vec()))))
+            }
+            ColumnarValue::Scalar(ScalarValue::Binary(None)) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(None)))
+            }
+            ColumnarValue::Array(array) => Ok(ColumnarValue::Array(array.clone())),
+            other => exec_err!("Unsupported argument type for st_asbinary: {:?}", other),
+        }
+    }
+}
--- a/crates/sail-function/src/scalar/geo/st_geogfromwkb.rs
+++ b/crates/sail-function/src/scalar/geo/st_geogfromwkb.rs
@@ -1,0 +1,103 @@
+use std::any::Any;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use datafusion::arrow::datatypes::{DataType, Field};
+use datafusion::common::cast::as_binary_array;
+use datafusion_common::{exec_err, Result, ScalarValue};
+use datafusion_expr::{
+    ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
+};
+
+use super::wkb_reader::validate_geography;
+
+/// ST_GeoGFromWKB - Convert WKB to Geography(4326)
+///
+/// Input: Binary containing WKB
+/// Output: Binary with type Geography(4326)
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct StGeogFromWKB {
+    signature: Signature,
+}
+
+impl Default for StGeogFromWKB {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl StGeogFromWKB {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::exact(vec![DataType::Binary], Volatility::Immutable),
+        }
+    }
+}
+
+impl ScalarUDFImpl for StGeogFromWKB {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn name(&self) -> &str {
+        "st_geogfromwkb"
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::Binary)
+    }
+
+    fn return_field_from_args(&self, _args: ReturnFieldArgs) -> Result<Arc<Field>> {
+        let mut metadata = HashMap::new();
+        metadata.insert(
+            "ARROW:extension:name".to_string(),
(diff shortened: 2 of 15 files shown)


crates/sail-plan/src/resolver/expression/function.rs
```diff
@@ -143,7 +145,40 @@ impl PlanResolver<'_> {
...
+                metadata.insert(
+                    "ARROW:extension:name".to_string(),
+                    "geoarrow.wkb".to_string(),
+                );
+                metadata.insert(
+                    "ARROW:extension:metadata".to_string(),
+                    r#"{"crs":"OGC:CRS84","edges":"spherical"}"#.to_string(),
+                );
+                metadata
+            }
```
james-willis: I know we will need to consider a better structure for this at some point.

james-willis: We can of course add methods to types and functions to provide this metadata, but in the functions case the return type might depend on the input arguments.

For example st_setsrid will return Geometry(4326) when the second argument is a scalar subquery, a literal, or a constant folding result  equal to 4326 (at least literal is definitely true in spark 4.1, need to test the rest). 

This might be challenging to do in the resolver. I'm looking for advice from someone who has a better understanding of sail internals to explain what we will know about the arguments at this point.


My inclination for this PR to leave this switch statement as is and implement something more sophisticated along side the st_setsrid function

crates/sail-function/src/scalar/geo/st_asbinary.rs
```diff
@@ -0,0 +1,70 @@
...
+            );
+        }
+
+        match &args[0] {
+            ColumnarValue::Scalar(ScalarValue::Binary(Some(b))) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(Some(b.to_vec()))))
+            }
+            ColumnarValue::Scalar(ScalarValue::Binary(None)) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(None)))
+            }
```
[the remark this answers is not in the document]
shehabgamin: @james-willis wdyt?

crates/sail-plan/src/resolver/expression/function.rs
shehabgamin: cc @linhr

crates/sail-function/src/scalar/geo/st_asbinary.rs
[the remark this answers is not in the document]
james-willis: we need the validation in the vectorized code path unfortunately. I wish spark was lazy about wkb validation but theyre not.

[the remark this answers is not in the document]
james-willis: this one actually is that simple but the other two need the validation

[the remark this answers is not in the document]
james-willis: done

crates/sail-function/src/scalar/geo/st_geogfromwkb.rs
```diff
@@ -0,0 +1,75 @@
...
+                    return exec_err!("Invalid WKB: {}", e);
+                }
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(Some(b.to_vec()))))
+            }
+            ColumnarValue::Scalar(ScalarValue::Binary(None)) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(None)))
+            }
+            other => exec_err!("Unsupported argument type for st_geogfromwkb: {:?}", other),
+        }
+    }
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-function/src/scalar/geo/st_geomfromwkb.rs
```diff
@@ -0,0 +1,75 @@
...
+                    return exec_err!("Invalid WKB: {}", e);
+                }
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(Some(b.to_vec()))))
+            }
+            ColumnarValue::Scalar(ScalarValue::Binary(None)) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Binary(None)))
+            }
+            other => exec_err!("Unsupported argument type for st_geomfromwkb: {:?}", other),
+        }
+    }
```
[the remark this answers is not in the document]
james-willis: done

crates/sail-plan/src/resolver/expression/mod.rs
```diff
@@ -519,4 +520,116 @@ mod tests {
...
+
+        assert!(
+            metadata_map.contains_key("ARROW:extension:name"),
+            "Expected ARROW:extension:name in metadata, got: {:?}",
+            metadata_map
+        );
+        assert_eq!(
+            metadata_map.get("ARROW:extension:name"),
+            Some(&"geoarrow.wkb".to_string())
+        );
```
linhr: nitpick: I feel the first assertion for `contains_key` is a bit redundant. The assertion for `get` with `Some(_)` as the expected value would be enough.

Same for other occurrences.

crates/sail-function/src/scalar/geo/wkb_reader.rs
```diff
@@ -0,0 +1,1251 @@
...
+        r#type: i32,
+        position: usize,
+    },
+    GeographyBoundsViolation {
+        message: String,
+        position: usize,
+    },
+}
+
+impl fmt::Display for WkbError {
```
linhr: We could derive `thiserror::Error` so that we don't need to implement `Display` manually. The codebase has many examples for this.

This is optional for this PR.

crates/sail-plan/src/resolver/expression/function.rs
linhr: DataFusion recently added `ScalarUDFImpl::return_field_from_args`. If we implement this method instead of `ScalarUDFImpl::return_type`, we would be able to add metadata for the return field.

We want to adopt this pattern in general for all our function implementations, since it is also useful for expressing nullability for the return type. (We haven't migrated many functions to use this pattern though.)

Having the logic in resolver is fine for this PR.

linhr: > when the second argument is a scalar subquery, a literal, or a constant folding result equal to 4326

The resolver is able to distinguish the logical expression type and do constant folding via `LiteralEvaluator`. Once we determine the kind of the second argument, we can then pass the "resolved" information to `StSetSrid::try_new()` when creating the UDF instance.

linhr (approved):
This looks really nice! We can merge this once we fix the merge conflict.

> Is there a good way to do kind of E2E testing, ie query -> result (and in this case schema)?

You could write BDD tests in `python/pysail/tests/spark/function/features`. The `query result` and `query schema` steps would be handy. This could be a follow-up PR.

crates/sail-plan/src/resolver/expression/function.rs
james-willis: Thanks for all the help. I've implemented the changes for `return_field_from_args`

crates/sail-function/src/scalar/geo/wkb_reader.rs
james-willis: thanks for the help cleaning up my rust. done.

crates/sail-plan/src/resolver/expression/function.rs
```diff
@@ -277,3 +292,58 @@ impl PlanResolver<'_> {
...
+            let data_type = arg.get_type(&empty_schema).unwrap_or(DataType::Null);
+            Arc::new(Field::new(format!("arg_{}", i), data_type, true))
+        })
+        .collect();
+
+    // Extract literal scalar values from expressions
+    let mut scalar_values = Vec::new();
+
+    for arg in args {
+        if let expr::Expr::Literal(scalar_value, _) = arg {
```
james-willis, line 322: I'm unsure if I need to handle other scalar types here.

crates/sail-plan/src/resolver/expression/function.rs
```diff
@@ -277,3 +292,58 @@ impl PlanResolver<'_> {
...
+
+    // Try to extract metadata, but don't fail if it doesn't work
+    if let Ok(field) = udf.return_field_from_args(return_field_args) {
+        Ok(field
+            .metadata()
+            .iter()
+            .map(|(k, v)| (k.clone(), v.clone()))
+            .collect())
+    } else {
+        Ok(vec![])
```
james-willis, line 347: Is this too permissive?

linhr: This should be fine for now.

crates/sail-plan/src/resolver/expression/function.rs
linhr: This should be fine for now.

crates/sail-plan/src/resolver/expression/function.rs
```diff
@@ -143,7 +143,22 @@ impl PlanResolver<'_> {
             argument_display_names.iter().map(|x| x.as_str()).collect(),
             is_distinct,
         )?;
-        Ok(NamedExpr::new(vec![name], func))
+
+        // Extract metadata from UDF if it implements return_field_from_args
```
linhr, line 147: This is Ok for now. But I just realized we can revisit the current approach of storing `metadata` and I created an issue #1449 to track this. I guess once we make sure that field information (e.g., metadata, nullability) is preserved in logical plan creation, we may not need to explicitly manage metadata in `NamedExpr`.

linhr (approved):
Thanks for the nice changes!

PR #1402, opened 2026-02-23
fix: convert Int64 to Int32 in CSV schema inference to match Spark behavior

davidlghellin:
part #501


linhr:
Thanks for looking into this!

Sorry I didn't get a chance to respond to this PR. I'm now considering a new interface for reading files so that we can have full control over schema inference etc. We can keep this PR open and the logic in this PR can hopefully be incorporated when the new design is ready.

davidlghellin:
> Thanks for looking into this!
> 
> Sorry I didn't get a chance to respond to this PR. I'm now considering a new interface for reading files so that we can have full control over schema inference etc. We can keep this PR open and the logic in this PR can hopefully be incorporated when the new design is ready.

Thanks for the update! Happy to wait for the new design

PR #1407, opened 2026-02-23
refactor: use datafusion-spark UDFs for bitwise NOT and shiftrightunigned

davidlghellin:


python/pysail/tests/spark/function/features/bitwise.feature
```diff
@@ -0,0 +1,595 @@
...
+
+    Scenario: shiftrightunsigned with negative shift ANSI off
+      Given config spark.sql.ansi.enabled = false
+      When query
+        """
+        SELECT shiftleft(16, -1) AS r1, shiftright(16, -1) AS r2
+        """
+      Then query result
+        | r1 | r2 |
+        | 0  | 0  |
```
lonless9: It seems the Scenario doesn't match this query?

davidlghellin: You're right — I was trying a few extra functions and this one must have slipped in while I was cleaning up.

lonless9 (approved):
LGTM!

linhr:
Is this comment addressed? <https://github.com/lakehq/sail/pull/1407#discussion_r2844900818> It's a bit unclear to me, but I haven't read the entire file so I may be missing something.

python/pysail/tests/spark/function/features/bitwise.feature
lonless9: It seems you mistakenly deleted other tests? https://github.com/lakehq/sail/pull/1407#issuecomment-3950564606

davidlghellin:
> Is this comment addressed? [#1407 (comment)](https://github.com/lakehq/sail/pull/1407#discussion_r2844900818) It's a bit unclear to me, but I haven't read the entire file so I may be missing something.

Sorry for the late reply — yes, that comment has been addressed. The mismatched scenario name was fixed and the accidentally deleted tests were restored in the updated version. The current file has all 72 scenarios intact. Apologies for not responding sooner.

shehabgamin:
@linhr @davidlghellin @lonless9 I actually think we shouldn't use the UDF unless the existing behavior is incorrect. There are no optimizers for these UDFs so there will be a big performance penalty.

davidlghellin:
> @linhr @davidlghellin @lonless9 I actually think we shouldn't use the UDF unless the existing behavior is incorrect. There are no optimizers for these UDFs so there will be a big performance penalty.

Good point, thanks — no reason to pay the optimizer penalty without a concrete correctness issue. Closing.

PR #1408, opened 2026-02-23
fix: handle schemeless URLs in `parse_url` for PATH and FILE components

davidlghellin:
Spark's `java.net.URI` treats schemeless strings (e.g. `'notaurl'`) as relative URIs where the entire input becomes the path component. The Rust `url` crate rejects these with `RelativeUrlWithoutBase`, and the current implementation maps all such errors to `NULL` — but Spark returns the input string for `PATH` and `FILE`.


davidlghellin:
image

PR #1410, opened 2026-02-24
feat: add SQL support for CALL procedures and lakehouse maintenance commands

lonless9:


lonless9:
Some of the assumptions underlying this PR are invalid. Gonna close this and revisit after addressing catalog-related issues.

PR #1413, opened 2026-02-25
feat: poc plugin grpc

davidlghellin:


davidlghellin:
basic use:

```sh
cargo run -p sail-cli -- spark server --port 50051 --plugin-endpoint grpc://localhost:50052

PYTHONPATH=sail-plugins/python hatch run python sail-plugins/python/examples/string_functions.py

export SPARK_REMOTE="sc://localhost:50051"
hatch run python3 -c "
spark = SparkSession.builder.remote('sc://localhost:50051').getOrCreate()
spark.sql(\"SELECT upper_flight('hello'), word_count('one two three')\").show()
"
```

image

PR #1432, opened 2026-02-27
docs: extract comment-only updates from cache-manager work

pomykalakyle:
## Summary
- Split out documentation-only changes from the cache manager branch to make review focused.
- Keep only doc comment updates in execution and plan resolver modules.
- Drop functional/cache-manager implementation changes from this PR.

### DIFF
--- a/crates/sail-execution/src/task_runner/core.rs
+++ b/crates/sail-execution/src/task_runner/core.rs
@@ -74,9 +74,6 @@ impl TaskRunner {
     }
 
     /// Deserializes and prepares a physical plan for execution on this node.
-    ///
-    /// Decodes the plan from protobuf, rewrites parquet adapters, rewrites shuffle
-    /// nodes as needed, and executes the resulting plan on this worker.
     fn execute_plan<T: Actor>(
         &mut self,
         ctx: &mut ActorContext<T>,


crates/sail-execution/src/task_runner/core.rs
```diff
@@ -73,6 +73,11 @@ impl TaskRunner {
         }
     }
 
+    /// Deserializes and prepares a physical plan for execution on this node.
+    ///
+    /// Decodes the plan from protobuf, rewrites parquet adapters, and wraps the plan
+    /// with the appropriate output node (ShuffleWriteExec for normal jobs).
+    /// If the plan's root is a CacheWriteExec, injects the worker's local cache store.
```
linhr: This line around caching needs to be removed.

Actually, I wonder if we should skip this paragraph, since details such as parquet adapters are likely to change frequently in the future, and we don't want to have the extra burden of maintaining the accuracy of the docs for internal functions. As AI is pretty capable of understanding the code itself, the documentation for internal functions may not be critical in many circumstances.

crates/sail-execution/src/driver/actor/handler.rs
```diff
@@ -509,6 +509,11 @@ impl DriverActor {
         }
     }
 
+    /// Assigns pending tasks to available workers and dispatches them for execution.
+    ///
+    /// Gets task assignments from the task assigner, builds task definitions from the job
+    /// scheduler, and dispatches each task to either the driver or a remote worker via gRPC.
+    /// Tasks that fail to build a definition are reported as failed.
```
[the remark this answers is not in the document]
linhr: I feel the original version is fine. The minor ambiguity is not too confusing.

But here is the thing. In many cases, even for human developers, it's just easier to skim through the function implementation than reading both the documentation and the code. Documentation can be vague, while the code is always precise. For AI model, the same holds. It could just read the code to understand everything without consuming extra tokens on less precise and potentially outdated documentation.

crates/sail-execution/src/task_runner/core.rs
pomykalakyle: done

crates/sail-execution/src/driver/actor/handler.rs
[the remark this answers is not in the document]
pomykalakyle: That's fair - code is self documenting. Most of these docstrings are not meant to have the actual details of the function in them, instead are one line summaries so that if I can't tell what the function does from the name itself the one line docstring will tell me what is up with it (this is especially useful on hover)

crates/sail-execution/src/task_runner/core.rs
```diff
@@ -73,6 +73,10 @@ impl TaskRunner {
         }
     }
 
+    /// Deserializes and prepares a physical plan for execution on this node.
+    ///
+    /// Decodes the plan from protobuf, rewrites parquet adapters, rewrites shuffle
+    /// nodes as needed, and executes the resulting plan on this worker.
```
linhr: Let's actually remove this paragraph. The last part is incorrect since the task runner is also used for driver tasks.

pomykalakyle: Sorry about misunderstanding! This is fixed.

PR #1448, opened 2026-02-28
feat: implement bucketed Parquet read with shuffle elimination

davidlghellin:
WIP

## Known limitations
- Distributed execution: codec support is implemented but only single-node execution has been benchmarked. Multi-worker validation is pending.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6871,6 +6871,7 @@ dependencies = [
  "sail-build-scripts",
  "sail-common",
  "sail-common-datafusion",
+ "sail-function",
  "sail-sql-analyzer",
  "serde",
  "thiserror",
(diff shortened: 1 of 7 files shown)


davidlghellin:
https://github.com/davidlghellin/LakeBench/blob/feat/nix_benchmark/examples/sail_comparison.ipynb

```sh
nix develop
python -m jupyter notebook examples/sail_comparison.ipynb
```

local (0.5.1) is sail with changes

image

image

linhr:
Nice exploration! The result looks really promising.

I wonder if we should follow Spark's convention in writing bucketed Parquet files. It seems like Spark doesn't use Parquet metadata to store bucket information. Instead, this information is stored as table metadata in the catalog (`BucketSpec`), and the file name convention is used to infer the bucket ID (see `BucketingUtils`).

I feel that having Parquet metadata (`sail.*`) implies a contract with the user and we then have some maintenance responsibility to ensure the compatibility from now on. So it might be easier to follow the existing convention from Spark. Using existing convention also means that the file written by Sail is interoperable with other systems including the JVM Spark.

But I can see that customizing the file name is not yet possible due to the DataFusion writer in use. We plan to implement our own listing table writer so that we can have better control over file names. Maybe we can revisit the solution when that is ready.

PR #1482, opened 2026-03-06
fix: support non-literal expressions in window RANGE frame boundaries

davidlghellin:
Ibis generates `CAST(0 AS INTERVAL SECOND)` for window RANGE boundaries,
which failed because `resolve_window_boundary` only accepted literals.
Now it resolves arbitrary constant expressions via type coercion +
constant folding + LiteralEvaluator.

part #501

### DIFF
--- a/crates/sail-plan/src/resolver/expression/window.rs
+++ b/crates/sail-plan/src/resolver/expression/window.rs
@@ -238,25 +238,18 @@ impl PlanResolver<'_> {
         if let datafusion_expr::Expr::Literal(scalar, _) = resolved {
             return Ok(scalar);
         }
-        // Apply type coercion and constant folding before evaluating.
+        // Apply type coercion so that expressions like `CAST(0 AS INTERVAL SECOND)`
+        // have compatible types before physical evaluation.
         let props = ExecutionProps::new();
         let context = SimplifyContext::new(&props).with_schema(schema.clone());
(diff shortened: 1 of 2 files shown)


crates/sail-plan/src/resolver/expression/window.rs
```diff
@@ -205,31 +210,63 @@ impl PlanResolver<'_> {
...
+            ))
+        })?;
+        let simplified = simplifier.simplify(coerced).map_err(|e| {
+            PlanError::invalid(format!(
+                "window boundary must be a constant expression: {e}"
+            ))
+        })?;
+        if let datafusion_expr::Expr::Literal(scalar, _) = simplified {
+            return Ok(scalar);
+        }
```
linhr: I'm curious if all these are needed. Could you try just calling `resolve_expression` and pass the logical expression directly to `LiteralEvaluator`? Let me know if I miss anything here!

davidlghellin: Tried removing everything and using just resolve_expression → LiteralEvaluator. The coerce step is still needed because expressions like CAST(0 AS INTERVAL SECOND) produce type mismatches that create_physical_expr can't handle (e.g., Int32 * Int64). But simplify() and the intermediate literal check were indeed redundant — removed those.

PR #1484, opened 2026-03-07
fix: namespace handling for catalog providers

linhr:
When working with catalog providers, it's usually incorrect to use `Namespace` methods to convert the namespace to string (a potentially quoted SQL object name), since each catalog provider has its own contract for working with database names.

This PR corrects various issues for namespace handling for catalog providers. The methods for string conversion is removed from `Namespace` to prevent misuse.

This PR also fixes a few small issues for the AWS Glue catalog, especially around the Iceberg format.

### DIFF
--- a/crates/sail-catalog-glue/src/provider.rs
+++ b/crates/sail-catalog-glue/src/provider.rs
@@ -472,10 +472,16 @@ impl CatalogProvider for GlueCatalogProvider {
         &self,
         database: &Namespace,
         table: &str,
-        options: CreateTableOptions,
+        mut options: CreateTableOptions,
     ) -> CatalogResult<TableStatus> {
         let client = self.get_client().await?;
         let format_lower = options.format.to_lowercase();
+
+        // Skip location or path options since the location is available in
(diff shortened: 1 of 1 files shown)


crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -149,6 +149,18 @@ impl IcebergRestCatalogProvider {
         Ok((client, merged_catalog_config))
     }
 
+    /// Converts a `Namespace` into a string representation for the REST API URL.
+    /// The Iceberg REST API separates namespace components with `\x1F`.
+    fn namespace_string(database: &Namespace) -> String {
+        // TODO: The separator is actually configurable and we should support it as an option.
```
shehabgamin, line 155: Maybe copilot could do this? 😄

linhr: Yeah this could be a separate PR and should be straightforward. I didn't do this since we need to decide on an option name for this, which would become our contract to the user and cannot be changed arbitrarily.

PR #1486, opened 2026-03-07
fix: align string functions with Spark semantics

davidlghellin:
- base64(NULL): handle Null input type in return_type and invoke_with_args
- char/chr(-1): return empty string for invalid codepoints instead of error
- left/right with negative length: return empty string instead of trimming
- substring with negative start: count from end of string


linhr:
@davidlghellin Feel free to resolve the Copilot comments if they are addressed or if they are invalid. And let me know if the latest comments need to be taken care of. I'll take a look at the PR shortly afterwards!

PR #1490, opened 2026-03-07
feat: implement `ANALYZE TABLE` as no-op and list partition values from filesystem

davidlghellin:
part #501


linhr:
For the same reason I mentioned in <https://github.com/lakehq/sail/pull/1494#issuecomment-4065634723>, maybe we can revisit `ANALYZE TABLE` later?

I feel it's fine that some DDL/DML tests are failing in Ibis for now. Then in the future we can consider implementing this properly.

PR #1493, opened 2026-03-09
feat: reconstruct simple CASE form to enable hash-based evaluation

davidlghellin:
Sail's SQL analyzer flattens `CASE col WHEN val THEN result` into `CASE WHEN col = val THEN result` (expr: None), which prevents DataFusion's built-in LiteralLookupTable optimization from activating.

This adds a logical optimizer rule that detects the flattened pattern and reconstructs the simple CASE form (expr: Some(col)), enabling DataFusion's O(1) HashMap lookup instead of O(n) linear evaluation.


davidlghellin:
```sh
❯ cargo bench -p sail-logical-optimizer --bench case_expr
    Finished `bench` profile [optimized] target(s) in 0.28s
     Running benches/case_expr.rs (target/release/deps/case_expr-a90aa295a8ee1097)
Gnuplot not found, using plotters backend
case_10_branches/flattened_linear/100000
                        time:   [4.6325 ms 4.6737 ms 4.7202 ms]
                        change: [-6.1987% -4.9117% -3.5399%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
case_10_branches/optimized_hash/100000
                        time:   [596.46 µs 607.08 µs 621.25 µs]
                        change: [-1.5782% +0.5340% +2.3359%] (p = 0.62 > 0.05)
                        No change in performance detected.
Found 17 outliers among 100 measurements (17.00%)
  2 (2.00%) high mild
  15 (15.00%) high severe

case_50_branches/flattened_linear/100000
                        time:   [13.077 ms 13.206 ms 13.353 ms]
                        change: [-6.8624% -5.3515% -3.8805%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
case_50_branches/optimized_hash/100000
                        time:   [867.90 µs 876.24 µs 885.72 µs]
                        change: [-15.997% -14.853% -13.761%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  3 (3.00%) high mild
  13 (13.00%) high severe

Benchmarking case_200_branches/flattened_linear/100000: Collecting 100 samples in estimated 9.0736 s (200 itecase_200_branches/flattened_linear/100000
                        time:   [44.035 ms 44.832 ms 45.834 ms]
                        change: [-2.9259% -0.9164% +1.7570%] (p = 0.45 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
Benchmarking case_200_branches/optimized_hash/100000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.1s, enable flat sampling, or reduce sample count to 60.
Benchmarking case_200_branches/optimized_hash/100000: Collecting 100 samples in estimated 6.1479 s (5050 itercase_200_branches/optimized_hash/100000
                        time:   [1.1947 ms 1.2049 ms 1.2181 ms]
                        change: [-4.7233% -2.6362% -0.7214%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  5 (5.00%) high severe

Benchmarking case_500_branches/flattened_linear/100000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.5s, or reduce sample count to 50.
Benchmarking case_500_branches/flattened_linear/100000: Collecting 100 samples in estimated 9.5156 s (100 itecase_500_branches/flattened_linear/100000
                        time:   [93.953 ms 94.587 ms 95.314 ms]
                        change: [-4.6116% -3.0624% -1.6512%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe
Benchmarking case_500_branches/optimized_hash/100000: Collecting 100 samples in estimated 9.4775 s (10k iteracase_500_branches/optimized_hash/100000
                        time:   [922.39 µs 930.53 µs 940.19 µs]
                        change: [-3.1393% -1.7374% -0.1990%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe
```

shehabgamin:
@davidlghellin Nice find!!

Looks like there are some merge conflicts

crates/sail-logical-optimizer/src/lib.rs
```diff
@@ -15,5 +17,7 @@ pub fn default_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
 
 pub fn default_optimizer_rules() -> Vec<Arc<dyn OptimizerRule + Send + Sync>> {
     let Optimizer { rules } = Optimizer::default();
+    let mut rules = rules;
+    rules.push(Arc::new(case_expr::ReconstructSimpleCaseExpr));
     rules
```
lonless9: If we want to establish our own custom rules, we need to ensure that we fully list all the built-in logical optimizers in DataFusion and use tests to guarantee that the application order of our optimizers matches that of the upstream. We also need to ensure that we can detect changes in the number and names of optimizer rules when the DataFusion version is upgraded. Additionally, we need to ensure that newly added custom optimizers do not break other rules due to differing responsibilities after being inserted into the rule list. This means we need to have a general understanding of the implementation of other optimization rules. You can refer to how our physical optimizer is set up.

crates/sail-logical-optimizer/src/case_expr.rs
```diff
@@ -0,0 +1,302 @@
...
+use datafusion::optimizer::{OptimizerConfig, OptimizerRule};
+
+/// Reconstructs the simple CASE form from flattened equality conditions.
+///
+/// Sail's SQL analyzer converts `CASE col WHEN val THEN result` into
+/// `CASE WHEN col = val THEN result` (with `expr: None`).
+/// DataFusion has a built-in `LiteralLookupTable` optimization that uses
+/// a HashMap for O(1) per-row evaluation, but it only activates when
+/// `expr: Some(...)`. This rule detects the flattened pattern and
+/// reconstructs the simple form to enable that optimization.
```
lonless9: The comments could be refined. Besides we are currently working on some refinements in testing. The current join reorder is a preliminary example, but it should be migrated in the coming days, so the testing in this area is currently in an unstable state.
Ideally, we would have a switch for this optimizer, and then in BDD tests, we can demonstrate different execution plans by turning this optimizer on and off. Of course we can also simply use the explain codegen mode (which we use to print the output of each optimizer) to see what this optimizer does.

lonless9 (commented):
Great! 🚀 
I have listed some possible improvements below. 
We are currently doing some refactoring of the tests, so let's hear @linhr 's thoughts.

crates/sail-logical-optimizer/src/lib.rs
davidlghellin: I'll also investigate whether this can be fixed at the SQL analyzer level instead — currently the analyzer expands `CASE col WHEN val THEN` ... into `CASE WHEN col = val THEN` ..., which prevents DataFusion's `LiteralLookupTable` optimization from activating. Fixing it at the source would make the optimizer rule unnecessary.

lonless9: Yeah when you check the relevant upstream crates, you can see that upstream recommends folding new optimization rules into the existing ones as much as possible. Therefore, it would be better if this issue could be resolved from aspects other than the optimizer.

PR #1494, opened 2026-03-09
feat: add `ALTER TABLE` support

davidlghellin:
part #501


linhr:
Thanks for looking into this! We are in the process of redesigning our catalog and table provider internals and setting up better test infrastructure for integration tests with external catalog services. Maybe we can revisit this PR after the setup is in a better shape?

PR #1519, opened 2026-03-14
feat: support Arrow-native PySpark UDFs and UDF/UDTF keyword arguments

tamirkifle:


crates/sail-python-udf/src/python/spark.py
```diff
@@ -507,6 +507,64 @@ def __call__(self, args: list[pa.Array], _num_rows: int) -> pa.Array:
...
 
 
+class PySparkScalarArrowUdf:
+    def __init__(
+        self,
+        udf: Callable[..., Any],
+        config,
+    ):
+        self._udf = udf
+        self._serializer = ArrowStreamPandasUDFSerializer(
```
linhr: Is there a reason to use `ArrowStreamPandasUDFSerializer` for all these Arrow UDFs? My understanding is that the UDF only works with Arrow arrays so no Pandas is involved. And I'm curious if the logic around `ChunkedArray` is needed if we do not interact with Pandas.

Let me know if I missed anything.

crates/sail-spark-connect/src/proto/expression.rs
```diff
@@ -235,8 +235,13 @@ impl TryFrom<Expression> for spec::Expr {
...
             }
-            ExprType::NamedArgumentExpression(_) => {
-                Err(SparkError::todo("named argument expression"))
+            ExprType::NamedArgumentExpression(named) => {
+                // Convert named argument (used for Python UDF kwargs)
+                let value = named.value.required("named argument value")?;
+                Ok(spec::Expr::NamedArgument {
+                    key: named.key,
+                    value: Box::new((*value).try_into()?),
+                })
```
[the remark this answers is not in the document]
linhr: We don't need Rust unit tests for this.

crates/sail-plan/src/resolver/expression/window.rs
```diff
@@ -138,21 +139,43 @@ impl PlanResolver<'_> {
...
                 )?;
-                let function = match function.eval_type {
-                    spec::PySparkUdfType::GroupedAggPandas => {
+                let (function, arguments) = match function.eval_type {
+                    spec::PySparkUdfType::GroupedAggPandas
+                    | spec::PySparkUdfType::GroupedAggPandasIter
+                    | spec::PySparkUdfType::GroupedAggArrow
+                    | spec::PySparkUdfType::GroupedAggArrowIter
+                    | spec::PySparkUdfType::WindowAggPandas
+                    | spec::PySparkUdfType::WindowAggArrow => {
```
linhr, line 151: It seems the Arrow UDFs also have the `PySparkGroupAggKind::Pandas` kind. Is this intended?

crates/sail-plan/src/resolver/query/aggregate.rs
```diff
@@ -35,6 +54,55 @@ impl PlanResolver<'_> {
...
+                            .func
+                            .inner()
+                            .as_any()
+                            .downcast_ref::<PySparkGroupAggregateUDF>()
+                            .is_some()
+                        {
+                            if pyspark_agg_name.is_none() {
+                                // Strip the "@hash" suffix to get the Python function name
+                                let full = agg.func.name();
+                                let human = full.split('@').next().unwrap_or(full);
```
linhr: The name extraction logic needs to be moved to a function in `crates/sail-python-udf/src/name.rs`. Otherwise we may forget to make changes here if we ever change the `@` naming to something else.

crates/sail-plan/src/resolver/query/aggregate.rs
```diff
@@ -35,6 +54,55 @@ impl PlanResolver<'_> {
...
+        for proj in &projections {
+            if let Some(name) = find_volatile_in_aggregate_context(&proj.expr) {
+                return Err(PlanError::AnalysisError(format!(
+                    "Non-deterministic expression {name} should not appear in an aggregate query",
+                )));
+            }
+        }
+
+        // Spark CheckAnalysis: GroupedAgg Pandas/Arrow UDFs cannot be mixed with regular
+        // (non-UDF) aggregate functions in the same .agg() call.
```
linhr, line 68: It might be better to move this code block to a function.

crates/sail-spark-connect/src/proto/expression.rs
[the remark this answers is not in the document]
linhr: My understanding is that this is fixed in `resolve_expression_function`. Could you take another look? And feel free to resolve this comment if the logic works as expected.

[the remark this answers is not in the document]
tamirkifle: Yes, fixed in 64ea33af — resolve_expression_function now calls extract_kwargs(arguments) first, peeling out any embedded NamedArgument entries before they reach resolve_expressions_and_names. Resolving this thread.

crates/sail-plan/src/resolver/query/aggregate.rs
tamirkifle: Done in 8798dc3a — added get_udf_display_name to name.rs next to get_udf_name, re-exported it from lib.rs, and replaced the inline split('@') in aggregate.rs.

crates/sail-plan/src/resolver/expression/window.rs
tamirkifle: Good catch, fixed in f2e46a74 — GroupedAggArrow, GroupedAggArrowIter, and WindowAggArrow now get PySparkGroupAggKind::Arrow, derived from eval_type before constructing the UDAF.

crates/sail-plan/src/resolver/query/aggregate.rs
tamirkifle: Done in b8f377a6 — moved to fn check_no_mixed_grouped_agg_udf(projections: &[NamedExpr]) -> PlanResult<()>.

crates/sail-python-udf/src/python/spark.py
tamirkifle: Fixed in dc0fc613 — removed the serializer and Pandas fallback from all three Arrow wrappers. Now raises TypeError for non-pa.Array output. Kept the ChunkedArray.combine_chunks() guard since pyarrow compute ops can return ChunkedArray without any Pandas involvement.

linhr (approved):
Thanks!

PR #1526, opened 2026-03-16
feat: expand catalog traits

lonless9:


lonless9:
Expecting the redesign.

PR #1535, opened 2026-03-18
feat: add CLI command to run PySpark script

linhr:
1. Add the `sail spark run` command to run a single PySpark script.
2. Refactor the implementation of all `sail spark` commands to consolidate the server initialization and cleanup logic.

### DIFF
--- a/crates/sail-cli/src/spark/mcp_server.rs
+++ b/crates/sail-cli/src/spark/mcp_server.rs
@@ -61,7 +61,7 @@ pub fn run_spark_mcp_server(
             let _telemetry = runtime
                 .handle()
                 .primary()
-                .block_on(async { telemetry::TelemetryGuard::new(&config) });
+                .block_on(async { telemetry::TelemetryGuard::try_new(&config) })?;
 
             _run_mcp_server(settings, x.clone())
         }
--- a/crates/sail-cli/src/spark/server.rs
+++ b/crates/sail-cli/src/spark/server.rs
@@ -31,12 +31,12 @@ pub(super) mod telemetry {
     use sail_telemetry::telemetry::{init_telemetry, shutdown_telemetry, ResourceOptions};
 
     pub struct TelemetryGuard {
-        /// A marker to prevent struct creation without calling [`TelemetryGuard::new()`].
+        /// A marker to prevent struct creation without calling [`TelemetryGuard::try_new()`].
         _marker: (),
     }
 
     impl TelemetryGuard {
-        pub fn new(config: &AppConfig) -> Result<Self, Box<dyn std::error::Error>> {
+        pub fn try_new(config: &AppConfig) -> Result<Self, Box<dyn std::error::Error>> {
             let resource = ResourceOptions { kind: "server" };
             init_telemetry(&config.telemetry, resource)?;
             Ok(Self { _marker: () })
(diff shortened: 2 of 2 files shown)


crates/sail-cli/src/spark/run.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::python::Modules;
+use crate::spark::server::with_spark_connect_server;
+
+pub fn run_pyspark_script(file: String) -> Result<(), Box<dyn std::error::Error>> {
+    // We follow the same setup as `run_pyspark_shell`.
+    // Please refer to the comments in that function for details.
+    let (tx, rx) = oneshot::channel::<()>();
+    let address = (IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 0);
+    let shutdown = async {
+        let _ = rx.await;
```
lonless9, line 16: It seems we have three identical copies.

crates/sail-cli/src/runner.rs
```diff
@@ -48,6 +49,22 @@ enum SparkCommand {
...
+            default_value = "-"
+        )]
+        file: String,
+        #[arg(
+            short = 'C',
+            long,
+            help = "The directory to change to before running the script"
+        )]
+        directory: Option<String>,
+    },
```
lonless9, line 67: What if users want to use `sys.argv` or `argparse` to get th trailing arguments? Will they get `sail spark run app.py` instead?

crates/sail-cli/src/runner.rs
```diff
@@ -48,6 +49,22 @@ enum SparkCommand {
         about = "Start the PySpark shell with a Spark Connect server running in the background"
     )]
     Shell,
+    #[command(about = "Run a PySpark script and exit")]
+    Run {
```
lonless9, line 53: Will we have `spark-submit` alias as well?

crates/sail-cli/src/python/spark_run.py
```diff
@@ -0,0 +1,31 @@
+import sys
+from pathlib import Path
+
+from pyspark.sql import SparkSession
```
lonless9, line 4: Besides will we support some arguments like `--venv <path>` or `--requirements <req.txt>` to have better isolation?

linhr: The user can run `sail` command inside a virtual environment for this. I probably won't go further to support `requirements.txt` since dependency management ideally should be done outside the CLI.

crates/sail-cli/src/runner.rs
linhr: I feel `spark-submit` serves a different purpose. The CLI is the entrypoint for the workload itself, while how to submit the workload is handled separately (e.g., creating a pod to invoke the CLI, which can be seen in our K8s example).

crates/sail-cli/src/runner.rs
linhr: Oh good point! This is not supported but can be added in the future. I'd imagine `sail spark run -f script.py -- <args>` to accept arguments for the script itself.

crates/sail-cli/src/spark/run.rs
linhr: Yeah this is the leftover piece of duplicated code after the refactoring, and can be further improved in a follow-up PR.

PR #1539, opened 2026-03-19
refactor: remove redundant arg count checks in invoke_with_args

davidlghellin:
## Summary
- Remove redundant argument count validation from invoke_with_args in 32 UDF implementations
- These checks are already performed by coerce_types or the Signature (user_defined, one_of, exact, String(N)) before invoke_with_args is called

## Changes
- Replace let [a, b] = args.as_slice() else { return Err(...) } with direct indexing let a = &args[0]
- Remove if args.len() != N { return Err(...) } blocks
- Remove inner function arg checks in make_scalar_function delegates
- Clean up unused imports where applicable

## Test plan
- cargo clippy --all-targets --all-features -- -D warnings passes with zero warnings
- cargo nextest run passes all 515 tests
- No behavioral changes — only removal of unreachable code paths

### DIFF
--- a/crates/sail-function/src/scalar/array/spark_array_item_with_position.rs
+++ b/crates/sail-function/src/scalar/array/spark_array_item_with_position.rs
@@ -11,7 +11,6 @@ use datafusion::arrow::datatypes::{
 use datafusion::common::cast::{as_large_list_array, as_list_array};
 use datafusion::common::{DataFusionError, Result};
 use datafusion::logical_expr::{ColumnarValue, ScalarUDFImpl, Signature, Volatility};
-use datafusion_common::exec_err;
 use datafusion_expr::ScalarFunctionArgs;
 
 #[derive(Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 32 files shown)


linhr:
I feel we still want to have the argument count check so that the function implementation is defensive against bugs in DataFusion. It's better to signal the error instead of panicking on out-of-bound indexing in case anything goes wrong inside DataFusion in the future.

But I can see that such guards can never be covered by tests if DataFusion works properly, so it decreases test coverage. I'd imagine using `unreachable!` expresses the intent the best, but I'm not sure if Codecov is able to exclude such lines. (cc @lonless9)

As a related note, I feel the `let [a, b] = args.as_slice() else { ... };` syntax is clear and succinct, and we should encourage this approach in function implementations. I can see that some functions checks `args.len()` and then use indexing to access individual arguments. I feel this is less elegant than pattern matching.

So maybe let's revisit the cleanup once we figure out the best approach?

lonless9:
> I'm not sure if Codecov is able to exclude such lines.

I tried adding some filtering rules to grcov, but codecov might have misstatements in coverage at the PR level. Code blocks on the base commit that are later deleted or filtered out may still be marked as missing. I guess it might only return to normal when merged into the main branch.

PR #1543, opened 2026-03-20
feat: implement SQL time travel support for Delta Lake and Iceberg

lonless9:
part of #171 #172

### DIFF
--- a/crates/sail-plan/src/resolver/query/time_travel.rs
+++ b/crates/sail-plan/src/resolver/query/time_travel.rs
@@ -153,7 +153,7 @@ impl PlanResolver<'_> {
         Ok(resolved)
     }
 
-    // TODO: Extract as general utilities and incorporated into LiteralEvaluator. 
+    // TODO: Extract as general utilities and incorporated into LiteralEvaluator.
     fn validate_time_travel_expression(&self, expr: &Expr, kind: &str) -> PlanResult<()> {
         if expr.any_column_refs() {
             return Err(Self::invalid_time_travel_message(
--- a/python/pysail/testing/spark/steps/iceberg_metadata.py
+++ b/python/pysail/testing/spark/steps/iceberg_metadata.py
@@ -181,7 +181,7 @@ def iceberg_tag_points_to_snapshot_index(
     _write_metadata(table_path, metadata)
 
 
-# FIXME: Remove this workaround once we get a proper solution. 
+# FIXME: Remove this workaround once we get a proper solution.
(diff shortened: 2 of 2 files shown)


python/pysail/tests/spark/iceberg/features/time_travel.feature
```diff
@@ -0,0 +1,145 @@
...
+          value STRING
+        )
+        USING iceberg
+        LOCATION {{ location.uri }}
+        """
+      Given statement
+        """
+        INSERT INTO iceberg_sql_time_travel_schema_test VALUES (1, 'base')
+        """
+      Given append JSON row {"id": 2, "value": "new", "extra": 10} to iceberg table in location with mergeSchema
```
linhr, line 117: Is this step a workaround for now? I assume we would ultimately use standard SQL DML statements to append data. Is it because `mergeSchema` is not supported in `INSERT INTO`? (I guess it would require custom SQL syntax such as SQL hints.)

crates/sail-plan/src/resolver/query/time_travel.rs
```diff
@@ -0,0 +1,356 @@
...
+        let schema = Arc::new(DFSchema::empty());
+        let resolved = self
+            .resolve_expression(expr, &schema, state)
+            .await
+            .map_err(|e| Self::invalid_time_travel_spec(kind, e))?;
+        self.validate_time_travel_expression(&resolved, kind)?;
+        Ok(resolved)
+    }
+
+    fn validate_time_travel_expression(&self, expr: &Expr, kind: &str) -> PlanResult<()> {
```
linhr, line 157: Evaluating constant expressions is a common requirement in query planning, and we have `LiteralEvaluator` for that purpose. I feel the validation and safeguard here (volatile expressions, subquery, limit) looks nice and can be extracted as general utilities and incorporated into `LiteralEvaluator`. 

This is future work.

python/pysail/tests/spark/iceberg/features/time_travel.feature
lonless9: Yes, `INSERT INTO` currently does not support such an API, but Spark 4.1 introduced `MERGE WITH SCHEMA EVOLUTION INTO ...` Therefore, we might design this as a "dialect" in the future. It is also highly likely that I will adopt a similar workaround for alter-related operations during the implementation of subsequent features. Let me add a FIXME here.

crates/sail-plan/src/resolver/query/time_travel.rs
lonless9: Let me add a todo here!

Issue #842, opened 2025-09-06
feat: support remaining lateral join queries


shehabgamin:
To find failing tests:
```
hatch run test-spark.spark-4.0.0:env \
scripts/spark-tests/run-tests.sh \
--pyargs pyspark.sql.tests.connect -v -k 'test_lateral_join'
```

wudihero2:
Hi @shehabgamin,
It seems that DataFusion currently only supports CROSS and INNER lateral joins.
Would it be acceptable for LakeSail to implement support for these first as well?
Please let me know if my understanding is incorrect. Thanks!

shehabgamin:
@wudihero2 Yeah def, makes sense to me! Do you want to take this on?

wudihero2:
@shehabgamin ,
Sounds good. Please assign this to me. Thanks!

PR #1565, opened 2026-03-27
feat: lateral join

wudihero2:
#842
Hi @shehabgamin, I am not sure whether it is okay to tag you, but could you please take a look at this PR and let me know if anything needs to be fixed?

Also, regarding the reports in this PR, I do not fully understand all of the numbers yet. Is there a way to rerun them? I am wondering whether rerunning them might be necessary.

Thank you!

## 1. `test_lateral_join_inside_subquery`

Query under test:

```sql
SELECT * FROM t1
WHERE c1 = (
  SELECT MIN(a)
  FROM t2, LATERAL (SELECT c1 AS a)
)

let df = ctx.sql("SELECT MIN(a) FROM t2, LATERAL (SELECT c1 AS a)").await?;
let plan = df.logical_plan().display_indent_schema();

Projection: max(a) [max(a):Int32;N]
  Aggregate: groupBy=[[]], aggr=[[max(a)]] [max(a):Int32;N]
    Cross Join: [c1:Int32;N, c2:Int32;N, a:Int32;N]
      TableScan: t2 [c1:Int32;N, c2:Int32;N]
      Subquery: [a:Int32;N]
        Projection: outer_ref(t2.c1) AS a [a:Int32;N]
          EmptyRelation: rows=1 []
```
This currently fails with:
```
Error: NotImplemented("Physical plan does not support logical expression OuterReferenceColumn")
```
It looks like this is blocked by datafusion/datafusion/optimizer/src/decorrelate.rs, specifically the _ if plan.contains_outer_reference() branch:
```
impl TreeNodeRewriter for PullUpCorrelatedExpr {
    type Node = LogicalPlan;

    fn f_down(&mut self, plan: LogicalPlan) -> Result<Transformed<LogicalPlan>> {
        match plan {
            ...
            _ if plan.contains_outer_reference() => {
                // unsupported cases: the plan expressions contain outer reference columns
                // (for example, window expressions)
                self.can_pull_up = false;
                Ok(Transformed::new(plan, false, TreeNodeRecursion::Jump))
            }
            _ => Ok(Transformed::no(plan)),
        }
    }
}
```
As a workaround, I added an optimization in sail/crates/sail-logical-optimizer/src/lateral_join.rs to rewrite this logical plan.

I am not sure whether this is the best long-term approach. It may make more sense to implement this in DataFusion itself.

## 2. `SELECT * FROM t1, LATERAL (SELECT max(c2) AS m FROM t2 WHERE t1.c2 < t2.c2)`
Original logical plan:
```
Projection: max(t2.#3) AS m AS #4 [#4:Int32;N]
  Aggregate: groupBy=[[]], aggr=[[max(t2.#3)]] [max(t2.#3):Int32;N]
    Filter: outer_ref(t1.#1) < t2.#3 [#2:Int32;N, #3:Int32;N]
      Projection: t2.c1 AS #2, t2.c2 AS #3 [#2:Int32;N, #3:Int32;N]
        Projection: t2.#2 AS c1, t2.#3 AS c2 [c1:Int32;N, c2:Int32;N]
          SubqueryAlias: t2 [#2:Int32;N, #3:Int32;N]
            SubqueryAlias: t2 [#2:Int32;N, #3:Int32;N]
              Projection: #0 AS #2, #1 AS #3 [#2:Int32;N, #3:Int32;N]
                Projection: column1 AS #0, column2 AS #1 [#0:Int32;N, #1:Int32;N]
                  Values: (Int32(0), Int32(2)), (Int32(0), Int32(3)) [column1:Int32;N, column2:Int32;N]
```
Sail reports:
```
Schema error: No field named t2.column2. Valid fields are t2."#3"
```
DataFusion native reports:
```
NotImplemented("Physical plan does not support logical expression OuterReferenceColumn(...)")
```
DecorrelateLateralJoin attempts to decorrelate this plan, but can_pullup_over_aggregation currently only supports =.
Because this query uses <, it sets can_pull_up = false and returns Transformed::no, so the plan remains unchanged.
After that, other optimizer rules (such as OptimizeProjections) still rewrite the subquery plan, which appears to trigger the schema mismatch.
To make the behavior more stable, I collapse nested projections(fn collapse_projections) first and simplify the logical plan:
```
Projection: max(t2.#3) AS m AS #4 [#4:Int32;N]
  Aggregate: groupBy=[[]], aggr=[[max(t2.#3)]] [max(t2.#3):Int32;N]
    Filter: outer_ref(t1.#1) < t2.#3 [#2:Int32;N, #3:Int32;N]
      Projection: column1 AS #2, column2 AS #3 [#2:Int32;N, #3:Int32;N]
        SubqueryAlias: t2 [column1:Int32;N, column2:Int32;N]
          Values: (Int32(0), Int32(2)), (Int32(0), Int32(3)) [column1:Int32;N, column2:Int32;N]
```
After this rewrite, Sail now fails with the same error as native DataFusion:
```
Physical plan does not support logical expression OuterReferenceColumn
```
This seems better than the previous schema mismatch, since both Sail and DataFusion now fail for the same underlying reason.

## 3. Support wildcard expansion in resolve_wildcard_or_nested_field_wildcard

Add support for cases like:
```
SELECT t1.*
FROM t1
LATERAL JOIN (SELECT t1.*)
```

### DIFF
--- a/crates/sail-logical-optimizer/src/lib.rs
+++ b/crates/sail-logical-optimizer/src/lib.rs
@@ -18,6 +18,12 @@ pub fn default_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
 
 pub fn default_optimizer_rules() -> Vec<Arc<dyn OptimizerRule + Send + Sync>> {
     let Optimizer { rules } = Optimizer::default();
+    // Custom rules are prepended so they run before DataFusion's built-in rules.
+    // `DecorrelateLateralProjection` must run before `DecorrelateLateralJoin`
+    // because it handles the simple case where OuterRef only appears in
+    // Projection expressions (e.g. `LATERAL (SELECT t1.a + 1)`), rewriting
+    // it into a CrossJoin + Projection. The remaining complex cases (OuterRef
+    // in Filter/Aggregate) are left for DataFusion's `DecorrelateLateralJoin`.
     let mut custom: Vec<Arc<dyn OptimizerRule + Send + Sync>> =
         vec![Arc::new(DecorrelateLateralProjection::new())];
     custom.extend(rules);


shehabgamin:
> Hi @shehabgamin, I am not sure whether it is okay to tag you, but could you please take a look at this PR and let me know if anything needs to be fixed?

@wudihero2 Thanks so much for this contribution!!

Yep, totally okay to tag me. I will review thoroughly over the next 2 days.

Also, you can re-run the reports by adding the commit message "[spark tests]" or by adding the tests labels

crates/sail-logical-optimizer/src/lib.rs
```diff
@@ -15,5 +18,8 @@ pub fn default_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
 
 pub fn default_optimizer_rules() -> Vec<Arc<dyn OptimizerRule + Send + Sync>> {
     let Optimizer { rules } = Optimizer::default();
-    rules
+    let mut custom: Vec<Arc<dyn OptimizerRule + Send + Sync>> =
+        vec![Arc::new(DecorrelateLateralProjection::new())];
+    custom.extend(rules);
```
shehabgamin, line 29: Order matters here, see:
https://github.com/apache/datafusion/blob/c17c87c575be916c6501df8a588e3c2614a9dc6f/datafusion/optimizer/src/optimizer.rs#L268-L311

You can also checkout how Sail handles this for physical optimizes 
https://github.com/wudihero2/sail/blob/3c19b6fe869060ceaefa399d33a6c4443ea08bc0/crates/sail-physical-optimizer/src/lib.rs

crates/sail-logical-optimizer/src/lateral_join.rs
```diff
@@ -0,0 +1,173 @@
...
+        if !proj.expr.iter().any(contains_outer_ref) {
+            return Ok(Transformed::no(plan));
+        }
+        if !proj.input.all_out_ref_exprs().is_empty() {
+            return Ok(Transformed::no(plan));
+        }
+
+        let left_field_count = join.left.schema().fields().len();
+
+        let LogicalPlan::Join(join) = plan else {
```
shehabgamin, line 69: Why do we do this again?

crates/sail-logical-optimizer/src/lateral_join.rs
```diff
@@ -0,0 +1,173 @@
...
+        let right_unwrapped = match Arc::unwrap_or_clone(join.right) {
+            LogicalPlan::Subquery(sq) => Arc::unwrap_or_clone(sq.subquery),
+            other => other,
+        };
+        let right_inner = match right_unwrapped {
+            LogicalPlan::SubqueryAlias(sa) => Arc::unwrap_or_clone(sa.input),
+            other => other,
+        };
+
+        let LogicalPlan::Projection(proj) = right_inner else {
```
shehabgamin, line 88: Is this guaranteed to always be a projection?

crates/sail-logical-optimizer/src/lateral_join.rs
```diff
@@ -0,0 +1,173 @@
+use std::sync::Arc;
+
+use datafusion::optimizer::{ApplyOrder, OptimizerConfig, OptimizerRule};
+use datafusion_common::tree_node::{Transformed, TransformedResult, TreeNode};
```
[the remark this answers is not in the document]
wudihero2: `TransformedResult` is actually used at `.transform(...).data()?`.
The `.data()` method is provided by the `TransformedResult` trait to unwrap `Result<Transformed<T>>` into `Result<T>`. Removing the import causes a
compilation error.

crates/sail-logical-optimizer/src/lib.rs
wudihero2: Yes, the ordering is important — `DecorrelateLateralProjection` is placed before
DataFusion's built-in rules so it runs first. It handles the simpler case where
`OuterRef` only appears in `Projection` expressions (for example, `SELECT t1.a + 1`
over an `EmptyRelation`), rewriting it into a `CrossJoin` + `Projection` that does
not require DataFusion's dependent-join decorrelation.

If you have any concerns about the current ordering, please feel free to let me know.
I’d be happy to clarify further or add a comment to document the intent more explicitly.

crates/sail-logical-optimizer/src/lateral_join.rs
wudihero2: This is needed because of Rust's ownership rules. The first match at line 31 is
a borrow (`&plan`) for read-only access. Here at line 69 we need to move out of
`plan` to take ownership of `join.left` and `join.right` for `Arc::unwrap_or_clone`.

crates/sail-logical-optimizer/src/lateral_join.rs
wudihero2: It's guaranteed by the early-return at line 54-56: if `after_alias` is not a Projection, we return `Transformed::no(plan)` and never reach this point.

wudihero2:
> @wudihero2 This is really well done!!
> 
> I just have a couple minor comments/questions.

@shehabgamin I’ve made some changes based on the review comments and also replied to a few of them. Could you please take another look at the PR when you have a chance? Thank you!

crates/sail-logical-optimizer/src/lib.rs
shehabgamin: @wudihero2 Got it, thanks for explaining :)

Yeah a simple comment in the code should be good.

crates/sail-logical-optimizer/src/lateral_join.rs
shehabgamin: Ahh i missed that, makes sense.

wudihero2:
@shehabgamin Done!

PR #1566, opened 2026-03-27
feat: new function datafusion 53

davidlghellin:


davidlghellin:
solved in https://github.com/lakehq/sail/pull/1897 json_tuple

PR #1572, opened 2026-03-29
feat: implement `sentences` string function

rafafrdz:
Implement the Spark SQL `sentences(str[, lang, country])` function that splits text into an array of sentences, each being an array of words. The lang and country parameters are accepted but currently ignored (locale-independent splitting).

Ref #508

### DIFF
--- a/crates/sail-function/src/scalar/string/spark_sentences.rs
+++ b/crates/sail-function/src/scalar/string/spark_sentences.rs
@@ -61,7 +61,7 @@ impl ScalarUDFImpl for SparkSentences {
 
     fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
         let ScalarFunctionArgs { args, .. } = args;
-        // We only use the first argument (str); lang and country are ignored.
+        // TODO: Support locale-aware sentence splitting using the lang and country parameters.
         let arr = match &args[0] {
             ColumnarValue::Array(arr) => arr.clone(),
             ColumnarValue::Scalar(scalar) => scalar.to_array()?,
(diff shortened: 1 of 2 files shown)


crates/sail-function/src/scalar/string/spark_sentences.rs
```diff
@@ -0,0 +1,260 @@
...
+            let text = if let Some(a) = string_arr {
+                a.value(i)
+            } else if let Some(a) = large_string_arr {
+                a.value(i)
+            } else {
+                return Err(generic_internal_err(
+                    SparkSentences::NAME,
+                    "Could not downcast argument to string array",
+                ));
+            };
```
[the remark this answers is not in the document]
rafafrdz: Fixed — now casting to `Utf8` upfront via `arrow::compute::cast`, which handles Utf8, LargeUtf8, and Utf8View uniformly.

crates/sail-function/src/scalar/string/spark_sentences.rs
```diff
@@ -0,0 +1,260 @@
...
+            builder.append(true);
+        }
+    }
+    let array: ListArray = builder.finish();
+    Ok(Arc::new(array))
+}
+
+/// Split text into sentences, then each sentence into words.
+/// Approximates Java's BreakIterator behavior:
+/// - Sentences are split on `.`, `!`, `?` followed by whitespace or end of string
```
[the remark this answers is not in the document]
rafafrdz: Right, updated the doc comment to match the actual behavior.

crates/sail-plan/src/function/scalar/string.rs
```diff
@@ -284,7 +285,7 @@ pub(super) fn list_built_in_string_functions() -> Vec<(&'static str, ScalarFunct
         ("right", F::binary(expr_fn::right)),
         ("rpad", F::var_arg(expr_fn::rpad)),
         ("rtrim", F::var_arg(rev_args(expr_fn::rtrim))),
-        ("sentences", F::unknown("sentences")),
+        ("sentences", F::udf(SparkSentences::new())),
```
linhr, line 288: Could you register this UDF in `crates/sail-execution/src/codec.rs` as well?

crates/sail-function/src/scalar/string/spark_sentences.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+    if !current_word.is_empty() {
+        words.push(current_word);
+    }
+
+    words
+}
+
+#[cfg(test)]
```
linhr: We usually don't write Rust unit tests for functions. Let's move all these tests to `test_sentences.txt`. Or you can write BDD tests in `.feature` files.

crates/sail-function/src/scalar/string/spark_sentences.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(DataType::List(Arc::new(Field::new_list_field(
+            DataType::List(Arc::new(Field::new_list_field(DataType::Utf8, true))),
+            true,
+        ))))
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let ScalarFunctionArgs { args, .. } = args;
+        // We only use the first argument (str); lang and country are ignored.
```
linhr: Could you add a TODO comment here?

Or maybe we should return an error if additional arguments are provided? Silently ignoring them may cause confusion for the user.

crates/sail-plan/src/function/scalar/string.rs
rafafrdz: Done — registered `SparkSentences` in both `try_decode_udf()` and `try_encode_udf()` in `codec.rs`.

crates/sail-function/src/scalar/string/spark_sentences.rs
rafafrdz: Done — removed the Rust unit tests. The `test_sentences.txt` doctest file already covers the key scenarios.

crates/sail-function/src/scalar/string/spark_sentences.rs
rafafrdz: Added a TODO comment: `// TODO: Support locale-aware sentence splitting using the lang and country parameters.` I kept accepting the extra arguments rather than erroring since Spark itself accepts them silently — returning an error would break compatibility with existing Spark SQL queries that pass locale arguments.

PR #1578, opened 2026-03-30
refactor: reorganize Spark Python tests into topic subdirectories

pomykalakyle:
Reorganizing test suite to improve readability. Currently has a huge number of test files directly in the python spark tests to the point where it's hard to find things. 

## Summary
- Move 37 root-level test files in `python/pysail/tests/spark/` into logical subdirectories: `table_format/` (delta, iceberg, glue), `datasource/`, `query/`, `function/`, `dataframe/`, `session/`, and `udf/`
- Fix `Path(__file__)`-based data lookups in `test_tpcds.py` and `test_tpch.py` that broke due to the extra directory nesting
- Update iceberg utility imports to reflect the new `table_format/iceberg/` path


linhr:
Thanks for taking a look at this. This is going to conflict with ongoing PRs such as #1576, so would you mind closing this for now?

But I can see that the test organization does need improvements. Some thoughts:

1. We are going to rewrite all function tests via BDD, so all the function tests will be cleaned up at that time.
2. Delta and Iceberg are complex and important features. They might deserve their own top-level directories.
3. We are moving away from the concept of "table format", and have more tests around catalog providers. So Glue tests will get a better location when that happens.
4. The classification of DataFrame, queries, and UDT/UDTF tests are kind of blurred. We might get more ideas around test organization as we add tests coverage for the cluster mode. 
5. It is fine to have some testing at the root level, such as TPC tests.

I am going to incorporate the idea in this PR into future changes I'm going to make as I update our test setup.

PR #1631, opened 2026-04-09
test: pytest container

lonless9:
Also implemented `Describe Database` and fixed a `Drop Database` issue.


python/pysail/tests/spark/catalog_integration/conftest.py
```diff
@@ -0,0 +1,101 @@
...
+    catalog_list: str,
+    extra_env: dict[str, str] | None = None,
+) -> tuple[SparkConnectServer, str, dict[str, str | None]]:
+    """Start a Sail Spark Connect server with the given catalog configuration.
+
+    Returns ``(server, remote_url, saved_env)`` where *saved_env* contains
+    the original environment values so the caller can restore them later.
+    """
+    # Deferred import so pysail._native is not loaded at module collection time.
+    from pysail.spark import SparkConnectServer
```
linhr, line 35: This is likely not needed. Previously I was trying to see if we can load config during import (#1302), but that feature was reverted in #1319 (as this would be a breaking behavior change) and was never released, so such local import would not be needed. We now still load application configuration from environment variables on every server start. 

No need to make changes in this PR though, since I plan to do some consolidation for config loading in Sail server fixtures. The note is just an FYI.

python/pysail/tests/spark/catalog_integration/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+    this_dir = os.path.dirname(os.path.abspath(__file__))
+    markexpr: str = getattr(config.option, "markexpr", "") or ""
+
+    remaining: list[pytest.Item] = []
+    deselected: list[pytest.Item] = []
+    for item in items:
+        if str(item.fspath).startswith(this_dir):
+            item.add_marker(pytest.mark.catalog_integration)
+            if "catalog_integration" not in markexpr:
+                deselected.append(item)
```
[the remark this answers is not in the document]
linhr: This comment may need attention.

python/pysail/tests/spark/catalog_integration/iceberg_rest/features/namespace.feature
```diff
@@ -0,0 +1,261 @@
+Feature: Iceberg REST catalog namespace (database) operations
```
linhr, line 1: Do we want test cases with multi-level namespace names?

python/pysail/tests/spark/catalog_integration/iceberg_rest/features/view.feature
```diff
@@ -0,0 +1,146 @@
...
+      DROP VIEW iceberg_view_test.nonexistent_drop_view
+      """
+
+  Scenario: Drop non-existent view with IF EXISTS does not raise error
+    Given statement
+      """
+      DROP VIEW IF EXISTS iceberg_view_test.nonexistent_drop_view
+      """
+
+  Scenario: List views in a namespace
```
linhr, line 124: This scenario does not list views.

Maybe we can do another pass of AI review locally for the `.feature` files and see if any issues can be identified and fixed? I've skimmed through them manually but I likely won't catch all the minor things.

python/pysail/tests/spark/catalog_integration/glue/features/table.feature
```diff
@@ -0,0 +1,265 @@
...
+  Scenario: Invalid DDL is rejected
+    When query
+      """
+      CREATE TABLE unsupported_type (col_union UNIONTYPE<INT, STRING>)
+      USING parquet
+      LOCATION 's3://bucket/unsupported'
+      """
+    Then query error .*
+
+  Scenario Outline: Create table with <fmt> storage format
```
linhr: Oh the `Scenario Outline` is nice!

This seems to generate tests such as `test_create_table_with_fmt_storage_format[parquet]`. Maybe we can remove `<fmt>` from the title so that `_fmt_` won't appear in the test name. The test name has the parameter at the end already.

linhr (approved):
This is exciting!

I have a few comments and the Copilot comments look reasonable as well. But overall this looks good!

python/pysail/tests/spark/catalog_integration/iceberg_rest/features/namespace.feature
lonless9: We can pay more attention when adding integration tests later.

python/pysail/tests/spark/catalog_integration/iceberg_rest/features/view.feature
lonless9: I made some loops. It might help complete this part better when removing unit tests.

python/pysail/tests/spark/catalog_integration/glue/features/table.feature
lonless9: Done!

python/pysail/tests/spark/catalog_integration/conftest.py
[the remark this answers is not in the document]
lonless9: Let's temporarily use this behavior to see how to adjust it later.

linhr (approved):

PR #1643, opened 2026-04-10
fix: add remote codec support for relaxed tz cast

pomykalakyle:
## Summary
- register `RelaxedTzCastExec` in the remote physical plan proto
- add encode and decode support in `RemoteExecutionCodec`
- align the execution node API with the surrounding codec access pattern via `input()`


linhr (approved):
Thanks for the follow-up!

PR #1670, opened 2026-04-14
feat: add try_* datetime functions, unify under safe flag pattern

davidlghellin:
Unify Spark{Date,Time,Timestamp,MakeInterval,VariantGet} under a single
`safe: bool` flag (matching #1647's SparkVariantGet pattern). Each struct
exposes both the strict and safe variants of its conversion via the flag,
removing the dedicated SparkTryToDate/SparkTryToTime/SparkTryToTimestamp
structs (~500 lines of duplicated logic).


davidlghellin:
Closing in favor of fresh, per-function PRs on current `main`. This branch is ~128 commits behind the DF54 bump and conflicts hard in `proto`/`codec`/datetime UDFs; the `safe`-flag design is sound but `main` already adopted that pattern (#1879, abs/pmod/ceil). Reopening `try_to_date`/`try_to_time`/… smaller, each JVM-validated. Design + `.feature` scenarios carry over.

PR #1722, opened 2026-04-18
fix: arrays_zip NULL and zero-arg handling

davidlghellin:
Fixes two correctness bugs in `arrays_zip` to match Spark's `ArraysZip` contract,
expands test coverage from 18 to 42 scenarios, and applies local arrow-check
hygiene (with_capacity + uniform-length fast-path)


linhr (approved):
Nice!

PR #1729, opened 2026-04-19
fix: correct ceil/floor edge-case bugs and ANSI overflow behavior

davidlghellin:
## Summary

- `ceil(NULL)` / `floor(NULL)` → NULL (was: coercion error)
- `ceil(x, scale)` where `scale > input_scale` → value unchanged (was: multiplied by 10^diff)
- `ceil(NaN|±Infinity, scale)` → NULL (was: cast overflow error)
- `ceil(x, -38)` → error (was: silent overflow producing garbage)
- ANSI mode: overflow on `ceil`/`floor` with negative scale now errors correctly

## Test plan

- Feature scenarios covering NULL, NaN/Inf, multi-row, ANSI, boundary, aliases
- All validated against Spark JVM


crates/sail-execution/src/proto/codec.rs
```diff
@@ -2436,8 +2442,8 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             }
             "spark_to_chrono_fmt" => Ok(Arc::new(ScalarUDF::from(SparkToChronoFmt::new()))),
             "spark_expm1" | "expm1" => Ok(Arc::new(ScalarUDF::from(SparkExpm1::new()))),
-            "spark_ceil" | "ceil" => Ok(Arc::new(ScalarUDF::from(SparkCeil::new()))),
-            "spark_floor" | "floor" => Ok(Arc::new(ScalarUDF::from(SparkFloor::new()))),
+            "spark_ceil" | "ceil" => Ok(Arc::new(ScalarUDF::from(SparkCeil::new(false)))),
+            "spark_floor" | "floor" => Ok(Arc::new(ScalarUDF::from(SparkFloor::new(false)))),
```
shehabgamin: @davidlghellin Are these two lines needed at all? Might be a previous bug and looks like they can be removed

shehabgamin (commented):
@davidlghellin Feel free to tag me for re-review, thanks!

crates/sail-execution/src/proto/codec.rs
davidlghellin: Good catch — removed.

crates/sail-function/src/scalar/math/spark_ceil_floor.rs
```diff
@@ -22,23 +24,31 @@ use crate::error::{
...
 use crate::scalar::math::utils::decimal::round_decimal_base;
 
+fn extract_call_args<'a>(
+    name: &str,
+    args: &'a [ColumnarValue],
+) -> Result<(&'a ColumnarValue, &'a Option<i32>)> {
+    match args.len() {
+        1 => Ok((&args[0], &None)),
+        2 => match &args[1] {
+            ColumnarValue::Scalar(ScalarValue::Int32(value)) => Ok((&args[0], value)),
```
shehabgamin, line 34: We should be flexible on the integer value here. Any INT/UINT should be acceptable.

shehabgamin (commented):
Noticed 1 more small thing, pr is looking pretty solid overall though!

crates/sail-function/src/scalar/math/spark_ceil_floor.rs
shehabgamin: Never mind, disregard this comment. This is handled by coerce_types

shehabgamin (approved):
@davidlghellin Please disregard my previous review. This looks great!!

PR #1742, opened 2026-04-22
feat: make Python UDF output type and UDTF return type optional in pr…

shehabgamin:


linhr:
The UDTF return type issue was fixed in #1740. This PR reveals the underlying problem after having optional output types, but does not make substantial progress on feature implementation. I feel this PR can be closed.

Issue #1574, opened 2026-03-29
[Feature Request] Delta Lake: support user-defined commit metadata


keen85:
Hello @lonless9,
[OSS Spark Delta Lake implementation](https://docs.delta.io/delta-batch/#set-user-defined-commit-metadata), [Databricks Delta Lake implementation](https://docs.databricks.com/aws/en/delta/custom-metadata) as well as [delta-rs](https://delta-io.github.io/delta-rs/api/transaction/#deltalake.CommitProperties) allow specifying custom commit metadata that will end up in the commit file (`_delta_log/0001.json`) under key `commitInfo.userMetadata`.

In Spark Delta Lake, there are two ways for specifying the user-defined commit metadata:
1. **indirectly via Spark Session Config**:
```python
spark.conf.set("spark.databricks.delta.commitInfo.userMetadata", "my general custom metadata")  # this will be applied to all following Delta Lake transactions - if not overwritten by DataFrameWrite option

df.write.format("delta").mode("overwrite").save("/tmp/delta/people10m")
```
2. **explicitly via `DataFrameWriter` option**:
```python
df.write.format("delta") \
  .mode("overwrite") \
  .option("userMetadata", "my specific custom metadata") \
  .save("/tmp/delta/people10m")
```

I'd really appreciate if Sail would support this as well.

We use this feature in order to "label" commits with some internal process id and a runtime id to make it easier to cross reference commits with logging data later on for better traceability.

Maybe should be part of #171

keen85:
solved in `pysail>=0.6.1`

PR #1752, opened 2026-04-23
feat: support user-defined metadata for Delta Lake

lonless9:
#1574
part of #171 

Session level configuration might need some design consideration.


linhr (approved):
Nice!

PR #1768, opened 2026-04-23
feat: add Hive Metastore catalog provider with HA and Kerberos support

santosh-d3vpl3x:
## Summary

Add a new `sail-catalog-hms` crate that connects Sail to an external Hive Metastore service over Thrift. The provider supports plain and Kerberos-secured connections, HA endpoint failover, and configurable SASL QOP.

### New crate: `sail-catalog-hms`

**Core provider** (`provider.rs`, `convert.rs`, `data_type.rs`):
- Full `CatalogProvider` implementation: databases, tables, views, partition columns
- `alter_table` implemented via fetch-mutate-alter RPC pattern (fetches existing table, applies property changes, calls `alter_table` — matching Spark's approach)
- HA URI lists with automatic failover on transport errors
- Stable error-prefix classification for retry logic (`"transport error"`, `"gssapi error"`, `"dns error"`) instead of fragile library-specific substring matching
- Per-endpoint `_HOST` expansion for Kerberos principals
- Mutation-result normalization (AlreadyExists/NotFound) on retried creates/drops
- Configurable connect timeout, comma-separated URI parsing, `thrift://` scheme normalization (case-insensitive)

**Kerberos / SASL transport** (`security/gssapi.rs`, `security/thrift_sasl.rs`):
- GSSAPI via `libloading` (runtime-loaded, no compile-time Kerberos dependency)
- Full SASL handshake state machine with Thrift SASL framing (START/OK/COMPLETE)
- Configurable QOP floor (`min_sasl_qop = auth | auth_int | auth_conf`)
- Strongest-available layer selection during negotiation
- `GssapiFrameProtector` for post-handshake `wrap`/`unwrap` with `auth_int` / `auth_conf`
- Polled `SaslReadHalf` / `SaslWriteHalf` for both wrapped and passthrough frames
- GSSAPI buffer leak fixes: `GssBuf::release()` calls `gss_release_buffer` after copying data from `gss_init_sec_context`, `gss_wrap`/`gss_unwrap`, and `gss_display_status` output buffers
- Async mutex fix: `rotate_client_after_failure` drops the lock before awaiting client build, reacquires only to update state
- Attribution: GSSAPI session implementation inspired by [hdfs-native](https://github.com/Kimahriman/hdfs-native) (Apache License 2.0)

**Type conversions** (`data_type.rs`):
- Bidirectional Arrow <-> Hive type mapping (structs, maps, arrays, decimals including Decimal32/Decimal64/Decimal128/Decimal256)

### Changes to existing crates

- **`sail-catalog`**: New `hive_format` module (extracted from Glue, shared HMS format detection + `_HOST` expansion). `TextInputFormat` fallback returns `TextFile` instead of `Csv`. Unit tests moved inline per Rust convention.
- **`sail-catalog-glue`**: Deleted `format.rs`, now uses shared `hive_format`
- **`sail-common`**: `CatalogType::HiveMetastore` variant with config aliases (`hive_metastore`, `hms`, `hive-metastore`), `min_sasl_qop`, `connect_timeout_secs`
- **`sail-session`**: HMS catalog wiring

### Devcontainer and CI

- Devcontainer: Added `docker-outside-of-docker` feature, `krb5-user`, `libgssapi-krb5-2`
- CI: Ignored test step installs `krb5-user` + `libgssapi-krb5-2`; runs when PR has `run catalog tests` label or commit message contains `[catalog test]`

### Documentation

- HMS catalog user guide (`docs/guide/catalog/hms.md`)
- HMS development notes with GSSAPI bindgen regeneration instructions (`docs/development/recipes/hms.md`)

## Configuration

```bash
# Plain HMS
export SAIL_CATALOG__LIST='[{type="hive_metastore", name="sail", uris=["127.0.0.1:9083"]}]'

# HA with failover
export SAIL_CATALOG__LIST='[{type="hms", name="sail", uris=["hms1.internal:9083","hms2.internal:9083"], thrift_transport="framed", connect_timeout_secs=10}]'

# Kerberos
export SAIL_CATALOG__LIST='[{type="hms", name="sail", uris=["hms.internal:9083"], auth="kerberos", kerberos_service_principal="hive-metastore/_HOST@EXAMPLE.COM", min_sasl_qop="auth_int", thrift_transport="framed"}]'
```

Configuration options:
- `type` (required): `hive_metastore` or aliases `hms`, `hive-metastore`
- `name` (required): catalog name
- `uris` (required): list of HMS endpoints (`host:port` or `thrift://host:port`, comma-separated)
- `thrift_transport` (optional): `buffered` (default) or `framed`
- `auth` (optional): `none` (default) or `kerberos`
- `kerberos_service_principal` (required when `auth = "kerberos"`): e.g. `hive-metastore/_HOST@REALM`
- `min_sasl_qop` (optional): `auth` (default), `auth_int`, or `auth_conf`
- `connect_timeout_secs` (optional): per-endpoint timeout, default `5`

## Testing

**Unit tests** (33 in `sail-catalog-hms`, plus tests in `sail-catalog` and `sail-common`):
- Config parsing (including alias variants), auth validation, QOP floor selection
- URI normalization, `_HOST` expansion, failover retry classification
- SASL handshake (multi-step, early-complete, server-error)
- Security layer negotiation, GSSAPI library loading
- Type conversions, table/view building

**Integration tests** (11 ignored tests, all green with testcontainers):
- Plain HMS: database/table/view CRUD (4 tests)
- HA: dead-primary failover (1 test)
- Kerberos: database/table round-trip, missing-credentials rejection (2 tests)
- Kerberos HMS harness uses a self-contained MIT KDC + Kerberized HMS stack

## What is not included (future work)

- TLS support (can be added via `volo::net::tls::TlsMakeTransport`)
- Delegation-token authentication
- Partition operations (add/drop/list/rename partitions)
- Table rename
- Alter table schema (add/drop/rename columns)
- Statistics (reading Hive stats, writing Spark-computed stats)
- Function support (create/drop/alter/list Hive permanent functions)
- Schema-in-properties workaround for HMS case-sensitivity


crates/sail-catalog-hms/src/security/gssapi.rs
```diff
@@ -0,0 +1,630 @@
...
+    if major == bindings::GSS_S_COMPLETE {
+        return Ok(());
+    }
+
+    let mut context = 0;
+    let mut message = GssBuf::empty();
+    let ret = unsafe {
+        libgssapi()?.gss_display_status(
+            &mut minor,
+            major,
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. Added a `GssBuf::release()` helper that calls `gss_release_buffer` and invoked it after copying the error message string.

docs/guide/catalog/hms.md
```diff
@@ -0,0 +1,124 @@
...
+
+- Hive ACID or transactional HMS APIs such as transaction heartbeats, locks, or write ID allocation.
+- Iceberg-in-HMS behavior.
+- Delegation-token authentication.
+
+This provider supports Hive Metastore in its pure form:
+
+- Flat databases only. Nested namespaces are not supported.
+- Tables and views stored in Hive Metastore.
+- Generic Hive storage formats: `parquet`, `csv` or `textfile`, `json`, `orc`, `avro`, and `delta` with the alias `deltalake`.
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. Removed the duplicated block and consolidated into the earlier support/limitation lists.

crates/sail-catalog/src/hive_format.rs
```diff
@@ -0,0 +1,170 @@
...
+            Self::Orc
+        } else if haystack.contains("Avro") || haystack.contains("avro") {
+            Self::Avro
+        } else if haystack.contains("OpenCSVSerde") {
+            Self::Csv
+        } else if haystack.contains("LazySimpleSerDe") || haystack.contains("LazySimpleSerde") {
+            Self::TextFile
+        } else if haystack.contains("TextInputFormat") {
+            Self::Csv
+        } else {
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. The `TextInputFormat` fallback now returns `TextFile` instead of `Csv`. Merged the `LazySimpleSerDe` and `TextInputFormat` branches since they both map to `TextFile`.

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1291 @@
...
+            }
+        }
+
+        let endpoint_count = self.endpoints.len();
+        let mut last_error = None;
+        for step in 1..endpoint_count {
+            let next_index = (failed_index + step) % endpoint_count;
+            match self.build_client_for_endpoint(next_index).await {
+                Ok(client) => {
+                    state.active_index = next_index;
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. Refactored `rotate_client_after_failure` to drop the mutex lock before awaiting `build_client_for_endpoint()`, then reacquire it only to update `active_index`/`client` after a successful build.

crates/sail-catalog-hms/src/security/gssapi.rs
```diff
@@ -0,0 +1,630 @@
...
+            true
+        };
+
+        self.flags |= flags_out;
+
+        let out_token = unsafe {
+            if out.value.is_null() {
+                None
+            } else {
+                Some(slice::from_raw_parts(out.value.cast(), out.length).to_vec())
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. After copying the output token to a Vec, the GSSAPI buffer is released via `GssBuf::release()` which calls `gss_release_buffer`.

crates/sail-catalog-hms/src/security/gssapi.rs
```diff
@@ -0,0 +1,630 @@
...
+
+        Ok((out_token, complete))
+    }
+
+    fn wrap(&mut self, encrypt: bool, buf: &[u8]) -> CatalogResult<Vec<u8>> {
+        let mut minor = 0;
+        let mut buf_in = GssBuf::from(buf);
+        let mut buf_out = GssBuf::empty();
+        let major = unsafe {
+            libgssapi()?.gss_wrap(
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Fixed in 885fdb70. Both `wrap()` and `unwrap()` now call `buf_out.release()` after `to_vec()` to free the GSSAPI-allocated output buffer. The `GssBuf::release()` helper calls `gss_release_buffer` and nulls out the pointer to prevent double-free.

santosh-d3vpl3x (commented):
Reviewed and addressed all 6 Copilot review comments in commits 885fdb70 and 6d383523:

1. **GSSAPI buffer leaks** (3 comments) — Added `GssBuf::release()` helper calling `gss_release_buffer`. Invoked after copying data from `gss_init_sec_context`, `gss_wrap`/`gss_unwrap`, and `gss_display_status` output buffers.
2. **Async mutex across await** — Refactored `rotate_client_after_failure` to drop the lock before awaiting client build, reacquiring only to update state.
3. **TextInputFormat fallback** — Changed `HiveDetectedFormat::detect` to return `TextFile` instead of `Csv` for `TextInputFormat`.
4. **Duplicated docs** — Removed the redundant support/limitation paragraph from `hms.md`.

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+workspace = true
+
+[dependencies]
+sail-catalog = { path = "../sail-catalog" }
+sail-common = { path = "../sail-common" }
+sail-common-datafusion = { path = "../sail-common-datafusion" }
+
+arrow = { workspace = true }
+async-trait = { workspace = true }
+bitflags = "2.10.0"
```
shehabgamin:
```suggestion
bitflags = { workspace = true }
```

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+chrono = { workspace = true }
+datafusion = { workspace = true }
+hive_metastore = "0.2.0"
```
shehabgamin:
```suggestion
hive_metastore = { workspace = true }
```

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+libloading = "0.8.9"
```
shehabgamin:
```suggestion
libloading = { workspace = true }
```

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+pilota = "0.11.10"
```
shehabgamin:
```suggestion
pilota = { workspace = true }
```

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+tokio = { workspace = true }
+volo = "0.10.6"
```
shehabgamin:
```suggestion
volo = { workspace = true }
```

crates/sail-catalog-hms/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+volo-thrift = "0.10.8"
```
shehabgamin:
```suggestion
volo-thrift = { workspace = true }
```

shehabgamin (commented):
@santosh-d3vpl3x Super exciting!

I’m still reviewing, but I noticed that the added dependencies need to be moved to the workspace:
https://github.com/lakehq/sail/blob/main/Cargo.toml

crates/sail-catalog-hms/src/data_type.rs
```diff
@@ -0,0 +1,255 @@
...
+                }
+            }
+            Err(CatalogError::InvalidArgument(
+                "Map type must have key and value fields".to_string(),
+            ))
+        }
+        DataType::Dictionary(_, value_type) => arrow_to_hive_type(value_type),
+        DataType::Union(_, _)
+        | DataType::RunEndEncoded(_, _)
+        | DataType::Decimal32(_, _)
```
shehabgamin: nitpick: We can support Decimal32 and Decimal64 if we want:
```
DataType::Decimal32(precision, scale)  | DataType::Decimal64(precision, scale) | DataType::Decimal128(precision, scale) | DataType::Decimal256(precision, scale)
```

crates/sail-catalog-hms/src/data_type.rs
```diff
@@ -0,0 +1,255 @@
...
+        DataType::Int16 => Ok("smallint".to_string()),
+        DataType::Int32 => Ok("int".to_string()),
+        DataType::Int64 => Ok("bigint".to_string()),
+        DataType::UInt8 => Ok("tinyint".to_string()),
+        DataType::UInt16 => Ok("smallint".to_string()),
+        DataType::UInt32 => Ok("int".to_string()),
+        DataType::UInt64 => Ok("bigint".to_string()),
+        DataType::Float16 | DataType::Float32 => Ok("float".to_string()),
+        DataType::Float64 => Ok("double".to_string()),
+        DataType::Decimal128(precision, scale) | DataType::Decimal256(precision, scale) => {
```
shehabgamin: Do we need to validate precision and scale for Decimal256?

crates/sail-catalog-hms/src/data_type.rs
```diff
@@ -0,0 +1,255 @@
...
+            Ok(format!("decimal({precision},{scale})"))
+        }
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => Ok("string".to_string()),
+        DataType::Binary
+        | DataType::FixedSizeBinary(_)
+        | DataType::LargeBinary
+        | DataType::BinaryView => Ok("binary".to_string()),
+        DataType::Date32 | DataType::Date64 => Ok("date".to_string()),
```
shehabgamin, line 48: Does date64 map cleanly to date?

crates/sail-catalog/src/hive_format.rs
```diff
@@ -0,0 +1,171 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct HiveCatalogFormat {
+    pub logical_format: &'static str,
+    pub storage_format: HiveStorageFormat,
+}
+
+impl HiveCatalogFormat {
+    pub fn from_format(format: &str) -> CatalogResult<Self> {
+        match format.trim().to_ascii_lowercase().as_str() {
```
shehabgamin, line 82: Missing iceberg?

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+}
+
+impl HmsCatalogProvider {
+    #[expect(
+        clippy::panic,
+        reason = "HmsCatalogProvider::new is the infallible constructor"
+    )]
+    pub fn new(name: String, config: HmsCatalogConfig, runtime: RuntimeHandle) -> Self {
+        Self::try_new(name, config, runtime)
+            .unwrap_or_else(|error| panic!("failed to build HMS provider config: {error}"))
```
shehabgamin: Possible to avoid a panic?

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+            sasl_qop_min,
+            endpoints,
+            state: Mutex::new(HmsClientState {
+                active_index: 0,
+                client: None,
+            }),
+        })
+    }
+
+    fn should_retry(error: &CatalogError) -> bool {
```
shehabgamin, line 260: Should we match on the error type (+ maybe message) returned from HMS instead of string match on `CatalogError`?

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+        }
+    }
+
+    async fn alter_table(
+        &self,
+        _database: &Namespace,
+        _table: &str,
+        _options: AlterTableOptions,
+    ) -> CatalogResult<()> {
+        Ok(())
```
shehabgamin: Should we throw an error here?

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+                    ))),
+                }
+            }
+        })
+        .await
+    }
+
+    async fn list_hms_tables(&self, database: &Namespace) -> CatalogResult<Vec<Table>> {
+        let db_name = validate_namespace(database)?;
+        let names = self
```
shehabgamin, line 549: I think we can instead use `get_table_objects_by_name`:
https://github.com/Xuanwo/hive_metastore_rs/blob/970cbd5fd63b420f28e16ba8d8b1a3aec378d2d0/src/hms/service_ThriftHiveMetastore.rs#L332

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+    async fn list_databases(
+        &self,
+        prefix: Option<&Namespace>,
+    ) -> CatalogResult<Vec<DatabaseStatus>> {
+        let prefix = match prefix {
+            Some(prefix) => Some(validate_namespace(prefix)?),
+            None => None,
+        };
+
+        let databases = self
```
shehabgamin: `get_databases`?
https://github.com/Xuanwo/hive_metastore_rs/blob/970cbd5fd63b420f28e16ba8d8b1a3aec378d2d0/src/hms/service_ThriftHiveMetastore.rs#L54

crates/sail-catalog-hms/src/provider.rs
```diff
@@ -0,0 +1,1310 @@
...
+        return Err(CatalogError::InvalidArgument(
+            "HMS `uris` must contain at least one non-empty endpoint".to_string(),
+        ));
+    }
+    Ok(result)
+}
+
+fn normalize_hms_uri(uri: &str) -> CatalogResult<String> {
+    let normalized = uri
+        .strip_prefix("thrift://")
```
shehabgamin, line 116: Should ideally use case-insensitive stripping

shehabgamin (commented):
Second round of review!

crates/sail-catalog-hms/Cargo.toml
santosh-d3vpl3x: These dependencies (`bitflags`, `hive_metastore`, `libloading`, `pilota`, `volo`, `volo-thrift`) are not currently declared in the workspace `Cargo.toml` as shared dependencies, so they cannot use `{ workspace = true }`. They are unique to the HMS crate. If the workspace adopts them later we can switch.

santosh-d3vpl3x: Would you recommend making these workspace deps?

shehabgamin: @santosh-d3vpl3x For Sail, we put all dependencies in the workspace. Dependencies declared in `[workspace.dependencies]` are only compiled for the member crates that actually list them in their own `Cargo.toml`, and centralizing them ensures every crate resolves to the same version

santosh-d3vpl3x: Done — all six dependencies now use `{ workspace = true }`.

crates/sail-catalog-hms/src/data_type.rs
santosh-d3vpl3x: Good call. All four decimal variants (Decimal32, Decimal64, Decimal128, Decimal256) are now handled in a single match arm with shared validation.

crates/sail-catalog-hms/src/data_type.rs
santosh-d3vpl3x: Yes — all decimal variants now validate that precision ≤ 38 and scale is within range, since HMS caps at DECIMAL(38,*) regardless of the Arrow width.

crates/sail-catalog-hms/src/data_type.rs
santosh-d3vpl3x: Good question. HMS DATE is always days-since-epoch, so Date64 (millisecond precision) maps cleanly — the sub-day precision is simply unused. Added a comment in the code explaining the rationale.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Fixed — `new()` now returns `CatalogResult<Self>` and delegates to `try_new()`, so panics are no longer possible.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Good point. This was addressed in two ways:

1. `should_retry` now matches on stable self-owned prefixes (`"transport error"`, `"gssapi error"`, `"dns error"`) instead of fragile library-specific substrings.
2. Each error source injects its own consistent prefix — `hms_client_error` for volo-thrift transport errors, `check_gss_ok` for GSSAPI errors, and `build_client_for_endpoint` for DNS errors.

String matching is still used (rather than enum matching) because GSSAPI and DNS errors lose their type information across boundaries — GSSAPI errors get stringified into `io::Error` in the SASL transport, and DNS errors bypass volo-thrift entirely.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Already using it — `list_hms_tables` fetches names with `get_all_tables` then batch-fetches with `get_table_objects_by_name`.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Already using it — `list_databases` uses `get_databases` (with a pattern prefix) for filtered queries and `get_all_databases` for unfiltered queries.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Fixed — `normalize_hms_uri` now lowercases the input before `strip_prefix`, so `THRIFT://`, `Thrift://`, etc. all work.

crates/sail-catalog/src/hive_format.rs
santosh-d3vpl3x: lets do that in followup. needs to have a deeper look and scope is already big enough.

crates/sail-catalog-hms/src/provider.rs
santosh-d3vpl3x: Actually this slipped through. Better approach is to do a merger of existing table registered in metastore with whatever that is altered. I will check.

crates/sail-catalog/src/hive_format.rs
shehabgamin: Makes sense!

crates/sail-catalog/tests/hive_format_tests.rs
linhr, line 1: These are unit tests so the convention is to put them in `crates/sail-catalog/src/hive_format.rs`. The `tests` directory is often for integration tests in Rust.

crates/sail-common/src/config/application.rs
```diff
@@ -492,6 +492,17 @@ pub enum CatalogType {
         #[serde(skip_serializing_if = "Option::is_none")]
         endpoint_url: Option<String>,
     },
+    #[serde(alias = "hms")]
```
linhr: Let's also add `#[serde(alias = "hive-metastore")]` for flexibility.

crates/sail-common/src/config/application.rs
```diff
@@ -492,6 +492,17 @@ pub enum CatalogType {
...
+    HiveMetastore {
+        name: String,
+        uris: Vec<String>,
+        warehouse: Option<String>,
+        thrift_transport: Option<String>,
+        auth: Option<String>,
+        kerberos_service_principal: Option<String>,
+        sasl_qop_min: Option<String>,
```
linhr: Does it make sense to name this `min_sasl_qop`? People may think `min` is for "minutes" but this actually means "minimum" here.

docs/development/recipes/hms.md
linhr, line 1: Let's move this under `docs/development/recipes`.

crates/sail-catalog-hms/src/security/gssapi_bindings.rs
```diff
@@ -0,0 +1,1649 @@
+/* automatically generated by rust-bindgen 0.69.4 */
```
linhr, line 1: How was this generated? Would there be a chance that this needs to be updated? A brief note in the dev guide would be helpful.

crates/sail-catalog-hms/src/security/gssapi.rs
linhr, line 1: Did we borrow any code from the [hdfs-native](https://github.com/kimahriman/hdfs-native) project? If yes we may need to have an acknowledgement, possibly in `security/mod.rs`.

linhr (commented):
Overall the structure looks very good! And the test setup seems quite solid.

I've gone through most of the files and there are a few big files remaining for me to review, so here are my first round of comments in the meantime.

crates/sail-catalog/tests/hive_format_tests.rs
santosh-d3vpl3x: Moved inline. The `tests/` directory retains only integration tests (Kerberos harness, failover, etc.).

crates/sail-common/src/config/application.rs
santosh-d3vpl3x: Done.

crates/sail-common/src/config/application.rs
santosh-d3vpl3x: Renamed to `min_sasl_qop` across config, provider, security module, and tests.

docs/development/recipes/hms.md
santosh-d3vpl3x: Done.

crates/sail-catalog-hms/src/security/gssapi_bindings.rs
santosh-d3vpl3x: Added regeneration instructions in `security/mod.rs` and `docs/development/recipes/hms.md`. The bindings are portable across architectures since the GSSAPI C ABI uses only standard types.

crates/sail-catalog-hms/src/security/gssapi.rs
santosh-d3vpl3x: Added attribution in `security/mod.rs`.

santosh-d3vpl3x:
reviewers: please add the `run catalog tests` label to trigger the ignored HMS/Kerberos integration tests in CI.

crates/sail-catalog-hms/src/security/gssapi.rs
santosh-d3vpl3x: @Kimahriman I wonder if the krb implementation from hdfs-native can be released as standalone crate? Will it be useful that way?

shehabgamin (approved):
Well done!!

crates/sail-catalog-hms/src/security/gssapi.rs
Kimahriman: I'm not sure how much is actually reusable beyond the bindings themselves which are just automatically generated. Probably not enough reuse to warrant extracting to a standalone crate based on a quick look at this. I have no issues just copying whatever you need here.

PR #1769, opened 2026-04-23
feat: align `abs` UDF with Spark

davidlghellin:
## Summary
Aligns Sail's `abs` UDF with Spark JVM semantics on multiple fronts:

### 1. Integer overflow honours ANSI mode
TINYINT/SMALLINT/INT/BIGINT MIN now:
- Wrap to themselves under ANSI=false (matching Java's `Math.abs(int)`)
- Raise `[ARITHMETIC_OVERFLOW]` under ANSI=true

### 2. Interval/Duration overflow always errors (regardless of ANSI mode)
Verified against Spark JVM 4.x: `abs(interval_MIN)` raises
`ARITHMETIC_OVERFLOW` unconditionally — interval abs is always-checked,
unlike integer abs. Sail was previously wrapping silently in both modes;
now matches Spark.

### 3. String coercion (Utf8 → Float64)
Spark implicitly casts string operands to double before calling abs.
Sail rejected strings at planning time. Added `coerce_types` that maps
`Utf8` / `LargeUtf8` / `Utf8View` to `Float64`, so `abs('-5')` now returns
`5.0` like Spark.

### 4. Idempotence simplify (`abs(abs(x)) → abs(x)`)
Added planner-time collapse of nested abs calls. Mathematically valid
because `|x| ≥ 0` always, so `||x|| = |x|`. Safe under both ANSI modes —
inner abs's error/wrap behavior is preserved by the outer-equivalent
single call.

### 5. All-null input short-circuit
Skip the kernel pass when every input row is NULL. Returns
`new_null_array` directly. Placed AFTER the arity guard so malformed
calls still error.

### 6. Error helper consistency
Replaces raw `exec_err!` / `internal_err!` calls with
`invalid_arg_count_exec_err` / `unsupported_data_type_exec_err` helpers
(matches convention in other scalar UDF modules).


crates/sail-function/src/scalar/math/spark_abs.rs
```diff
@@ -55,121 +64,397 @@ impl ScalarUDFImpl for SparkAbs {
...
         {
             Ok(arg_types[0].clone())
         } else {
-            internal_err!("Unsupported data type {} for function abs", arg_types[0])
+            Err(unsupported_data_type_exec_err(
+                "abs",
+                "Numeric, Interval, or Duration type",
+                &arg_types[0],
+            ))
+        }
```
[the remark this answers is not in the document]
davidlghellin: validation in coerce_types

crates/sail-function/src/scalar/math/spark_abs.rs
```diff
@@ -55,121 +64,397 @@ impl ScalarUDFImpl for SparkAbs {
...
             }
             ColumnarValue::Scalar(ScalarValue::IntervalMonthDayNano(interval)) => {
-                Ok(ColumnarValue::Scalar(ScalarValue::IntervalMonthDayNano(
-                    interval.map(|x| x.wrapping_abs()),
-                )))
+                let r = match interval {
+                    Some(x) => Some(x.checked_abs().ok_or_else(|| {
+                        exec_datafusion_err!(
+                            "[INTERVAL_ARITHMETIC_OVERFLOW] interval overflow on abs(interval month-day-nano)"
+                        )
```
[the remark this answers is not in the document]
davidlghellin: solved

crates/sail-function/src/scalar/math/spark_abs.rs
```diff
@@ -55,121 +64,408 @@ impl ScalarUDFImpl for SparkAbs {
...
+                | DataType::Decimal128(_, _)
+                | DataType::Decimal256(_, _)
+                | DataType::UInt8
+                | DataType::UInt16
+                | DataType::UInt32
+                | DataType::UInt64
+                | DataType::Null
+        ) {
+            return datafusion::functions::math::abs::AbsFunc::new().invoke_with_args(args);
+        }
```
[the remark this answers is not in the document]
davidlghellin: Good catch in general,  UDF pipeline coerce_types validates arity before invoke_with_args is ever called.

crates/sail-function/src/scalar/math/spark_abs.rs
```diff
@@ -55,121 +64,408 @@ impl ScalarUDFImpl for SparkAbs {
...
         }
     }
 
     fn simplify(&self, args: Vec<Expr>, info: &SimplifyContext) -> Result<ExprSimplifyResult> {
-        match info.get_data_type(&args[0])? {
-            DataType::Interval(_) | DataType::Duration(_) => Ok(ExprSimplifyResult::Original(args)),
+        // Idempotence: abs(abs(x)) = abs(x).
+        if args.len() == 1 {
+            if let Expr::ScalarFunction(inner) = &args[0] {
+                if inner.func.name() == self.name() {
```
linhr: I wonder if we should use `.as_any().downcast_ref::<Self>()` and look at `ansi_mode` as well. We would like the simplification only if the ANSI mode match. In SQL, such extreme case of ANSI mode difference would not be present, but I feel it's still good to be strict in the code.

Also, as a general question for `simplify`, is simplification applied iteratively until the result does not change? If not, should we handle the case of `abs(abs(abs(x)))` here?

python/pysail/tests/spark/function/features/abs.feature
```diff
@@ -0,0 +1,893 @@
...
+        | result      |
+        | 5           |
+        | 0           |
+        | 5           |
+        | NULL        |
+        | -2147483648 |
+
+    @sail-bug
+    # Vectorized abs path is correct — same root cause as the scalar interval
+    # scenarios above (see `Rule: Interval values` comment at L457): Sail
```
linhr: Let's avoid mentioning line number in the comment since it will go outdated when this file changes.

python/pysail/tests/spark/function/features/abs.feature
```diff
@@ -0,0 +1,893 @@
...
+    # widens Spark interval subranges to DAY TO SECOND at the type layer.
+    # The vectorized kernel succeeds; the rendered/expected interval format
+    # diverges. Fix path: preserve subrange in Spark→Arrow Field metadata.
```
linhr: As a reference, the interval display difference would hopefully be fixed in #1895.

linhr (commented):
This looks good overall! And thanks for the comprehensive tests!

crates/sail-function/src/scalar/math/spark_abs.rs
davidlghellin: Yes, DataFusion applies `simplify` bottom-up to a fixed point — so `abs(abs(abs(x)))` is handled automatically in two passes without any special-casing. We added a plan snapshot test to confirm this:

Scenario: EXPLAIN triple-nested abs collapses to single spark_abs

```
EXPLAIN SELECT abs(abs(abs(v))) FROM VALUES (CAST(-3 AS INT)) AS t(v)
```
Result:

```
ProjectionExec: expr=[#2@0 as abs(abs(abs(v)))]
  ProjectionExec: expr=[spark_abs(column1@0) as #2]
    DataSourceExec: partitions=1, partition_sizes=[1]
```

python/pysail/tests/spark/function/features/abs.feature
davidlghellin: Thanks! Updated the comment to reference #1895.

python/pysail/tests/spark/function/features/abs.feature
```diff
@@ -0,0 +1,756 @@
...
+        | 5           |
+        | NULL        |
+        | -2147483648 |
+
+    @sail-bug
+    # Vectorized abs path is correct — same root cause as the scalar interval
+    # scenarios in `Rule: Interval values`: Sail widens Spark interval subranges
+    # to DAY TO SECOND at the type layer. The vectorized kernel succeeds; the
+    # rendered/expected interval format diverges.
+    # Fix path: preserve subrange in Spark→Arrow Field metadata (see #1895).
```
linhr: #1895 was closed. So let's just remove the reference here.

linhr (commented):
I feel let's move `abs_simplify.feature` back to `abs.feature` so the BDD tests for a single function stay in one file. Also `abs.yaml` and `abs_simplify.yaml` seem to contain duplicated content, so the latter should be removed.

python/pysail/tests/spark/function/features/abs.feature
davidlghellin: done

linhr (approved):
Nice! Thanks!

PR #1784, opened 2026-04-25
fix: add schema_of_variant_agg to codec and improve variant field validation

lonless9:


crates/sail-function/src/scalar/variant/utils/helper.rs
```diff
@@ -16,16 +16,30 @@ pub fn try_field_as_variant_array(field: &Field) -> datafusion_common::Result<()
         return Ok(());
     }
 
-    ensure(
-        is_variant_field(field),
-        "field does not have extension type VariantType",
-    )?;
+    if !is_variant_field(field) {
```
linhr: I don't quite understand the logic here. This is essentially:

```rust
if !is_variant_field(field) && !is_variant_storage_type(field.data_type()) {
    return exec_err!("...");
}
```

But would we want `||` instead of `&&` for the negated conditions?

linhr: I feel the `ensure` function should just be removed since it's just a wrapper of an `if` statement.

lonless9: DataFusion proto serializes only the return_type/name/nullable of ScalarFunctionExpr's return field, without preserving Arrow field metadata, so there would be err `field does not have extension type VariantType`. But the storage struct is still legal variant.

linhr: I see!

linhr (approved):

PR #1786, opened 2026-04-26
feat: support custom file extensions for Parquet listings

### LINKED ISSUE #1785: Allow reading Parquet files with non-standard file extensions (e.g. `.hive`)
## Problem

Sail's Parquet listing format hard-codes `.parquet` as the suffix used to filter
files when listing a directory or reading a single file path. Files written
under non-standard suffixes are silently ignored, even when the user points
directly at a single file.

The most common case I've hit is `.hive` files — Parquet files written by
Hive-managed tables. Today the only workarounds are:

1. Rename every file to `*.parquet` (often not feasible — files may be
   produced by another system, or under active write).
2. Read the file with another tool, which defeats the point of using Sail.

There's an existing TODO at
[`crates/sail-data-source/src/formats/listing.rs:61`](https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/formats/listing.rs#L61)
that anticipates this need at the global-config level.

## Reproduce

```python
# /dir/ contains a single Parquet file: data.hive
spark.read.parquet("/dir/")        # returns empty result — file is filtered out
spark.read.parquet("/dir/data.hive")  # same — filter is applied even for a single file

```

The filter is applied at
[`list_all_files` in `crates/sail-data-source/src/listing.rs#L188-L246`](https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/listing.rs#L188-L246)
via `path.ends_with(file_extension)`.

## Proposal

Add a per-read `extension` option (with alias `fileExtension`, plus the
`option.*` table-property forms) to the Parquet listing format that overrides
the default `.parquet` suffix.

| Value          | Effect                                                       |
|----------------|--------------------------------------------------------------|
| not set        | `.parquet` (existing behavior, unchanged)                    |
| `".hive"` etc. | Only files ending in that suffix are listed                  |
| `""` (empty)   | Extension filter disabled — every listed file is a candidate |

Three call shapes that should work:

```python
spark.read.option("extension", ".hive").parquet("/dir/")
spark.read.option("fileExtension", ".hive").parquet("/dir/file.hive")
```

```sql
CREATE TABLE t USING parquet OPTIONS (fileExtension '.hive') LOCATION '/dir/';
```

## Out of scope (for follow-up issues)

- Same option for csv/json/text/binary formats.
- Session- or global-level configuration for the extension (covered by the
  existing TODO linked above).

I have a working implementation and am happy to open a PR if the maintainers
agree with this direction.

Shaked98:
Closes #1785.

## What this changes

Sail currently hard-codes `.parquet` as the suffix used to filter files when listing a directory (or even when reading a single file path) for Parquet reads. Files written under non-standard suffixes — most commonly `.hive` files emitted by Hive-managed tables — cannot be read at all without renaming them first.

This PR adds a per-read `extension` option (with alias `fileExtension`, plus the `option.*` table-property forms) to the Parquet listing format that overrides the default filter.

```python
# Read a directory whose Parquet files use the .hive suffix
spark.read.option("extension", ".hive").parquet("/dir/")

# Read a single non-standard-suffix file
spark.read.option("fileExtension", ".hive").parquet("/dir/file.hive")
```

```sql
CREATE TABLE t USING parquet OPTIONS (fileExtension '.hive') LOCATION '/dir/';
```

Semantics:

| Value          | Effect                                                        |
|----------------|---------------------------------------------------------------|
| not set        | `.parquet` (existing behavior, unchanged)                     |
| `".hive"` etc. | Only files ending in that suffix are listed                   |
| `""` (empty)   | Extension filter disabled — every listed file is a candidate  |

## Design notes for reviewers

- **Generic hook on `ListingFormat`.** Added `fn file_extension_override(...) -> Result<Option<String>>` to the trait with a default `Ok(None)` so all other formats (csv/json/text/avro/binary) are untouched. Parquet is the only impl that returns `Some(...)` in this PR.
- **Override applied in `ListingTableFormat::create_provider`** before `resolve_listing_schema` is called, because that function reads `listing_options.file_extension` to filter the file list at [`crates/sail-data-source/src/listing.rs#L188-L246`](https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/listing.rs#L188-L246).
- **`rust_type: String` + `parse_string`, deliberately not `Option<String>` + `parse_optional_string`.** `parse_optional_string` collapses `""` to `None`, which would make it impossible to express "disable filtering". Plain `String` lets `""` flow through to `path.ends_with("")`, which matches every file.
- **No `TableParquetOptions` field for this.** The field controls listing, not Parquet decoding, so `BuildPartialOptions` seeds it as `None` (lets the YAML default apply), and `ParquetReadOptions::into_table_options` discards it via `extension: _`.
- **Hook re-resolves options.** `ParquetListingFormat::file_extension_override` calls `resolve_parquet_read_options` again (the `FileFormat` path also resolves them). This duplication is intentional — keeping the trait method standalone avoids changing existing trait signatures.

## Files changed

- `crates/sail-data-source/src/options/data/parquet.yaml` — new `extension` option with both snake_case and camelCase keys for `option`/`table_property` origins.
- `crates/sail-data-source/src/formats/parquet/options.rs` — seed `extension: None` in `BuildPartialOptions`; discard it in `into_table_options`; new unit test.
- `crates/sail-data-source/src/formats/listing.rs` — new trait method with default impl; apply override in `create_provider`.
- `crates/sail-data-source/src/formats/parquet/mod.rs` — implement the hook for Parquet.

## Tests

New `test_resolve_parquet_read_extension` in `formats::parquet::options::tests` covers four cases:

- default (no option) → `".parquet"`
- snake_case (`extension=".hive"`) → `".hive"`
- camelCase alias (`fileExtension=".pq"`) → `".pq"`
- empty string (`extension=""`) → `""`

Local verification:

- `cargo check -p sail-data-source` — clean
- `cargo test -p sail-data-source --lib formats::parquet::options::tests::test_resolve_parquet_read_extension` — `1 passed`
- `cargo clippy -p sail-data-source -- -D warnings` — clean
- `cargo fmt --check` — clean

## Out of scope

- Other listing formats (csv/json/text/binary). The trait shape allows it, but each has its own option YAML; keeping this PR focused on Parquet.
- Session-/global-level configuration for the extension. The existing TODO at [`crates/sail-data-source/src/formats/listing.rs:61`](https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/formats/listing.rs#L61) calls this out separately.
- Binary format's `path_glob_filter` is a different mechanism (lives inside the `FileFormat`); not changed.


linhr (commented):
Thanks for the contribution! The solution looks good to me.

Could you add a test case in `python/pysail/tests/spark/datasource/test_parquet.py` to confirm this is working end to end? For example, a Parquet file written with a custom extension can be read successfully.

(The diff in Spark test reports is due to flaky tests unrelated to this PR. We can ignore them.)

Shaked98:
Added in the latest commit ([4e5418f9](https://github.com/Shaked98/sail/commit/4e5418f9)). The test writes a Parquet file as data.hive via pandas, then verifies it can be read four ways: DataFrame .option("extension", ".hive") on the directory, DataFrame .option("fileExtension", ".hive") on the single file path, SQL CREATE TABLE ... USING parquet OPTIONS (fileExtension '.hive'), and extension="" to disable filtering. CI will exercise it; let me know if any of the assertions need to be tweaked for the Spark connect protocol.

shehabgamin:
@Shaked98 I think https://github.com/lakehq/sail/pull/1825 might actually resolve this!

linhr:
> @Shaked98 I think #1825 might actually resolve this!

@shehabgamin I think this PR would still be valuable since we could use the `extension` option to filter files. Let me take a look at the failed test and see if we can merge this PR before #1825.

linhr (approved):
@Shaked98 Thanks for the contribution!

I temporary skipped the SQL test and I'll look into it as part of #1825. The code changes look good overall!

Shaked98:
Thanks for the approval! Happy to contribute :)

PR #1795, opened 2026-04-27
refactor: expand the usage of `OptionLayer`

linhr:
This PR is a continuation of the prior work around data source options.

1. `OptionLayer` is now used in more places in logical planning.
2. We now have only "table properties" for catalog tables, while "options" (`Vec<OptionLayer>`) is a concept for data sources.
3. Options specified via the `OPTIONS` clause when creating the table is now stored with the `option.` prefix as a table property. (Previously, a few catalog providers use the `options.` (plural) prefix which is incorrect.
4. We can see that we now correctly propagate the table properties from the existing table for Delta Lake merge and delete operations.


linhr:
@copilot

1. Address Copilot review comments.
2. Investigate the 3 failed Delta Lake Python tests.

lonless9 (approved):
🚀

PR #1796, opened 2026-04-27
feat: improve checkpointing for Delta Lake

lonless9:
Support `delta.checkpoint.writeStatsAsStruct` and `delta.checkpoint.writeStatsAsJson`

Picked from #1773
#171


linhr (approved):
This is great! 🚀

PR #1797, opened 2026-04-27
perf: implement preimage hook on last_day for Parquet filter pushdown

davidlghellin:
The preimage of `last_day(ts) = D` is the closed-open month containing D: `ts ∈ [first_of_month(D), first_of_month(D) + 1 month)`. This fits `PreimageResult::Range` exactly, and `udf_preimage` derives the rewrite for all 5 comparison operators plus `!=` from the equality preimage.

## Correctness invariant

The preimage rule REPLACES the original predicate (no outer FilterExec re-checks the UDF). Returning a Range for an unsatisfiable D would over-approximate to the full month and include rows where last_day(x) ≠ D — silent correctness bug. The hook returns PreimageResult::None for such inputs so the predicate falls through to per-row UDF evaluation, which correctly yields zero rows.

A dedicated plan snapshot scenario (last_day(d) = '2024-04-15') asserts the predicate remains UNCHANGED in this case, locking in the satisfiability check as a CI canary against future regressions.


linhr (approved):
This is great!

I feel this PR also serves as a good example of how to add optimizer hooks to function implementations. Great work!

PR #1807, opened 2026-04-28
fix: case-insensitive file extension matching for listing table formats

shehabgamin:
Thanks for reporting @djouallah !

### DIFF
--- a/python/pysail/tests/spark/datasource/test_csv.py
+++ b/python/pysail/tests/spark/datasource/test_csv.py
@@ -241,10 +241,11 @@ def test_csv_read_uppercase_extension_directory(spark, tmp_path, ext):
 
 
 def test_csv_read_uppercase_extension_compressed(spark, sample_pandas_df, tmp_path):
-    # Compressed-extension matching should also be case-insensitive: a file
-    # named `*.CSV.GZ` must still be discovered as a gzipped CSV.
-    path = tmp_path / "csv_upper_gz"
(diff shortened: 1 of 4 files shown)


python/pysail/tests/spark/datasource/test_csv.py
```diff
@@ -213,3 +213,38 @@ def test_csv_format_path(spark, tmp_path):
...
+        Row(name="Bob", age="40"),
+    ]
+
+
+def test_csv_read_uppercase_extension_compressed(spark, sample_pandas_df, tmp_path):
+    # Compressed-extension matching should also be case-insensitive: a file
+    # named `*.CSV.GZ` must still be discovered as a gzipped CSV.
+    path = tmp_path / "csv_upper_gz"
+    path.mkdir()
+    sample_pandas_df.to_csv(str(path / "sample.CSV.GZ"), index=False, compression="gzip")
```
linhr: This test works but it's not very useful since `sample.CSV.GZ` is actually the directory name rather than the file name.

linhr (approved):

linhr (approved):

python/pysail/tests/spark/datasource/test_csv.py
linhr: Actually I didn't realize `to_csv` is a Pandas method so it indeed writes a single file. So I was wrong. This works equally well as writing file manually.

PR #1399, opened 2026-02-22
feat: add `from_json` function

kentkr:
This implements the `from_json` udf  (#219). It's current implementation targeted passing the `sail-spark-connect` tests, so it's incomplete. The following features still need to be implemented:
- [x] support for the schema input to be a string like a spark DDL or spark-connect json type
- [x] the top level data structure being a map or array, not just a struct
- [x] most data types not required for the `sail-spark-connect` tests
- [x] integrating the third optional `options` arg

I’d really appreciate feedback and design suggestions, especially since:
1) this is my first ever rust contribution 
2) copying `PlanResolver` and removing `ctx`/`config` isn't great and 
3) the implementation was mostly copied from [comet](https://github.com/apache/datafusion-comet/blob/main/native/spark-expr/src/json_funcs/from_json.rs) (shout out @davidlghellin for the recommendation)

Credit to [datacomet](https://github.com/apache/datafusion-comet/blob/main/native/spark-expr/src/json_funcs/from_json.rs) and @davidlghellin  for giving me inspo on this.

PR #1818, opened 2026-04-28
feat: add from_json

twsl:
My rust skills are very limited, but to move #219 forward, i tried to add support for `from_json`.
Feel free to take over this PR during review.

edit: didn't saw https://github.com/lakehq/sail/pull/1399, feel free to close

### DIFF
--- a/crates/sail-execution/src/codec.rs
+++ b/crates/sail-execution/src/codec.rs
@@ -144,7 +144,7 @@ use sail_function::scalar::geo::st_geogfromwkb::StGeogFromWKB;
 use sail_function::scalar::geo::st_geomfromwkb::StGeomFromWKB;
 use sail_function::scalar::hash::spark_murmur3_hash::SparkMurmur3Hash;
 use sail_function::scalar::hash::spark_xxhash64::SparkXxhash64;
-use sail_function::scalar::json::{SparkSchemaOfJson, SparkToJson};
+use sail_function::scalar::json::{SparkFromJson, SparkSchemaOfJson, SparkToJson};
 use sail_function::scalar::map::str_to_map::StrToMap;
 use sail_function::scalar::math::rand_poisson::RandPoisson;
 use sail_function::scalar::math::randn::Randn;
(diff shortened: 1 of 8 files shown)


crates/sail-execution/src/codec.rs
```diff
@@ -2291,6 +2295,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             || node_inner.is::<SparkExpm1>()
             || node_inner.is::<SparkFloor>()
             || node_inner.is::<SparkFromCSV>()
+            || node_inner.is::<SparkFromJson>()
```
shehabgamin: @twsl This will conflict with below, i believe it can be removed

linhr:
@twsl Thanks for the contribution! I skimmed through the PR and the structure looks good! (I'll defer to @shehabgamin for the detailed review.)

@kentkr I can see that you've spent time on #1399 and you'd like to continue with it. I feel this PR (#1818) has very good test cases, but the implementation builds `ScalarValue` for each row, so the performance could be improved by using array builders (the `FieldBuilder` approach that was borrowed from Comet to #1399). Maybe we merge #1818 first and then #1399 can come afterwards to incorporate the performance improvement via `FieldBuilder`. With the extensive test cases from #1818 in place, we will be confident of the correctness of any follow-up improvements. What do you think?

kentkr:
@linhr sounds good, please go ahead and merge #1818 first. I started a new job recently so I haven't been able to get back to #1399. Should be able to pick it up again in a week or two. Happy to follow up with the `FieldBuilder` performance improvements once things settle down on my end

crates/sail-execution/src/codec.rs
shehabgamin: @twsl In case ya missed it:
https://github.com/lakehq/sail/pull/1818/changes#r3157810542

shehabgamin (approved):
@twsl Well done! Once the issue gets resolved in `codec.rs` this is good to merge. Thanks so much for the contribution!

crates/sail-execution/src/codec.rs
twsl: Wouldnt this then also apply to SparkFromCSV?

shehabgamin: @twsl Good catch, yes that's a bug haha

shehabgamin: You can remove that as well if you'd like :) Nice find!

shehabgamin: I can take care of it in a separate PR actually, so it doesn't block this one.

twsl:
i think there is still a line in `crates/sail-plan/src/function/scalar/variant.rs` that needs to be updated and was missed:
```
("try_parse_json", F::unknown("try_parse_json")),
```
please double check @kentkr @shehabgamin

PR #1822, opened 2026-04-29
feat: implement ALTER TABLE SET/UNSET TBLPROPERTIES for Iceberg

lonless9:
#172


lonless9:
@copilot+claude-sonnet-4.6 apply changes based on the comments in [this thread](https://github.com/lakehq/sail/pull/1822#pullrequestreview-4203095373)

crates/sail-data-source/src/options/data/iceberg.yaml
```diff
@@ -113,5 +120,1122 @@
...
+  supported: false
+  rust_type: String
+  scopes: [write]
+  origins:
+    - type: table_property
+      keys: [write.delete.format.default]
+      case_sensitive: true
+      parser: crate::options::parsers::parse_string
+
+- key: write_parquet_properties
```
linhr, line 826: Each property would have its own key when supported. And Iceberg supports per-column property which we don't support during option parsing yet. (I have rough ideas for defining `parser` to parse per-column properties as a map.)

I think you have "group" keys just for convenience for now (since they are not supported anyway)?

linhr (approved):
Great work!

crates/sail-data-source/src/options/data/iceberg.yaml
lonless9: Sure they are placeholders for now. 😄

PR #1829, opened 2026-04-30
feat: add variant_explode and variant_explode_outer functions

twsl:
For #862

- Implemented SparkVariantExplodeUdf for exploding variant types into a list of structs containing position, key, and value.
- Updated the codec to register the new UDFs for variant_explode and variant_explode_outer.
- Enhanced the explode functionality to handle variant types, including arrays and objects, with appropriate tests.
- Added compatibility entries for the new functions in generator.json and variant.json.
- Created feature tests for variant_explode and variant_explode_outer to validate behavior with various input types.

@davidlghellin mind checking this? quickly threw this together


crates/sail-execution/src/codec.rs
```diff
@@ -3573,3 +3517,52 @@ impl RemoteExecutionCodec {
         Ok(message.encode_to_vec())
     }
 }
+
+#[cfg(test)]
```
linhr, line 3521: Let's remove the unit tests here. We don't rely on unit tests for codec.

crates/sail-function/src/scalar/explode.rs
```diff
@@ -86,3 +99,173 @@ impl ScalarUDFImpl for Explode {
         )
     }
 }
+
+#[cfg(test)]
```
linhr: Let's remove the unit tests here.

crates/sail-function/src/scalar/explode.rs
```diff
@@ -70,12 +76,19 @@ impl ScalarUDFImpl for Explode {
...
-            &[DataType::List(f)]
-            | &[DataType::LargeList(f)]
-            | &[DataType::FixedSizeList(f, _)]
-            | &[DataType::Map(f, _)] => Ok(f.data_type().clone()),
-            _ => plan_err!("{} should only be called with a list or map", self.name()),
+        match self.kind {
+            ExplodeKind::VariantExplode | ExplodeKind::VariantExplodeOuter => {
+                // variant_explode accepts a variant (struct) and returns struct<pos, key, value>
+                // but the actual return type is handled by the rewriter
+                Ok(arg_types.first().cloned().unwrap_or(DataType::Null))
```
linhr: If we don't modify `ExplodeKind` we won't need to worry about the return type here.

python/pysail/tests/spark/function/features/variant_explode.feature
```diff
@@ -0,0 +1,404 @@
...
+Feature: variant_explode and variant_explode_outer
+
+  Rule: variant_explode analysis and schema
+
+    Scenario: variant_explode returns pos key value columns
+      When query
+        """
+        SELECT pos, key, value
+        FROM (SELECT parse_json('[1]') AS v) t
+        LATERAL VIEW variant_explode(v) ve AS pos, key, value
```
linhr, line 31: Spark doesn't seem to understand `variant_explode` or `variant_explode_outer` in `LATERAL VIEW`. I think it's good for us to support this though.

(This is just a note.)

crates/sail-plan/src/resolver/tree/explode.rs
```diff
@@ -163,3 +217,119 @@ impl TreeNodeRewriter for ExplodeRewriter<'_> {
         Ok(Transformed::yes(out))
     }
 }
+
+#[cfg(test)]
```
linhr: Let's remove the unit tests.

We can use snapshot tests to record query plans. Look for the `query plan matches snapshot` step in `.feature` files for examples. Right now we record physical plans instead of logical plans, but it's a good enough approximation for now.

<https://docs.lakesail.com/sail/latest/development/build/python.html#updating-test-snapshots>

crates/sail-function/src/scalar/variant/spark_variant_explode.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    Ok(ListArray::new(
+        variant_explode_item_field(),
+        OffsetBuffer::new(offsets.into()),
+        Arc::new(struct_arr),
+        None,
+    ))
+}
+
+#[cfg(test)]
```
linhr: Let's remove the unit tests here since the logic is covered by BDD tests already.

crates/sail-plan/src/resolver/tree/explode.rs
```diff
@@ -75,6 +88,47 @@ impl TreeNodeRewriter for ExplodeRewriter<'_> {
             ExplodeKind::PosExplodeOuter => (true, true, false),
             ExplodeKind::Inline => (false, false, true),
             ExplodeKind::InlineOuter => (false, true, true),
+            ExplodeKind::VariantExplode | ExplodeKind::VariantExplodeOuter => {
```
linhr: I think we can support variant explode in `crates/sail-plan/src/function/generator.rs` without adding `ExplodeKind` or modifying the rewriter.

The idea is that you call `SparkVariantExplodeUdf` on the argument and then call `Explode` on the result. This way you don't need to duplicate the inline explode logic in this function.

You can refer to how `stack` is implemented (via `F::custom`) as an example.

linhr (commented):
The idea looks good overall!

I feel we can implement this without changing the rewriter. See my comment for details.

Also we typically favor end-to-end tests over Rust unit tests, since Rust unit tests are much harder to maintain and are slower to run. The BDD tests in this PR are good.

twsl:
hope i addressed everything

linhr (approved):
This looks great now! Thanks for your contribution!

PR #1837, opened 2026-05-01
feat: improve Spark and LakeSail interop conformance across HMS and object store paths

santosh-d3vpl3x:
## Summary

Consolidate HMS test infrastructure to a single S3-backed container and fix CI failures.

### Problem

The HMS interop tests had two separate Docker containers (plain HMS + S3 HMS) and two JVM Spark sessions, causing:

1. **CI permission failures** — Docker bind-mounted host temp dirs on Linux enforce host ownership, so the non-root HMS container can't create subdirectories
2. **PySpark 4.x JVM session conflict** — only one Hive-enabled Spark session can exist per process; `reference_spark_s3` had to kill `reference_spark`, breaking any test that still needed it
3. **`file://` paths not prod-relevant** — real deployments use distributed storage (S3), not local filesystem

### Solution

- Remove the plain HMS container and all its fixtures (`hms_container`, `hms_endpoint`, `hms_remote`, `hms_spark`, `reference_spark`, `hms_database`)
- All tests now use the S3-capable HMS container (MinIO + S3A), which handles both `s3://` and `file://` paths
- Single JVM Spark session (`reference_spark_s3`) eliminates session conflicts
- Removed `file://`-specific tests (relative path resolution, URI authority) that don't apply to prod setups
- 1 HMS container, 1 MinIO container, 1 Sail server, 1 JVM Spark session — no conflicts possible

### Test results

- 22 passed, 0 failed, 5 skipped (pre-existing skips in unrelated modules)
- `hatch fmt --check` clean
- `cargo +nightly fmt --check` clean
- `cargo +nightly clippy --all-targets -- -D warnings` clean

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,6 +148,9 @@ jobs:
     uses: ./.github/workflows/catalog-tests.yml
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+    needs:
+      - setup
+      - rust-build
 
   codecov-notify:
     name: Codecov Notify
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,10 +77,11 @@ jobs:
         with:
           script: |
             const labels = context.payload.pull_request?.labels?.map(label => label.name) || [];
-            core.setOutput('run_catalog_tests', labels.includes('run catalog tests') ? 'true' : 'false');
-            core.setOutput('run_spark_tests', labels.includes('run spark tests') ? 'true' : 'false');
-            core.setOutput('run_hms_interop_tests', labels.includes('hms interop tests') ? 'true' : 'false');
-            core.setOutput('run_ibis_tests', labels.includes('run ibis tests') ? 'true' : 'false');
+            const hasAnyLabel = (...names) => names.some(name => labels.includes(name));
+            core.setOutput('run_catalog_tests', hasAnyLabel('catalog tests', 'run catalog tests') ? 'true' : 'false');
+            core.setOutput('run_spark_tests', hasAnyLabel('spark tests', 'run spark tests') ? 'true' : 'false');
+            core.setOutput('run_hms_interop_tests', hasAnyLabel('hms interop tests', 'run hms interop tests') ? 'true' : 'false');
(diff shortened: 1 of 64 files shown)


.github/workflows/build.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       baseline_run_id: ${{ steps.baseline-workflow.outputs.workflow_run_id }}
       run_catalog_tests: ${{ ((github.event_name == 'pull_request' && (steps.message.outputs.run_catalog_tests == 'true' || steps.label.outputs.run_catalog_tests == 'true')) || github.event_name == 'push') && 'true' || 'false' }}
       run_spark_tests: ${{ ((github.event_name == 'pull_request' && (github.event.action == 'opened' || steps.message.outputs.run_spark_tests == 'true' || steps.label.outputs.run_spark_tests == 'true')) || github.event_name == 'push') && 'true' || 'false' }}
+      run_hms_interop_tests: ${{ ((github.event_name == 'pull_request' && (steps.message.outputs.run_hms_interop_tests == 'true' || steps.label.outputs.run_hms_interop_tests == 'true')) || github.event_name == 'push') && 'true' || 'false' }}
```
linhr: Let's run HMS tests as part of catalog tests (triggered via the `run catalog tests` label).

Right now all Python catalog tests in `python/pysail/tests/spark/catalog_integration` are skipped for some reasons, but since your HMS tests are outside of that directory, those can be run with the `catalog_integration` marker. Then we don't need the `hms_interop` marker.

You can copy `pytest_collection_modifyitems()` from `python/pysail/tests/spark/catalog_integration/conftest.py` to `python/pysail/tests/spark/hms/conftest.py` to skip the tests by default so that we don't need to have `-m "not hms_interop"` in the command for unit tests. (We will look into a better mechanism in the future so it's fine to duplicate the code for now.)

**The goal is that we should be able to complete this PR without modifying any workflow files or adding new labels**, so that the setup is more manageable if we want to have tests for other catalog providers in the future. (I can see that there may be a need to run *only* HMS tests instead of the entire catalog test suite. Let me think about this and follow up separately.)

linhr (commented):
Nice! I'll defer to @shehabgamin for the detailed review. Here is just a comment around the high-level setup.

.github/workflows/build.yml
santosh-d3vpl3x: This was indeed the dilemma. I will merge them.

Issue #1851, opened 2026-05-05
Feature request: Add Glue Catalog databases and tables cache


aminebizid:
The access to AWS Glue catalog is too slow. We used to use these parameters in Databricks to cache the db and tables metadata:

```
spark.hadoop.aws.glue.cache.db.enable true
spark.hadoop.aws.glue.cache.db.size 1000
spark.hadoop.aws.glue.cache.db.ttl-mins 30
spark.hadoop.aws.glue.cache.table.enable true
spark.hadoop.aws.glue.cache.table.size 1000
spark.hadoop.aws.glue.cache.table.ttl-mins 30
```

Can we implement this feature.

Kr,
Amine

linhr:
Thanks for the feature request! I feel caching would indeed be helpful for remote catalog access.

Here are a few thoughts on the design.

1. Caching is a shared concern for all catalog providers (except the memory provider). We should implement the caching layer as a wrapper of any catalog provider, similar to how `RuntimeAwareCatalogProvider` manages the Tokyo runtime for catalog providers.
2. We need to have the configuration as the catalog provider options instead of global application configuration. Each catalog provider instance may want to enable caching separately.

Here is the configuration design I'm thinking about.

```rust
pub enum CatalogType {
    Glue {
        name: String,
        // other existing fields
        #[serde(flatten)]
        cache: CatalogCacheConfig,
    },
    // same cache config for other providers
}

// setting the cache size to 0 implies unbounded cache size
// setting TTL to 0 implies infinite caching
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(deny_unknown_fields)]
pub struct CatalogCacheConfig {
    pub database_cache_enabled: bool,
    #[serde(
        serialize_with = "serialize_non_zero",
        deserialize_with = "deserialize_non_zero"
    )]
    pub database_cache_size: Option<usize>,
    #[serde(
        serialize_with = "serialize_non_zero",
        deserialize_with = "deserialize_non_zero"
    )]
    pub database_cache_ttl_secs: Option<u64>,
    pub table_cache_enabled: bool,
    #[serde(
        serialize_with = "serialize_non_zero",
        deserialize_with = "deserialize_non_zero"
    )]
    pub table_cache_size: Option<usize>,
    #[serde(
        serialize_with = "serialize_non_zero",
        deserialize_with = "deserialize_non_zero"
    )]
    pub table_cache_ttl_secs: Option<u64>,
}
```

[...]

PR #1852, opened 2026-05-05
feat: implement caching for all remote catalogs

aminebizid:
This PR introduces a caching system for the AWS Glue catalog to improve the access performances.


crates/sail-catalog/src/provider/namespace.rs
```diff
@@ -81,3 +81,13 @@ impl Namespace {
...
+
+impl std::fmt::Display for Namespace {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.head)?;
+        for segment in &self.tail {
+            write!(f, ".{}", segment)?;
+        }
+        Ok(())
+    }
+}
```
linhr: We don't support `Display` for `Namespace` because we used to misuse it without thinking about SQL special character escape etc. We should use the `quote_namespace_if_needed` function so that the intent is explicit.

linhr (commented):
Thanks for having the PR quickly!

Would you like to consider this comment for the implementation? <https://github.com/lakehq/sail/issues/1851#issuecomment-4379735050>

Also I'm not sure if "cross-session" in the PR title is accurate since the implementation uses a cache per catalog provider instance (which makes sense and should be the right way to go IMO).

crates/sail-catalog/src/provider/namespace.rs
aminebizid: I've addressed the comments in the latest commits:

Removed the Display implementation for Namespace and used quote_namespace_if_needed instead to ensure safe SQL escaping.
Verified that all Glue catalog tests (including the new cache tests) pass successfully.
I also agree that 'cross-session' in the title was a bit imprecise, I'll update the PR title to: feat(cargo): implement caching for AWS Glue catalog provider.

crates/sail-catalog-glue/src/provider.rs
```diff
@@ -29,6 +29,12 @@ pub struct GlueCatalogConfig {
     pub region: Option<String>,
     /// Custom endpoint URL (optional). Useful for VPC endpoints or local development.
     pub endpoint_url: Option<String>,
+    /// Whether to enable database caching.
+    pub cache_db_enable: bool,
+    /// Whether to enable table caching.
+    pub cache_table_enable: bool,
+    /// Cache TTL in seconds.
+    pub cache_ttl_secs: u64,
```
linhr: I don't think we need to change the Glue catalog anymore. Could you revert changes to the `sail-catalog-glue` crate and remove the newly added unit tests in this crate?

crates/sail-catalog/src/provider/namespace.rs
```diff
@@ -81,3 +81,53 @@ impl Namespace {
...
+    }
+
+    #[test]
+    fn test_namespace_conversions() {
+        let vec_str = vec!["a", "b"];
+        let ns = Namespace::try_from(vec_str).unwrap();
+        
+        let vec_arc: Vec<Arc<str>> = ns.clone().into();
+        assert_eq!(vec_arc.len(), 2);
+        assert_eq!(vec_arc[0].as_ref(), "a");
```
linhr, line 122: We can also add `assert_eq!(vec_arc[1].as_ref(), "b");`.

crates/sail-common/src/config/application.rs
```diff
@@ -575,3 +591,70 @@ impl ClusterConfigEnv {
...
+    pub table_cache_size: Option<usize>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "serialize_non_zero",
+        deserialize_with = "deserialize_non_zero"
+    )]
+    pub table_cache_ttl_secs: Option<u64>,
+}
+#[cfg(test)]
```
linhr: Remove the unit tests here.

crates/sail-flight/src/session.rs
```diff
@@ -88,3 +85,22 @@ pub fn create_flight_session_manager(
         ))
     })
 }
+
+#[cfg(test)]
```
linhr: Remove the unit tests here.

crates/sail-session/src/session_factory/server.rs
```diff
@@ -258,3 +264,166 @@ impl ServerSessionFactory {
             .maximum_buffered_record_batches_per_stream;
     }
 }
+
+#[cfg(test)]
```
linhr: Remove the unit tests here.

crates/sail-session/Cargo.toml
```diff
@@ -48,3 +48,6 @@ tokio = { workspace = true }
 chrono = { workspace = true }
 indexmap = { workspace = true }
 futures = { workspace = true }
+
+[dev-dependencies]
+pyo3 = { workspace = true, features = ["auto-initialize"] }
```
linhr: Remove the dev dependency. And we can't enable the `auto-initialize` in this project.

docs/guide/catalog/index.md
```diff
@@ -43,11 +43,178 @@ In the next few pages, we will explore the different catalog providers supported
...
+| `drop_database`      | database list cache + table + view caches |
+| `create_table`       | table list cache for the database         |
+| `drop_table`         | table list cache for the database         |
+| `alter_table`        | table list cache for the database         |
+| `create_view`        | view list cache for the database          |
+| `drop_view`          | view list cache for the database          |
+
+Read-only operations (`get_database`, `get_table`, `get_view`) are **not** cached and always hit the remote catalog directly. This is a deliberate design choice because individual lookups are fast and caching them would complicate invalidation logic.
+
+## Developer Guide: Adding Caching to a New Provider
```
linhr: Let's remove the developer guide entirely. The details are readily available by looking at the code, while the docs can get outdated over time.

crates/sail-session/src/catalog.rs
```diff
@@ -206,3 +233,98 @@ pub fn create_catalog_manager(
...
+
+pub fn create_catalog_manager(
+    config: &AppConfig,
+    runtime: RuntimeHandle,
+) -> Result<CatalogManager> {
+    let providers = create_catalog_providers(config, runtime)?;
+    create_catalog_manager_with_providers(config, providers)
+}
+
+#[cfg(test)]
```
linhr: Remove the unit tests here.

crates/sail-session/src/catalog.rs
```diff
@@ -5,28 +5,30 @@ use datafusion::common::{plan_datafusion_err, Result};
...
 use sail_catalog_onelake::OneLakeCatalogProvider;
 use sail_catalog_system::{SystemCatalogProvider, SYSTEM_CATALOG_NAME};
 use sail_catalog_unity::UnityCatalogProvider;
-use sail_common::config::{AppConfig, CatalogType};
+use sail_common::config::{AppConfig, CatalogCacheConfig, CatalogType};
 use sail_common::runtime::RuntimeHandle;
 use secrecy::ExposeSecret;
 
-pub fn create_catalog_manager(
+pub fn create_catalog_providers(
```
linhr: I'm not sure if we need to separate these functions.

crates/sail-session/src/session_factory/server.rs
```diff
@@ -56,29 +58,33 @@ pub trait ServerSessionMutator: Send {
...
 
+#[derive(Clone)]
 pub struct ServerSessionFactory {
     config: Arc<AppConfig>,
     runtime: RuntimeHandle,
     system: Arc<Mutex<ActorSystem>>,
-    mutator: Box<dyn ServerSessionMutator>,
+    mutator: Arc<dyn ServerSessionMutator>,
     runtime_env: RuntimeEnvFactory,
+    catalog_providers: HashMap<String, Arc<dyn CatalogProvider>>,
```
linhr: I feel having `catalog_providers` in `ServerSessionFactory` doesn't seem right.

crates/sail-session/src/session_factory/server.rs
```diff
@@ -38,7 +40,7 @@ pub struct ServerSessionInfo {
     pub session_manager: ActorHandle<SessionManagerActor>,
 }
 
-pub trait ServerSessionMutator: Send {
+pub trait ServerSessionMutator: Send + Sync {
```
linhr, line 41: Is there a reason for the mutator to be `Sync`? I guess it's due to your changes to the session factory etc. but I'm not quite sure if those are needed.

crates/sail-spark-connect/src/session_manager.rs
```diff
@@ -70,25 +70,22 @@ fn create_spark_session_factory(
...
+    let session_factory =
+        create_spark_session_factory(config.clone(), runtime.clone(), system.clone())?;
     let factory = {
-        let config = config.clone();
-        let runtime = runtime.clone();
-        let system = system.clone();
-        Box::new(move || {
-            create_spark_session_factory(config.clone(), runtime.clone(), system.clone())
-        })
+        Box::new(move || Box::new(session_factory.clone()) as Box<dyn SessionFactory<ServerSessionInfo>>)
```
linhr: The idea was that `create_spark_session_factory` is called in the closure so we don't need `Arc` for mutator. I don't quite understand why we need to change this to make the cache wrapper works. Let me know if I missed anything.

crates/sail-catalog/src/provider/options.rs
```diff
@@ -9,21 +9,21 @@ use sail_common_datafusion::catalog::{
 use serde::{Deserialize, Serialize};
 
 /// Options for creating a database in a catalog.
-#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd, Serialize, Deserialize, Default)]
```
linhr: Let's remove `Default` for all these options. People may have different ideas about those defaults (e.g., whether `if_not_exists` should be `true` or `false`) so it's good to just avoid them to avoid misuse.

crates/sail-common/src/config/application.rs
```diff
@@ -416,12 +417,17 @@ pub enum CacheType {
...
         serialize_with = "serialize_non_empty_string",
         deserialize_with = "deserialize_non_empty_string"
     )]
     pub default_catalog: Option<String>,
+    #[serde(default)]
     pub default_database: Vec<String>,
+    #[serde(default)]
     pub global_temporary_database: Vec<String>,
+    #[serde(default)]
     pub list: Vec<CatalogType>,
```
linhr, line 427: I'm not sure if we need these since the configuration is always backed by the default configuration in the `.toml` file.

crates/sail-catalog/src/provider/cached.rs
```diff
@@ -0,0 +1,377 @@
...
+use crate::provider::{
+    AlterTableOptions, CatalogProvider, CreateDatabaseOptions, CreateTableOptions,
+    CreateViewOptions, DropDatabaseOptions, DropTableOptions, DropViewOptions, Namespace,
+};
+
+pub struct CachedCatalogProvider {
+    inner: Arc<dyn CatalogProvider>,
+    db_list_cache: Option<Cache<Option<Namespace>, Vec<DatabaseStatus>>>,
+    table_list_cache: Option<Cache<Namespace, Vec<TableStatus>>>,
+    view_list_cache: Option<Cache<Namespace, Vec<TableStatus>>>,
```
linhr, line 19: I can see that this PR only supports caching list operations and changes to individual databases or tables invalidates the entire cache. Maybe we can support more fine-grained cache management in the future. It's fine to consider this out of scope for this PR.

linhr (commented):
Nice progress! I think we are heading towards the right direction.

This PR however introduces a lot more code changes than I expected. Could you review the diff around Glue catalog and session factory? I few most of these changes are not needed, but let me know if I missed anything.

Also we don't usually want Rust unit tests since they are difficult to maintain and slow to run. We have end-to-end tests in Python to cover the functionalities. I've pointed out a few places where the tests should be removed. The Rust tests around `CachedCatalogProvider` are good though.

And let's fix CI errors.

crates/sail-catalog/src/provider/cached.rs
aminebizid: I remember when using Databricks, I used to restart the whole cluster after catalog changes to not wait for the ttl. This is why I opt for something like event driven invalidation. And since we cache views...

crates/sail-session/src/session_factory/server.rs
aminebizid: I tested removing the Sync bound on ServerSessionMutator, but the compiler confirmed it is required. Since ServerSessionFactory implements SessionFactory: Send and contains an Arc<dyn ServerSessionMutator>, the trait must be Sync to allow safe sharing between threads.

error[E0277]: (dyn ServerSessionMutator + 'static) cannot be shared between threads safely

crates/sail-spark-connect/src/session_manager.rs
aminebizid: let factory = {
        let config = config.clone();
        let runtime = runtime.clone();
        let system = system.clone();
        Box::new(move || {
            create_spark_session_factory(config.clone(), runtime.clone(), system.clone())
        })
    };


In the original implementation, create_spark_session_factory was called every time a new session was created. Since create_spark_session_factory calls ServerSessionFactory::try_new (line 77), which in turn calls create_catalog_providers (in server.rs), new cache instances would be created for every single session. By moving the factory creation outside the closure and cloning it instead, we ensure that the CachedCatalogProvider (and its underlying Moka cache) is shared across all sessions, making the cache persistent and effective.

linhr: Thanks for the context. Sharing catalog providers across sessions is a breaking behavior change so I'd prefer to leave it out of this PR. And cross-session cache sharing can be done differently without sharing catalog providers themselves. (We have a pattern for this for file statistics and metadata etc. which can be used for catalog caches as well. We can revisit this once this PR is merged.)

linhr:
@aminebizid Thanks for addressing the comments. Could you consider this note so that this PR won't need to touch the session factory? <https://github.com/lakehq/sail/pull/1852#discussion_r3198725102>

Also it seems that CI is failing so you may want to make sure linting and testing works locally.

<https://docs.lakesail.com/sail/latest/development/build/rust.html>
<https://docs.lakesail.com/sail/latest/development/documentation/>

linhr:
Nice to see the CI is now successful!

Just to clarify a few points.

> In the original implementation, create_spark_session_factory was called every time a new session was created <https://github.com/lakehq/sail/pull/1852#discussion_r3196501017>

This is not true. The closure `Box::new(move || { create_spark_session_factory(config.clone(), runtime.clone(), system.clone()) })` is stored in `SessionManagerOptions` and the closure is called only once in `SessionManagerActor::new()`.

> Sharing catalog providers across sessions is a breaking behavior change <https://github.com/lakehq/sail/pull/1852#discussion_r3198725102>

For example, the memory catalog provider tracks databases and tables via `DashMap` which should not be shared across sessions.

So my hope is that all the changes to session factory are reverted for this PR. (Right now there are 22 changed files, but my guess is that only a few files would be involved to support per-session catalog cache.) It's possible to support global caching by tracking the caches (indexed by the name of catalog providers) in `ServerSessionFactory`, similar to how file statistics/metadata caches are tracked in `RuntimeEnvFactory`. But let's consider this out of scope for now so that it's easier to merge this PR. (As a follow-up we could support multiple caching modes, including both the global mode and per-session mode.)

Issue #1683, opened 2026-04-15
[Feature Request] Support writing to Delta Lake tables with `GeneratedColumns`


keen85:
## Situation

Sail 0.6.0 prevents writing to Delta Lake tables having `GeneratedColumns` feature enabled (https://github.com/lakehq/sail/pull/1557).
While this ensures consistency, it currently blocks our organization from adopting Sail.

In version 0.5.3, writing to these tables was still possible as long as the `DataFrame` already contained the required generated columns. I was able to work around the missing native support by manually calculating the columns based on the table's metadata before writing:
```python
from deltalake import DeltaTable as DeltaTableRust
from pyspark.sql import types as T
from pyspark.sql import functions as F

table_schema = T.StructType.fromJson(json.loads(DeltaTableRust(table_path).schema().to_json()))
generation_expr = table_schema["generated_column"].metadata["delta.generationExpression"]
df = df.withColumn("generated_column", F.expr(generation_expr))
```

Since this manual approach is no longer permitted in 0.6.0, we are forced to fall back to delta-rs for appends. We would greatly appreciate it if Sail could handle these columns - either by allowing pre-calculated columns (as before) or, ideally, by implementing native support and evaluating the Spark SQL column expression at runtime transparently.

## Implementation Specification

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#generated-columns

## Performance consideration (but this does not need to be part of the initial feature implementation and can come later)

Spark is also able to make use of the generation expression for partition pruning when the generated column is used for partitioning a Delta table and a query predicated uses the column that is also used in the generation expression. This can have a significant performance impact on query performance.
See details [here](https://docs.delta.io/delta-batch/#-deltausegeneratedcolumns:~:text=Delta%20Lake%20may%20be%20able%20to%20generate%20partition%20filters%20for%20a%20query%20whenever%20a%20partition%20column%20is%20defined%20by%20one%20of%20the%20following%20expressions)

shehabgamin:
cc @lonless9

keen85:
@lonless9 , I just had some time to test #1701.
Unfortunately for me it did not work: I can insert data in a Delta Lake table with generated column with no error - but the generated column is not populated on insert:

```python
import shutil
from pathlib import Path
import deltalake as delta_rs

# create an empty Delta Lake table using delta-rs
path_for_dummy_table = Path(...)
if path_for_dummy_table.exists():
    shutil.rmtree(path_for_dummy_table)

[...]

PR #1855, opened 2026-05-05
fix: allow generated columns for datasource target & introduce snapshot/commit context

lonless9:
#1683


crates/sail-execution/proto/sail/plan/physical.proto
```diff
@@ -490,6 +490,36 @@ message PhysicalSinkMode {
   }
 }
 
+enum DeltaColumnMappingMode {
+  DELTA_COLUMN_MAPPING_MODE_NONE = 0;
+  DELTA_COLUMN_MAPPING_MODE_NAME = 1;
+  DELTA_COLUMN_MAPPING_MODE_ID = 2;
+}
+
+message DeltaSnapshotContextNode {
```
linhr: I feel we can just name these `DeltaSnapshotContext`, `DeltaCommitContext`, and `DeltaWriteContext`. The term "node" is for physical plan nodes specifically.

linhr (approved):
This is great!

crates/sail-execution/proto/sail/plan/physical.proto
lonless9: Done!

PR #1856, opened 2026-05-05
feat: align coalesce with pyspark

twsl:
Improves spark parity

- Align explicit partition management with Spark semantics by distinguishing `coalesce`, round-robin `repartition(n)`, and hash `repartition(exprs)`
- Add a custom non-shuffle coalesce path so reducing partitions preserves rows, does not increase partition count, and avoids unnecessary shuffles
- Implement round-robin repartition execution so `repartition(n)` actually redistributes rows across output partitions, including cases with identical input rows
- Rewrite explicit repartition nodes to the appropriate physical operators and eliminate no-op coalesce cases during optimization
- Extend plan serialization and distributed job-graph planning so the new repartitioning and coalesce behavior works in cluster mode as well as local execution
- Add Spark-facing coverage for plan shape, partition counts, row preservation, `COALESCE` hint behavior, partition grouping, and cluster-mode correctness


linhr:
Thanks for this PR. How is the implementation different from #1782? Or is it fixing the same problems? Could you elaborate the design in the PR description?

twsl:
just noticed this function wasnt fully supported and tried to align it with spark behaviour, so i can finally replace pyspark in my applications. I updated the description of this pr. I'd say they are basically aiming to solve the same issue but haave minor differences in implementation.

linhr:
> just noticed this function wasnt fully supported and tried to align it with spark behaviour, so i can finally replace pyspark in my applications. I updated the description of this pr. I'd say they are basically aiming to solve the same issue but haave minor differences in implementation.

Got it. Thanks for the context!

This PR involves changes to distributed query planner so it will take a bit of time to review. I'll try to go through it in the next one or two days. In the meantime, you can fix the CI issues, and see if the Copilot comments are reasonable to consider.

twsl:
sure, tbh i intended to create a draft first, but forgot about it. Will address these proposals

twsl:
hope i addressed everything

python/pysail/tests/spark/execution/features/repartition.feature
```diff
@@ -0,0 +1,39 @@
+@sail-only
+Feature: Explicit repartition end-to-end behavior
+
+  Rule: Round-robin repartition via DataFrame API
```
linhr: For tests around the DataFrame API, we usually do it in Python so that we don't have to define BDD steps for it. The Python tests are often more readable than `.feature` files in this case.

This is just a note. We can keep the tests without making further changes for now. I can help with the adjustments separately.

crates/sail-execution/src/plan/shuffle_write.rs
```diff
@@ -140,23 +158,26 @@ impl ExecutionPlan for ShuffleWriteExec {
             );
         }
         let stream = self.plan.execute(partition, context)?;
-        // TODO: Revisit this
-        let shuffle_partitioning = match &self.shuffle_partitioning {
-            Partitioning::UnknownPartitioning(size) => Partitioning::RoundRobinBatch(*size),
-            shuffle_partitioning => shuffle_partitioning.clone(),
-        };
-        // TODO: Support metrics in batch partitioner
```
linhr, line 148: Could you add back this TODO?

crates/sail-execution/src/plan/shuffle_write.rs
```diff
@@ -17,10 +17,28 @@ use datafusion::physical_plan::{
 };
 use futures::future::try_join_all;
 use futures::StreamExt;
+use sail_physical_plan::repartition::RowRoundRobinPartitioner;
 
 use crate::plan::ListListDisplay;
 use crate::stream::writer::{TaskStreamSinkState, TaskStreamWriter, TaskWriteLocation};
 
+enum ShufflePartitioner {
+    Hash(BatchPartitioner),
```
linhr: I feel this could be named `Batch` (to make it agnostic about what's inside the partitioner), even if we only construct a hash partitioner in the code.

crates/sail-common-datafusion/src/session/repartition.rs
```diff
@@ -0,0 +1,32 @@
...
+            buffer_size: buffer_size.max(1),
+        }
+    }
+
+    pub fn buffer_size(&self) -> usize {
+        self.buffer_size
+    }
+}
+
+impl SessionExtension for RepartitionBufferConfig {
```
linhr, line 28: I feel we could make this extension more general to manage all config options for execution, and we could potentially expose a config option to the user in `AppConfig`. Let me think about it more thoroughly.

This is just a note for the future. No need to make any changes in this PR.

crates/sail-physical-plan/src/repartition.rs
```diff
@@ -36,21 +113,260 @@ impl ExplicitRepartitionExec {
...
+        for _ in 0..output_partitions {
+            let (sender, receiver) = channel::<Result<RecordBatch>>(buffer_size);
+            senders.push(sender);
+            receivers.push(Some(receiver));
+        }
+
+        for input_partition in 0..input_partition_count {
+            let input = self.input.execute(input_partition, context.clone())?;
+            let output_senders = senders.clone();
+            tokio::spawn(async move {
```
linhr: Let's use `SpawnedTask` from DataFusion to ensure proper task cleanup.

crates/sail-physical-plan/src/repartition.rs
```diff
@@ -36,21 +109,152 @@ impl ExplicitRepartitionExec {
...
+                    output_senders,
+                    input_partition,
+                    input_partition_count,
+                    output_partitions,
+                )
+                .await;
+            });
+        }
+
+        state.receivers = Some(receivers);
```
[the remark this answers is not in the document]
linhr: I think the current implementation is fine.

crates/sail-physical-plan/src/coalesce.rs
```diff
@@ -0,0 +1,408 @@
...
+    use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
+    use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+    use datafusion::physical_plan::{DisplayAs, ExecutionPlan, PlanProperties};
+    use datafusion_common::{internal_datafusion_err, Result, Statistics};
+    use futures::{stream, StreamExt};
+
+    use super::{coalesced_input_range, CoalesceExec};
+
+    #[derive(Debug)]
+    struct TestExec {
```
linhr: I feel we don't need this `TestExec`. We could use `MemTable` and call `scan` to get a physical plan that returns the batches. You could define a helper function to create the plan from `Vec<Vec<i32>>`.

linhr (commented):
This looks really nice!

Besides the few minor comments I have, here are two important notes:

1. I think `CoalesceExec` needs to be in its own stage in the job graph for distributed processing, since it has non-conventional behavior for executing input plan partitions. You can refer to the comments around `InputMode` in `crates/sail-execution/src/job_graph/mod.rs` for more details. I feel we may need a new `InputMode` similar to "rescale" in Flink. But since this PR is large already, I'd recommend tackling it in a follow-up PR. For this PR, you can return a "not implemented" error when creating the job graph if `CoalesceExec` is present. (And let's use `@pytest.mark.skip` for the test you added in `python/pysail/tests/spark/test_execution.py`.) Then I can create an issue for the design discussion.
2. As I mentioned in one of the comment, We need to use `SpawnedTask` to manage tasks in `ExplicitRepartitionExec`, which means we would need to track the handles in the state. #1782 seems to have the logic we need but I haven't confirmed the correctness. You can take a look but you would need to review the code carefully. (Note: We'll not merge #1782. I feel your implementation in this PR is much cleaner overall!)

twsl:
tried to cover everything, also added a proposal for `CoalesceExec` stage. let me know what you think

linhr:
Thanks for the update! I'll take a look this week.

linhr (approved):
This is great! The new input mode for rescale looks good to me.

Here are some thoughts:
1. All round-robin repartition usages are now row-based instead of batch-based in cluster mode. This is a performance degradation when row-base partitioning is not necessary. In TPC-H and TPC-DS benchmarks we have many round-robin repartition nodes. We should keep the batch-based repartition as the default and use the row-based approach for explicit `repartition()` and `coalesce()` method calls.
2. For `ExplicitRepartitionExec` the spawned task handle is now kept in the state but ideally they should be moved to the returned stream so that the task is cancelled when the stream is dropped. Keeping the task handle in the state may make the task run longer when `ExplicitRepartitionExec` is still accessible. We might want to see how DataFusion handles this in `RepartitionExec`.
3. The coalesce BDD tests should preferably be written in Python so that we don't need to define the steps for it. This follow the same idea as my earlier [comment](https://github.com/lakehq/sail/pull/1856#discussion_r3240038352).

Since this PR is already long, I'm going to merge this as is and we'll track the follow-up work in this issue: #1988.

PR #1859, opened 2026-05-06
fix: truncate large gRPC error messages to prevent metadata size overflow

pomykalakyle:
## Summary

gRPC has a hard limit of 16384 bytes for the total trailer metadata size. Sail's error path puts the message into both `grpc-message` (percent-encoded, worst case ~3x expansion) and `grpc-status-details-bin` (base64-encoded protobuf with the message embedded again, ~4/3x expansion). A long Python traceback can exceed the limit, in which case gRPC drops the response and the client sees `RESOURCE_EXHAUSTED` instead of the real error.

## Fix

Cap the message at 3700 bytes (combined worst-case encoding ~16021 bytes, leaving margin for other trailers), respecting UTF-8 boundaries, appending `\n[truncated]`, and preserving a trailing newline if the original had one (Python traceback formatting invariant). Applied to `SparkThrowable`, `SendError`, and `InternalError` `Status` conversions.

## History

Picks up the work from #1755 (Copilot-authored, never merged). Closes the gap left by the revert in #1628 of the earlier #1488 attempt, which was reverted because over-aggressive truncation at 1024 bytes broke ~20 parity tests. The 3700-byte cap here is high enough that real Spark error messages fit without truncation in the common case.

### DIFF
--- a/crates/sail-spark-connect/src/error.rs
+++ b/crates/sail-spark-connect/src/error.rs
@@ -280,6 +280,44 @@ impl SparkThrowable {
     }
 }
 
+/// The maximum length of a gRPC status message in bytes.
+///
+/// gRPC has a hard limit of 16384 bytes for the total metadata size.
+/// When error details are present, the message is encoded in two headers:
+/// - `grpc-message`: percent-encoded (worst case 3x expansion for non-ASCII bytes)
+/// - `grpc-status-details-bin`: base64-encoded protobuf (~4/3x expansion)
+///
+/// Combined worst-case encoding overhead is approximately 4.33x the raw message length
+/// (3x + 4/3x). A cap of 3700 bytes keeps the combined encoded size under 16384 bytes
+/// even in the worst case (3700 * 4.33 ≈ 16,021), with some margin for other headers.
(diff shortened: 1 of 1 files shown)


crates/sail-spark-connect/src/error.rs
```diff
@@ -291,10 +329,16 @@ impl From<SparkThrowable> for Status {
         let mut details = ErrorDetails::new();
         details.set_error_info(class, "org.apache.spark", metadata);
 
+        // Truncate the message if it exceeds the maximum length.
+        // gRPC has a hard limit of 16384 bytes for total metadata size.
+        // Both `grpc-message` and `grpc-status-details-bin` encode the message,
+        // so long messages (e.g., Python tracebacks) can easily exceed the limit.
+        let message = truncate_grpc_message(throwable.message());
```
linhr, line 336: Maybe it's better to keep the last few lines of the traceback. But this requires parsing the message and the logic can easily get complex. So this is fine for now.

linhr (approved):
Thanks for the follow-up on this!

I guess we could go even further to see if it's possible to adjust the gRPC metadata size limit, or avoid message duplication in the metadata. I'm not sure if these are possible though. So message truncation seems a reasonable approach for now.

PR #1865, opened 2026-05-07
feat: add HMS table metadata parity

santosh-d3vpl3x:
## Summary
- Add Spark/HMS metadata conversion coverage, including schema/provider/path metadata for Spark datasource tables
- Add HMS Spark roundtrip tests for Sail-to-Spark, Spark-to-Sail, and S3 metastore coverage
- Support table-level `ALTER TABLE ... SET LOCATION` as catalog metadata (error improvement deferred to follow-up)
- Hardwire `type_name()` to `"EXTERNAL"` for all `TableKind::Table` variants — Sail treats persistent tables as external by default
- Reject user-supplied `spark.sql.*` and `EXTERNAL` properties on write; filter them on read

## Scope notes
- **EXTERNAL semantics**: Sail-created generic HMS tables surface as `EXTERNAL` in HMS/Spark metadata. This is intentional — Sail treats all persistent tables as external.
- **Bare `decimal` default change**: `hive_type_to_arrow("decimal")` now returns `Decimal128(10, 0)` (Hive convention) instead of the previous `Decimal128(38, 18)` (Spark's internal `SYSTEM_DEFAULT`). This is correct for HMS interop — Hive normalizes bare `decimal` to `(10, 0)` before storage.
- **Timestamp timezone normalization**: Non-UTC timestamp timezones are normalized to UTC through the Hive round-trip (`arrow_to_hive_type` maps all timestamps to `"timestamp"` which HMS treats as UTC). This matches Spark's behavior.


santosh-d3vpl3x:
@shehabgamin possible to enable `run catalog tests`?

.github/workflows/rust-tests.yml
```diff
@@ -79,7 +79,12 @@ jobs:
         if: ${{ inputs.run_catalog_tests == 'true' }}
         env:
           LLVM_PROFILE_FILE: ${{ github.workspace }}/target/llvm-profiles/rust-slow/sail-%p-%m.profraw
-        run: cargo nextest run --run-ignored ignored-only -j 6 --no-fail-fast
+        run: |
+          cargo nextest run --run-ignored ignored-only -j 6 --no-fail-fast \
```
shehabgamin: @santosh-d3vpl3x I'm assuming you forgot to undo this 😄

crates/sail-sql-analyzer/src/statement.rs
```diff
@@ -1920,6 +1920,7 @@ fn from_ast_alter_table_operation(
                 if_exists: if_exists.is_some(),
             })
         }
+        AlterTableOperation::SetLocation { .. } => Ok(spec::AlterTableOperation::Unknown),
```
shehabgamin: ?

python/pysail/tests/spark/hms/__init__.py
shehabgamin, line 1: Haven't seen this before on GitHub:

> Whitespace-only changes.

crates/sail-delta-lake/src/spec/statistics.rs
```diff
@@ -368,11 +368,8 @@ fn null_count_stats_schema(schema: &StructType) -> Option<StructType> {
             let data_type = match &field.data_type {
                 DataType::Array(_) | DataType::Map(_) | DataType::Variant(_) => DataType::LONG,
                 DataType::Struct(inner) => {
-                    if let Some(inner_schema) = null_count_stats_schema(inner) {
-                        DataType::from(inner_schema)
-                    } else {
-                        return None;
-                    }
+                    let inner_schema = null_count_stats_schema(inner)?;
```
shehabgamin: Probably better to retain the previous behavior for flexibility. Maybe add a debug log in the else statement

crates/sail-common-datafusion/src/catalog/status.rs
```diff
@@ -87,7 +87,7 @@ impl TableKind {
 
     pub fn type_name(&self) -> &str {
         match self {
-            TableKind::Table { .. } => "MANAGED",
+            TableKind::Table { .. } => "EXTERNAL",
```
shehabgamin: This should probably stay as managed, but i'm not fully sure. cc @linhr

crates/sail-catalog-unity/tests/rest_integration_test.rs
```diff
@@ -1060,6 +1060,7 @@ async fn test_get_table() {
         sort_by,
         bucket_by,
         properties,
+        ..
```
shehabgamin: Could we fully expand these please? Helps give context to AI

crates/sail-catalog-memory/src/provider.rs
```diff
@@ -307,7 +307,11 @@ impl CatalogProvider for MemoryCatalogProvider {
             .get_mut(table)
             .ok_or_else(|| CatalogError::NotFound(CatalogObject::Table, table.to_string()))?;
         match &mut status.kind {
-            TableKind::Table { properties, .. } => match options {
+            TableKind::Table {
+                location: _,
+                properties,
+                ..
```
shehabgamin, line 316: Could we fully expand these please? Helps give context to AI

.github/workflows/rust-tests.yml
santosh-d3vpl3x: yes indeed!

crates/sail-catalog-memory/src/provider.rs
```diff
@@ -232,6 +232,7 @@ impl CatalogProvider for MemoryCatalogProvider {
                 sort_by,
                 bucket_by,
                 properties,
+                is_external: true,
```
shehabgamin, line 237: Actually I'm not 100% sure if this would be considered external or not. I think it is, @linhr would know better than me

crates/sail-catalog/src/command.rs
```diff
@@ -395,6 +395,10 @@ impl CatalogCommand {
                 // update the catalog metadata, so we never end up with the two layers
                 // out of sync.
                 if let (Some(location), Some(format)) = (location, format) {
+                    if !is_lakehouse_format(&format) {
```
shehabgamin, line 403: Could you please elaborate on why these changes are necessary? I don't fully understand

crates/sail-plan/src/catalog.rs
```diff
@@ -150,22 +150,14 @@ impl CatalogObjectDisplay for SparkCatalogObjectDisplay {
...
-            TableKind::View { .. } => "VIEW",
-            TableKind::TemporaryView { .. } => "TEMPORARY",
-            TableKind::GlobalTemporaryView { .. } => "TEMPORARY",
-        };
-        let is_temporary = match status.kind {
-            TableKind::Table { .. } | TableKind::View { .. } => false,
-            TableKind::TemporaryView { .. } | TableKind::GlobalTemporaryView { .. } => true,
-        };
+        let table_type = status.kind.type_name().to_string();
+        let is_temporary = status.kind.is_temporary();
```
shehabgamin, line 154: Nice!

shehabgamin (commented):
First round of review 😃

crates/sail-delta-lake/src/spec/statistics.rs
santosh-d3vpl3x: Agreed!

crates/sail-sql-analyzer/src/statement.rs
santosh-d3vpl3x: Files extracted out of bigger PR. Should be fine to reset to `origin/main`.

crates/sail-common-datafusion/src/catalog/status.rs
santosh-d3vpl3x: TL;DR MANAGED vs EXTERNAL is mostly informational in Sail today. We preserve and report the HMS-recorded type for parity, but we do not yet implement distinct managed-table lifecycle semantics, and HMS DROP TABLE remains metadata-only regardless of type. Changing the default would require some follow-on code changes to preserve current behavior, but that would still be mostly mechanical plumbing rather than a true semantic redesign. So it will be low value in nature.

The main reason for these changes is metadata parity rather than new managed-table semantics.

Today Sail does not have a real managed-vs-external lifecycle split: there is no create-time knob for managed tables, planner read/write paths ignore the distinction, and DROP TABLE is not keyed off table type. But we still need to preserve the HMS-recorded type for interoperability and correct catalog surfaces.

Concretely:
- Sail-created tables still default to EXTERNAL.
- Tables created by other engines (for example Spark through HMS) can now surface as MANAGED or EXTERNAL based on the metastore metadata.
- DESCRIBE EXTENDED / catalog APIs should reflect that recorded type instead of flattening everything to MANAGED or EXTERNAL incorrectly.

One actual behavioral choice here is HMS DROP TABLE: this branch makes it metadata-only (`delete_data = false`) regardless of type, so we are explicitly not claiming managed-table deletion semantics yet.

crates/sail-catalog-memory/src/provider.rs
santosh-d3vpl3x: Probably this could help: https://github.com/santosh-d3vpl3x/sail/blob/884ee7fb4244fecca0bd9edbe9550ea48e77b590/crates/sail-catalog-memory/src/provider.rs#L273

crates/sail-catalog/src/command.rs
santosh-d3vpl3x: Added a comment to elaborate.

python/pysail/tests/spark/hms/test_spark_to_sail_roundtrip.py
```diff
@@ -0,0 +1,441 @@
...
+
+def _assert_sail_describes_spark_table(
+    hms_s3_spark: SparkSession,
+    table_fqn: str,
+    *,
+    table_type: str,
+) -> None:
+    properties = _describe_extended_properties(hms_s3_spark, table_fqn)
+
+    assert properties.get("Type") == table_type
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Not necessarily wrong but this review lacks complete context.

python/pysail/tests/spark/hms/test_hms_s3_roundtrip.py
```diff
@@ -0,0 +1,84 @@
...
+_S3_WAREHOUSE_PREFIX = "s3://hms-warehouse"
+
+
+def _assert_sail_describes_s3_managed_table(
+    hms_s3_spark: SparkSession,
+    table_fqn: str,
+    location_prefix: str,
+) -> None:
+    props = _describe_extended_properties(hms_s3_spark, table_fqn)
+    assert props["Type"] == "MANAGED"
```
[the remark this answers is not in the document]
santosh-d3vpl3x: Not necessarily wrong but this review lacks complete context.

python/pysail/tests/spark/hms/__init__.py
santosh-d3vpl3x: Looks like newly added file. With namespace packages, we can remove it.

santosh-d3vpl3x:
```
   INP001 File `python/pysail/tests/spark/hms/test_spark_to_sail_roundtrip.py` is part of an implicit namespace package. Add an `__init__.py`.
   --> python/pysail/tests/spark/hms/test_spark_to_sail_roundtrip.py:1:1
```
Doesn't look like a bad idea to keep it in reality.

crates/sail-catalog-memory/src/provider.rs
linhr: I think external tables make sense for memory catalog.

crates/sail-catalog/src/command.rs
linhr: We'll redesign this part in the near future so the fix looks good for now.

crates/sail-plan/src/catalog.rs
```diff
@@ -150,22 +150,14 @@ impl CatalogObjectDisplay for SparkCatalogObjectDisplay {
...
-        let table_type = match status.kind {
-            TableKind::Table { .. } => "MANAGED",
-            TableKind::View { .. } => "VIEW",
-            TableKind::TemporaryView { .. } => "TEMPORARY",
-            TableKind::GlobalTemporaryView { .. } => "TEMPORARY",
-        };
-        let is_temporary = match status.kind {
-            TableKind::Table { .. } | TableKind::View { .. } => false,
-            TableKind::TemporaryView { .. } | TableKind::GlobalTemporaryView { .. } => true,
-        };
```
linhr, line 162: Thanks for the clean up. This is a code duplication we missed.

python/pysail/tests/spark/hms/__init__.py
linhr: Yeah let's keep the empty `__init__.py`.

linhr (approved):
This is great!

PR #1879, opened 2026-05-09
refactor: unify `try_to_number` into `to_number` with safe mode flag

davidlghellin:


crates/sail-execution/src/codec.rs
```diff
@@ -2158,9 +2160,11 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_concat" | "concat" | "array_concat" => {
                 Ok(Arc::new(ScalarUDF::from(SparkConcat::new())))
             }
-            "spark_to_number" | "to_number" => Ok(Arc::new(ScalarUDF::from(SparkToNumber::new()))),
+            "spark_to_number" | "to_number" => {
+                Ok(Arc::new(ScalarUDF::from(SparkToNumber::new(false))))
+            }
             "spark_try_to_number" | "try_to_number" => {
-                Ok(Arc::new(ScalarUDF::from(SparkTryToNumber::new())))
+                Ok(Arc::new(ScalarUDF::from(SparkToNumber::new(true))))
```
linhr: I think we don't need these anymore now that we have `UdfKind::SparkToNumber`.

linhr (approved):
This is great!

We can merge this once the minor codec issue and the merge conflict are resolved.

crates/sail-execution/src/codec.rs
davidlghellin: done

linhr (approved):
Thanks!

Issue #1015, opened 2025-11-02
[Caching] Object Store Cache Using Foyer


shehabgamin:

wudihero2:
Hi @shehabgamin, do you have any thoughts on this?

When I use Trino with Iceberg, the number of S3 requests becomes very high and the cost gets out of control. In contrast, Trino can enable embedded Alluxio, which reduce the cost dramatically.

I have not studied the source code deeply, but based on how it behaves, my impression is that each Trino worker may maintain its own local file cache.

Based on that, do you think Sail could also use Foyer so that each worker caches its own files locally?

That said, I imagine there would be two main challenges:

1.Workers may need support for persistent local storage, such as fixed PVCs.
2.How should the cache key and value be designed? Would it make sense to design them around ObjectStore::get_range(path, range)?

Another possible approach would be to use an independent cache service, such as Alluxio, Liquid Cache, or [ClickHouse's distributed cache layer for S3](https://clickhouse.com/blog/building-a-distributed-cache-for-s3#a-new-layer-in-the-stack-the-network). But my feeling is that starting with local worker-level caching would probably be simpler.

shehabgamin:
@wudihero2 What I had in mind was adding a caching layer to the `sail-object-store` crate, similar to how we already have `RuntimeAwareObjectStore`:
https://github.com/lakehq/sail/blob/6993cf444cbd8d28ebdb4d76f701ee9792eb0b6b/crates/sail-object-store/src/layers/runtime.rs#L21

I think we could largely follow the approach in https://github.com/lancedb/ocra, but swap Moka for Foyer (and make sure to attribute any copied code in the comments). This would give us a local per-worker cache.

**Regarding PVC:**
PVCs can be handled via the pod template, so Sail doesn't need special handling for that, since it already supports a user-provided `worker_pod_template` that gets merged into the programmatically-built pod spec. See here:
https://github.com/lakehq/sail/blob/6993cf444cbd8d28ebdb4d76f701ee9792eb0b6b/crates/sail-execution/src/worker_manager/kubernetes.rs#L30

**Regarding cache key/value design:**
Yes, path and range seems good parts for the key. Some considerations:
1. Do we want to support TTL?
2. How is range merged and how to handle range overlap as a cache hit?

It's also worth looking at how OCRA handles this (it's been a while since I went through their codebase).

**Additional thoughts / questions:**
- The workers in Sail are auto-scaled, so the cache may be less effective once an idle worker is removed and new ones are created.
[...]

PR #1886, opened 2026-05-11
feat: implement a provider-agnostic caching layer for catalog listings

aminebizid:
This PR introduces a caching layer for catalog metadata listings to improve performance when querying external catalogs (Glue, HMS, Unity, etc.). It also includes configurable cache scopes to allow fine-grained control over how metadata is shared across sessions.

### Key Changes
- **CachingCatalogProvider**: A new decorator in sail-catalog that uses moka for high-performance async caching (LRU and TTL support). Updated to support trait objects (?Sized) for dynamic wrapping.
- **Cache Scoping**: Introduced cache_type to define the visibility of the cache:
global (default): Shared across all sessions (ideal for performance).
session: Private to the current session (ideal for security/isolation).
none: Disables the caching layer for that specific catalog.
- **Configuration**: Unified CatalogCacheConfig in sail-common. It supports Option types where 0 signifies unbounded/infinite caching (normalized automatically).
- Automatic Integration: All remote catalog providers are now automatically wrapped with the caching layer in the catalog manager.
- **Cache Invalidation**: Intelligent invalidation of database, table, and view caches upon write operations (CREATE/DROP/ALTER).
### Configuration Example
You can configure the cache scope and settings via environment variables:

```bash
# Example: Enabling a session-isolated cache for Glue
export SAIL_CATALOG__LIST='[{
    "type": "glue",
    "name": "my_catalog",
    "database_cache_type": "global",
    "table_cache_type": "session",
    "view_cache_type": "none",
    "database_cache_size": 2048,
    "table_cache_ttl_secs": 3600
}]'
```


crates/sail-catalog/src/provider/cache.rs
```diff
@@ -0,0 +1,687 @@
+#![allow(clippy::unwrap_used)]
```
linhr: We are not supposed to use `unwrap()` except in tests. Let's propagate the error properly.

crates/sail-catalog/src/provider/cache.rs
```diff
@@ -0,0 +1,687 @@
...
+            let c: &Cache<Namespace, Vec<TableStatus>> = c;
+            c.invalidate(database).await;
+        }
+        Ok(())
+    }
+}
+
+fn unwrap_catalog_error(e: std::sync::Arc<CatalogError>) -> CatalogError {
+    std::sync::Arc::try_unwrap(e).unwrap_or_else(|e| (*e).clone())
+}
```
linhr: This sounds strange to me.

crates/sail-catalog/src/error.rs
```diff
@@ -35,10 +35,10 @@ impl fmt::Display for CatalogObject {
     }
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Clone, Error)]
 pub enum CatalogError {
     #[error("error in DataFusion: {0}")]
-    DataFusionError(#[from] DataFusionError),
+    DataFusionError(std::sync::Arc<DataFusionError>),
```
linhr: These sound strange to me. Typically we won't need to `Clone` errors or use `Arc` for errors.

linhr (commented):
Thanks for the contribution! I'll take a look in more details shortly, but I've noticed a few unconventional patterns that indicate potential issues with the design. Could you take a look so that we don't need such workarounds at the Rust language level?

Also free feel to stop requesting more Copilot reviews. Typically one or two rounds of automated review would be enough. (But it's kind of strange that Copilot keeps generating new comments. I rarely saw this before.)

linhr:
And could you resolve all outdated Copilot comments if you believe those are fixed?

aminebizid:
Hi @linhr It's all marked as resolved. Could you check asap please, I need to work on another PR for Auth2 auth.
Kr,
Amine

aminebizid:
@linhr I also want to know if it's Ok for you this new config:
`export SAIL_CATALOG__LIST='[{type="glue", name="sail", region="us-west-2", database_cache_enabled=true, database_cache_size=100, database_cache_ttl_secs=3600, table_cache_enabled=true, table_cache_size=1000, table_cache_ttl_secs=3600}]'`

aminebizid:
@linhr last question:
Are you Ok to start working on a PR for data caching (like Databricks) ?

linhr:
> Could you check asap please, I need to work on another PR for Auth2 auth.

Thanks!

This PR will take some time to review and merge. I'd recommend working on top of the current `main` branch for other tasks you're interested in.

Also, could you share some background about the auth changes? It would be a good idea to start an issue to discuss. As the code itself can be mostly done by AI, it's often a good idea to align on the design first to reduce back and forth on the PR review.

> I also want to know if it's Ok for you this new config:

I have some rough thoughts around supporting global/session/none cache types, but let's consider this out of scope for this PR.

linhr:
> Are you Ok to start working on a PR for data caching (like Databricks) ?

This is a big topic. See #1015 for the ongoing discussion.

crates/sail-catalog/Cargo.toml
```diff
@@ -19,3 +19,6 @@ lazy_static = { workspace = true }
 async-trait = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
+moka = { workspace = true }
+sail-common = { path = "../sail-common" }
```
linhr: Move workspace crate to the top of `[dependencies]`.

crates/sail-catalog/src/error.rs
```diff
@@ -52,3 +52,9 @@ pub enum CatalogError {
     #[error("external error: {0}")]
     External(String),
 }
+
+impl CatalogError {
+    pub fn datafusion(err: DataFusionError) -> Self {
+        Self::DataFusionError(err)
+    }
+}
```
linhr: Remove this if it is unused.

crates/sail-common/src/config/application.rs
```diff
@@ -404,14 +404,68 @@ pub struct FileMetadataCacheConfig {
...
 #[serde(rename_all = "snake_case")]
 pub enum CacheType {
+    #[default]
     None,
     Global,
     Session,
 }
 
+fn default_catalog_cache_type() -> CacheType {
+    CacheType::Global
```
linhr: The default should be `None` for backward compatibility.

crates/sail-common/tests/catalog_cache_config.rs
linhr, line 1: Remove this test file.

crates/sail-common/src/config/application.rs
```diff
@@ -404,14 +404,68 @@ pub struct FileMetadataCacheConfig {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct CatalogCacheConfig {
+    #[serde(default = "default_catalog_cache_type")]
+    pub cache_type: CacheType,
```
linhr: Let's consolidate `cache_type`, `database_cache_enabled`, and `table_cache_enabled`.

We should have `database_cache_type` and `table_cache_type`. The default cache type should be `CacheType::None`. We should also have configuration for `view_cache`.

This means the cache bundle needs to be split so that we can support global/session cache for database/table/view separately.

crates/sail-common/src/config/application.rs
```diff
@@ -404,14 +404,68 @@ pub struct FileMetadataCacheConfig {
     pub size_limit: Option<u64>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
 #[serde(rename_all = "snake_case")]
 pub enum CacheType {
+    #[default]
```
linhr: I don't think we need these changes. It's better to avoid default here to prevent misuse. Also without `PartialEq` and `Eq` we can still compare the value by `matches!`.

linhr (commented):
Nice progress! I feel we're getting fairly close. Let's address these comments and then we can merge the PR.

aminebizid:
@linhr 
Done, example:
`export SAIL_CATALOG__LIST='[{type="glue", name="sail", region="us-west-2", database_cache_type="global", database_cache_size=100, database_cache_ttl_secs=3600, table_cache_type="session", table_cache_size=1000, table_cache_ttl_secs=3600}]'`

linhr:
Thanks!

I think you need to merge from `main` and fix lint and test errors. There was a PR that was just merged that requires adjustments to this PR.

aminebizid:
@linhr rebased.
The lint problem was caused by the previous merged PR (is_external).

linhr (approved):
Everything looks good now! Thanks for your contribution!

Issue #1568, opened 2026-03-28
[Bug] `SparkRuntimeException` when aggregating column with custom metadata


keen85:
### Description
Performing an [aggregation](https://spark.apache.org/docs/latest/api/python/reference/pyspark.pandas/api/pyspark.pandas.DataFrame.agg.html) on a column that has custom metadata attached via [`DataFrame.withMetadata()`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.withMetadata.html) the execution fails.

### Error Message
`SparkRuntimeException: Physical input schema should be the same as the one converted from logical input schema.`
Same error as in #1422.

### Environment
- Sail version: 0.5.3
- Platform: Windows 11

### Minimal Reproducible Example
```python
df_raw = spark.createDataFrame(
    schema="id INTEGER, some_string STRING",
    data=[
        (1, "A"),
        (2, "B"),
    ],
)
df_with_metadata = df_raw.withMetadata("some_string", {"foo": "bar"})

# ✅ works fine
df_raw.select(F.max("some_string")).show()

# ✅ works fine
df_with_metadata.show()

# ✅ works fine
print("Metadata:", df_with_metadata.schema["some_string"].metadata)

# ❌ fails: SparkRuntimeException: Physical input schema should be the same as the one converted from logical input schema.
df_with_metadata.select(F.max("some_string")).show()
```

<details>

<summary>stack trace</summary>

```python
---------------------------------------------------------------------------
SparkRuntimeException                     Traceback (most recent call last)
Cell In[50], [line 22](vscode-notebook-cell:?execution_count=50&line=22)
     19 print("Metadata:", df_with_metadata.schema["some_string"].metadata)
     21 # ❌ fails: SparkRuntimeException: Physical input schema should be the same as the one converted from logical input schema.
---> [22](vscode-notebook-cell:?execution_count=50&line=22) df_with_metadata.select(F.max("some_string")).show()

File c:\dev\GitHub\dataengineering-playground\.venv\Lib\site-packages\pyspark\sql\connect\dataframe.py:1153, in DataFrame.show(self, n, truncate, vertical)
   1152 def show(self, n: int = 20, truncate: Union[bool, int] = True, vertical: bool = False) -> None:
-> [1153](file:///C:/dev/GitHub/dataengineering-playground/.venv/Lib/site-packages/pyspark/sql/connect/dataframe.py:1153)     print(self._show_string(n, truncate, vertical))

[...]

PR #1895, opened 2026-05-13
fix: preserve Spark interval type qualifiers

shehabgamin:
Continuing in https://github.com/lakehq/sail/pull/1901


shehabgamin:
@claude Review thoroughly. Behavior should match Apache Spark exactly. ChatGPT wrote the code ;)

crates/sail-function/src/aggregate/preserve_input_metadata.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn return_type(&self, arg_types: &[DataType]) -> Result<DataType> {
+        self.inner.return_type(arg_types)
+    }
+
+    /// Copy the Sail interval extension metadata from the first input field
+    /// onto the inner UDAF's return field. Only the Sail interval keys are
+    /// propagated — other input metadata is left to the inner UDAF, which
+    /// already handles e.g. dictionary-encoded inputs via its default
+    /// `return_field` implementation.
+    fn return_field(&self, arg_fields: &[FieldRef]) -> Result<FieldRef> {
```
shehabgamin, line 84: This should actually handle all meta

crates/sail-common-datafusion/src/logical_expr.rs
```diff
@@ -14,3 +17,27 @@ impl ExprWithSource {
...
+/// this in place of `expr.alias(name)` anywhere we know the Field metadata
+/// that should flow downstream — `to_field` on the alias will then surface
+/// the metadata on the Logical Field. Pass `HashMap::new()` (or
+/// `Default::default()`) when there is no metadata to carry.
+///
+/// Note: DataFusion's physical planner drops Alias metadata for non-`Literal`
+/// inner expressions (see `datafusion-physical-expr/src/planner.rs`), so this
+/// alone won't carry metadata onto the physical schema for `BinaryExpr`,
+/// `AggregateFunction`, or `ScalarFunction` inner exprs — those need their
+/// own `return_field` / `return_field_from_args` to attach metadata.
```
linhr, line 32: This is a bug first noticed in <https://github.com/lakehq/sail/issues/1568#issuecomment-4200629573>. The best approach is to fix it upstream so that we don't have to patch it in our codebase.

PR #1913, opened 2026-05-16
feat: from_json builder approach

kentkr:
Update the `from_json` function to use a builder pattern rather than scalar values.

Brief discussion in #1818


crates/sail-function/src/scalar/json/from_json.rs
```diff
@@ -212,381 +212,599 @@ fn from_json_inner(args: &[ArrayRef], session_timezone: &str) -> Result<ArrayRef
...
 
-    let offset_buffer = OffsetBuffer::new(ScalarBuffer::from(offsets));
-    Ok(Arc::new(ListArray::new(
-        element_field.clone(),
-        offset_buffer,
-        values_array,
-        Some(validity.into()),
-    )))
+#[derive(Debug)]
+enum FieldBuilder {
```
shehabgamin, line 290: @kentkr Could you add a [Credit] comment pointing to the Comet code? You can see other places in our codebase for how to give attribution for code copied from other repos.

shehabgamin (commented):
@kentkr Thanks for the PR! I am having copilot do an initial review and then I will do a full review shortly after.

Just left a comment for 1 thing I noticed thus far 😃

kentkr:
@shehabgamin should be set for a full review

shehabgamin (approved):
@kentkr Well done!! Thanks for the contribution 😃

PR #1914, opened 2026-05-16
fix: stale expected_partitions in PythonDataSourceWriteCommitExec causes intermittent write failure

nis12ram:
## Problem

Writing to a Python datasource fails intermittently with:

```text
AnalysisException: Invalid partition id 2 in write result (expected < 2)
```

reproducible example (triggers on machines where CPU cores > input partition count):

```python
from pyspark.sql.datasource import DataSource, DataSourceArrowWriter
from collections.abc import Iterator

class NoopWriter(DataSourceArrowWriter):
    def write(self, iterator: Iterator) -> None:
        for _ in iterator:
            pass

class NoopDataSource(DataSource):
    @classmethod
    def name(cls) -> str:
        return "noop"

    def writer(self, schema, overwrite):
        return NoopWriter()

server = SparkConnectServer()
server.start()
spark.dataSource.register(NoopDataSource)

spark.range(100000).repartition(2).write.mode("overwrite").format("noop").save()
```

## Root Cause

`expected_partitions` is captured at plan-creation time in `table_format.rs`. After the optimizer inserts `RepartitionExec` and `CoalescePartitionsExec`, `with_new_children` on `CommitExec` preserves the stale value instead of re-deriving it from the actual `WriteExec` partition count.

## Fix
In `with_new_children`, traverse the new child tree to find `PythonDataSourceWriteExec`'s actual partition count rather than reusing the stale `self.expected_partitions`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -705,9 +705,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-glue"
-version = "1.145.0"
+version = "1.145.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95190a2c2e0be9088d0f26ad4931d91defed3a5f584559ecb71189de6d4bc238"
+checksum = "6eaa019d39389807e4681e6a040a2268fa5dce7c2be4407117d54f52c0c77bb5"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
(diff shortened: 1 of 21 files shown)


nis12ram:
@linhr let me know what you think 🙂

nis12ram:
Apologies — I accidentally ran the wrong Git commands and messed up the commits, and I wasn’t sure how to recover them properly.

So I recreated the feature branch and raised the PR again https://github.com/lakehq/sail/pull/1934

PR #1920, opened 2026-05-17
feat: improve Iceberg REST catalog support

shehabgamin:


crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -1209,24 +1286,62 @@ fn columns_to_nested_fields(
...
+                        CatalogError::InvalidArgument(format!(
+                            "Failed to parse default value as JSON for column '{name}': {e}"
+                        ))
+                    })?;
+                Some(Literal::try_from_json(json_default, &field_type)
+                    .map_err(|e| {
+                        CatalogError::InvalidArgument(format!(
+                            "Failed to convert default value to Iceberg literal for column '{name}': {e}"
+                        ))
+                    })?
```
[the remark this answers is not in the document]
shehabgamin: Nice catch

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -1036,30 +1072,43 @@ impl CatalogProvider for IcebergRestCatalogProvider {
...
         for (k, v) in properties {
+            let trimmed = v.trim();
+            if !trimmed.is_empty() {
+                if k.eq_ignore_ascii_case("path") {
+                    path = Some(trimmed.to_string());
+                } else if k.eq_ignore_ascii_case("location") {
+                    location = Some(trimmed.to_string());
+                }
+            }
             props.insert(k, v);
```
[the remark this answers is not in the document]
shehabgamin: intentional

crates/sail-common/src/config/application.rs
```diff
@@ -519,6 +519,13 @@ pub enum CatalogType {
         warehouse: Option<String>,
         #[serde(skip_serializing_if = "Option::is_none")]
         prefix: Option<String>,
+        #[serde(
+            alias = "namespace-separator",
+            alias = "namespaceSeparator",
+            alias = "namespaceseparator",
```
linhr, line 525: We don't usually define alias for catalog provider options.

We are more flexible for data source options due to Spark compatibility, but for catalog provider (and application config in general) we can have our own strict style.

crates/sail-iceberg/src/spec/types/values.rs
```diff
@@ -105,7 +106,318 @@ impl Datum {
...
+    Ok(acc)
+}
+
 impl Literal {
+    pub fn try_from_str(
+        value: &str,
+        data_type: &crate::spec::types::Type,
+    ) -> Result<Option<Self>, String> {
+        let literal = sail_literal_from_str(value, data_type)?;
+        Self::try_from_sail_literal(literal, data_type)
```
linhr, line 343: I don't quite understand why we need the Sail spec as a bridge when parsing Iceberg literals. IMO we should just define a parser in Chumsky or use the existing SQL parser and then convert SQL AST to the Iceberg spec.

linhr (approved):

crates/sail-common/src/config/application.rs
linhr: #1939

PR #1923, opened 2026-05-17
fix: qualify warehouse, database, and table locations for Spark parity

santosh-d3vpl3x:
## Summary

This PR aligns location qualification with Spark for Spark-managed catalogs while preserving catalog-owned semantics for native catalogs.  
It also removes UUID-based randomness from Spark-managed default table locations so managed table roots stay deterministic.

## What changed

- Added shared path qualification helpers for warehouse/database/table location handling.
- `CREATE DATABASE` policy split by provider:
  - Spark-managed catalogs (HMS/Glue/Memory): synthesize default DB location when `LOCATION` is omitted.
  - Native namespace catalogs (e.g. Iceberg REST/Unity): keep `location = None` unless explicitly provided.
- `CREATE TABLE` / `CTAS` policy split by provider:
  - Spark-managed catalogs: qualify locations with Spark-like rules and derive deterministic default table locations.
  - Native catalogs: preserve explicit relative `path/location` and defer default location choice to the catalog.
- Namespace metadata hydration for Iceberg REST now resolves location from:
  - `location`, else `warehouse`, else `path`.
  - This fixes external namespace provenance cases (e.g., Nessie/Gravitino-created namespaces).
- Path normalization follow-ups:
  - Preserve Windows path prefix/root semantics during normalization.
  - Added Windows coverage for drive-letter and UNC normalization.
  - Removed HOME-based tilde expansion from this PR to avoid env-dependent behavior.

## Why UUID removal is in scope

The UUID suffix previously prevented path collisions in tests by changing managed table location behavior.  
That introduces nondeterministic product paths in Spark-managed catalogs, which hurts:

- Spark parity
- Predictable storage layout and debugging
- Path-first interop use cases (DuckDB/Polars and similar tooling)

Test isolation should be handled in test fixtures (warehouse/db/table isolation), not in default location derivation.

## Validation

- Rust unit coverage for path qualification and provider-policy behavior.
- Spark/HMS local + S3 path parity tests.
- `CREATE TABLE` feature coverage for deterministic default locations and special-character table names.


linhr:
Thanks for looking into this! We encountered a related issue for the Iceberg REST catalog as @alexy was testing Nessie and Apache Gravitino.

How do we feel if we fallback to the `warehouse` property [here](https://github.com/lakehq/sail/blob/784b32d57f586b7473e8945871df52b5dc1ddede/crates/sail-catalog-iceberg/src/provider.rs#L675-L678) to prevent the global default `spark-warehouse/` from being used in `resolve_default_table_location`? I guess we can apply similar logic in HMS. The idea is to resolve the database location when _getting_ the database, while this PR addresses the issue when _creating_ the database (via `uses_spark_default_database_location` and `resolve_catalog_create_database`). I feel the former approach is more general since not all databases are created by Sail so they may lack the location anyway.

(I haven't looked at the PR closely so let me know if I missed any context. I'm just writing down an idea for discussion.)

crates/sail-plan/src/resolver/command/catalog/table.rs
```diff
@@ -203,55 +224,21 @@ impl PlanResolver<'_> {
...
-        // Note that this is different from how Spark handles table locations
-        // for the default catalog.
         let catalog_manager = self.ctx.extension::<CatalogManager>()?;
         let location = catalog_manager
             .get_database_by_qualifier(qualifier)
             .await?
             .location;
-        let (base, suffix) = match &location {
-            Some(loc) => (
-                loc.trim_end_matches(object_store::path::DELIMITER),
```
linhr, line 250: We may have to keep this UUID suffix. This is not the same behavior as Spark, but it ensures uniqueness of take locations during tests. Otherwise we would have a hard time making all Spark tests to work.

This won't affect how table locations are derived when catalogs are in use, so this won't impact production use cases.

python/pysail/tests/spark/hms/conftest.py
linhr, line 1: BTW all HMS tests are temporary skipped to unblock the previous release. You can revert the changes to this file in #1940. Let me take care of the fix for the test setup.

shehabgamin:
> Thanks for looking into this! We encountered a related issue for the Iceberg REST catalog as @alexy was testing Nessie and Apache Gravitino.
> 
> How do we feel if we fallback to the `warehouse` property [here](https://github.com/lakehq/sail/blob/784b32d57f586b7473e8945871df52b5dc1ddede/crates/sail-catalog-iceberg/src/provider.rs#L675-L678) to prevent the global default `spark-warehouse/` from being used in `resolve_default_table_location`? I guess we can apply similar logic in HMS. The idea is to resolve the database location when _getting_ the database, while this PR addresses the issue when _creating_ the database (via `uses_spark_default_database_location` and `resolve_catalog_create_database`). I feel the former approach is more general since not all databases are created by Sail so they may lack the location anyway.

@linhr @santosh-d3vpl3x

The [code snippet](https://github.com/lakehq/sail/blob/784b32d57f586b7473e8945871df52b5dc1ddede/crates/sail-catalog-iceberg/src/provider.rs#L675-L678) Heran shared should also check for `path` I think

santosh-d3vpl3x:
Took a while to find right shape for handling the paths properly.

santosh-d3vpl3x:
Good call on `path` and read time qualified path for databases, both are now handled.  Handling both non-sail-> catalog-> sail and sail -> catalog -> external paths become gentle to each other. 

We pass certain responsibilities to proper native catalog as they are very well geared to handle path qualification themselves.

For Nessie/Gravitino like issues, We should still add some regression tests that capture individual behaviors from those catalogs. We maybe able to then implement fallback logic well. Do we know what was the issue there and do we still need to handle it?

crates/sail-plan/src/resolver/command/catalog/table.rs
santosh-d3vpl3x: Good point on uniqueness in tests. I still think we should keep the UUID removal in this PR.

The UUID suffix changes real managed-table behavior whenever DB location is unset (warehouse fallback), so it is not only a test detail. For Spark-managed catalogs we should keep deterministic db.table -> path mapping for Spark parity and path-first interop (DuckDB/Polars/storage tooling like we use in our clusters). For native catalogs we now defer default location ownership to the catalog, so this UUID tradeoff is mainly about Spark-managed filesystem/object-store tables.

I’d prefer to solve collision risk in test fixture isolation (warehouse/db/table uniqueness) rather than injecting randomness into default table location derivation. If we still see flakes, we can take a focused follow-up on test setup.

santosh-d3vpl3x:
Can we enable spark and catalog tests?

crates/sail-plan/src/resolver/command/catalog/table.rs
linhr: I agree that it's less desirable to leak the UUID suffix convention externally. And I can see that it's not only a test setup and it has real world impact.

The unfortunate fact is that it affects parity tests borrowed from the Spark codebase. As you can see from the report that they are multiple tests failing now. Many Spark tests were written poorly so conflicts arise when we run them.

I feel we do want to remove randomness in table locations. So we either debug those tests and figure out solutions, or revert this logic for now and do some investigation in follow-up PRs. I feel the latter approach might be easier since this PR is very large already.

crates/sail-physical-plan/src/file_write.rs
```diff
@@ -54,3 +83,147 @@ pub async fn create_file_write_physical_plan(
...
+    options: Vec<OptionLayer>,
+    logical_schema: DFSchemaRef,
+}
+
+#[derive(Debug)]
+struct DeferredCatalogTableWriteExec {
+    state: SessionState,
+    input: Arc<dyn ExecutionPlan>,
+    info: DeferredCatalogTableWriteInfo,
+    properties: Arc<PlanProperties>,
```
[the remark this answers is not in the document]
linhr: We do have codec for `DeferredCatalogTableWriteExec`.

crates/sail-physical-plan/src/file_write.rs
```diff
@@ -54,3 +82,209 @@ pub async fn create_file_write_physical_plan(
...
+                .with_aggregate_functions(context.aggregate_functions().values().cloned().collect())
+                .with_window_functions(context.window_functions().values().cloned().collect())
+                .build();
+            let registry = context.extension::<TableFormatRegistry>()?;
+            let plan = registry
+                .get(&info.format)?
+                .create_writer(&session_state, sink)
+                .await?;
+            plan.execute(partition, context)
+        })
```
linhr: It's an anti-pattern to create session and physical plan inside `execute()`. This would cause surprising issues since the session created on the fly may not have full setup for extensions. Also this won't work when `execute()` is executed in worker which has different setup than the driver.

crates/sail-plan/src/resolver/command/catalog/mod.rs
```diff
@@ -1,3 +1,37 @@
 pub(super) mod database;
 pub(super) mod table;
 pub(super) mod view;
+
+use crate::error::{PlanError, PlanResult};
+
+fn validate_location_identifier(name: &str, kind: &str) -> PlanResult<()> {
```
linhr, line 7: Spark is actually stricter in table name validation. In my testing, for example, `@` is not allowed. For maximum compatibility with various object stores, if we have to derive directory name from table name, I feel we should restrict name to valid SQL identifiers (alphanumeric character, `-`, `_`). We can also find out how Spark handles this in its codebase.

```
pyspark.errors.exceptions.captured.AnalysisException: [INVALID_SCHEMA_OR_RELATION_NAME] `my@table` is not a valid name for tables/schemas. Valid names only contain alphabet characters, numbers and _. SQLSTATE: 42602
```

python/pysail/tests/spark/write_table/features/create_table.feature
```diff
@@ -103,7 +103,7 @@ Feature: CREATE TABLE default location
       | 2  |
     Then file tree in db_location matches
       """
-      📂 myu+0040table
+      📂 my@table
```
linhr: See also my comment for `validate_location_identifier`.

crates/sail-spark-connect/src/config.rs
```diff
@@ -362,3 +362,49 @@ impl TryFrom<&SparkRuntimeConfig> for PySparkUdfConfig {
         Ok(output)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
linhr: Tests like these are not super useful.

crates/sail-plan/src/config.rs
```diff
@@ -49,10 +58,84 @@ impl Default for PlanConfig {
...
+            default_warehouse_directory: qualify_warehouse_directory("spark-warehouse"),
             session_user_id: "".to_string(),
             ansi_mode: false,
             cross_join_enabled: true,
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
linhr: Most tests here are testing functionalities for `sail_common::path`.

crates/sail-plan/src/resolver/command/catalog/table.rs
```diff
@@ -566,3 +607,734 @@ fn extract_sort_int_arg(args: &[spec::Expr], index: usize, description: &str) ->
...
+        })
+    };
+    find("path").or_else(|| find("location"))
+}
+
+fn is_table_location_option(key: &str) -> bool {
+    key.eq_ignore_ascii_case("path") || key.eq_ignore_ascii_case("location")
+}
+
+#[cfg(test)]
```
linhr: All these tests should be removed as they are not maintainable. If the coverage drops as reported by Codecov, we should add end-to-end Python or BDD tests.

crates/sail-plan/src/resolver/command/catalog/table.rs
```diff
@@ -199,62 +217,86 @@ impl PlanResolver<'_> {
...
+        let catalog_manager = self.ctx.extension::<CatalogManager>()?;
+        let provider = catalog_manager.database_provider_by_qualifier(qualifier)?;
+        Ok(provider.uses_spark_default_table_location())
+    }
+
+    async fn qualify_table_location_options(
+        &self,
+        table: &spec::ObjectName,
+        options: &mut [(String, String)],
+    ) -> PlanResult<()> {
```
linhr, line 313: Could you share the reason why we need to "patch" the options? If we don't do this what would break?

This is just for my learning and information gathering for future refactoring.

crates/sail-delta-lake/src/operations/write/writer.rs
```diff
@@ -629,15 +629,18 @@ mod tests {
 
     #[cfg(debug_assertions)]
     #[tokio::test]
-    #[should_panic(expected = "input violated partition grouping contract")]
-    async fn debug_contract_panics_on_partition_key_regression() {
+    async fn debug_contract_tracks_closed_partition_keys() {
```
linhr, line 885: What's the reason for changing this test?

@lonless9 Could you take a look at this file and see if the changes are fine?

linhr (commented):
Thanks for having this! Since the PR is rather large, I'd suggest the following changes so that we can merge this quickly and follow up separately.

1. Let's add back the UUID suffix for default table names. As I mentioned in the comment, this is causing issue with Spark parity tests which are out of our control. We indeed want to remove UUID suffix in the long term, but fixing the test setup is non-trivial so let's consider it out of scope for now.
2. I don't quite understand the need for `DeferredCatalogTableWriteExec` and its implementation is flawed as I mentioned in the comment, so this physical node has to be removed. I thought the logic for resolving table location can be done entirely in `crates/sail-plan/src/resolver/command/write.rs`, but if it's not possible, let's create an issue to track the use case and consider it out of scope for this PR.

Also, moving forward let's heavily reduce the amount of Rust unit tests and favor end-to-end Python or BDD tests. I have pointed out a few places in my comments, but basically all the tests that requires excessive setup or mock need to be removed. Testing is good in general and I really appreciate the efforts. However, given that AI is so good at massive test generation, we have to make a tradeoff to leave out certain tests in our codebase, to avoid increasing CI time and maintenance burden in the long term.

And let's merge from `main` after making the changes so that we can get a clean diff of Spark parity tests for this PR.

Besides these notes, the approach makes sense to me overall!

linhr:
> For Nessie/Gravitino like issues, We should still add some regression tests that capture individual behaviors from those catalogs. We maybe able to then implement fallback logic well. Do we know what was the issue there and do we still need to handle it?

Yeah, catalog-specific tests for these would be valuable. Let's consider this out of scope as we're still manually evaluating/testing the integration with these catalogs and we would hopefully have more ideas soon. (cc @alexy)

crates/sail-delta-lake/src/operations/write/writer.rs
lonless9: This lowers the requirements for this test. I tried the original test, and it still passes in this PR. But keeping it this way is fine, as we may fully refactor the writer in the future.

linhr:
@santosh-d3vpl3x Would you be able to follow up with this PR this week? I feel this is largely in good shape and I'd like to have it merged, and the remaining items in my comments could be considered afterwards. We're making many major changes around catalog and table operations, so it would be good to merge this PR early to avoid merge conflicts accumulated over time. Let me know if there is anything I can help with!

santosh-d3vpl3x:
Thanks for feedback and review! I have been cleaning up stuff. I will push today-ish :)

santosh-d3vpl3x:
After digging into the failures here, I think this PR has crossed a boundary where it is mixing two separate problems:

1. path qualification / Spark parity
2. managed CTAS/write-path behavior

The regressions are a sign that the second problem is not just another policy branch inside the first. Some catalogs want provider-owned location behavior, but CTAS/write planning still needs a concrete write target early enough that this cannot be treated as a pure path-qualification change.

Because of that, I’m no longer convinced this PR is the right vehicle. I prefer to stop extending this PR and come back with a narrower follow-up that isolates the CTAS/write-path problem from the location-parity work. I think that will be much easier to reason about, test, and review.

santosh-d3vpl3x:
[atomic-ctas-maintainer-decision.pdf](https://github.com/user-attachments/files/29329752/atomic-ctas-maintainer-decision.pdf)

adding the further analysis here for future reference.

tl;dr introduce path parity with spark by moving to provider managed model, and add defaultFS and warehouse paths. CTAS makes things tricky, the pdf shades light on it.

PR #1924, opened 2026-05-17
fix: fixing create() table for Nessie, adding test for it

alexy:
fix: always set stage-create for Iceberg REST create, and add a test for it

When using Nessie for an Iceberg provider, I found that it insists on the stage-create parameter being present.  It should be set to false, but present.  Codex and I have also discovered extraneous trailing slashes that could be trimmed.  The integration test is also added.

### DIFF
--- a/crates/sail-catalog-iceberg/src/provider.rs
+++ b/crates/sail-catalog-iceberg/src/provider.rs
@@ -125,9 +125,9 @@ impl IcebergRestCatalogProvider {
                     .await?;
 
                 let uri = if let Some(uri) = config.overrides.remove(REST_CATALOG_PROP_URI) {
-                    uri
+                    uri.trim_end_matches('/').to_string()
                 } else {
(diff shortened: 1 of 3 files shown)


crates/sail-catalog-iceberg/tests/rest_integration_test.rs
```diff
@@ -117,6 +117,106 @@ async fn setup_catalog(
     (catalog, minio, mc, rest)
 }
 
+async fn setup_nessie_catalog(
```
shehabgamin: Can you move the tests to https://github.com/lakehq/sail/tree/main/python/pysail/tests/spark/catalog_integration/iceberg_rest

shehabgamin (commented):
@alexy This looks good to me, let's move the tests over and then we can merge this in

shehabgamin (approved):

PR #1929, opened 2026-05-18
feat: catalog managed Iceberg table

lonless9:
#172


crates/sail-iceberg/src/physical_plan/commit/commit_exec.rs
linhr, line 491: This `execute` method is nearly 600 lines now. Given that we've delivered many advanced features so far, I feel it might be a good time to look at the codebase holistically and plan some refactoring work.

linhr (approved):
Great work!

linhr (approved):

PR #1942, opened 2026-05-20
fix: propagate NULL from parent struct in field extraction

davidlghellin:


lonless9 (commented):
I manually tested and found that the SQL `VALUES ... CAST(NULL AS STRUCT...)` BDD cases pass even without the Rust fix. Could you add the dataframe test using explicit `StructField(..., metadata=...)`? 
Besides Spark preserves child field metadata for s.id, s['id'], and nested fields, and the old direct get_field path also preserved it. The new `CASE WHEN parent IS NULL THEN NULL ELSE get_field(...) END` path fixes values but drops child metadata.

davidlghellin:
https://github.com/lakehq/sail/pull/2060

Issue #325, opened 2024-12-11
Support duplicated field names in Spark schema


linhr:
Spark handles duplicated field names via `deduplicateFieldNames` in `ArrowUtils.scala`. We should incorporate the same logic for conversion between Arrow schema and Spark schema.

This would allow the test case `test_createDataFrame_duplicate_field_names` to pass.

PR #1950, opened 2026-05-21
fix: improve complex type handling in Arrow/Pandas conversion

lonless9:


crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -176,7 +184,11 @@ impl TryFrom<adt::Field> for sdt::StructField {
         let metadata = &field
             .metadata()
             .iter()
-            .filter(|(k, _)| !k.starts_with("udt.") && !k.starts_with("ARROW:extension:"))
+            .filter(|(k, _)| {
+                !k.starts_with("udt.")
```
linhr: The way to support UDT has changed in #1938 so we need to make changes in related files.

crates/sail-spark-connect/src/proto/data_type_arrow.rs
```diff
@@ -6,6 +6,12 @@ use serde::Deserialize;
 use crate::error::{SparkError, SparkResult};
 use crate::spark::connect::{data_type as sdt, DataType};
 
+const DAY_TIME_INTERVAL_START_FIELD_KEY: &str = "sail.spark.dayTime.interval.startField";
+const DAY_TIME_INTERVAL_END_FIELD_KEY: &str = "sail.spark.dayTime.interval.endField";
+const YEAR_MONTH_INTERVAL_START_FIELD_KEY: &str = "sail.spark.yearMonth.interval.startField";
+const YEAR_MONTH_INTERVAL_END_FIELD_KEY: &str = "sail.spark.yearMonth.interval.endField";
+const SAIL_SPARK_METADATA_PREFIX: &str = "sail.spark.";
```
linhr: I've discussed it with @shehabgamin that we'd like interval to be an Arrow extension type. The idea can be found in #1901 but we can implement it in this PR. We'd implement it with the `ExtensionType` trait similar to how we did the refactoring in #1958.

linhr (commented):
I haven't reviewed the PR but here are just two thoughts. I'll go through the PR once these are handled.

python/pysail/tests/spark/test_variant.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+from pyspark.sql import Row
+from pyspark.sql import functions as spark_functions
+from pyspark.sql import types as spark_types
```
linhr: For style consistency, these should be imported as `F` and `T` with `# noqa`.

python/pysail/tests/spark/delta/__snapshots__/features/operation_metrics.yaml
```diff
@@ -62,31 +62,16 @@ data:
...
     numOutputRows: 3
-    numRemovedBytes: 751
+    numRemovedBytes: 720
     numRemovedFiles: 1
     numTouchedRows: 5
     rewriteTimeMs: <time_ms>
     scanTimeMs: <time_ms>
     writeTimeMs: <time_ms>
 ---
-name: "test_dv_delete_reports_numdeletionvectorsadded_and_zero_numcopiedrows"
```
linhr, line 75: I noticed these two unused snapshots locally as well. Interestingly these didn't cause errors on CI.

crates/sail-python-udf/src/python/spark.py
```diff
@@ -340,6 +366,71 @@ def from_pyspark(self, data: Sequence[Any]) -> pa.Array:
...
+                metadata.append(None)
+                mask.append(True)
+                continue
+            mask.append(False)
+            if variant_val_type is not None and isinstance(x, variant_val_type):
+                internal = self._spark_data_type.toInternal(x)
+            elif isinstance(x, dict) and all(key in x for key in ["value", "metadata"]):
+                internal = x
+            else:
+                internal = self._spark_data_type.toInternal(x)
```
linhr: The first and third branches are the same, so we don't actually need to check type against `VariantVal`.

crates/sail-spark-connect/src/schema.rs
```diff
@@ -1,15 +1,100 @@
...
+        ),
+        ArrowDataType::Map(field, sorted) => ArrowDataType::Map(
+            normalize_schema_field_for_pandas(field, field.name().clone()),
+            *sorted,
+        ),
+        _ => data_type.clone(),
+    }
+}
+
+pub(crate) fn deduplicate_field_names(fields: &Fields) -> Vec<String> {
```
linhr: Does this fix #325?

Spark deduplicates field names when converting Spark data types to Arrow data types (during input processing), but here we deduplicate field names during output processing. Are we fixing the root cause here to match Spark behavior?

crates/sail-spark-connect/src/executor.rs
```diff
@@ -319,12 +324,207 @@ pub(crate) async fn read_stream(
...
+            schema,
+            columns,
+            &RecordBatchOptions::default().with_row_count(Some(batch.num_rows())),
+        )?)
+    } else {
+        Ok(RecordBatch::try_new(schema, columns)?)
+    }
+}
+
+fn normalize_fields_for_pandas(fields: &Fields, deduplicate: bool) -> Fields {
```
linhr: Are these functions the same as the ones in `schema.rs`?
* `normalize_fields_for_pandas`
* `normalize_field_for_pandas`
* `normalize_data_type_for_pandas`

crates/sail-plan/src/resolver/expression/cast.rs
```diff
@@ -96,6 +98,24 @@ impl PlanResolver<'_> {
...
+                DataType::LargeUtf8,
+            ),
+            (from, DataType::Utf8View, _) if is_variant_storage_type(&from) => {
+                ScalarUDF::new_from_impl(SparkVariantToJsonUdf::new()).call(vec![expr])
+            }
+            (from, to, is_try) if is_variant_storage_type(&from) => {
+                ScalarUDF::new_from_impl(SparkVariantGet::new(is_try)).call(vec![
+                    expr,
+                    lit("$"),
+                    lit(to.to_string()),
```
linhr: `to.to_string()` doesn't seem right since `to` is an Arrow data type but the function expects a Spark data type string. `data_type_to_simple_string` would be a better approximation here.

In the future we may want to refactor `SparkVariantGet` so that it accepts an Arrow type directly when constructing the function.

crates/sail-spark-connect/tests/gold_data/expression/interval.json
```diff
@@ -10,10 +10,25 @@
...
+                  "metadata": [
+                    [
+                      "ARROW:extension:name",
+                      "sail.spark.yearMonth.interval"
+                    ],
+                    [
+                      "ARROW:extension:metadata",
+                      "{\"startField\":0,\"endField\":1}"
+                    ]
+                  ]
```
linhr: We probably want to extend the spec to store the "logical" information of interval field types for literals, instead of having the "physical" Arrow metadata directly, since it's the responsibility of the resolver to map spec to Arrow representations.

crates/sail-plan/src/function/scalar/array.rs
```diff
@@ -21,8 +21,28 @@ use sail_function::scalar::misc::raise_error::RaiseError;
...
+}
+
+fn array_repeat_with_nullable_element(
+    element: expr::Expr,
+    count: expr::Expr,
+    output_type: DataType,
+    schema: &datafusion_common::DFSchemaRef,
+) -> PlanResult<expr::Expr> {
+    let element_type = make_nullable_array_type(&element.get_type(schema.as_ref())?);
+    let nullable_element = when(lit(true), cast(element, element_type)).end()?;
```
linhr, line 42: I don't quite understand `when(lit(true), ...).end()?`. If it's needed for some reason, could we add a comment?

crates/sail-function/src/scalar/collection/spark_concat.rs
```diff
@@ -325,3 +318,24 @@ fn cast_list_columnar_values(
...
+
+fn merge_list_types(left: &DataType, right: &DataType) -> Option<DataType> {
+    match (left, right) {
+        (DataType::List(left), DataType::List(right)) => {
+            let data_type =
+                merge_list_types(left.data_type(), right.data_type()).or_else(|| {
+                    left.data_type()
+                        .equals_datatype(right.data_type())
+                        .then(|| left.data_type().clone())
+                })?;
```
linhr, line 331: I don't quite understand the `or_else` part, could you add a comment to explain it if this is expected?

linhr (commented):
It's great to see the fixes for Spark parity improvements! Since this PR touches many diverse areas of the codebase, I'd suggest the following changes so that we can merge the PR faster.

1. Let's revert all changes to the `sail-sql-analyzer` crate. Modifying metadata for intervals and generated alias seems "semantic analysis" to me, so I'm not sure if they belong to this crate. We need a design discussion for this.
2. Let's revert Arrow extension types for intervals. Sorry for the churn, but I realized this is a more complex change that requires a design discussion and careful review, so we need to have a dedicated PR for it.
3. Let's revert changes for duplicated field name handling (`crates/sail-spark-connect/src/{executor,schema}.rs`). As I mentioned in the comment, I'm not sure if this is the best approach, and the code seems verbose, so we need to evaluate it in a separate PR.

(I have a few comments related to the changes to be reverted, and the comments are meant for future consideration.)

python/pysail/tests/spark/test_variant.py
lonless9: Done!

crates/sail-spark-connect/src/schema.rs
lonless9: Yes I think it is.

linhr (approved):
This looks great!

PR #1954, opened 2026-05-21
feat: safe guards for join reorder optimizer

lonless9:


crates/sail-physical-optimizer/src/join_reorder/mod.rs
```diff
@@ -24,12 +28,106 @@ mod graph;
 mod join_set;
 mod reconstructor;
 
-#[derive(Default)]
-pub struct JoinReorder {}
+#[derive(Debug, Clone)]
+pub struct JoinReorderOptions {
+    pub max_relations: usize,
+    pub emit_threshold: usize,
+    pub enable_non_inner: bool,
```
linhr: I assume join reorder for outer joins would be non-trivial, so how does the current algorithm behave if `enable_non_inner` is set to `true`?

linhr (approved):
Great work!

crates/sail-physical-optimizer/src/join_reorder/mod.rs
lonless9: This option is currently not exposed as a user-configurable one. After entering `visit_hash_join`, non-inner will be processed into an edge, but it will still act as a barrier, with its semantics and direction unchanged. The benefit is that DP can freely rearrange the inner join relationship sets on both sides of non-inner. 
This is a hook for follow-up work handover, but I confirm that enabling this option now is wrong because there is a lack of legality verification in the greedy fallback.

linhr: From your comment and the code it seems that "DP can freely rearrange the inner join relationship sets on both sides of non-inner" even if this option is `false`. So I wonder if we should remove this option for now to avoid misuse?

lonless9: Make sense to me! Let me do the adjustment.

PR #1957, opened 2026-05-22
feat: catalog managed Delta Lake table

lonless9:
#171


crates/sail-catalog-unity/src/provider.rs
```diff
@@ -735,21 +792,32 @@ impl CatalogProvider for UnityCatalogProvider {
...
-            }
-            Err(e) => Err(CatalogError::External(format!("Failed to drop table: {e}"))),
+        let url = format!(
+            "{}/tables/{}",
+            client.baseurl,
+            progenitor_client::encode_path(&full_name)
+        );
+        let response = client
+            .client
+            .delete(url)
```
linhr: Is there a reason why we make HTTP requests directly rather than calling `client.delete_table()`? Same question for `commit_table`.

crates/sail-common-datafusion/src/datasource.rs
```diff
@@ -32,6 +32,12 @@ pub const MERGE_ROW_INDEX_COLUMN: &str = "__sail_file_row_index";
 /// Value is one of the [`RowLevelOperationType`] integer constants.
 pub const OPERATION_COLUMN: &str = "__sail_operation_type";
 
+/// Internal write option carrying the fully qualified catalog table name as a JSON string.
+///
+/// Table formats use this to coordinate storage commits with catalog commits. It is injected by
+/// the planner and must not be persisted as a user table property.
+pub const CATALOG_TABLE_OPTION: &str = "__sail.catalog.table";
```
linhr: Just to confirm, this is used within Sail and would never be written to external catalogs (meaning that we can change it freely anytime)?

python/pysail/tests/spark/catalog_integration/unity/conftest.py
```diff
@@ -134,7 +203,274 @@ def unity_spark(
...
+def _location_to_path(location: str) -> Path:
+    parsed = urlparse(location)
+    if parsed.scheme == "file":
+        return Path(parsed.path)
+    if parsed.scheme:
+        pytest.fail(f"expected a local table location, got: {location}")
+    return Path(location)
+
+
+def _write_seed_delta_log(location: Path, table_id: str) -> None:
```
linhr, line 169: Would Unity Catalog be able to write initial Delta log for managed tables? It's interesting to see that we have to do this manually here.

(This is just a question.)

python/pysail/tests/spark/catalog_integration/unity/conftest.py
```diff
@@ -27,6 +34,33 @@
     from pyspark.sql import SparkSession
 
 DEFAULT_CATALOG = "sail_test_catalog"
+UNITY_CATALOG_IMAGE = os.environ.get(
+    "PYSAIL_UNITY_CATALOG_IMAGE",
+    "unitycatalog/unitycatalog:v0.4.0",
+)
+
+
+def _exec_postgres_sql(
```
linhr: Why do we need to modify the database directly (e.g., creating `uc_delta_commits`, deleting rows for cleanup)? Would Unity Catalog be able to handle table management all via the REST API?

(This is just a question.)

linhr (approved):
Great work!

crates/sail-catalog-unity/src/provider.rs
lonless9: Done!

crates/sail-common-datafusion/src/datasource.rs
lonless9: Currently moved it into `SourceInfo/SinkInfo`.

python/pysail/tests/spark/catalog_integration/unity/conftest.py
lonless9: This is for a testing case. Unity Catalog doesn't write the initial delta log.

python/pysail/tests/spark/catalog_integration/unity/conftest.py
lonless9: This was a wrong usage. We now use the embedded backend.

linhr (approved):

PR #1964, opened 2026-05-23
fix: parse Spark variant paths with Chumsky

CoderHariswar:
## Summary

Implemented a Chumsky-based parser for Spark variant paths used by `variant_get` and `try_variant_get`.

The parser validates Spark path syntax into an internal AST and converts that AST into the path format expected by `parquet-variant`. This makes Sail reject Spark-invalid paths like `$a[0]` while preserving valid Spark paths like `$.a[0]` and `$[0]`.

## Tests

- `cargo clippy -p sail-function --all-targets --all-features -- -D warnings`
- `cargo test -p sail-function spark_variant_get`

### DIFF
--- a/crates/sail-function/src/scalar/variant/spark_variant_get.rs
+++ b/crates/sail-function/src/scalar/variant/spark_variant_get.rs
@@ -117,8 +117,8 @@ fn invoke_variant_get(args: ScalarFunctionArgs, name: &str, safe: bool) -> Resul
         None
     };
 
-    // Parse the Spark path: strip leading "$." or "$"
-    // Validate: "$." alone is invalid (trailing dot with no field name)
+    // Parse and validate the Spark path expression, then convert it to
+    // parquet-variant path syntax for VariantPath construction.
     let clean_path = spark_path_to_variant_path(&path_str, name)?;
     let variant_path = if clean_path.is_empty() {
         VariantPath::default()
--- a/crates/sail-function/src/scalar/variant/spark_variant_get.rs
+++ b/crates/sail-function/src/scalar/variant/spark_variant_get.rs
@@ -9,7 +9,7 @@ use datafusion_common::{arrow_datafusion_err, exec_datafusion_err, Result, Scala
 use datafusion_expr::{
     ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
 };
-use parquet_variant::VariantPath;
+use parquet_variant::{VariantPath, VariantPathElement};
 use parquet_variant_compute::{variant_get, GetOptions, VariantType};
 
 use crate::error::{generic_exec_err, invalid_arg_count_exec_err, unsupported_data_type_exec_err};
@@ -117,16 +117,11 @@ fn invoke_variant_get(args: ScalarFunctionArgs, name: &str, safe: bool) -> Resul
         None
     };
 
-    // Parse and validate the Spark path expression, then convert it to
-    // parquet-variant path syntax for VariantPath construction.
-    let clean_path = spark_path_to_variant_path(&path_str, name)?;
-    let variant_path = if clean_path.is_empty() {
-        VariantPath::default()
-    } else {
-        VariantPath::try_from(clean_path.as_str())?
-    };
+    // Parse and validate the Spark path expression into a parquet-variant path.
+    let variant_path = spark_path_to_variant_path(&path_str, name)?;
 
     // Resolve the target type.
+
     // For Decimal/Timestamp, parquet-variant doesn't support direct extraction,
     // so we extract as an intermediate type and then cast.
     // Decimal: extract as Float64 then cast (loses precision beyond ~15 digits;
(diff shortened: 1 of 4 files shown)


CoderHariswar:
ghfailedlakesail1

image


@linhr The current Build / Setup failure appears unrelated to this PR. It fails in `dorny/paths-filter@v3` while fetching changed files for  the GitHub API with `Bad credentials`, before any build/test steps run.

Local checks passed:
- `cargo clippy -p sail-function --all-targets --all-features -- -D warnings`
- `cargo test -p sail-function spark_variant_get`

linhr:
> The current Build / Setup failure appears unrelated to this PR. It fails in `dorny/paths-filter@v3` while fetching changed files for the GitHub API with `Bad credentials`, before any build/test steps run.

This seems a transient GitHub issue.

crates/sail-function/src/scalar/variant/spark_variant_get.rs
```diff
@@ -424,6 +409,115 @@ fn spark_type_to_arrow(type_str: &str) -> Result<DataType> {
...
+                    if !out.is_empty() {
+                        out.push('.');
+                    }
+                    out.push_str(name);
+                }
+                SparkVariantPathSegment::Index(index) => {
+                    out.push('[');
+                    out.push_str(&index.to_string());
+                    out.push(']');
+                }
```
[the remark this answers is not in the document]
linhr: After parsing, I feel we should directly construct `parquet_variant::VariantPath` from the AST. In this way we don't need to call `VariantPath::try_from` (which is another round of parsing), and we won't need to worry about special characters in the field name.

crates/sail-function/src/scalar/variant/spark_variant_get.rs
```diff
@@ -626,4 +720,108 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
```
linhr, line 544: Let's remove unit tests for `SparkVariantGet` and use Python BDD tests instead.

We can have unit tests for the variant path parser though.

linhr (commented):
Thanks for having this!

I skimmed through the code and here are a few initial high-level comments.

crates/sail-function/src/scalar/variant/spark_variant_get.rs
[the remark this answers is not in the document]
linhr: Actually we can just return `parquet_variant::VariantPath` directly in the parser without defining our own AST `SparkVariantPath` and `SparkVariantPathSegment`.

crates/sail-function/src/scalar/variant/spark_variant_get.rs
```diff
@@ -424,6 +409,115 @@ fn spark_type_to_arrow(type_str: &str) -> Result<DataType> {
...
+        .ignore_then(none_of('\'').repeated().collect::<String>())
+        .then_ignore(just('\''));
+
+    let double_quoted_field_name = just('"')
+        .ignore_then(none_of('"').repeated().collect::<String>())
+        .then_ignore(just('"'));
+
+    let backtick_quoted_field_name = just('`')
+        .ignore_then(none_of('`').repeated().collect::<String>())
+        .then_ignore(just('`'));
```
linhr: How does Spark handle escape characters in quoted field names?

CoderHariswar:
@linhr I refactored the Spark path parser based on your approach to construct `parquet_variant::VariantPath` directly using `VariantPathElement`, so we no longer convert the parsed path into an intermediate string and reparse it with `VariantPath::try_from`.

I also removed the Rust UDF-level `SparkVariantGet` tests and kept the Rust coverage focused on the path parser.

crates/sail-function/src/scalar/variant/spark_variant_get.rs
CoderHariswar: > Let's remove unit tests for `SparkVariantGet` and use Python BDD tests instead.
> 
> We can have unit tests for the variant path parser though.

For the BDD test approach  can we add cases in `variant.feature` for paths like `$.a[0]`, `$.a[0].b`, and quoted fields like `$["a.b"]` / `$["a[0]"]` ?

linhr: Yeah feel free to add test cases in `variant.feature`!

CoderHariswar:
@linhr Added BDD coverage in `variant.feature` for Spark path compatibility, including `$.a[0]`, `$.a[0].b`, and quoted field names like `$["a.b"]` / `$["a[0]"]`.

Validated with:
- `hatch run test.spark-4.1.1:pytest python/pysail/tests/spark/function/test_features.py -k "variant"`

crates/sail-function/src/scalar/variant/spark_variant_get.rs
CoderHariswar: The quoted field names are supported, but escaped quotes/backticks inside them are not handled properly yet. I forgot to implement that logic.

I have added them, so paths like `$["a\"b"]`, `$['a\'b']`, and $.`a\`b` keep the escaped character as part of the field name instead of closing the quoted section early. Parser tests have been added as well

linhr (approved):
This looks great! Thanks for the contribution!

PR #1969, opened 2026-05-24
feat: add DataFrame checkpoint and local checkpoint support

lonless9:
Moved to #2270


crates/sail-cache/src/cached_relation.rs
```diff
@@ -0,0 +1,1757 @@
...
+            }
+            CachedRelationData::Pending(_) => Err(internal_datafusion_err!(
+                "cached relation materialization did not complete"
+            )),
+        }
+    }
+
+    fn is_exclusively_owned(&self) -> bool {
+        Arc::strong_count(&self.data) == 1
+    }
```
linhr, line 409: A method like this doesn't seem right. The usage is likely vulnerable to race conditions.

crates/sail-physical-plan/src/checkpoint.rs
```diff
@@ -0,0 +1,377 @@
...
+            // FIXME: Stream Arrow IPC to object storage instead of buffering an entire partition.
+            let mut batches = vec![];
+            while let Some(batch) = input.next().await {
+                batches.push(batch?);
+            }
+            let bytes = write_record_batches_file(&batches, input_schema.as_ref())?;
+            store
+                .put(&location, PutPayload::from(bytes))
+                .await
+                .map_err(|error| DataFusionError::ObjectStore(Box::new(error)))?;
```
linhr, line 291: This materializes everything in memory. We should stream the data and use multipart upload instead.

linhr (commented):
I feel we need to redo the implementation once we figure out the design.

I thought we were going to do checkpoint based on object storage only. We do not want to follow Spark's semantics around "local" checkpoint or serialization level. It seems the current implementation sends all data to the driver and stores it in the driver's memory or disk, which is not scalable so it does not have high practical value. (Of course, we do not want to store the data in the worker either, since the worker is stateless for the simplicity of auto-scaling.)

When we work with Codex to do the design, the priority is to achieving a simple but effective solution that meet the user's need in cloud environments. If Codex instead attempts to fully match Spark's behavior and favors feature parity in tests, the overall architecture may be complex and it's hard to reason about the performance characteristics.

I wonder if we should only implement remote eager checkpoint in the first iteration, with simple cleanup logic at the end of the session. I feel the cached relation lease is hard to reason about. Lazy checkpoint is a bit tricky as well, since we allow concurrent query planning and execution within the session, so which query is responsible for materialization requires careful design.

PR #1976, opened 2026-05-24
feat: add VACUUM command for Delta Lake tables

ma-ts:
Adds support for the `VACUUM` command on Delta Lake tables. This was built with AI assistance.

The SQL surface follows Spark's Delta Lake:

```sql
VACUUM my_table
VACUUM my_table RETAIN 240 HOURS
VACUUM my_table DRY RUN
VACUUM 's3://bucket/path' RETAIN 48 HOURS DRY RUN
```

`DRY RUN` returns one row per candidate file without touching anything. A plain `VACUUM` deletes unreferenced files older than the retention window and returns the table URI.

The implementation threads through the full stack — SQL parser (new `VACUUM`, `RETAIN`, `DRY`, `RUN` keywords), SQL analyzer, plan resolver, a new `FileVacuumNode` logical extension, a `create_vacuum_writer` method on the `TableFormat` trait, and a `DeltaVacuumExec` physical node that does the actual listing and deletion.

A few things worth calling out:

- Paths are compared in object-store-relative space on both sides, avoiding any URL round-trip or coordinate mismatch that could cause active files to be treated as candidates.
- UUID-based deletion vector files referenced by active `Add` actions are explicitly preserved.
- Tombstones (removed files) within the retention window are kept so concurrent time-travel reads still work.
- Tables with the `vacuumProtocolCheck` writer feature are rejected with a clear error until that feature is properly supported.
- The 168-hour minimum retention guard is enforced by default, overridable via the table property `delta.vacuum.retentionDurationCheck.enabled = false` (note: Spark surfaces this as a session config instead — that divergence is documented in the code).
- `EXPLAIN ANALYZE` exposes `num_files_to_delete`, `size_of_data_to_delete`, and `num_deleted_files` metrics.


linhr:
Thanks for having this, but unfortunately we are unable to accept the PR in this case. The `TableFormat` trait is under heavy redesign as we separate data source operations and table operations. So it would just be easier to implement table maintenance commands once we have a good foundation for general logical planning and physical planning for DDL and DML statements.

Also, features like this would require end-to-end testing written in Python or `.feature` files.

If you'd like to see `VACUUM` supported for your data processing workflows, please feel free to open an issue for a feature request. You are also welcome to join our Slack community. We would be happy to learn more about your use cases in general.

If you are interested in contributing to Sail, you can look for issues labeled as "good first issue".

Thanks for your understanding!

PR #1983, opened 2026-05-25
fix: render implicit idx=1 in regexp_extract_all display name

pomykalakyle:
## Summary

PySpark's \`regexp_extract_all(str, regexp, idx=None)\` defaults \`idx\` to \`1\` and renders the default into the column header. Sail currently omits it for 2-arg calls, producing \`regexp_extract_all(str, pattern)\` where Spark produces \`regexp_extract_all(str, pattern, 1)\`. The execution values are correct - only the display name diverges.

Adds \`regexp_extract_all\` to the existing \`append_start_pos_if_arglen_eq(2, 1, ...)\` arm in \`SparkPlanFormatter\`, matching how \`position\`, \`locate\`, and \`regexp_instr\` already synthesize their implicit defaults. 3-arg calls were already correct via the generic fallback and remain so.

Fixes the \`pyspark.sql.functions.regexp_extract_all\` doctest.


crates/sail-plan/src/formatter.rs
```diff
@@ -621,6 +621,7 @@ impl PlanFormatter for SparkPlanFormatter {
                 Ok(format!("{}({arguments})", name.to_lowercase()))
             }
             "position" | "locate" => Ok(append_start_pos_if_arglen_eq(2, 1, name, arguments)),
+            "regexp_extract_all" => Ok(append_start_pos_if_arglen_eq(2, 1, name, arguments)),
```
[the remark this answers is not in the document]
pomykalakyle: Added in e0e9716c - new 2-arg gold case in crates/sail-spark-connect/tests/gold_data/function/string.json that exercises the new arm directly (synthesized `, 1` in the column name).

crates/sail-spark-connect/tests/gold_data/function/string.json
```diff
@@ -2514,6 +2514,32 @@
         "success": "ok"
       }
     },
+    {
+      "input": {
+        "query": "SELECT regexp_extract_all('100-200, 300-400', '(\\\\d+)-(\\\\d+)');",
```
linhr: We should not add test cases to gold data JSON files since they will be overridden the next time the file is generated from scratch. We can add the test case to `.feature` file.

linhr (changes requested):
Nice fix! We can merge this once the test is done in the correct way.

crates/sail-spark-connect/tests/gold_data/function/string.json
pomykalakyle: Good catch, sorry about that. Moved the case to python/pysail/tests/spark/function/features/regexp_extract_all.feature in 4934fce1 (new scenario without an alias so the column header surfaces the synthesized `, 1`) and dropped the gold_data entry.

python/pysail/tests/spark/function/features/regexp_extract_all.feature
```diff
@@ -40,6 +40,15 @@ Feature: regexp_extract_all() extracts all regex capture group matches from stri
...
 
+    Scenario: regexp_extract_all without idx renders the synthesized idx=1 in the column name
+      When query
+      """
+      SELECT regexp_extract_all('1a 2b 14m', r'(\d+)([a-z]+)')
+      """
+      Then query result
+      | regexp_extract_all(1a 2b 14m, (\d+)([a-z]+), 1) |
+      | [1, 2, 14]                                      |
+
```
linhr, line 51: Due to limitations in feature file parsing, values with escape characters do not work well in the table. So we may have to avoid `\d` in the regex. We can use `[0-9]` here instead.

Reference: #1907

linhr: (This is a limitation in the test setup but does not affect the actual functionalities of Sail.)

linhr: I attempted the fix and it's working now.

linhr (approved):

PR #1993, opened 2026-05-27
feat: support check constraints for Delta Lake

lonless9:
#171


crates/sail-common-datafusion/src/column_features.rs
```diff
@@ -25,18 +25,22 @@ use datafusion::arrow::datatypes::Field;
...
+    /// Marker used in the planning pipeline to preserve a Delta NOT NULL
+    /// constraint when DataFusion expression rewrites make the output nullable.
+    NotNullConstraint,
 }
 
 impl ColumnFeatureKey {
     pub const fn as_str(self) -> &'static str {
         match self {
             Self::GenerationExpression => "delta.generationExpression",
+            Self::NotNullConstraint => "sail.column.notNull",
```
linhr, line 54: How does Spark store this column feature?

crates/sail-common/src/spec/plan.rs
```diff
@@ -1326,6 +1326,11 @@ pub enum AlterTableOperation {
         name: ObjectName,
         data_type: DataType,
     },
+    AddCheckConstraint {
+        name: Identifier,
+        expression: Expr,
+        expression_source: String,
```
linhr: We can use `ExprWithSource` for consistency. Even thought the `source` is required here, we can detect `None` source when resolving the spec.

crates/sail-plan/src/resolver/command/delta.rs
```diff
@@ -0,0 +1,956 @@
...
+        .metadata()
+        .get(DELTA_INVARIANTS_METADATA_KEY)
+        .and_then(|raw| {
+            serde_json::from_str::<serde_json::Value>(raw)
+                .ok()
+                .and_then(|value| expression_from_value(&value))
+        })
+}
+
+fn parse_delta_generation_expr(gen_expr_str: &str) -> PlanResult<spec::Expr> {
```
linhr: This function can be used to simplify SQL parsing code in two other places in this file.

linhr (approved):
Great work!

crates/sail-common-datafusion/src/column_features.rs
lonless9: > Delta/Spark stores this as `StructField(nullable = false)`

We are currently using it as a marker here, but it is persisted as Delta schema.

crates/sail-catalog-hms/thrift/hive_metastore.thrift
```diff
@@ -1240,14 +1240,14 @@ service ThriftHiveMetastore
   // number of partition columns - the unspecified values are considered the same
   // as "".
   list<Partition> get_partitions_ps(1:string db_name 2:string tbl_name
-  	3:list<string> part_vals, 4:i16 max_parts=-1)
+    3:list<string> part_vals, 4:i16 max_parts=-1)
```
linhr, line 1243: Since we've touched this file, could you just change all tab with space? I found there are remaining occurrences around `get_table_objects_by_name_req`.

linhr (approved):

PR #2004, opened 2026-05-28
feat: support default column values for Delta Lake

lonless9:
#171


crates/sail-sql-parser/src/ast/expression.rs
```diff
@@ -234,6 +234,7 @@ pub enum AtomExpr {
     DateLiteral(Date, StringLiteral),
     TimeLiteral(Time, StringLiteral),
     Null(Null),
+    Default(Default),
```
linhr: I'm not sure if we need to have it here. Currently `DEFAULT` is parsed as an identifier and the plan resolver can still handle it correctly in the context of insert options. But if we consider `DEFAULT` as an expression, would it cause issues with the following query?

```python
spark.sql("SELECT DEFAULT FROM VALUES (1) AS t(DEFAULT)").show()
```

crates/sail-common-datafusion/src/column_features.rs
```diff
@@ -31,12 +36,15 @@ use datafusion::arrow::datatypes::Field;
...
     GenerationExpression,
+    /// SQL expression used when a write explicitly or implicitly requests the column default.
+    CurrentDefault,
 }
 
 impl ColumnFeatureKey {
     pub const fn as_str(self) -> &'static str {
         match self {
             Self::GenerationExpression => "delta.generationExpression",
+            Self::CurrentDefault => "CURRENT_DEFAULT",
```
linhr, line 63: How does Spark track the default value in metadata?

crates/sail-delta-lake/src/table_format.rs
```diff
@@ -385,7 +412,15 @@ impl TableFormat for DeltaTableFormat {
             },
         )
         .await
-        .map_err(|e| DataFusionError::External(Box::new(e)))?;
+        {
+            Ok(table) => table,
+            Err(DeltaTableError::InvalidTableLocation(message))
+                if message.contains("No commit files found in _delta_log") =>
```
linhr, line 464: We probably need a `FIXME` comment here since this is not a robust way to classify errors.

linhr (commented):
This looks good overall!

crates/sail-sql-parser/src/ast/expression.rs
lonless9: Yes we need to resolve it as column named Default. Added a test for this and it passes now.

crates/sail-common-datafusion/src/column_features.rs
lonless9: Added the comment at the line above.

crates/sail-delta-lake/src/table_format.rs
lonless9: Done!

crates/sail-sql-analyzer/src/expression.rs
```diff
@@ -509,6 +509,163 @@ pub fn from_ast_expression(expr: Expr) -> SqlResult<spec::Expr> {
     }
 }
 
+pub(crate) fn expr_with_default_column_values(expr: spec::Expr) -> spec::Expr {
```
linhr, line 513: This is "semantic analysis" logic that should not be present in SQL analyzer. We should keep `default` as an unresolved attribute and the plan resolver is going to decide how it should be interpreted given the context (the actual logical plan node and the schema). In this way we won't need to duplicate the logic if the Spark Connect protocol (or any other user-facing interface other than SQL) supports `default` in the future.

We've seen such incorrect design many times. Maybe we should document it as a rule in `AGENTS.md` so that AI won't make the same mistake in the future?

No need to make changes in this PR, but let's add a FIXME for this.

linhr (approved):
This is great!

PR #2007, opened 2026-05-29
feat: support identity columns for Delta Lake

lonless9:
#171


crates/sail-common-datafusion/src/catalog/status.rs
```diff
@@ -230,6 +230,7 @@ pub struct TableColumnStatus {
     pub comment: Option<String>,
     pub default: Option<String>,
     pub generated_always_as: Option<String>,
+    pub identity: Option<sail_common::spec::TableColumnIdentity>,
```
linhr: The status is considered "resolved" so ideally we should not depend on the spec. We can define `CatalogTableColumnIdentity` in `crates/sail-common-datafusion/src/catalog/mod.rs` and have "resolved" start and step values of type `i64`, while `spec::TableColumnIdentity` has `Option<i64>`.

crates/sail-catalog/src/provider/options.rs
```diff
@@ -47,6 +48,7 @@ pub struct CreateTableColumnOptions {
     pub comment: Option<String>,
     pub default: Option<String>,
     pub generated_always_as: Option<String>,
+    pub identity: Option<TableColumnIdentity>,
```
linhr: According to the other comment I have, we should not rely on spec here either. We can use `CatalogTableColumnIdentity` instead.

crates/sail-delta-lake/src/physical_plan/planner/context.rs
```diff
@@ -47,6 +48,7 @@ pub struct DeltaPlannerConfig {
     /// Delta commit (new tables) even when the physical planner strips the arrow
     /// field metadata set at logical-plan construction time.
     pub generation_expressions: HashMap<String, String>,
+    pub identity_columns: HashMap<String, TableColumnIdentity>,
```
linhr: Same here, we can use `CatalogTableColumnIdentity`.

crates/sail-delta-lake/src/physical_plan/writer_options.rs
```diff
@@ -28,6 +29,8 @@ pub struct DeltaWriterExecOptions {
     pub replace_where: Option<String>,
     #[serde(default)]
     pub generation_expressions: HashMap<String, String>,
+    #[serde(default)]
+    pub identity_columns: HashMap<String, TableColumnIdentity>,
```
linhr: Same here, use `CatalogTableColumnIdentity`.

crates/sail-common/src/spec/plan.rs
```diff
@@ -926,6 +926,17 @@ pub struct TableColumnDefinition {
...
     pub generated_always_as: Option<String>,
+    /// Delta identity column metadata.
+    pub identity: Option<TableColumnIdentity>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TableColumnIdentity {
+    pub start: i64,
+    pub step: i64,
```
linhr: I feel we should have `Option<i64>` here rather than "resolving" the default value in SQL analyzer.

crates/sail-sql-analyzer/src/statement.rs
```diff
@@ -1446,6 +1449,7 @@ struct ColumnDefinitionOptions {
     not_null: bool,
     default: Option<Expr>,
     generated_always_as: Option<Expr>,
+    identity: Option<spec::TableColumnIdentity>,
```
linhr: For consistency we should have AST here, and `from_ast_identity_column` would be called afterwards.

crates/sail-sql-analyzer/src/statement.rs
```diff
@@ -1472,17 +1476,113 @@ impl TryFrom<Vec<ColumnDefinitionOption>> for ColumnDefinitionOptions {
...
+                }
+            }
+        }
+    }
+    let step = step.unwrap_or(1);
+    if step == 0 {
+        return Err(SqlError::invalid(
+            "INCREMENT BY value for identity column cannot be 0",
+        ));
+    }
```
linhr: This validation should be done when resolving the spec.

crates/sail-common/src/spec/plan.rs
```diff
@@ -926,6 +926,17 @@ pub struct TableColumnDefinition {
...
+    pub allow_explicit_insert: bool,
+    pub high_water_mark: Option<i64>,
```
linhr: Is `high_water_mark` something that can be "specified"? Should we only have it in the "resolved" struct?

linhr (approved):
Great work!

My comments are minor design suggestions that do not affect correctness.

linhr (approved):

Issue #1741, opened 2026-04-22
Support for LIKE/ILIKE with ALL/ANY/SOME


shehabgamin:
Follow up after https://github.com/lakehq/sail/pull/1718

https://spark.apache.org/docs/4.1.1/sql-ref-syntax-qry-select-like.html#examples

radmirnovii:
made fix for NOT semantics in https://github.com/lakehq/sail/pull/2014

PR #2014, opened 2026-05-31
fix: support LIKE/ILIKE with ALL/ANY/SOME

radmirnovii:
fix LIKE/ILIKE ALL/ANY/SOME matching Spark's `LikeAll`/`LikeAny` semantics: `NOT` is pushed into each pattern

### DIFF
--- a/crates/sail-spark-connect/tests/gold_data/expression/like.json
+++ b/crates/sail-spark-connect/tests/gold_data/expression/like.json
@@ -6,29 +6,54 @@
         "success": {
           "unresolvedFunction": {
             "functionName": [
-              "ilike"
+              "and"
             ],
             "arguments": [
               {
-                "unresolvedAttribute": {
-                  "name": [
-                    "a"
-                  ],
-                  "planId": null,
-                  "isMetadataColumn": false
-                }
-              },
-              {
                 "unresolvedFunction": {
                   "functionName": [
-                    "all"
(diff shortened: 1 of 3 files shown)


linhr:
Thanks for your contribution. I feel this approach is not desirable for the following reasons:

1. It expand the `LIKE` expression to boolean expressions in the SQL analyzer, which may cause inefficient execution. For example, for `a not like some ('foo%', 'b%')`, the expression `a` is evaluated multiple times (if it's not a simple column reference). If you look at Spark's implementation, you will find an expression named `notlikeany` in the plan instead of the boolean rewrite.
2. Semantic analysis (boolean expression expansion) should not be present in the SQL analyzer. We should just translate the query to the spec, and the spec is responsible for semantic analysis.

As we can see in the current gold data JSON file, we will see `ANY` being translated as a function call `any()` in the spec. So we need to identify such function inside `like` and resolve it to something more efficient, such as a dedicated scalar UDFs so that we can evaluate multiple patterns at once. 

And I feel we don't have to push `NOT` into `LIKE` in the first iteration. This may be something to consider later for optimization.

As agentic coding makes it so easy to generate code nowadays, it's critical to align on the design first before code changes to streamline the review process. You can use #1741 for the discussion.

Thanks for your understanding, and I appreciate your interest in contributing to Sail!

PR #2017, opened 2026-06-01
feat: support grouping_id aggregate function

lonless9:
#309


python/pysail/tests/spark/delta/__snapshots__/features/operation_metrics.yaml
```diff
@@ -72,21 +72,6 @@ data:
     scanTimeMs: <time_ms>
     writeTimeMs: <time_ms>
 ---
-name: "test_dv_delete_reports_numdeletionvectorsadded_and_zero_numcopiedrows"
```
lonless9: I forgot to delete these so nothing to do with this PR.

linhr (approved):
Great work!

PR #2018, opened 2026-06-01
fix: propagate spark.sql.ansi.enabled to DataFusion session config (enables pmod ANSI errors)

davidlghellin:
Verified end-to-end: `pmod(x, 0)` now errors with REMAINDER_BY_ZERO under ANSI on (was silently NULL). Same plumbing covers any future `datafusion_spark::*` UDF that reads the flag.

### DIFF
--- a/crates/sail-plan/src/lib.rs
+++ b/crates/sail-plan/src/lib.rs
@@ -31,11 +31,39 @@ pub async fn execute_logical_plan(ctx: &SessionContext, plan: LogicalPlan) -> Re
     Ok(df)
 }
 
+/// Write the current ANSI flag into the DataFusion `SessionContext` so that
+/// upstream UDFs reading `ConfigOptions.execution.enable_ansi_mode` see the
+/// value the client requested via `SET spark.sql.ansi.enabled = …`. No-op if
+/// the flag is already at the target value (avoids needless write-lock churn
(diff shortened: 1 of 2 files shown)


linhr (commented):
I feel the patch means that `resolve_and_execute_plan` has a side-effect of modifying the session context, and concurrent plan resolver invocation would be hard to reason about.

In my opinion the DataFusion `execution.enable_ansi_mode` option is not a good way to allow downstream project to configure ANSI behavior for Spark functions. It actually doesn't work in our cluster settings since the worker's session context is different from the driver's one.

Let's define our own `ScalarUDFImpl` for pmod and accept `ansi_mode` in the constructor, similar to our other functions. Then we can delegate to DataFusion's `spark_mod` function within `invoke_with_args`.

davidlghellin:
Thanks @linhr, both points are well taken — especially the distributed one. 

Agreed on the approach: I'll close this PR and reimplement with a dedicated ScalarUDFImpl for pmod that takes ansi_mode in its constructor (same pattern as SparkAbs and friends), baking the flag into the plan at planning time so it serializes through the codec to workers, and delegating to datafusion_spark's spark_pmod/spark_mod inside invoke_with_args. Will open a fresh PR along those lines.

PR #2031, opened 2026-06-04
feat: enable ANSI mode for `negative`

davidlghellin:
## Summary

Makes Spark's unary minus (`-x` / `negative(x)`) honor `spark.sql.ansi.enabled`,
matching Spark's overflow semantics. Previously Sail delegated to DataFusion's
`Negative`, which does not implement Spark's ANSI behavior.

## Behavior

| Case | ANSI on | ANSI off |
|---|---|---|
| `-INT_MIN` / `-LONG_MIN` / `-SHORT_MIN` / `-BYTE_MIN` | `ARITHMETIC_OVERFLOW` | wraps to itself (two's complement) |
| ordinary numeric (`-5`, `-1.5`, floats) | negated value | negated value |
| `NaN` / `±Infinity` / `-0.0` | unchanged sign rules (`-NaN=NaN`, `-Inf=-Infinity`) | same |
| numeric string (`-'5'`) | `-5.0` (coerced to DOUBLE) | `-5.0` |
| non-numeric string (`-'abc'`) | `CAST_INVALID_INPUT` error | `NULL` |
| day-time interval | negated interval | negated interval |

## Implementation

- New `SparkNegative` UDF wrapping the upstream `datafusion_spark` negative,
  with `ansi_mode` captured at planning time and serialized via the codec so
  distributed workers see the value the client requested.
- Strings are coerced to `DOUBLE` before negating (cast honors ANSI mode);
  `Duration` keeps the existing `NegateDuration` path.
- Negated **numeric literals** are folded to a literal at plan time
  (`-3` → `lit(-3)`) so functions that require a constant argument (e.g.
  `ceil`/`floor` target scale, which validate at planning, before the optimizer
  would fold the UDF call) still see a literal. Overflow (`-INT_MIN`) can't fold
  and falls through to the runtime UDF.

## Tests

`negative.feature` covers the integral MIN overflow boundary (all four widths,
ANSI on/off), ordinary/float/decimal negation, special floats (NaN, Infinity),
day-time intervals, string coercion, and double negation. All scenarios
validated against Spark JVM (`SPARK_REMOTE="local"`).

## Known limitation

Negating the maximum `DECIMAL(38,0)` errors in Spark (a runtime value-boundary
quirk) where Sail succeeds; tracked as `@sail-bug`.


crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -105,13 +108,11 @@ fn spark_minus(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
-            Ok(ScalarUDF::from(NegateDuration::new()).call(vec![arg]))
-        } else {
-            Ok(Expr::Negative(Box::new(arg)))
-        }
+        Ok(spark_unary_negate(
+            arg,
+            function_context.plan_config.ansi_mode,
+            function_context.schema,
+        ))
     } else {
```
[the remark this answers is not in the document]
davidlghellin: Applied for floats — they never overflow and behave the same in both ANSI modes, so they now use native `Expr::Negative`; numeric literals were already folded. Integral ANSI-off has to stay on the UDF though: native `Expr::Negative` raises on `-INT_MIN` where Spark wraps. Added a multi-row float-column test to cover the native path.

python/pysail/tests/spark/function/features/negative.feature
```diff
@@ -0,0 +1,250 @@
...
+        | NaN    |
+
+    Scenario: negate Infinity returns negative Infinity
+      When query
+        """
+        SELECT -CAST('Infinity' AS DOUBLE) AS result
+        """
+      Then query result
+        | result    |
+        | -Infinity |
```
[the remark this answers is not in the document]
davidlghellin: The harness compares Spark's `_show_string` output, and the same rendering runs for both Spark JVM and Sail (Sail implements Spark Connect), so the textual form is deterministic and identical across engines — these are validated green on both. The textual scenarios intentionally check rendering parity. Per the suggestion, I also added predicate-based scenarios (`isnan(-NaN)`, `-Infinity = ...`) for the value-level checks.

crates/sail-function/src/scalar/math/spark_negative.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let mut config = (*args.config_options).clone();
+        config.execution.enable_ansi_mode = self.ansi_mode;
+        let args = ScalarFunctionArgs {
+            config_options: Arc::new(config),
+            ..args
+        };
+        self.inner.invoke_with_args(args)
```
linhr: I feel cloning the config on every batch would degrade the performance. I wonder if we should reach out to the upstream to make `spark_negative` a `pub` function we can use.

linhr (commented):
I haven't reviewed this PR. Just sharing a thought for discussion.

crates/sail-function/src/scalar/math/spark_negative.rs
davidlghellin: Agreed, the per-batch `config_options` clone isn't great. The reason it's there is that `datafusion_spark's` free spark_negative helper is private, so the only way to drive ANSI mode is to wrap the UDF and mutate `enable_ansi_mode` in the config on every `invoke`.

We already have the clean version for `pmod`: `datafusion_spark` exposes `spark_pmod` as a `pub fn`, so `SparkPmod` just calls `spark_pmod(&args.args, self.ansi_mode)` with no config work. The ideal fix is to ask upstream to make `spark_negative pub` the same way — then `SparkNegative` drops the clone and becomes a one-liner that bakes `ansi_mode` at plan time, consistent with `pmod`.

I'll also check the other ANSI-aware functions in case any rely on the same per-batch config workaround.

linhr: Or maybe we store an `Arc<ConfigOptions>` within the UDF and assign it to `ScalarFunctionArgs` on each batch. Then we don't need upstream changes.

This could be a general pattern to make it easier to reason about UDFs that depend on session configuration (which are not shared in the driver and workers). By storing a "snapshot" of the configuration when creating the UDF, we can be sure that the UDF behavior is determined during query planning, so any later modification to the session configuration would not affect the UDF behavior.

@davidlghellin

davidlghellin: Thanks @linhr — much cleaner. An `Arc<ConfigOptions>` snapshot is just a refcount bump per batch (no deep clone) and freezes the config at planning time, with no upstream change needed. I'll rework `SparkNegative` to use it. Really appreciate the pointer — still learning where these `Arc` snapshot patterns fit. 🙏

linhr (approved):
Great! Thanks!

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -579,6 +580,70 @@ fn spark_pmod(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
+        }
+        // A negated numeric literal folds to a literal so constant-arg functions
+        // (e.g. `ceil`/`floor` target scale) still see a constant; overflow
+        // (`-INT_MIN`) can't fold and falls through to the runtime UDF.
+        _ => match negate_literal(&arg) {
+            Some(folded) => folded,
+            None => ScalarUDF::from(SparkNegative::new(ansi_mode)).call(vec![arg]),
+        },
+    }
+}
```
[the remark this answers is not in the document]
linhr: This should be fine.

Issue #231, opened 2024-09-29
Spark SQL Functions Coverage and Parity - XML


shehabgamin:
- [ ] from_xml
- [ ] schema_of_xml
- [ ] to_xml
- [x] xpath
- [x] xpath_boolean
- [x] xpath_double
- [x] xpath_float
- [x] xpath_int
- [x] xpath_long
- [x] xpath_number
- [x] xpath_short
- [x] xpath_string

anarefolio:
Hi @shehabgamin  Can i work on to_xml?

shehabgamin:
@anarefolio Yes! 😃

PR #2032, opened 2026-06-05
feat: implement to_xml Spark function

anarefolio:
## feat: implement `to_xml`

This PR introduces the implementation of the to_xml functionality, closely
replicating Spark's to_xml capabilities. Serializes a struct value to an XML string, with an optional second argument for configuration options.

Ref: [Spark docs](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_xml.html)

---
### What's supported

Options accepted via the map argument (keys are case-insensitive):

| Option | Default |
|---|---|
| `rowTag` | `ROW` |
| `declaration` | `version="1.0" encoding="UTF-8" standalone="yes"` |
| `attributePrefix` | `_` |
| `valueTag` | `_VALUE` |
| `arrayElementName` | `item` |
| `nullValue` | *(omit null fields by default)* |
| `timestampFormat` | `yyyy-MM-dd'T'HH:mm:ss.SSSXXX` |
| `timestampNTZFormat` | `yyyy-MM-dd'T'HH:mm:ss.SSS` |
| `dateFormat` | `yyyy-MM-dd` |

A few things worth noting:
- Null fields are silently omitted unless `nullValue` is set
- Null rows (the whole struct is NULL) return SQL NULL
- UTC timestamps render with `Z` not `+00:00`; non-UTC sessions render the local offset (e.g. `-04:00`)
- Fields prefixed with `_` become XML attributes on the root tag; prefix is stripped from the attribute name
- `_VALUE` fields are written as inline text content on the parent element, not as child elements
- Scalar arrays wrap each element as `<fieldName><item>val</item></fieldName>`; inner tag name is configurable via `arrayElementName`
- Pretty-printing is hardcoded to 4-space indent, same as Spark
- `VariantType` input is accepted but not serialized — no Spark behavior to replicate here

## MapType support

Map fields inside structs are now fully serialized, empirically verified against Spark 4.0.2:
- Map keys become child XML tag names: `map('k1', 1, 'k2', 2)` : `<m><k1>1</k1><k2>2</k2></m>`
- Null map values are silently omitted unless `nullValue` is set
- Empty map produces a self-closing tag: `<m/>`
- Map value is a struct: struct contents nested inside the key tag
- Map value is an array: key repeats as sibling tags, one per element (no `<item>` wrapper)
- Non-string keys or illegal XML name start characters (digit, `-`, `.`): runtime error matching Spark's `XMLStreamException`

## Files changed

- `to_xml.rs`: core implementation + unit tests
- `mod.rs`: module registration
- `xml.rs`: wires up the real UDF replacing `F::unknown("to_xml")`
- `codec.rs`: encode/decode for distributed execution
- `physical.proto`: proto definition
- `to_xml.feature`: integration test scenarios

Part of https://github.com/lakehq/sail/issues/231


crates/sail-execution/src/codec.rs
```diff
@@ -2551,6 +2556,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             || node_inner.is::<UrlDecode>()
             || node_inner.is::<UrlEncode>()
             || node_inner.is::<Xpath>()
+            || node_inner.is::<SparkToXml>()
```
linhr: This line should be removed.

crates/sail-function/src/scalar/xml/to_xml.rs
```diff
@@ -0,0 +1,1484 @@
...
+            } else {
+                Some(values.value(i).to_string())
+            };
+        }
+    }
+    None
+}
+
+#[cfg(test)]
+mod tests {
```
linhr, line 934: Could you remove all Rust unit tests that involves Arrow arrays? All such functionalities needs to use end-to-end tests in `.feature` files.

linhr (approved):
Looks good to me! Could you resolve the merge conflicts? Then we can merge this PR.

linhr (approved):
Thanks!

Issue #472, opened 2025-05-24
PySpark DataFrame Doctest: sameSemantics


shehabgamin:
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.sameSemantics`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.sameSemantics.html

linhr:
This is a tricky feature since we need to take care of attribute names, commutative expressions etc. The JVM Spark implementation scatters the logic in various places such as `QueryPlan.normalizeExpressions()` and the `CommutativeExpression` trait.

To get the canonical plan, Spark simply assign all expression IDs to 0 and erase attribute names. Although this is sufficient for semantic comparison, we cannot do so since otherwise DataFusion may complain about invalid `LogicalPlan`. So we have to build a "canonical" `LogicalPlan` properly while having consistent ways to assign attribute names, which means a ton of work. I guess this would become easier if we have our own logical representation before DataFusion's `LogicalPlan`. But such intermediate representation to assist query plan analysis is not being planned yet.

Also, as I play with the JVM Spark implementation I feel it's buggy.

```
>>> df1 = spark.sql("SELECT (a + b) + c FROM VALUES (1e308, 1e308, -1e308) AS t(a, b, c)")
>>> df2 = spark.sql("SELECT a + (b + c) FROM VALUES (1e308, 1e308, -1e308) AS t(a, b, c)")
>>> df1.show()
+-------------+
|((a + b) + c)|
+-------------+
|     Infinity|
+-------------+

>>> df2.show()
+-------------+
|(a + (b + c))|
+-------------+
|      1.0E308|
+-------------+

>>> df1.sameSemantics(df2)
True
```

Floating point arithmetics are not associative, but Spark consider these the same semantics although the results are different. While "false negative" is acceptable for semantics comparison, such case is "false positive" which is a more severe correctness issue.

linhr:
Given the analysis above, the `semanticHash` method is non-trivial to implement as well.

PR #2036, opened 2026-06-06
feat: implement sameSemantics

nis12ram:
issue: https://github.com/lakehq/sail/issues/472

### DIFF
--- a/crates/sail-function/src/scalar/hash/mod.rs
+++ b/crates/sail-function/src/scalar/hash/mod.rs
@@ -1,3 +1,5 @@
 pub mod spark_murmur3_hash;
 pub mod spark_xxhash64;
 pub(crate) mod utils;
+
+pub use utils::spark_compatible_murmur3_hash;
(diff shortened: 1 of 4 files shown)


nis12ram:
@linhr
Could you please suggest the best approach here? I’m a little confused about whether we should use the analyzed plan or the optimized plan as the input to canonicalize for the sameSemantics implementation.

I’d also appreciate your guidance on the scope of the canonicalization steps that should be covered to achieve semantic coverage.
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/QueryPlan.scala#L680

Sorry for the trouble

linhr:
This seems a very tricky feature and we haven't heard feature requests from the user. I'd imagine mapping logical plan to canonical representation based on its semantic is highly non-trivial. I'd suggest looking at other tasks if you'd like. Thanks for your interest in contributing to Sail!

nis12ram:
@linhr, got your point.

Just resolving my own confusion, and for anyone implementing it in the future, Spark uses the analyzed logical plan as input to the canonicalization step.

Closing this PR.

PR #2038, opened 2026-06-06
chore: prepare v0.6.4

linhr:


pyproject.toml
```diff
@@ -32,9 +32,7 @@ classifiers = [
 [project.optional-dependencies]
 test = [
     # The dependencies for testing the installed package.
-    # PySpark 4.1.0 does not work on Windows.
-    #   https://issues.apache.org/jira/browse/SPARK-54745
-    "pyspark-client>=4.0,<4.2,!=4.1.0",
+    "pyspark-client==4.1.1",
```
linhr, line 35: `test_udt.txt` fails in PySpark 4.1.2 so we pin the PySpark version for now. PySpark no longer allows unpickling UDT classes so we need to adjust the tests. It's good to pin the version anyway for test reproducibility.

See also: <https://issues.apache.org/jira/browse/SPARK-56463>

.github/workflows/release.yml
```diff
@@ -190,6 +190,7 @@ jobs:
           # This is to avoid unexpected PyPy wheels when building inside containers.
           # https://github.com/PyO3/maturin/issues/2852
           args: --release --out dist --interpreter python3.11
+          rustup-components: rustfmt
```
linhr, line 193: This is somehow needed for generating HMS client from Thrift within the codebase.

PR #2058, opened 2026-06-12
fix: column default metadata handling for Hive Metastore catalog writes

### LINKED ISSUE #2055: Bug: String/integer literals in write expressions misinterpreted as column identifiers
## Description

When executing write operations (both INSERT INTO ... VALUES and DataFrame.writeTo(...).createOrReplace()), Sail's plan resolver misinterprets string and integer literals as column identifiers, causing AnalysisException: attribute ObjectName([Identifier("...")]) is missing from the schema: cannot resolve attribute.

### Environment

 - Sail: built from main (commit 1e0dd69, 2026-06-11)
 - PySpark client: 4.1.1
 - Catalog: hive_metastore type, connected to a running HMS instance

### Steps

Start Sail with an HMS-backed catalog:
```sh
 $ export SAIL_CATALOG__LIST='[{name="spark_catalog", type="hive_metastore", uris=["localhost:9083"]}]'
 $ export SAIL_CATALOG__DEFAULT_CATALOG=spark_catalog
 $ sail spark server --ip 127.0.0.1 --port 50051
```
Then via PySpark:
```python
 
 spark = SparkSession.builder.remote("sc://127.0.0.1:50051").getOrCreate()
 
 # Case 1: INSERT INTO with VALUES
 spark.sql("INSERT INTO my_db.my_table VALUES (1, 'hello')")
 # → AnalysisException: attribute ObjectName([Identifier("hello")]) is missing
 #   from the schema: cannot resolve attribute
 
 # Case 2: DataFrame writeTo API
 df = spark.sql("SELECT 1 as id, 'hello' as text")
 df.writeTo("my_db.my_table").using("iceberg").createOrReplace()
 # → Same error: literal "hello" treated as column identifier
```
### Expected Behavior

String literal 'hello' and integer literal 1 should be resolved as literal values, not as column references (ObjectName([Identifier(...)])).

### Actual Behavior

The expression resolver within the write command path treats all literals in the input plan as identifiers, then fails to find them in the schema.

## Analysis

I traced the code path:

 1. crates/sail-spark-connect/src/proto/plan.rs — TryFrom<WriteOperationV2> correctly deserializes the write plan including all modes
 2. crates/sail-plan/src/resolver/command/write_v2.rs — resolve_command_write_to() builds the WritePlanBuilder correctly
 3. The bug appears to be in how the input DataFrame's expressions are resolved when they flow through the write resolution path — literals lose their type information and get treated as unresolved identifiers

shehabgamin:
Closes https://github.com/lakehq/sail/issues/2055


shehabgamin:
@santosh-d3vpl3x Would you like to review this?

lonless9 (approved):
Looks good to me! 
I removed some tests, and there is still some follow-up work to be done, so this part can be addressed in a later PR.

PR #2059, opened 2026-06-12
fix: data source temporary views

shehabgamin:
- Add CREATE TEMPORARY VIEW ... USING support, including global temp views.
- Handle typed schemas, comments, and Spark-compatible path option behavior.
- Preserve backing format/options and route inserts into backed temp views.


crates/sail-delta-lake/src/physical_plan/dv_writer_exec.rs
```diff
@@ -908,6 +909,7 @@ async fn scan_file_for_matching_rows(
     let file_group = FileGroup::from(vec![partitioned_file]);
     let file_scan_config = FileScanConfigBuilder::new(object_store_url, file_source)
         .with_file_groups(vec![file_group])
+        .with_expr_adapter(Some(Arc::new(SchemaEvolutionPhysicalExprAdapterFactory {})))
```
shehabgamin, line 912: @lonless9 I think this was missing from here. Let me know if it should be removed though!

crates/sail-data-source/src/formats/parquet/read.rs
```diff
@@ -155,6 +156,7 @@ impl ReadFormat for ParquetReadFormat {
             .with_file_groups(input.file_groups)
             .with_constraints(input.constraints)
             .with_statistics(input.statistics)
+            .with_expr_adapter(Some(Arc::new(SchemaEvolutionPhysicalExprAdapterFactory {})))
```
shehabgamin, line 159: @linhr @lonless9 I think this was missing from here. Let me know if it should be removed though!

crates/sail-data-source/src/listing/planner.rs
```diff
@@ -428,19 +428,25 @@ async fn do_collect_statistics_and_ordering(
...
+        )),
+        path: meta.location.clone(),
     };
 
     if let Some(cached) = file_statistic_cache
         .as_ref()
         .and_then(|x| x.get(&cache_key))
     {
-        if cached.is_valid_for(meta) {
+        // Skip a cached entry whose column count differs to avoid an out-of-bounds panic
```
shehabgamin, line 446: This was a bad bug that broadly impacted any code paths that relied on `list_files_for_scan`.

linhr (approved):
This is great! I have a few thoughts on the spec design and I'll follow up separately.

PR #2070, opened 2026-06-15
refactor: logical planning for `MERGE`

linhr:
This PR introduces `TableFormat::create_merger` to support logical planning for the `MERGE` operation. We no longer need the intermediate logical node `MergeIntoNode`. The `sail-plan-lakehouse` crate is now removed.


crates/sail-logical-plan/src/merge.rs
```diff
@@ -1858,7 +1679,7 @@ fn build_merge_projection(
         else_expr: Some(Box::new(copy_op)),
     });
     projections.push(op_expr.alias(OPERATION_COLUMN));
-    projections.push(lit(ScalarValue::UInt64(None)).alias(MERGE_SOURCE_METRIC_COLUMN));
+    projections.push(lit(ScalarValue::Int64(None)).alias(MERGE_SOURCE_METRIC_COLUMN));
```
linhr, line 1682: This change is for fixing test failures after the refactor. Here is a note from Codex:

> The real bug was in MERGE metric typing: normal MERGE rows emitted `UInt64(NULL)` for `__sail_merge_source_metric`, while the source-count row is signed, so DataFusion coerced the union to `Decimal128(20, 0)`. The Delta writer only accepts `UInt64`/`Int64`, so MERGE execution failed. I changed the placeholder to `Int64(NULL)` in `crates/sail-logical-plan/src/merge.rs`.

lonless9 (approved):

PR #2078, opened 2026-06-16
feat: introduce catalog execution context

lonless9:


python/pysail/tests/spark/delta/test_delta_io.py
```diff
@@ -73,6 +74,150 @@ def test_delta_io_create_table_with_sql(self, spark, delta_test_data, expected_p
...
+                  id BIGINT,
+                  name STRING
+                )
+                USING DELTA
+                LOCATION '{escape_sql_string_literal(str(delta_path))}'
+                """
+            )
+
+            rows = spark.sql(f"SELECT id, name FROM {table_name} ORDER BY id").collect()  # noqa: S608
+            assert rows == [Row(id=1, name="one")]
```
linhr, line 217: I don't quite understand why `CREATE OR REPLACE TABLE` keeps existing data. How does JVM Spark behave for Delta Lake or Iceberg tables?

crates/sail-delta-lake/src/session_extension.rs
```diff
@@ -18,7 +19,21 @@ const DEFAULT_MAX_ENTRIES: u64 = 1024;
 pub(crate) struct TableCacheKey {
     pub(crate) table_url: String,
     pub(crate) version: i64,
-    pub(crate) catalog_table: Vec<String>,
+    pub(crate) lakehouse_table: Option<LakehouseExecutionContext>,
```
linhr, line 22: It might be expensive to include the large struct `LakehouseExecutionContext` in the cache key. (I haven't thought about the alternatives though. So I'm just pointing this out in case we need attention for future improvements.)

crates/sail-object-store/src/registry.rs
```diff
@@ -51,11 +53,36 @@ impl DynamicObjectStoreRegistry {
...
             },
             Arc::new(LoggingObjectStore::new(Arc::new(LocalFileSystem::new()))),
         );
         Self { stores, runtime }
     }
+
+    pub fn register_session_store(
+        &self,
+        url: &Url,
+        session_fingerprint: &str,
```
linhr, line 66: Is the "fingerprint" for creating object stores that uses pre-signed URLs from the catalog? Have we thought about how this works in workers?

linhr (approved):
This is exciting! 🚀

I have a few comments for future consideration, but feel free to merge this PR and we can keep refining the design.

Also it seems this PR supports materialization for `CREATE TABLE`. Maybe we should have `feat:` for the title and highlight the changes?

python/pysail/tests/spark/delta/test_delta_io.py
lonless9: This PR hasn't implement the full `CREATE OR REPLACE TABLE` part and the testing name is misleading. This is actually the side-effect of adopt/validate location during `CREATE TABLE ... LOCATION existing_lakehouse_path`, but is wrong for `CREATE OR REPLACE TABLE`. I will fix this issue shortly.

crates/sail-object-store/src/registry.rs
lonless9: Yes that is the direction for separating operation-scpoed object store sessions. For workers we still need a follow-up mechansim. Iceberg REST vended credentials and remote signing would be examples.

PR #2080, opened 2026-06-16
fix: docker moby error

twsl:
Fixes the docker moby error and updated the node feature. Up for discussion, whether the devcontainer lock file should be included or not

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,8 +4,10 @@
     "dockerfile": "Dockerfile"
   },
   "features": {
-    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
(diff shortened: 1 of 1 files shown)


shehabgamin (approved):
Thanks, this looks good to me!

With regards to the following, cc @linhr 

> Up for discussion, whether the devcontainer lock file should be included or not

linhr:
> Up for discussion, whether the devcontainer lock file should be included or not

I'm leaning toward leaving out the devcontainer lock file for now. We don't exercise the devcontainer setup in CI regularly so reproducibility is not a top concern. By leaving out the lock file, the developer can try out latest images sooner and identify things when they break.

The support for devcontainer is largely a community-driven effort. We really appreciate PRs like this one to keep it up to date!

PR #2085, opened 2026-06-18
fix: support try_make_interval

pomykalakyle:
## Summary

- Route `try_make_interval` through the existing `SparkMakeInterval` UDF since this already contains the relevant 'try' logic
- Accept `try_make_interval` in the execution codec as the same interval UDF.
- Update datetime function gold data so the existing `try_make_interval` cases succeed.

### DIFF
--- a/crates/sail-plan/src/function/scalar/datetime.rs
+++ b/crates/sail-plan/src/function/scalar/datetime.rs
@@ -1039,7 +1039,7 @@ pub(super) fn list_built_in_datetime_functions() -> Vec<(&'static str, ScalarFun
         ("to_unix_timestamp", F::custom(to_unix_timestamp)),
         ("to_utc_timestamp", F::custom(to_utc_timestamp)),
         ("trunc", F::binary(trunc)),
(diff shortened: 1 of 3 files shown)


crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -1039,7 +1039,7 @@ pub(super) fn list_built_in_datetime_functions() -> Vec<(&'static str, ScalarFun
         ("to_unix_timestamp", F::custom(to_unix_timestamp)),
         ("to_utc_timestamp", F::custom(to_utc_timestamp)),
         ("trunc", F::binary(trunc)),
-        ("try_make_interval", F::unknown("try_make_interval")),
+        ("try_make_interval", F::udf(SparkMakeInterval::new())),
```
shehabgamin, line 1042: I'm not sure that this is correct. Where do you see the "relevant 'try'" logic here?
https://github.com/apache/datafusion/blob/283648001bbcf4e32a09ca9d796ffc682c68c329/datafusion/spark/src/function/datetime/make_interval.rs

PR #2097, opened 2026-06-18
fix: align lakehouse CREATE and REPLACE TABLE semantics

lonless9:


crates/sail-sql-analyzer/src/statement.rs
```diff
@@ -1215,8 +1225,7 @@ pub fn from_ast_statement(statement: Statement) -> SqlResult<spec::Plan> {
 
 struct TableDefinition {
     external: bool,
-    or_replace: bool,
-    if_not_exists: bool,
```
linhr, line 1219: For style consistency we need to keep the `bool` fields here and resolve them to `spec::CreateTableMode` inside `from_ast_table_definition`. This is because the helper struct `TableDefinition` defined in this file should contain only AST data.

linhr (approved):
Great work! 🚀

PR #2106, opened 2026-06-20
refactor: simplify query planning and codec for lambda functions

### LINKED ISSUE #2101: Refactor Lambda Integration
**Current implementation:**

1. Physical optimizer finds higher-order expressions.
2. It wraps them in DistributedHigherOrderExpr.
3. Codec serializes that wrapper.
4. Worker decodes the wrapper.
5. Wrapper delegates execution to the inner DataFusion expression.

**What we should likely do (https://github.com/lakehq/sail/pull/2057#pullrequestreview-4529690002):**

1. Sail builds and optimizes the normal physical plan.
2. At distributed stage serialization time, Sail handles higher-order expressions in `try_encode_plan`.
3. Codec sends the same higher-order function, lambda, variables, and schema context.
4. Worker decodes a normal DataFusion higher-order expression.
5. DataFusion evaluates it directly.

shehabgamin:
Closes https://github.com/lakehq/sail/issues/2101


crates/sail-logical-optimizer/src/lib.rs
```diff
@@ -32,6 +35,6 @@ pub fn default_optimizer_rules() -> Vec<Arc<dyn OptimizerRule + Send + Sync>> {
     // `ResolveLambdaVariables` must run after the built-in rules: constant
     // folding can change the type or nullability of higher-order function
     // arguments, and the lambda variable fields must be refreshed to match.
-    custom.push(Arc::new(ResolveLambdaVariables::new()));
+    custom.push(Arc::new(ResolveLambdaVariables));
```
shehabgamin, line 38: Turns out this is still needed, even with an added analyzer rule.

linhr (approved):
This is great!

I made some minor adjustments for the codec.

PR #2108, opened 2026-06-20
chore: reorganize Python unit tests

linhr:
1. Move Python unit test files to `dataframe/` and a few other directories.
2. Ensure consistent naming and organization for function tests.


python/pysail/tests/spark/test_dml.py
linhr, line 1: This file is removed since we won't support DML operations for the listing files data source. DML operations are only meaningful for lakehouse formats.

PR #2110, opened 2026-06-21
chore: use hashmap instead of array traversal for merge plan

ctydw123:
Came across this part of the code arbitrarily, and noticed the usage of the array is a bit weird. Based on the usage pattern, the fields should be HashMaps rather than Vecs.


lonless9 (approved):
LGTM!

PR #2114, opened 2026-06-22
fix: align EXPLAIN EXTENDED, and COST output with Spark

nis12ram:
## Summary
Updated `explain`'s `extended` and `cost` modes to align more with Spark's behaviour.

## Extended mode
Separated `analyzed logical` and `optimized logical` in the same way Spark does by prepending the output schema of the analyzed logical plan in front of it.

Screenshot 2026-06-23 161448

## Cost mode
Updated from `Parsed Logical - Analyzed Logical - Physical (stat) `to `Optimized Logical - Physical (stat)` to align more with Spark's behaviour, which is `Optimized Logical (stat) - Physical`

Screenshot 2026-06-23 161526

### DIFF
--- a/crates/sail-plan/src/explain.rs
+++ b/crates/sail-plan/src/explain.rs
@@ -82,6 +82,7 @@ pub struct ExplainString {
 
 struct CollectedPlan {
     initial_logical: LogicalPlan,
+    analyzed_logical: LogicalPlan,
     optimized_logical: LogicalPlan,
     physical_plan: Option<Arc<dyn ExecutionPlan>>,
     physical_error: Option<String>,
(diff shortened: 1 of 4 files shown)


crates/sail-plan/src/explain.rs
```diff
@@ -417,9 +434,8 @@ async fn explain_from_collected(
         }
         ExplainKind::Extended => [
             render_section("Parsed Logical Plan", &logical_simple),
-            // TODO: Spark expects distinct analyzed vs optimized plans
```
nis12ram, line 420: To implement this, prepended the output schema of the analyzed logical plan in front of it, tried to do it in the same way as Spark does.

**For the schema, considered only Arrow data type names and did not map them to their Spark counterparts, as mapping to Spark counterparts can confuse users because the displayed analyzed plan itself contains Arrow data types.**

@linhr, can you please share your thoughts on this approach, to separate analyzed and optimized plans?

Thanks in advance

Image

linhr: Yes, I think it makes sense to show Arrow data type instead of Spark data type. The Arrow data type is what we use internally so the user gets an accurate understanding of how we do query planning.

linhr: You would need to update the snapshots to make CI working.

<https://docs.lakesail.com/sail/latest/development/build/python.html#updating-test-snapshots>

nis12ram:
@linhr Thanks again for your help.

Applied the changes.

linhr (approved):
This looks nice! Thanks!

PR #2120, opened 2026-06-23
feat: add `pivot` (rewrite to aggregate with per-value FILTER)

davidlghellin:
Implement resolve_query_pivot (previously a todo) by rewriting a pivot into a grouped aggregate that computes each aggregate with a FILTER per pivot value, matching Spark semantics. Supports explicit values and value inference (distinct scan during planning, capped at 10000 like spark.sql.pivotMaxValues), explicit or implicit grouping, Spark column naming (value, value_aggName, lowercase null), and NULL pivot values via an IS NULL predicate.

Fixes test_numeric_aggregation (DataFrame .pivot()) for Spark 3.5.7 and 4.1.1. Defers multi-column and struct (multi-literal) pivots with clear errors.


crates/sail-plan/src/resolver/query/pivoting.rs
```diff
@@ -19,10 +26,211 @@ use crate::resolver::PlanResolver;
...
+
+    /// Infer pivot values by collecting the distinct values of the pivot column
+    /// from the input, sorted ascending (matching Spark's `pivot(column)`).
+    async fn infer_pivot_values(
+        &self,
+        input: &LogicalPlan,
+        pivot_column: &expr::Expr,
+    ) -> PlanResult<Vec<ScalarValue>> {
+        // Spark rejects pivots whose distinct-value count exceeds the configured maximum
+        // (`spark.sql.pivotMaxValues`, default 10000), defined as `DATAFRAME_PIVOT_MAX_VALUES`
```
shehabgamin, line 161: We can actually support this as a config instead of hardcoding

shehabgamin (commented):
@davidlghellin Well done! Let's support the config for max pivot values and then this should mostly be in good shape

crates/sail-plan/src/resolver/query/pivoting.rs
davidlghellin: done

shehabgamin (approved):
Well done!

Issue #229, opened 2024-09-29
Spark SQL Functions Coverage and Parity - Lambda 


shehabgamin:
- [ ] aggregate
- [ ] array_sort
- [ ] exists
- [ ] filter
- [ ] forall
- [ ] map_filter
- [ ] map_zip_with
- [ ] reduce
- [ ] transform
- [ ] transform_keys
- [ ] transform_values
- [ ] zip_with

keen85:
@shehabgamin, are there any plans for tackling this in somewhat near future?

shehabgamin:
> [@shehabgamin](https://github.com/shehabgamin), are there any plans for tackling this in somewhat near future?

@keen85 Would you like to give it a shot?

keen85:
@shehabgamin, I did some analysis which PySpark functions we use in our project and `transform` and some others from this list showed up quite frequently.
So this is blocking us from `sail` adoption.
That is why I ask.

Unfortunately, I can't program Rust myself.

shehabgamin:
@keen85 Got it. Do you happen to know which exact functions are used?

keen85:
@shehabgamin, concerning **lambda** functions?
Several times `pyspark.sql.functions.transform` and a few `pyspark.sql.functions.array_sort`.

shehabgamin:
@keen85 Got it, will take a look and evaluate the scope.

shehabgamin:
@keen85 Looks like we can work with DataFusion to support this upstream hopefully!
https://github.com/apache/datafusion/pull/17220#issuecomment-3577605059

shehabgamin:
Follow-up to ^ here:
https://github.com/apache/datafusion/pull/18921

hugopendlebury:
Datafusion have just merged PR 21679 which adds support for lambdas and includes an initial implementation of transform or as they call it array_transform.

I wouldn't mind having a go at this issue. Would it be possible to give me a high level overview of the starting point and which files / crates I would need to get familiar with ?

lonless9:
Thanks @hugopendlebury . Although the upstream has pushed forward the relevant PR, I guess we won't be able to start the related work until after the datafusion 54.0.0 release in about a month. As for the starting point, I think @davidlghellin  has a better idea.

davidlghellin:
Hi, thanks! 
[...]

PR #2121, opened 2026-06-23
feat: support `array_sort` array higher-order function with comparator lambda

davidlghellin:
related https://github.com/lakehq/sail/issues/229

### DIFF
--- a/crates/sail-function/src/scalar/array/spark_array_sort.rs
+++ b/crates/sail-function/src/scalar/array/spark_array_sort.rs
@@ -1,0 +1,505 @@
+/// Spark-compatible `array_sort(array, comparator)` higher-order function.
+///
+/// Unlike `filter`/`transform` (element-wise maps evaluated once over the
+/// flattened values), `array_sort`'s lambda is a *comparator* `(left, right) ->
+/// int`: it must be invoked over pairs of elements and its result drives the
+/// reordering of each sublist.
+///
+/// Spark semantics: see `ArraySort` in
+/// `sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/higherOrderFunctions.scala`
+/// — the comparator receives both elements (NULLs included; the lambda decides
+/// their order, there is no NULLS-LAST special-casing in this form), must return
+/// `IntegerType`, and a `NULL` comparator result raises an error
+/// (`comparatorReturnsNull`). The sort is stable (`java.util.Arrays.sort`) and
+/// the result preserves the input array type.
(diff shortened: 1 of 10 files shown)


crates/sail-function/src/scalar/array/spark_array_sort.rs
```diff
@@ -0,0 +1,505 @@
...
+                .map(|o| o.as_usize())
+                .collect(),
+            other => return exec_err!("{} expected list, got {other}", self.name()),
+        };
+
+        // Enumerate ordered pairs `(i, j)`, `i != j`, within each sublist. Using
+        // ordered pairs (rather than the upper triangle) avoids assuming the
+        // user comparator is antisymmetric and mirrors the actual comparator
+        // calls a comparison sort would make.
+        let mut left_idx: Vec<u32> = Vec::new();
```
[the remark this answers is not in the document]
davidlghellin: Fixed — switched to a per-comparison stable sort (the same model as
Spark's `ArraySort`: `java.util.Arrays.sort` + per-pair `f.eval`), so it's
now O(n log n) comparisons and O(1) extra memory, with no n² pair matrix.

crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -254,19 +262,58 @@ fn array_sort_spark(array: expr::Expr, asc: expr::Expr) -> PlanResult<expr::Expr
...
-        ));
+    if rest.is_empty() {
+        // array_sort(array) without lambda - ascending order, NULLs last (Spark behavior).
+        return array_sort_spark(array, lit(true));
     }
 
-    // array_sort(array) without lambda - sorts in ascending order with NULLs last (Spark behavior)
-    array_sort_spark(array, lit(true))
+    let lambda = rest.one()?;
+    let expr::Expr::Lambda(lambda) = lambda else {
```
shehabgamin: @davidlghellin Direct expr matching can be a bit unreliable, since there may be aliases or other variations. What if we add the expectations when there is a direct expr match, and otherwise fall back to a more lenient approach? Future work could then focus on improving expr matching more broadly.

davidlghellin: Done — expectations now apply only on a direct `Expr::Lambda` match, with a lenient pass-through for other shapes (matching `aggregate`/`filter`). 

Thanks, I'll keep the broader expr-matching improvement in mind as future work.

shehabgamin (approved):
Well done!

PR #2123, opened 2026-06-24
feat: lambda aggregate

shehabgamin:


crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -157,39 +166,49 @@ fn transform(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
 
 fn exists(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     let (array, lambda) = input.arguments.two()?;
-    let expr::Expr::Lambda(lambda) = lambda else {
```
shehabgamin, line 160: Checking for expr type here is unreliable

crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -157,39 +166,49 @@ fn transform(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
...
     Ok(expr::Expr::HigherOrderFunction(HigherOrderFunction::new(
         Arc::clone(&SPARK_ARRAY_EXISTS_UDF),
-        vec![array, expr::Expr::Lambda(lambda)],
+        vec![array, lambda],
     )))
 }
 
 fn forall(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     let (array, lambda) = input.arguments.two()?;
-    let expr::Expr::Lambda(lambda) = lambda else {
```
shehabgamin, line 179: Checking for expr type here is unreliable

crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -157,39 +169,67 @@ fn transform(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
...
+            (array, zero, merge, finish)
+        }
+        4 => args.four()?,
+        n => {
+            return Err(PlanError::AnalysisError(format!(
+                "`aggregate` expects 3 or 4 arguments, got {n}"
+            )));
+        }
+    };
+    let (func, merge) = match merge {
```
shehabgamin, line 212: Checking for expr type here is unreliable but we have a fallback. Future work can explore a better option if needed

crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -104,36 +119,33 @@ fn array_lambda_with_index(
...
-        (
-            Arc::clone(udf_index_first),
-            Lambda {
-                params: vec![index],
-                body,
-            },
-        )
-    } else {
-        (Arc::clone(udf), lambda)
+    let (func, lambda) = match lambda {
```
shehabgamin, line 122: Checking for expr type here is unreliable but we have a fallback. Future work can explore a better option if needed

PR #2127, opened 2026-06-24
fix: validate aggregate/reduce lambda arity

davidlghellin:
The merge lambda of `aggregate`/`reduce` requires exactly 2 parameters and the finish lambda exactly 1. The UDF binding only rejects lambdas with too many parameters, so a merge with too few (e.g. `x -> x`) silently bound to a prefix and returned a wrong result instead of erroring like Spark (INVALID_LAMBDA_FUNCTION_CALL.NUM_ARGS_MISMATCH). Add an explicit arity check in the `aggregate` planner builder.

Also:
- mark exists/filter/forall/transform as supported in collection.json
- tag all higher-order-function feature files with @lambda_hof
- expand aggregate.feature with arity, reduce-alias, Inf/NaN/all-NULL and ANSI-overflow (@sail-bug) scenarios, all validated against Spark JVM
- reuse the fold's gather buffers across depth iterations (with_capacity, no per-depth alloc, no clone)

### DIFF
--- a/crates/sail-plan/src/function/scalar/lambda.rs
+++ b/crates/sail-plan/src/function/scalar/lambda.rs
@@ -191,13 +191,14 @@ fn forall(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
 fn expect_lambda_arity(role: &str, expr: &expr::Expr, arity: usize) -> PlanResult<()> {
     let expr::Expr::Lambda(lambda) = expr else {
         return Err(PlanError::AnalysisError(format!(
-            "`aggregate` expects a lambda function as the {role} argument"
+            "expects a lambda function as the {role} argument"
         )));
     };
     if lambda.params.len() != arity {
         // Mirrors Spark's `INVALID_LAMBDA_FUNCTION_CALL.NUM_ARGS_MISMATCH` wording.
+        // No function name: `aggregate` and `reduce` share this builder.
         return Err(PlanError::AnalysisError(format!(
-            "Invalid lambda function call. The `aggregate` {role} lambda function expects {arity} arguments, but got {}",
+            "Invalid lambda function call. The {role} lambda function expects {arity} arguments, but got {}",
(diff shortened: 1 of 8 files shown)


python/pysail/tests/spark/function/features/exists.feature
```diff
@@ -341,3 +342,12 @@ Feature: exists higher-order function
...
       Then query error .*
+
+  Rule: Lambda only accepts one parameter
+
+    Scenario: two-parameter lambda is rejected as type error
+      When query
+        """
+        SELECT exists(array(1, 2, 3), (x, i) -> x > i) AS result
+        """
+      Then query error .*
```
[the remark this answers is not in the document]
davidlghellin: clean

crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -209,6 +231,8 @@ fn aggregate(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
             )));
         }
     };
+    expect_lambda_arity("merge", &merge, 2)?;
+    expect_lambda_arity("finish", &finish, 1)?;
```
shehabgamin, line 238: Direct `expr` matching can be a bit unreliable, since there may be aliases or other variations. What if we add the expectations when there is a direct `expr` match, and otherwise fall back to a more lenient approach? Future work could then focus on improving `expr` matching more broadly.

shehabgamin (commented):
@davidlghellin Thanks for the follow up! Wdyt?

crates/sail-plan/src/function/scalar/lambda.rs
davidlghellin: Done 👍 only enforces on a direct `Expr::Lambda` match now, lenient otherwise. 
Lesson learned

shehabgamin (approved):
Looks great!

PR #2139, opened 2026-06-26
fix: round-trip Iceberg manifest lower/upper bounds through Avro

alexy:
## Problem

`IntBytesMapEntry.value` in the manifest `_serde.rs` is a `Vec<u8>` typed as Avro
`bytes` in the manifest schema, but serde's default `Vec<u8>` serialization emits an
Avro **array**. That array fails to resolve against the `bytes` schema, so the
nullable `lower_bounds`/`upper_bounds` maps are silently written as `null` — every
column lower/upper bound is dropped on a manifest write → parse round-trip.

## Fix

Force byte-string (de)serialization on the entry `value` (tolerating the legacy
array form on read for backward compatibility), so bounds survive the round-trip.

## Testing

- `cargo test -p sail-iceberg` — 79 passed.
- Confirmed a `DataFile` lower bound now round-trips through
  `ManifestWriter::to_avro_bytes_v2()` → `Manifest::parse_avro()` with the bound
  preserved (previously `lower_bounds().get(&1)` came back `None`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UbNyd333y17hMC55Pk4CRk

### DIFF
--- a/crates/sail-iceberg/src/datasource/pruning.rs
+++ b/crates/sail-iceberg/src/datasource/pruning.rs
@@ -35,7 +35,6 @@ use crate::utils::transform::apply_transform;
 /// Pruning statistics over Iceberg DataFiles
 pub struct IcebergPruningStats {
     files: Vec<DataFile>,
-    #[expect(unused)]
     arrow_schema: Arc<ArrowSchema>,
     /// Arrow field name -> Iceberg field id
     name_to_field_id: HashMap<String, i32>,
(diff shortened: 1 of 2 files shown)


alexy:
Marking this as draft per the maintainer feedback — thanks for the steer.

Flagging the substantive finding independent of this patch: `sail-iceberg` silently drops manifest `lower_bounds`/`upper_bounds` on read. `IntBytesMapEntry.value: Vec<u8>` is typed as Avro `bytes`, but serde serializes `Vec<u8>` as an Avro *array*, so the nullable bounds maps resolve to null and every column bound is lost on a write→parse round-trip.

Enabling the round-trip then surfaces a second issue: partition-transform pruning compares an `Int32` transformed bound against a `Timestamp`/`Date` predicate (`test_partition_transform_pruning`), tripping DataFusion's type assertion. My commits add a defensive type guard, but the proper fix likely belongs with your catalog/Iceberg work.

Please treat this as a bug report rather than a merge request — happy to hand off the repro or align with however you'd prefer to fix it.

alexy:
Consolidating into the LakeCat tracking PR #2145, which bundles these changes alongside the rest of the lakecat branch and LAKECAT-SAIL.md. Closing in favor of #2145.

PR #2134, opened 2026-06-25
fix: make `TableUpdate`/`ViewUpdate` discriminated enums in Iceberg REST catalog

alexy:
## Summary
- `models::TableUpdate`/`models::ViewUpdate` (generated by `sail-catalog-iceberg`'s OpenAPI client) were flat structs with every field of every update variant marked required, because the **upstream** Apache Iceberg REST spec declares them as bare `anyOf` with no discriminator (unlike `BaseUpdate`, which has one). Any real update — e.g. a `set-properties` commit — failed to deserialize into `CommitTableRequest`/`CommitViewRequest` with `missing field uuid`.
- `crates/sail-catalog-iceberg/spec/iceberg-rest-catalog.yaml` is a pinned, verbatim snapshot of `apache/iceberg`'s own spec (see the `SOURCE:` header) and is **not modified** here — this is a gap in the upstream spec itself, not a local authoring mistake.
- Instead, hand-patch the two generated files (`table_update.rs`, `view_update.rs`) into the `#[serde(tag = "action")]` enum shape the generator would emit if the spec carried a discriminator, mirroring the already-correct `base_update.rs`. This follows the repo's existing "Post-Generation Manual Steps" convention for working around `openapi-generator`'s Rust/OpenAPI-3.1 gaps (documented as a new step 4 in `docs/development/recipes/iceberg-openapi-generation.md`, so it survives the next `generate-client.sh` run).
- `docs/development/TABLE-UPDATE.md` describes the defect and the patch in full, including the variant/action list for both enums.
- Adds regression tests.

## Test plan
- [x] `cargo test -p sail-catalog-iceberg --lib` (24 passed)
- [x] `cargo +nightly fmt --check`
- [x] `cargo clippy -p sail-catalog-iceberg --all-targets --all-features -- -D warnings` (stable, matching CI's lint job)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2141, opened 2026-06-26
feat: improve catalog providers

linhr:
- Add shared catalog credential plumbing and use it across Iceberg REST, Unity Catalog, and OneLake providers to support automatic credential refresh, while ensuring connection reuse for the HTTP client.
- Support proper OneLake authentication by using the credential utilities from Azure storage in `object_store`.
- Rework the OneLake catalog provider to delegate to protocol-specific providers, with selectable `api = "delta"` or `api = "iceberg"` configuration options.
- Improve Unity Catalog compatibility for OneLake responses.
- Fix more issues (after #2134) for the generated Iceberg REST catalog client code around table commit.
- Refactor configuration management for Iceberg REST and Unity Catalog.
- Add `User-Agent` header for catalog requests.

We don't yet have integration tests for OneLake, so the related functionalities are tested manually. The existing integration tests for Iceberg REST and Unity Catalog are still working.

### DIFF
--- a/crates/sail-catalog-iceberg/src/provider.rs
+++ b/crates/sail-catalog-iceberg/src/provider.rs
@@ -10,11 +10,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::hash::{DefaultHasher, Hash, Hasher};
 use std::sync::Arc;
 
 use percent_encoding::percent_decode_str;
+use sail_catalog::credentials::CatalogCredentials;
 use sail_catalog::error::{CatalogError, CatalogObject, CatalogResult};
 use sail_catalog::lakehouse::{
     BeginTableAccessRequest, LakehouseCapability, LakehouseCommitOutcome, LakehouseCommitRequest,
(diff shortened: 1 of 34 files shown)


crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -59,217 +62,191 @@ const REST_REMOTE_SIGNING_ENABLED_KEY: &str = "s3.remote-signing-enabled";
...
             .get(REST_CATALOG_PROP_PREFIX)
             .map(|value| value.trim())
             .filter(|value| !value.is_empty())
     }
 
-    fn is_namespace_separator_key(key: &str) -> bool {
-        matches!(
-            key.trim().to_ascii_lowercase().as_str(),
-            "namespace-separator" | "namespaceseparator" | "namespace_separator"
-        )
```
linhr, line 103: The normalization logic for the "namespace separator" property key is removed. The Iceberg docs never mention that other spelling for the key is allowed. So we require that the key is strictly `namespace-separator` to be consistent with other properties.

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -283,126 +260,11 @@ impl IcebergRestCatalogProvider {
...
-            table = crate::apis::urlencode(table)
-        );
-        let body = serde_json::json!({
-            "identifier": {
-                "namespace": Vec::<String>::from(database.clone()),
-                "name": table,
-            },
-            "requirements": requirements,
-            "updates": updates,
-        });
```
linhr, line 336: We no longer need to construct the commit request manually thanks to the generated model bug fix by @alexy in #2134. (It's perfect timing!) The bug fix turns out to require some adjustments though, which is why I made further changes for `crates/sail-catalog-iceberg/src/models` in this PR.

The integration tests in `python/pysail/tests/spark/catalog/iceberg_rest/test_commit.py` show that this is working.

cc @lonless9

crates/sail-catalog-unity/build.rs
```diff
@@ -35,15 +35,100 @@ fn build_unity_catalog() -> Result<(), Box<dyn std::error::Error>> {
             "Self::new_with_client(baseurl, client.build().unwrap())",
             "Ok(Self::new_with_client(baseurl, client.build()?))",
         )
+        .replace(
```
linhr, line 38: The patch in this file is pretty hacky. I'm thinking about implementing our own codegen logic so that we can have better control over the generated code for bug fixes and overrides.

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -59,217 +62,191 @@ const REST_REMOTE_SIGNING_ENABLED_KEY: &str = "s3.remote-signing-enabled";
...
-            //  We may need to support them in the future.
-            match key.as_str() {
-                "oauth-access-token" => {
-                    client_config.oauth_access_token = Some(value.to_string());
-                }
-                "bearer-access-token" => {
-                    client_config.bearer_access_token = Some(value.to_string());
-                }
-                _ => {}
-            }
```
linhr, line 147: We no longer store tokens in the properties since they are not documented properties according to Iceberg. They are now handled when creating the catalog provider.

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -59,217 +62,191 @@ const REST_REMOTE_SIGNING_ENABLED_KEY: &str = "s3.remote-signing-enabled";
...
-    fn is_namespace_separator_key(key: &str) -> bool {
-        matches!(
-            key.trim().to_ascii_lowercase().as_str(),
-            "namespace-separator" | "namespaceseparator" | "namespace_separator"
-        )
+    fn namespace_separator(&self) -> CatalogResult<String> {
+        let separator = self
+            .properties
+            .get(REST_CATALOG_PROP_NAMESPACE_SEPARATOR)
+            .map(|value| value.trim());
```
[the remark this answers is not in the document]
linhr: This comment is incorrect. Although `namespace_separator` is the user-facing key in Sail application config, it gets translated to the `namespace-separator` property in `crates/sail-session/src/catalog.rs`.

crates/sail-catalog-unity/src/provider.rs
```diff
@@ -1003,7 +982,7 @@ impl CatalogProvider for UnityCatalogProvider {
             .await
             .map_err(|e| CatalogError::External(format!("Failed to load client: {e}")))?;
         let (catalog_name, schema_name) = self.get_catalog_and_schema_name(database)?;
-        let full_name = Self::get_full_table_name(&catalog_name, &schema_name, table);
+        let full_name = format!("{catalog_name}.{schema_name}.{table}");
```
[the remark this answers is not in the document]
linhr: Fixed.

docs/guide/catalog/onelake.md
```diff
@@ -12,25 +12,85 @@ OneLake catalog can be configured using the following options:
 - `type` (required): The string `onelake`.
 - `name` (required): The name of the catalog.
 - `url` (required): The OneLake item location.
+- `api` (optional): The OneLake catalog API. Use `delta` for the Unity Catalog endpoint or `iceberg` for the Iceberg REST catalog endpoint. The default is `delta`.
```
shehabgamin: Platform will need to add this field as well. Should only be a few lines of code to add it.

crates/sail-catalog-iceberg/src/provider.rs
shehabgamin: @linhr This change was initially added because of our configuration setup:
https://github.com/lakehq/sail/blob/91370a09c2d2077492ee1eb766e2023c8f013b38/crates/sail-common/src/config/application.rs#L533

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -59,217 +62,191 @@ const REST_REMOTE_SIGNING_ENABLED_KEY: &str = "s3.remote-signing-enabled";
...
+            base_path,
+            user_agent: Some(SAIL_USER_AGENT.to_string()),
+            client: http_client,
+            basic_auth: None,
+            oauth_access_token: None,
+            bearer_access_token: None,
+            api_key: None,
+        };
+        if let Some(credential) = credentials.retrieve().await? {
+            client_config.bearer_access_token = Some(credential);
```
shehabgamin, line 166: This looks good.

Seems that we don't need to worry about setting `oauth_access_token`. According to the following link, `OAuth` is deprecated:
https://editor.swagger.io/?url=https://raw.githubusercontent.com/apache/iceberg/main/open-api/rest-catalog-open-api.yaml

crates/sail-catalog-onelake/src/credentials.rs
```diff
@@ -0,0 +1,63 @@
...
+        match self {
+            Self::Static { bearer_token } => Ok(Some(bearer_token.clone())),
+            Self::Dynamic {
+                workspace,
+                provider,
+            } => {
+                let provider = provider
+                    .get_or_try_init(|| async {
+                        let mut builder = MicrosoftAzureBuilder::from_env()
+                            .with_account("onelake")
```
shehabgamin, line 29: Does the value of `with_account` matter?

crates/sail-catalog-onelake/src/credentials.rs
```diff
@@ -0,0 +1,63 @@
...
+                            .with_container_name(workspace)
+                            .with_use_fabric_endpoint(true);
+
+                        if std::env::var("AZURE_STORAGE_TOKEN").is_err() {
```
shehabgamin, line 33: Do we not need to do anything with `AZURE_STORAGE_TOKEN` if it exists?

crates/sail-catalog-unity/build.rs
shehabgamin: Should we update the OpenAPI yaml file?

Not sure when we last updated, but as of 2 months ago there are new changes here:
https://github.com/unitycatalog/unitycatalog/blob/main/api/all.yaml

Additionally, this `delta.yaml` seems to be a new file as of April:
https://github.com/unitycatalog/unitycatalog/blob/main/api/delta.yaml

shehabgamin (approved):
Well done!

Approved, although it would be good to double-check the changes made to `namespace_separator`.

docs/guide/catalog/onelake.md
shehabgamin: @linhr This isn't outdated, it only says so because the text was changed.

crates/sail-catalog-iceberg/src/provider.rs
linhr: Yeah the code removal is intended. See also my note regarding the Copilot comment: <https://github.com/lakehq/sail/pull/2141#discussion_r3480391229>.

crates/sail-catalog-unity/build.rs
linhr: Yeah let me update the YAML files in future PRs.

crates/sail-catalog-iceberg/src/provider.rs
linhr: It seems `oauth_access_token` and `bearer_access_token` are just names from the code generator. All we actually need is a bearer token for the HTTP header. In some future PR, I may just keep a single `bearer_token` option for the catalog config while keep these names as aliases for backward compatibility.

crates/sail-catalog-onelake/src/credentials.rs
linhr: The account name is always "onelake".

<https://learn.microsoft.com/en-us/fabric/onelake/onelake-access-api>

crates/sail-catalog-onelake/src/credentials.rs
linhr: `AZURE_STORAGE_TOKEN` is already handled by `MicrosoftAzureBuilder::from_env()`. This `if` is meant for the fallback to `AZURE_ACCESS_TOKEN` (when `AZURE_STORAGE_TOKEN` is not specified) which is not handled by the builder.

BTW I cannot find any documentation around `AZURE_ACCESS_TOKEN`. This could be a hallucination from the existing AI-generated OneLake implementation before this PR. Here we keep the handling logic for backward compatibility. We could also decide to remove it if we want.

PR #2148, opened 2026-06-27
fix: delta lake legacy protocol check blocking writes to delta-spark-created column-mapping tables

### LINKED ISSUE #2147: Delta Lake legacy protocol check blocking writes to delta-spark-created column-mapping tables
## Problem

Writes to delta-spark-created tables with legacy protocol versions (e.g. `minWriterVersion=5` with column mapping mode `name`) fail with:

```
Unsupported table features required: [ChangeDataFeed]
```

even when Change Data Feed is not enabled on the table. Reported in #2041.

## Blast radius

The blast radius covers tables created by `delta-spark` with column mapping mode feature enabled (`minWriterVersion=5`). Column mapping support was implemented in PR #985 (shipped in v0.5.0), so any user who created or interacted with column-mapping tables via delta-spark and then upgraded pysail past v0.5.3 is affected.

Specifically:
- Any table at `minWriterVersion=4` or higher is rejected if `ChangeDataFeed` is not in sail's supported writer feature set — regardless of whether CDF is enabled.
- Tables at `minWriterVersion=5` (the version delta-spark uses for column mapping mode `name`) are the most common case, since column mapping is a widely-used feature.
- `CheckConstraints` was also affected but was fixed in v0.6.4 by PR #1993 (which added `CheckConstraints` to the supported writer feature set). `ChangeDataFeed` remains unsupported and unfixed.

## Root cause

`ProtocolChecker::required_writer_features` ([`crates/sail-delta-lake/src/transaction/protocol.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/transaction/protocol.rs) `required_writer_features`) expands legacy writer versions (2–6) into their full implied feature sets and requires the writer to support every implied feature. For example, `minWriterVersion=5` implies `{AppendOnly, Invariants, CheckConstraints, ChangeDataFeed, GeneratedColumns, ColumnMapping}`. Since `ChangeDataFeed` writer support is not implemented (commented out in the `INSTANCE` static), every v5 table is rejected — regardless of whether CDF is actually enabled.

This behavior is inherited from **delta-rs** (credited at the top of `protocol.rs`). However, **delta-spark — the reference implementation — does not expand implied features for legacy versions.** Its [`protocolCheck`](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala) only checks (a) the version number is supported and (b) explicitly-listed `writerFeatures` (empty for legacy) are supported. Feature activation is validated separately at operation time via [`FeatureAutomaticallyEnabledByMetadata`](https://github.com/delta-io/delta/blob/master/spark/src/main/scala/org/apache/spark/sql/delta/TableFeature.scala).

## Historical context

- **PR #985** (v0.5.0, Nov 2025): Implemented column mapping for Delta Lake. This is when sail started supporting `delta.columnMapping.mode=name` tables.
- **PR #1557** (v0.6.0, Mar 2026): Refactored the protocol checker. Before this PR, the `INSTANCE` writer feature set falsely claimed `ChangeDataFeed` support, which masked the latent bug in the implied-feature expansion. PR #1557 made the feature claims honest (commenting out unimplemented features like `ChangeDataFeed`), which exposed the bug: the implied-feature expansion was always stricter than delta-spark, but was previously hidden by false claims.
- **PR #1993** (v0.6.4, Jun 2026): Added `CheckConstraints` to the supported writer feature set, fixing v3 tables with check constraints. `ChangeDataFeed` remains unsupported, so v4+ tables still fail.

## Impact

- Unblocks writes to delta-spark-created column-mapping tables (v5) and any legacy-protocol table (v2–6) with inactive unsupported features
- Aligns sail with delta-spark's protocol checking behavior
- No table migration required — purely a server-side code change

## Reproduction

Save as `reproduce_issue_2041.py` and run:

```bash
pip install pyspark delta-spark pysail
python reproduce_issue_2041.py /tmp/delta_cm_table
```

```python
"""Reproduction script for lakehq/sail#2041.

Creates a Delta table with delta-spark (column mapping mode=name), then
attempts to append data with pysail. On pysail >= 0.6.0, the append fails
with "Unsupported table features required: [ChangeDataFeed]".
"""
from __future__ import annotations

import argparse
import json
import sys
from pathlib import Path


def create_table_with_delta_spark(path: str) -> None:
    from delta import configure_spark_with_delta_pip

    print(f"[1] Creating Delta table at {path} with delta-spark (columnMapping.mode=name)...")

    spark = (
        SparkSession.builder.appName("reproduce_issue_2041")
        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
        .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog")
        .config("spark.sql.session.timeZone", "UTC")
    )
    spark = configure_spark_with_delta_pip(spark).getOrCreate()

    df = spark.createDataFrame([(1, "alice"), (2, "bob")], schema="id INT, name STRING")

    (
        df.write.format("delta")
        .mode("overwrite")
        .option("delta.columnMapping.mode", "name")
        .save(path)
    )

    print(f"    Table created. Row count: {spark.read.format('delta').load(path).count()}")
    spark.stop()


def print_protocol(path: str) -> None:
    log_file = Path(path) / "_delta_log" / "00000000000000000000.json"
    print(f"\n[2] Protocol from {log_file}:")
    with log_file.open("r", encoding="utf-8") as f:
        for line in f:
            obj = json.loads(line)
            if "protocol" in obj:
                p = obj["protocol"]
                print(f"    minReaderVersion: {p.get('minReaderVersion')}")
                print(f"    minWriterVersion: {p.get('minWriterVersion')}")
                print(f"    readerFeatures:   {p.get('readerFeatures', 'none (legacy)')}")
                print(f"    writerFeatures:   {p.get('writerFeatures', 'none (legacy)')}")
            if "metaData" in obj:
                config = obj["metaData"].get("configuration", {})
                print(f"    delta.enableChangeDataFeed: {config.get('delta.enableChangeDataFeed', 'not set')}")
                print(f"    delta.columnMapping.mode:   {config.get('delta.columnMapping.mode', 'not set')}")


def write_with_pysail(path: str) -> None:

    print(f"\n[3] Attempting to append data with pysail to {path}...")

    server = SparkConnectServer("127.0.0.1", 0)
    server.start(background=True)

    try:
        spark = (
            SparkSession.builder.appName("pysail_write")
            .remote(f"sc://localhost:{port}")
            .config("spark.sql.session.timeZone", "UTC")
            .create()
        )

        df = spark.createDataFrame([(3, "charlie"), (4, "dave")], schema="id INT, name STRING")

        df.write.format("delta").mode("append").save(path)

        count = spark.read.format("delta").load(path).count()
        print(f"    SUCCESS: Append completed. Total row count: {count}")
        spark.stop()
    except Exception as e:
        print(f"    FAILED: {e}")
        raise
    finally:
        server.stop()


def main() -> None:
    parser = argparse.ArgumentParser(description="Reproduce lakehq/sail#2041")
    parser.add_argument("path", type=str, help="Path for the Delta table")
    args = parser.parse_args()

    path = args.path
    create_table_with_delta_spark(path)
    print_protocol(path)

    try:
        write_with_pysail(path)
    except Exception:
        print("\n[RESULT] Reproduced: pysail cannot write to a delta-spark-created")
        print("         column-mapping table due to the legacy protocol check.")
        print("         See https://github.com/lakehq/sail/issues/2041")
        sys.exit(1)

    print("\n[RESULT] No issue: pysail successfully wrote to the table.")


if __name__ == "__main__":
    main()
```

**Expected output on pysail >= 0.6.0 (before fix):**

```
[1] Creating Delta table at /tmp/delta_cm_table with delta-spark (columnMapping.mode=name)...
    Table created. Row count: 2

[2] Protocol from /tmp/delta_cm_table/_delta_log/00000000000000000000.json:
    minReaderVersion: 2
    minWriterVersion: 5
    readerFeatures:   none (legacy)
    writerFeatures:   none (legacy)
    delta.enableChangeDataFeed: not set
    delta.columnMapping.mode:   name

[3] Attempting to append data with pysail to /tmp/delta_cm_table...
    FAILED: Unsupported table features required: [ChangeDataFeed]

[RESULT] Reproduced: pysail cannot write to a delta-spark-created
         column-mapping table due to the legacy protocol check.
         See https://github.com/lakehq/sail/issues/2041
```

**Expected output after fix (or on pysail <= 0.5.3):**

```
[3] Attempting to append data with pysail to /tmp/delta_cm_table...
    SUCCESS: Append completed. Total row count: 4

[RESULT] No issue: pysail successfully wrote to the table.
```

zemin-piao:
Fixes https://github.com/lakehq/sail/issues/2147

## Proposed fix

Stop expanding implied features for legacy versions (1–6). Just check the version number is supported, matching delta-spark. Feature activation is already validated at operation time by existing guards:

- **Append-only**: `can_commit` checks `delta.appendOnly` property ([`protocol.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/transaction/protocol.rs) `can_commit`)
- **Change Data Feed**: [`transaction/mod.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/transaction/mod.rs) rejects writes when `delta.enableChangeDataFeed=true` (the guard at line 889)
- **Column mapping**: `verify_column_mapping()` in [`snapshot/mod.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/snapshot/mod.rs) validates `delta.columnMapping.mode`
- **Deletion vectors**: [`transaction/mod.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/transaction/mod.rs) rejects when `delta.enableDeletionVectors=true` without protocol support

The change is ~10 lines in `required_writer_features` and `required_reader_features`:

```rust
// Before:
match protocol.min_writer_version() {
    0 | 1 => Ok(None),
    2 => Ok(Some(WRITER_V2.clone())),
    3 => Ok(Some(WRITER_V3.clone())),
    4 => Ok(Some(WRITER_V4.clone())),
    5 => Ok(Some(WRITER_V5.clone())),
    6 => Ok(Some(WRITER_V6.clone())),
    7 => Ok(Some(protocol.writer_features()...)),
    ...
}

// After:
match protocol.min_writer_version() {
    0..=6 => Ok(None),
    7 => Ok(Some(protocol.writer_features()...)),
    ...
}
```

The `WRITER_V2` through `WRITER_V6` and `READER_V2` static sets become dead code and are removed.

Additionally, the CDF guard in [`transaction/mod.rs`](https://github.com/lakehq/sail/blob/main/crates/sail-delta-lake/src/transaction/mod.rs) (line 889) must be updated to reject ALL CDF-enabled tables, not just v7+ tables. Previously the guard only rejected v7+ tables, relying on `can_write_to_protocol` to reject legacy v4–6 tables via implied-feature expansion. Now that legacy versions no longer expand implied features, the guard must cover legacy tables too.

## Safety

The only behavioral change: inactive unsupported features no longer block writes to legacy-protocol tables. Active unsupported features are still rejected by the existing operation-time guards. No silent data corruption.

### DIFF
--- a/crates/sail-delta-lake/src/transaction/mod.rs
+++ b/crates/sail-delta-lake/src/transaction/mod.rs
@@ -880,13 +880,13 @@ fn validate_effective_commit_target(
         return Err(TransactionError::TableFeaturesRequired(TableFeature::DeletionVectors).into());
     }
 
-    // TODO(cdf-writes): Data-changing operations still do not emit AddCDCFile actions. Commit-time
-    // protocol checks currently reject changeDataFeed tables before these writes can land, but once
-    // the feature is enabled here we need an operation-aware validation instead of relying on that.
-    if table_property_enabled(&metadata, "delta.enableChangeDataFeed")
-        && !protocol_supports_legacy_change_data_feed(&protocol)
-        && !protocol_has_writer_feature(&protocol, &TableFeature::ChangeDataFeed)
-    {
+    // TODO(cdf-writes): Data-changing operations still do not emit AddCDCFile actions. Until CDF
+    // write support is implemented, reject all writes to tables with CDF enabled, regardless of
+    // whether the protocol support is legacy (writer v4-6) or explicit (writer v7+ feature).
+    // Previously this guard only rejected v7+ tables, relying on `can_write_to_protocol` to reject
+    // legacy tables via implied-feature expansion. Now that legacy versions no longer expand
+    // implied features (matching delta-spark), this guard must cover legacy tables too.
+    if table_property_enabled(&metadata, "delta.enableChangeDataFeed") {
         return Err(TransactionError::TableFeaturesRequired(TableFeature::ChangeDataFeed).into());
     }
 
--- a/crates/sail-delta-lake/src/transaction/protocol.rs
+++ b/crates/sail-delta-lake/src/transaction/protocol.rs
@@ -27,48 +27,6 @@ use crate::spec::{
 };
 use crate::table::DeltaSnapshot;
 
-static READER_V2: LazyLock<HashSet<TableFeature>> =
-    LazyLock::new(|| HashSet::from_iter([TableFeature::ColumnMapping]));
-static WRITER_V2: LazyLock<HashSet<TableFeature>> =
-    LazyLock::new(|| HashSet::from_iter([TableFeature::AppendOnly, TableFeature::Invariants]));
-static WRITER_V3: LazyLock<HashSet<TableFeature>> = LazyLock::new(|| {
-    HashSet::from_iter([
-        TableFeature::AppendOnly,
-        TableFeature::Invariants,
-        TableFeature::CheckConstraints,
-    ])
-});
-static WRITER_V4: LazyLock<HashSet<TableFeature>> = LazyLock::new(|| {
-    HashSet::from_iter([
-        TableFeature::AppendOnly,
-        TableFeature::Invariants,
-        TableFeature::CheckConstraints,
-        TableFeature::ChangeDataFeed,
-        TableFeature::GeneratedColumns,
-    ])
-});
-static WRITER_V5: LazyLock<HashSet<TableFeature>> = LazyLock::new(|| {
-    HashSet::from_iter([
-        TableFeature::AppendOnly,
-        TableFeature::Invariants,
-        TableFeature::CheckConstraints,
-        TableFeature::ChangeDataFeed,
-        TableFeature::GeneratedColumns,
-        TableFeature::ColumnMapping,
-    ])
-});
-static WRITER_V6: LazyLock<HashSet<TableFeature>> = LazyLock::new(|| {
-    HashSet::from_iter([
-        TableFeature::AppendOnly,
-        TableFeature::Invariants,
-        TableFeature::CheckConstraints,
-        TableFeature::ChangeDataFeed,
-        TableFeature::GeneratedColumns,
-        TableFeature::ColumnMapping,
-        TableFeature::IdentityColumns,
-    ])
-});
-
 pub struct ProtocolChecker {
     reader_features: HashSet<TableFeature>,
     writer_features: HashSet<TableFeature>,
@@ -107,9 +65,13 @@ impl ProtocolChecker {
         &self,
         protocol: &Protocol,
     ) -> Result<Option<HashSet<TableFeature>>, TransactionError> {
+        // Legacy reader versions (1, 2) do not carry explicit feature lists. The version
+        // number implies a feature set (e.g. reader v2 implies columnMapping), but —
+        // matching delta-spark's `protocolCheck` — we do not expand the implied set here.
+        // Whether a legacy-implied feature is actually *active* is validated separately
+        // at operation time (e.g. `verify_column_mapping`).
         match protocol.min_reader_version() {
-            0 | 1 => Ok(None),
-            2 => Ok(Some(READER_V2.clone())),
+            0..=2 => Ok(None),
             3 => Ok(Some(
                 protocol
                     .reader_features()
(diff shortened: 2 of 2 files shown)


lonless9:
LGTM! 🚀 
Once it is marked as ready we can merge this PR!

lonless9 (approved):

PR #2151, opened 2026-06-28
feat: add SHOW FUNCTIONS and DESCRIBE FUNCTION support

lonless9:


crates/sail-plan/data/functions/scalar/array.yaml
```diff
@@ -0,0 +1,535 @@
...
+  signatures:
+    - "array(expr, ...)"
+  usage: "array(expr, ...) - Returns an array with the given elements."
+  examples: |2-
+    
+        Examples:
+          > SELECT array(1, 2, 3);
+           [1,2,3]
+      
+  since: "1.1.0"
```
linhr, line 15: Here are some thoughts on the YAML file design.

1. `scope` and `source` may convey overlapping information and we may want to skip this field by default for built-in functions.
2. `group` can be inferred from the file name.
3. I feel we don't need `class_name` as they are specific to JVM Spark and we don't have a good way to make sure they are up to date.
4. I feel `usage` could be renamed to `description`. The function usage consists of signatures, description, and examples.
5. `arguments` are not specified for all functions. We either treat them as required, or describe the arguments for the signatures in `description`.
6. We don't have to indent `examples` for better display in SQL output. This is a separate concern that we take care of outside of this file. And `examples` should be turned into structured data in follow-up PRs.

In summary, we would like to have consistent, minimum, and structured function metadata. And we don't have to follow exactly the same representation as JVM Spark.

crates/sail-plan/src/function/mod.rs
```diff
@@ -70,3 +96,51 @@ pub(super) fn transform_count_star_wildcard_expr(arguments: Vec<Expr>) -> Vec<Ex
         _ => arguments,
     }
 }
+
+#[cfg(test)]
```
linhr: The unit tests can be removed.

crates/sail-plan/src/function/mod.rs
```diff
@@ -48,6 +52,28 @@ pub fn is_built_in_generator_function(name: &str) -> bool {
     BUILT_IN_GENERATOR_FUNCTIONS.contains_key(name)
 }
 
+fn list_built_in_function_names_raw() -> Vec<&'static str> {
```
linhr: Could we remove `_raw` from the name of this function? Let me know if I missed anything.

crates/sail-plan/src/function/mod.rs
```diff
@@ -48,6 +52,28 @@ pub fn is_built_in_generator_function(name: &str) -> bool {
...
+    let mut names = BUILT_IN_SCALAR_FUNCTIONS
+        .keys()
+        .chain(BUILT_IN_GENERATOR_FUNCTIONS.keys())
+        .chain(BUILT_IN_TABLE_FUNCTIONS.keys())
+        .copied()
+        .collect::<Vec<_>>();
+    names.extend(aggregate::list_built_in_aggregate_function_names());
+    names.extend(window::list_built_in_window_function_names());
```
linhr: Why do we need these two? Could we use `.chain` on the `keys()` directly?

crates/sail-plan/src/function/metadata.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub class_name: &'static str,
+    pub is_public: bool,
+}
+
+include!(concat!(env!("OUT_DIR"), "/function_metadata.rs"));
+
+lazy_static! {
+    static ref BUILT_IN_FUNCTION_METADATA_BY_NAME: HashMap<&'static str, BuiltInFunctionMetadata> =
+        HashMap::from_iter(
+            BUILT_IN_FUNCTION_METADATA
```
linhr, line 26: I wonder if we should directly define `BUILT_IN_FUNCTION_METADATA` as a `phf::Map` via `phf_map!` in the generated code. This could be future work.

crates/sail-plan/src/function/metadata.rs
```diff
@@ -0,0 +1,65 @@
...
+                .iter()
+                .map(|metadata| (metadata.name, *metadata))
+        );
+}
+
+pub(crate) fn built_in_function_status(name: &str) -> FunctionStatus {
+    let Some(metadata) = BUILT_IN_FUNCTION_METADATA_BY_NAME.get(name) else {
+        return FunctionStatus::built_in(name.to_string());
```
linhr: Is there a reason for this name-only default? Shall we treat unknown name as an error?

python/pysail/tests/spark/catalog/test_function.py
```diff
@@ -0,0 +1,143 @@
+"""Tests for Spark Catalog function APIs."""
```
linhr, line 1: Could you add tests around UDFs registered by the user?

crates/sail-plan/build.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[serde(default)]
+    deprecated: Option<String>,
+    #[serde(default)]
+    class_name: Option<String>,
+    #[serde(default)]
+    signatures: Vec<String>,
+}
+
+impl FunctionMetadataEntry {
+    fn validate(&self) -> Result<(), Box<dyn std::error::Error>> {
```
linhr, line 36: Some of the validation logic should just be reflected in more strict `struct FunctionMetadataEntry` definition, such as by removing `Option` and changing string fields to `enum`.

crates/sail-plan/build.rs
```diff
@@ -0,0 +1,236 @@
...
+                .into());
+            }
+            entries.push(entry);
+        }
+    }
+    entries.sort_by(|left, right| left.name.cmp(&right.name));
+    Ok(entries)
+}
+
+fn build_function_metadata(
```
linhr, line 199: We should redo this using the `quote` library. Manipulating strings for code generation is fragile.

linhr (approved):
This looks nice overall!

I have some comments but we can address only small ones for this PR. We can save the design question and bigger changes for the future.

crates/sail-plan/src/function/mod.rs
lonless9: Done!

crates/sail-plan/src/function/mod.rs
lonless9: Done!

crates/sail-plan/src/function/mod.rs
lonless9: Done!

python/pysail/tests/spark/catalog/test_function.py
lonless9: Done!

crates/sail-plan/src/function/metadata.rs
lonless9: Done!

PR #2152, opened 2026-06-28
feat: write one file per output partition to match Spark behaviour

nis12ram:
# Problem

Spark outputs one file per output partition, but Sail does not behave similarly.

Specifically, the Listing formats (Parquet, Avro, CSV, JSON, Arrow, and Text) behave differently and mostly output **4 data files**.

## Reason

In DataFusion, compute parallelism and write parallelism are decoupled:
https://github.com/apache/datafusion/issues/7767

DataFusion uses two configs to determine write parallelism:

* `datafusion.execution.minimum_parallel_output_files` (default: `4`)
* `datafusion.execution.soft_max_rows_per_output_file` (default: `50000000`)

The default value of `minimum_parallel_output_files` (`4`) is the reason why 4 data files are produced in most cases.

# Approach

For Listing formats, these configs are set when `DataSinkExec` is created.

The proposed prototype is to create a wrapper `OutputFileCountExec` around `DataSinkExec` that updates `minimum_parallel_output_files` to match the output partition count.

## Exact Approach

The output partition count is captured at **planning time** from:

```rust
physical_input.output_partitioning().partition_count()
```

It is then injected into the `SessionConfig` at **execution time** by:

1. Cloning the `TaskContext`.
2. Modifying the `minimum_parallel_output_files` config.
3. Delegating execution to the inner plan.

The downstream `row_count_demuxer()` then reads the modified config value and creates the corresponding number of parallel output streams.

## How to Test

A simple way to verify the behavior is:

```python
df =  spark.range(int(1e5))
df.repartition(50).write.format("csv").save(path)
```

-------------------------

 **This PR is raised as a prototype to explore how one file per output partition can be achieved. Test cases and distributed execution support are intentionally skipped.**

 **@linhr, I hope this prototype will be helpful in the process of building a custom `DataSink`.**

 **And sorry if I misunderstood anything or presented any part incorrectly.**

### DIFF
--- a/crates/sail-data-source/src/listing/planner.rs
+++ b/crates/sail-data-source/src/listing/planner.rs
@@ -18,7 +18,7 @@ use datafusion::logical_expr::{Expr, LogicalPlan, TableScan, UserDefinedLogicalN
 use datafusion::physical_expr::create_lex_ordering;
 use datafusion::physical_expr_common::sort_expr::LexOrdering;
 use datafusion::physical_plan::empty::EmptyExec;
-use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::{ExecutionPlan, ExecutionPlanProperties};
 use datafusion::physical_planner::{ExtensionPlanner, PhysicalPlanner};
 use datafusion_common::stats::Precision;
 use datafusion_common::{internal_err, plan_err, project_schema, Statistics};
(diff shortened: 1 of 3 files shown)


linhr:
@nis12ram For better context, we have been investigating our own data sink implementation so that it can better match the Spark behavior and be reusable for all formats including Delta Lake and Iceberg. Since the design is somewhat involved, we may not be able to accept contributions around file readers/writers at this point.

If the file output convention is important for your use cases, feel free to create issues so that we can better prioritize the feature requests.

But also feel free to continue working on this PR if you're interested so that it can be used as a reference. The prototype would be super valuable for reaching the right design decisions.

PR #2161, opened 2026-06-30
fix: reject boolean input in product/kurtosis/skewness to match Spark

davidlghellin:
These float aggregates accepted BOOLEAN input (coercing true/false to 1.0/0.0) where Spark rejects it with DATATYPE_MISMATCH. The root cause was a blind `Cast(arg AS Float64)` in the sail-plan aggregate/window builders that ran before any type check.

- Give ProductFunction/KurtosisFunction/SkewnessFunc a `user_defined` signature + a shared `coerce_single_arg_to_float64` helper that coerces numeric, decimal and numeric-string inputs to DOUBLE (matching Spark's implicit casts) and rejects the rest (BOOLEAN, DATE, ...).
- Drop the blind cast from the aggregate and window builders so coerce_types governs the accepted types.

Tests (values validated against Spark JVM 4.1):
- kurtosis.feature / skewness.feature: numeric, int, string, decimal, null, single/empty -> NULL, group-by, window, zero-variance, two-value, sign, and boolean rejection. Run against both Sail and JVM.
- product.feature: SQL form, tagged @sail-only (Spark has no `product` SQL routine); test_product.py covers the DataFrame `F.product` form on both engines.
- Rust unit test for the coerce helper covering numeric/decimal/string (Utf8/LargeUtf8/Utf8View) acceptance and boolean/arity rejection.

One @sail-bug documents a pre-existing, unrelated scientific-notation double formatting gap (`1e18` vs Spark `1.0E18`); out of scope here.


shehabgamin:
@davidlghellin I feel it's okay to support features that Spark doesn't support.

Coercing true/false to 1.0/0.0 seems helpful imo

davidlghellin:
> @davidlghellin I feel it's okay to support features that Spark doesn't support.
> 
> Coercing true/false to 1.0/0.0 seems helpful imo

Makes sense, thanks! I'll keep the coercion and tag those scenarios as sail-only. I'll get to it tomorrow.

linhr:
> > @davidlghellin I feel it's okay to support features that Spark doesn't support.
> > Coercing true/false to 1.0/0.0 seems helpful imo
> 
> Makes sense, thanks! I'll keep the coercion and tag those scenarios as sail-only. I'll get to it tomorrow.

Allowing cast from `true/false` to `1.0/0.0` is not expanding Spark's behavior, but it's a diverging behavior from Spark's type promotion rule in ANSI mode (<https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html#type-promotion-and-precedence>). So if the PR explicitly fails with an error to detect such behavior (in ANSI mode), that's actually the right thing to do.

PR #2167, opened 2026-07-01
feat: add jdbc write support for postgresql, mysql and sql server

### LINKED ISSUE #2164: Proposal: write support for the JDBC data source
## Problem

The JDBC data source is read-only (connectorX). There is no way to `df.write` from Sail.

## Proposal

Add a `DataSourceArrowWriter` to the same source. PostgreSQL uses Arrow-native ADBC `adbc_ingest`; MySQL and SQL Server use a SQLAlchemy fallback (no ADBC ingest driver exists for them). Supports `append` and `overwrite` (staging-swap).

## Open questions for maintainers

1. **Dependencies** — the write path adds `adbc-driver-postgresql`, `psycopg`, `sqlalchemy`, `pandas` to the `pysail[jdbc]` extra; the MySQL/SQL Server DBAPI drivers (`pymysql` / `pymssql`) are installed by the user. Acceptable?
2. **Overwrite semantics** — PostgreSQL offers `atomic` (staging DROP+RENAME) and `truncate`; other dialects use a `DELETE` + `INSERT … SELECT` swap. Reasonable?
3. **Dialect scope** — PostgreSQL / MySQL / SQL Server to start, more via SQLAlchemy later. OK?

ErikBPF:
## What
Write (sink) support for the `jdbc` data source — same `JdbcDataSource` / `format("jdbc")` as the reader.

## Design
- **PostgreSQL** → ADBC `adbc_ingest` (Arrow-native binary COPY). **MySQL / SQL Server** → SQLAlchemy fallback (no ADBC ingest driver exists for them); rows inserted via SQLAlchemy-core from Arrow so nullable ints / bigints keep integer fidelity (no pandas float coercion).
- Save modes follow Spark's built-in JDBC writer (Spark 4.x docs): `append` is at-least-once; `overwrite` is `atomic` (staging + one-txn swap) or `truncate` (non-atomic). Concurrent overwrites to one table unsupported — documented, matching Spark (its writer takes no cross-partition lock).
- Schema-qualified targets supported; DSN scrubbed from errors; SQL Server URL params (`encrypt`, `applicationIntent`, …) parsed and applied.

## Test coverage
| Area | How verified | Where |
| --- | --- | --- |
| URL/DSN, quoting, injection-escaping, option validation, staging-name isolation, cred scrubbing, SQL Server URL parsing | unit (no DB) | `test_jdbc_unit.py` (24) |
| Reads: partitioned, filter pushdown, unicode, data-types, joins, nulls, injection-safety, schema-qualified | live PostgreSQL (testcontainers) | `test_jdbc.py -m integration` |
| PG writes: append, overwrite atomic/truncate, empty/nulls, concurrent-atomic no-corruption, abort-cleanup, schema-qualified append/atomic/truncate | live PostgreSQL | `test_jdbc.py -m integration` |
| MySQL + SQL Server: append, overwrite, **integer fidelity (NULL + bigint>2^53)** | live `mysql:8.4`, `mssql/server:2022` | `test_jdbc_dialects.py -m integration` |

<details><summary><b>Results</b> — 24 unit + 42 PostgreSQL + 6 dialect = 72 passed (verbatim)</summary>

```
# test_jdbc_unit.py
============================== 24 passed in 0.81s ==============================

# test_jdbc.py -m integration  (postgres:16-alpine) — write path shown
test_write_append_basic[wt_append_basic] PASSED
test_write_overwrite_replaces[wt_overwrite] PASSED
test_write_empty_df[wt_empty_df] PASSED
test_write_null_values[wt_nulls] PASSED
test_write_append_twice[wt_append_twice] PASSED
test_write_overwrite_atomic_replaces[wt_atomic_overwrite] PASSED
test_write_overwrite_truncate_replaces[wt_truncate_overwrite] PASSED
test_write_atomic_empty_df[wt_atomic_empty] PASSED
test_write_atomic_null_values[wt_atomic_nulls] PASSED
test_concurrent_atomic_overwrites_no_corruption[wt_concurrent_atomic] PASSED
test_write_atomic_abort_cleanup[wt_abort_atomic] PASSED
test_write_append_schema_qualified PASSED
test_write_overwrite_atomic_schema_qualified PASSED
test_write_overwrite_truncate_schema_qualified PASSED
============================== 42 passed in 9.83s ==============================

# test_jdbc_dialects.py -m integration  (mysql:8.4, mssql/server:2022)
test_mysql_write_append PASSED
test_mysql_write_overwrite PASSED
test_mysql_integer_fidelity PASSED
test_mssql_integer_fidelity PASSED
test_mssql_write_append PASSED
test_mssql_write_overwrite PASSED
======================== 6 passed in 35.40s =========================
```
CI: Lint green. `hatch fmt --check` clean.
</details>

## Notes
`atomic` overwrite replaces the table object (grants/RLS not preserved — use `truncate`). Target table must exist. Install: `pip install pysail[jdbc]`; MySQL/SQL Server writes also need `pymysql` / `pymssql`.

Closes #2164.


shehabgamin:
@ErikBPF Thanks for this contribution! Please see Copilot's review and feel free to tag me and when you'd like me to take another look.

ErikBPF:
Thanks @shehabgamin! I've addressed all three of Copilot's points:

- **`batchsize` validation** — now checked as a positive integer on the driver (before fan-out) and in each write engine's constructor. A `0` crashed at executor runtime (`range` step 0) and a negative value silently skipped every insert while still reporting rows written.
- **PostgreSQL `batchsize`** — the ADBC path now slices the Arrow table into `batchsize` chunks within a single transaction (via a small helper shared with the SQLAlchemy path). Previously it ingested the whole partition in one call, so the option had no effect.
- **Staging retry-safety** — the SQLAlchemy (MySQL/SQL Server) path now drops the staging table (`DROP TABLE IF EXISTS`) before creating it, so a retried task no longer fails on a leftover from a prior attempt and can't carry partial rows forward. (The PostgreSQL path already used `CREATE TABLE IF NOT EXISTS`; this brings the two paths in line.)

One related note while I was in there: the PostgreSQL `atomic` overwrite is **at-least-once** under Spark task retry / speculative execution — a re-run partition re-ingests into the shared staging, so the swapped-in table can gain duplicates (same as `append`, and as Spark's own JDBC writer). I documented this rather than trading away the zero-copy `RENAME` swap; disabling speculation gives exactly-once overwrite. Happy to implement per-partition staging instead if you'd prefer exactly-once by default.

Tests added: `batchsize <= 0` rejected on the driver and in both engines, and sqlite staging create is idempotent (unit); small-batchsize overwrite writes all rows (Postgres integration). Full jdbc suite green on my fork — 33 unit/meta + 49 integration (Postgres + MySQL 8.4 + SQL Server 2022).

Ready for another look whenever you have a moment. Happy to split anything out if that's easier to review.

scripts/spark-tests/check-jdbc-matrix.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
shehabgamin: I'm not sure we want to commit this file. What's the motivation behind keeping it long term?

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +293,597 @@ def _build_where(filters: list[str]) -> str:
...
+    """
+
+    def __init__(self, engine) -> None:
+        self._engine = engine
+
+    def write(self, iterator: Iterator[pa.RecordBatch]) -> WriterCommitMessage:
+        from pyspark import TaskContext  # noqa: PLC0415
+
+        ctx = TaskContext.get()
+        pid = ctx.partitionId() if ctx is not None else 0
```
shehabgamin, line 1491: I'm not sure if this will work, `TaskContext` does not seem to be initialized in:
https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/formats/python/executor.rs

Can we test and see if every partition gets `pid=0`? If so, we'd need to make changes to `executor.rs`

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +293,597 @@ def _build_where(filters: list[str]) -> str:
...
+    # Private per-mode write helpers (executor side)
+    # ------------------------------------------------------------------
+
+    def _prepare_atomic(self, qstaging: str, qtarget: str) -> None:
+        """Create the staging table (matching the target) if it does not exist."""
+        import psycopg  # noqa: PLC0415
+
+        try:
+            with psycopg.connect(self.dsn, autocommit=True) as conn, conn.cursor() as cur:
+                cur.execute(f"CREATE TABLE IF NOT EXISTS {qstaging} (LIKE {qtarget} INCLUDING ALL)")
```
shehabgamin, line 521: Would this not work with tables that have a serial column?

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +293,597 @@ def _build_where(filters: list[str]) -> str:
...
+    return url, connect_args
+
+
+def _sqlalchemy_url(dsn: str) -> tuple[str, dict]:
+    """Translate a stripped JDBC DSN into a ``(SQLAlchemy URL, connect_args)`` pair.
+
+    ``connect_args`` is empty for MySQL and carries the mapped SQL Server
+    connection options (see :func:`_parse_sqlserver_url`).
+    """
+    scheme, sep, rest = dsn.partition("://")
```
shehabgamin, line 833: What if we have something like ` jdbc:mysql://host/db`?

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +293,597 @@ def _build_where(filters: list[str]) -> str:
...
+        instance = instance or None
+
+    return userinfo, hostport, instance, port
+
+
+def _parse_sqlserver_url(rest: str) -> tuple[str, dict[str, object]]:
+    """Parse the JDBC SQL Server tail (after ``sqlserver://``) into a
+    ``(sqlalchemy_url, connect_args)`` pair for the ``mssql+pymssql`` dialect.
+
+    Form: ``[user:pass@]host[\\instance][:port][;key=value[;...]]``. Known params:
```
shehabgamin, line 754: What about forms such as ` jdbc:sqlserver://host;databaseName=db;user=alice;password=s3cret`?

shehabgamin (commented):
First round of questions / comments :)

python/pysail/tests/spark/datasource/test_version_gating.py
linhr, line 1: Remove this.

scripts/spark-tests/check-jdbc-matrix.sh
linhr, line 1: Remove this.

ErikBPF:
Thanks @shehabgamin, @linhr — all points addressed. Still green: 32 unit + 51 integration
(Postgres / MySQL / SQL Server), lint clean, and the datasource modules collect cleanly
(skip, not error) on Spark 3.5.7 / 4.0.1.

**@linhr** — removed both `scripts/spark-tests/check-jdbc-matrix.sh` and
`test_version_gating.py`. The version-gate regression they guarded is still covered: each
module self-gates on import, and CI already runs the full Spark matrix.

**@shehabgamin:**

- **pid=0 (`jdbc.py:833`)** — confirmed: Sail doesn't populate `TaskContext`, so
  `partitionId()` is 0 for every partition. Only the SQLAlchemy overwrite path relied on it
  (per-partition staging names collided → silent data loss on multi-partition overwrite);
  staging names are now unique per call. The PostgreSQL path uses per-*run* staging and is
  unaffected. No `executor.rs` change needed; added a regression test.
- **serial columns (`jdbc.py:416`)** — confirmed a hard failure: `CREATE TABLE staging
  (LIKE target INCLUDING ALL)` copies the serial default, so `commit()`'s `DROP TABLE
  target` is refused (staging depends on the target's owned sequence). The atomic swap now
  detaches owned sequences → drop → rename → re-owns them and `setval`s past the loaded max.
  Verified on Postgres 16; integration test added.
- **SQL Server `;user=;password=` (`jdbc.py:590`)** — those were dropped as unknown params;
  now mapped into the URL userinfo (percent-encoded; credentials in the authority still win).
- **`jdbc:mysql://host/db` (`jdbc.py:633`)** — already parses (user/port optional); added
  tests to keep it that way.

Also closed two older Copilot items: truncate mode now runs lock + TRUNCATE + sentinel in a
single transaction with `pg_advisory_xact_lock` (no partial-commit window), and credentials
embedded in a URL no longer leak into the "Invalid JDBC URL" error messages.

On where this sits today and where it can go: the PostgreSQL path is deliberately
native-optimized — ADBC binary `COPY` for ingest, a zero-copy `RENAME` swap for atomic
overwrite, advisory locks for truncate. MySQL and SQL Server currently go through SQLAlchemy
generically — row-wise `INSERT` and a `DELETE` + `INSERT … SELECT` staging swap — which is
correct but leaves throughput and atomicity on the table; each could move to its own native
bulk path. Plus two JVM-Spark parity gaps I kept out of this PR to stay scoped: no table
auto-creation (now a clear fail-fast error rather than a raw one) and a custom `overwriteMode`
option instead of Spark's `truncate` boolean.

If useful, happy to open an `[EPIC] JDBC Write Sink Feature Parity` issue to track these as a
roadmap — per-sink native optimization (MySQL bulk load, SQL Server bcp), auto-create, option
parity, and future sinks like Oracle. Just say the word.

ErikBPF:
@shehabgamin @linhr — while checking this against Spark 4.1.2's JDBC writer,
I found several observable parity gaps and prepared a stacked implementation for
reviewer decision.

The stacked draft now handles the full set together:

- preserves `errorIfExists`, `ignore`, `append`, and `overwrite` through Sail's
  Python write planner;
- moves JDBC DDL from physical planning into an execution-time preparation hook;
- makes `ignore` a true no-op without evaluating the input;
- creates missing targets from the DataFrame schema;
- makes default overwrite drop/recreate from the new schema;
- supports Spark's `truncate=true` behavior on MySQL and SQL Server while
  retaining PostgreSQL's Spark behavior (drop/recreate);
- uses strict per-dialect Spark type mappings and existing-column resolution;
- keeps `sail.jdbc.overwriteMode=atomic|truncate` as opt-in Sail extensions.

https://github.com/ErikBPF/sail/pull/1

I kept this separate from #2167 so you can decide whether it belongs in the
original PR. It deliberately excludes the unrelated `pyodbc` optimization from
an earlier experiment.

The planner forwards Sail's complete logical `SinkMode` internally because the
Python API itself only passes an `overwrite` boolean. A generic execution hook
then handles table existence/DDL once before any input is evaluated.

Performance-only differences such as a larger default `batchsize` are not parity
problems. The question is whether to merge #2167 as explicit append/overwrite
support with documented limitations, or expand planning/overwrite semantics before
calling JDBC writes drop-in compatible.

Local verification: affected Rust crates compile; JDBC unit tests pass on Spark
4.0.1 and 4.1.1; Ruff passes; all 74 PostgreSQL/MySQL/SQL Server integration
tests pass, including native Spark 4.1.2 differential coverage. A dedicated CI
job now runs that live-dialect stack.

erikbogado-nstech:
@shehabgamin @linhr Gentle ping — no rush, just checking in on next steps.

Where things stand:

- All points from the last review round are addressed (details in my July 9/14 comments); CI is fully green and the branch merges cleanly.
- The four open threads on `jdbc.py` (TaskContext `pid`, serial columns, the MySQL and SQL Server URL forms) each have a fix plus a regression test — happy to also reply inline on each thread if that makes tracking easier.
- The Spark-parity follow-up is staged separately in ErikBPF/sail#1, pending your call on whether it belongs in this PR or as a follow-up.

Happy to rebase, split the PR, or adjust scope — whatever makes review easiest. Just let me know how I can help move this forward.

linhr:
@ErikBPF @erikbogado-nstech Unfortunately, we don't have enough bandwidth recently to carefully review this PR. The code involves topics that we are not familiar with, especially the nuances around different database clients and potential security/performance concerns around bulk ingestion, so we must do a careful manual review before deciding to own this piece of the code within our codebase.

I noticed all your comments in this PR are AI-generated, so it's not yet clear to us the effort you put into supervising the AI coding agent for the design and implementation of the code. This is another reason why this PR would require a lot more scrutiny.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+    """Split a possibly schema-qualified name into ``(schema_or_None, table)``.
+
+    ADBC ingest and SQLAlchemy take the schema as a separate argument, so the parts
+    must stay apart. ``"public.orders"`` -> ``("public", "orders")``.
+    """
+    schema, sep, table = qualified.rpartition(".")
+    return (schema, table) if sep else (None, qualified)
+
+
+def _staging_name_atomic(dbtable: str, run_id: str) -> str:
```
shehabgamin, line 401: Long PostgreSQL table names can cause overwrite cleanup to drop the real target. The generated names append suffixes without respecting PostgreSQL’s 63-byte identifier limit.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+    """Per-run atomic staging table name, in the target's schema so the RENAME keeps
+    it there. The run-id suffix isolates concurrent writers.
+    """
+    schema, table = _split_schema(dbtable)
+    staging = f"{table}{_STAGING_PREFIX}{run_id}"
+    return f"{schema}.{staging}" if schema else staging
+
+
+def _staging_name_truncate_sentinel(dbtable: str, run_id: str) -> str:
```
shehabgamin, line 410: Long PostgreSQL table names can cause overwrite cleanup to drop the real target. The generated names append suffixes without respecting PostgreSQL’s 63-byte identifier limit.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+        ``truncate`` drops the sentinel (rows are already in the target).
+        """
+        total = sum(r.rows_written for r in results)
+
+        if self.overwrite_mode == "append":
+            return total
+
+        import psycopg  # noqa: PLC0415
+
+        if self.overwrite_mode == "atomic":
```
shehabgamin, line 654: As a consequence of:
- https://github.com/lakehq/sail/pull/2167/changes#r3626919110
- https://github.com/lakehq/sail/pull/2167/changes#r3626919110

This can execute DROP TABLE on the real table. Longer-but-still-valid names also truncate away the run token, allowing concurrent jobs to share staging. PostgreSQL documents this truncation in its identifier rules: https://www.postgresql.org/docs/current/sql-syntax-lexical.html

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+            # rollback): TRUNCATE and the sentinel insert must commit together, else a crash
+            # between them leaves the target truncated but unmarked, and another partition
+            # would TRUNCATE again after rows were ingested — wiping them.
+            with psycopg.connect(self.dsn) as conn, conn.cursor() as cur:
+                cur.execute(f"SELECT pg_advisory_xact_lock({lock_key})")
+                cur.execute(f"CREATE TABLE IF NOT EXISTS {qsentinel} (done BOOLEAN)")
+                cur.execute(f"SELECT COUNT(*) FROM {qsentinel}")  # noqa: S608
+                row = cur.fetchone()
+                if row[0] == 0:  # type: ignore[index]
+                    cur.execute(f"TRUNCATE {qtarget}")
```
shehabgamin: This issues plain TRUNCATE target, which includes inheritance descendants unless `ONLY` is specified. Spark deliberately emits `TRUNCATE TABLE ONLY`:
https://github.com/apache/spark/blob/e6d66abda0ac420f9f8448b6b10e6efbb3c67fbd/sql/core/src/main/scala/org/apache/spark/sql/jdbc/PostgresDialect.scala#L185-L192

python/pysail/spark/datasource/jdbc.py
```diff
@@ -462,6 +1176,96 @@ def reader(self, schema: pa.Schema) -> JdbcDataSourceReader:  # noqa: ARG002
...
+        batch_size = int(opts.get("batchsize", "65536"))
+        if batch_size <= 0:
+            msg = f"Option 'batchsize' must be a positive integer, got {batch_size}."
+            raise ValueError(msg)
+
+        import uuid  # noqa: PLC0415
+
+        run_id = uuid.uuid4().hex[:12]
+
+        if subprotocol == "postgresql":
```
shehabgamin, line 2095: Sail passes only an `overwrite` boolean:
https://github.com/lakehq/sail/blob/main/crates/sail-data-source/src/formats/python/table_format.rs#L325

This PR then turns false into append...

Spark handles these modes distinctly in:
https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala#L64-L89

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+    def _qualified(prep, schema: str | None, name: str) -> str:
+        return f"{prep.quote(schema)}.{prep.quote(name)}" if schema else prep.quote(name)
+
+    def _reflect_table(self, engine, table_name: str):
+        """Reflect *table_name* from the live database into an ``sa.Table``."""
+        import sqlalchemy as sa  # noqa: PLC0415
+
+        return sa.Table(table_name, sa.MetaData(), schema=self.schema, autoload_with=engine)
+
+    def _insert_arrow(self, engine, sa_table, table_obj: pa.Table) -> None:
```
shehabgamin: Does SQLAlchemy ignore unknown keys? If yes, then SQLAlchemy would silently discard unmatched DataFrame columns.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+        """
+        import sqlalchemy as sa  # noqa: PLC0415
+
+        if table_obj.num_rows == 0:
+            return
+        with engine.begin() as conn:
+            for chunk in _iter_arrow_chunks(table_obj, self.batch_size):
+                conn.execute(sa.insert(sa_table), chunk.to_pylist())
+
+    def _create_staging_like_target(self, engine, staging: str):
```
shehabgamin: This retains only type, nullability, and primary-key status, dropping defaults, computed columns, identity, autoincrement metadata, etc...

python/pysail/spark/datasource/jdbc.py
```diff
@@ -462,6 +1176,96 @@ def reader(self, schema: pa.Schema) -> JdbcDataSourceReader:  # noqa: ARG002
...
+        if subprotocol == "postgresql":
+            if overwrite:
+                overwrite_mode = opts.get("overwriteMode", "atomic")
+                valid = {"atomic", "truncate"}
+                if overwrite_mode not in valid:
+                    msg = f"Invalid overwriteMode {overwrite_mode!r}. Valid values: {sorted(valid)}"
+                    raise ValueError(msg)
+            else:
+                overwrite_mode = "append"
+            if not _pg_table_exists(conn_str, dbtable):
```
shehabgamin: The table-creation and overwrite contract seems to be opposite to Spark’s:
- https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala#L27C1-L100C1

- https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L856-L911

- https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JDBCOptions.scala#L171

Spark creates a missing target from the DataFrame schema. For an existing target, Spark’s default overwrite drops and recreates it from the new schema; this implementation retains/clones the old schema and ignores Spark’s standard truncate=true option in favor of overwriteMode. Schema-evolution writes therefore fail or produce a different table.

python/pysail/spark/datasource/jdbc.py
shehabgamin: `LIKE ... INCLUDING AL`L creates a fresh sequence for identity columns, but copying explicit IDs into staging does not advance it. `_owned_sequences()` only handles `deptype = 'a'` serial sequences, so after the swap the identity sequence can still start at 1 and collide with the loaded rows. Please re-synchronize identity sequences too and add coverage for both `GENERATED ALWAYS` and `BY DEFAULT`.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+    netloc = f"{userinfo}{host_segment}"
+    if port is not None:
+        netloc += f":{port}"
+
+    url = f"mssql+pymssql://{netloc}/{database}" if database else f"mssql+pymssql://{netloc}"
+
+    connect_args: dict[str, object] = {}
+    encrypt = params.get("encrypt")
+    if encrypt is not None:
+        connect_args["encryption"] = "require" if encrypt.lower() == "true" else "off"
```
shehabgamin: Valid `encrypt=strict` silently disables encryption. This maps only `"true"` to `"require"` and every other value, including JDBC's `"strict"` and malformed values to `"off"`.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+        except Exception as e:
+            safe_msg = _safe_error(e, self.dsn)
+            msg = f"Truncate-mode advisory-lock failed for {self.dbtable!r}: {safe_msg}"
+            raise RuntimeError(msg) from e
+
+    # ------------------------------------------------------------------
+    # Executor side
+    # ------------------------------------------------------------------
+
+    def write_partition(self, partition_id: int, batches: Iterable[pa.RecordBatch]) -> PartitionResult:
```
shehabgamin, line 574: `batchsize` does not bound memory, and `numPartitions` is ignored.

See Spark's for full parity:
https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala#L596

erikbogado-nstech:
Hi, @linhr! 
First of all, thanks for the response. I am a non-english native person, so I usually prefer to pass my comments on a AI to make them more coherent, but I can see how this causes a signaling problem. That said, most of the code here in this PR is produced by AI, but this is not necessarily a bad thing. On doing this I made a lot of research on current spark JDBC implementation and did my best to copy behavior ipsis litteris. I plan and AI write. I do a lot of planning docs and extra test scripts (like the ones you asked me to remove) to validate, but I can understand that even them are not enough sometimes. If you want I can break this pr into multiple small ones with thin value slices.

Ive been using Sail on postgres>Icerberg and Icerberg>icerberg flows and Im completly amazed with the performance.  I have a lot of contribution desires related do jdbc sinks, polaris integration and full iceberg support. Can even help with activities outside this scope. Please just tell me how I can be of service :)

python/pysail/spark/datasource/jdbc.py
shehabgamin: Also there could be PostgreSQL staging creation races across normal partitions given that every partition concurrently runs the same `CREATE TABLE IF NOT EXISTS`.

The Existing concurrency tests use different run IDs and do not exercise this path.

python/pysail/spark/datasource/jdbc.py
```diff
@@ -255,6 +300,674 @@ def _build_where(filters: list[str]) -> str:
...
+        collected = [b for b in batches if b.num_rows > 0]
+        table_obj = pa.Table.from_batches(collected) if collected else None
+
+        staging: str | None = None
+        rows = 0
+        engine = self._create_engine()
+        try:
+            if table_obj is not None and table_obj.num_rows > 0:
+                if self.overwrite:
```
shehabgamin: A random name is created here, , but that name reaches driver-side abort only after a successful task result. Failure after table creation leaves an undiscoverable staging table. Cleanup should happen in the task exception path or use driver-derivable name

shehabgamin (changes requested):
Next round of review.

I'm going to put the PR into draft mode. Can you please thoroughly go through Spark's JDBC code, and then implement changes accordingly:
https://github.com/apache/spark/tree/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc

Thanks!

linhr:
> Hi, @linhr! First of all, thanks for the response. I am a non-english native person, so I usually prefer to pass my comments on a AI to make them more coherent, but I can see how this causes a signaling problem. That said, most of the code here in this PR is produced by AI, but this is not necessarily a bad thing. On doing this I made a lot of research on current spark JDBC implementation and did my best to copy behavior ipsis litteris. I plan and AI write. I do a lot of planning docs and extra test scripts (like the ones you asked me to remove) to validate, but I can understand that even them are not enough sometimes. If you want I can break this pr into multiple small ones with thin value slices.
> 
> Ive been using Sail on postgres>Icerberg and Icerberg>icerberg flows and Im completly amazed with the performance. I have a lot of contribution desires related do jdbc sinks, polaris integration and full iceberg support. Can even help with activities outside this scope. Please just tell me how I can be of service :)

Thanks for your response. The clarity helps a lot.

To be honest, nearly all the code by Sail maintainers is written by AI nowadays, but I can see that it actually poses a challenge for external contributors. We have no visibility into how the code was delivered, as we have no access to the session histories of the agent. So we have to reverse engineer the design from the code which causes delay in review and may sometimes actually take longer than working with coding agents ourselves. Your use cases are valuable, and this is the kind of the context we're looking for.

ErikBPF:
Hi, @shehabgamin and @linhr!

I changed the PR as requested and added differential tests to adhere more
closely to the behavior of both Spark and pgJDBC. I tried to keep the scope
narrow (and kind of failed), so I moved some Iceberg changes to a future
contribution. All changes were reviewed adversarially by multiple agents and
run on a separate Kubernetes cluster for testing. Here is an AI-generated
summary of the changes:

### What changed

- Preserved all four save modes through the Python data-source planner:
  error-if-exists, `ignore`, `append`, and `overwrite`. The public
  PySpark Connect spelling for the first mode is `mode("error")`; omitting
  `.mode(...)` has the same default behavior.
- Moved existence checks and JDBC DDL from physical planning to a one-time
  execution preparation step. In particular, `ignore` now returns without
  evaluating the input.
- Matched Spark's existing/missing-table flow:
  - missing targets are created from the incoming schema for every save mode;
  - default overwrite drops and recreates the table from the incoming schema;
  - `truncate=true` uses each Spark dialect's supported truncate behavior;
  - append/truncate resolve incoming columns by name against the target schema
    while preserving DataFrame column order.
- Matched Spark's per-dialect create mappings and reject unsupported automatic
  types before destructive DDL.
- Matched `batchsize=1000`, positive-value validation, and write-side
  `numPartitions` coalescing.
- Added `createTableColumnTypes`, `createTableOptions`, `tableComment`,
  `isolationLevel`, `queryTimeout`, and native driver declarations for the
  supported backends.
- Kept whole-job atomic/truncate replacement as explicit PostgreSQL-only Sail
  extensions under `sail.jdbc.overwriteMode`; these are not presented as Spark
  behavior.
- Mapped PostgreSQL connection properties only where pgJDBC and libpq have
  equivalent semantics: `ApplicationName`, `connectTimeout`, `sslmode`,
  `sslcert`, `sslkey`, `sslrootcert`, and `options`. URL values retain
  precedence. Custom JVM drivers, Kerberos/JDBC connection providers, and
  JVM-specific properties remain explicit boundaries.

### Compatibility references

Spark 4.1.2 runtime behavior is the oracle, with source pinned to the commit
used during review rather than relying only on moving `latest` documentation:

- [Spark JDBC options documentation](https://spark.apache.org/docs/latest/sql-data-sources-jdbc.html)
- [`JdbcRelationProvider` save-mode flow](https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcRelationProvider.scala)
- [`JdbcUtils` schema resolution, batching, partitions, transactions, and statement options](https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/jdbc/JdbcUtils.scala)
- Spark dialect implementations:
  [PostgreSQL](https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/jdbc/PostgresDialect.scala),
  [MySQL](https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/jdbc/MySQLDialect.scala), and
  [SQL Server](https://github.com/apache/spark/blob/dccff5579c4e2f200dc4e7f1bb5d41ec1edccce0/sql/core/src/main/scala/org/apache/spark/sql/jdbc/MsSqlServerDialect.scala)
- PostgreSQL particulars:
  [identifier limits](https://www.postgresql.org/docs/current/sql-syntax-lexical.html),
  [`TRUNCATE` locking, `ONLY`, `CASCADE`, and transaction behavior](https://www.postgresql.org/docs/current/sql-truncate.html), and
  [identity columns](https://www.postgresql.org/docs/current/ddl-identity-columns.html)
- [pgJDBC connection properties](https://jdbc.postgresql.org/documentation/use/)
- [MySQL identifier limits](https://dev.mysql.com/doc/refman/8.4/en/identifier-length.html)
- SQL Server JDBC
  [connection properties](https://learn.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?view=sql-server-ver17) and
  [TLS behavior](https://learn.microsoft.com/en-us/sql/connect/jdbc/connecting-with-ssl-encryption?view=sql-server-ver17)

One useful finding came from testing source/runtime behavior rather than prose:
Spark's JDBC guide says PostgreSQL does not support the truncate optimization,
but Spark 4.1.2's `PostgresDialect` and live execution take that path and emit
`TRUNCATE TABLE ONLY` (optionally `CASCADE`). Sail now follows the observed
Spark 4.1.2 behavior, including inheritance and foreign-key cases.

### Direct answers to the open review threads

- **PostgreSQL 63-byte identifiers / unsafe cleanup:** fixed. Generated staging
  and sentinel identifiers are limited by UTF-8 byte length, keep the run token,
  and include a hash when the target prefix must be shortened. Cleanup derives
  the same bounded name; it cannot truncate back to the real target. Tests cover
  63-byte ASCII names, multibyte names, distinct run IDs, live writes, and
  cleanup.
- **`TRUNCATE` and inherited tables:** fixed. PostgreSQL paths emit
  `TRUNCATE TABLE ONLY`, with `CASCADE` only when requested, matching
  `PostgresDialect`.
- **Save modes:** fixed in the shared planner path. Sail now carries the actual
  `SinkMode` instead of reducing it to an `overwrite` boolean. Differential
  tests cover all four semantics for existing and missing tables, including
  implicit default mode and explicit `mode("error")`. PySpark Connect 4.1.1
  rejects the classic-client alias `mode("errorifexists")` locally while
  serializing the request, before Sail can receive it; that alias remains a
  client-side boundary rather than a JDBC implementation gap.
- **Unknown/reordered DataFrame columns:** SQLAlchemy does not get to decide
  this implicitly. Sail resolves incoming names against inspected target names
  first, rejects missing/extra/duplicate/ambiguous columns, and preserves the
  DataFrame's order. Case-sensitive and insensitive cases are covered.
- **Cloned defaults/identity/computed metadata:** Spark's default overwrite does
  not clone the old table; it drops and recreates from the incoming schema.
  Sail now does the same. Metadata-preserving atomic replacement remains an
  explicit Sail extension rather than the default Spark path.
- **Missing-table creation and schema-changing overwrite:** fixed to follow
  `JdbcRelationProvider`/`JdbcUtils`: create from the DataFrame schema when
  absent; drop and recreate from the new schema for default overwrite.
- **PostgreSQL serial and identity sequences:** the opt-in atomic path now
  handles owned serial sequences and internal identity sequences, then advances
  each sequence past the loaded maximum. Live tests cover `SERIAL`,
  `GENERATED ALWAYS`, and `GENERATED BY DEFAULT`.
- **SQL Server `encrypt=strict`:** it is rejected rather than silently weakened
  to encryption off. The pymssql/FreeTDS backend accepts only JDBC values whose
  semantics it can preserve (`true` and `false`); unsupported TLS properties
  and unknown driver properties fail explicitly.
- **`batchsize` / memory:** partitions are consumed batch-by-batch and sliced
  into zero-copy chunks no larger than `batchsize`; the implementation no
  longer materializes a whole partition as one Arrow table or pandas frame.
- **`numPartitions`:** the physical planner now caps write partitions with
  narrow coalescing and never increases partition count. This was compiled into
  a fresh native extension and exercised against PostgreSQL.
- **Staging failure cleanup:** task-side exceptions clean the deterministic
  per-run staging name immediately, and driver-side abort derives and removes
  the same name even when no task result returned. Failed cleanup can still
  leave an identifiable `__sail_stg_` table, which is safe to remove.
- **Concurrency boundary:** normal Spark-compatible overwrite no longer uses
  the staging algorithm. The staging/rename algorithm is an opt-in Sail
  extension; concurrent overwrites of the same target remain unsupported and
  are documented as such.

### Tests added

- Unit/regression coverage for:
  - save-mode preparation and unchanged-state failure paths;
  - no input evaluation for `ignore`;
  - existing/missing targets and schema-changing overwrite;
  - reordered, case-varied, missing, extra, and duplicate columns;
  - strict dialect type mappings and unsupported-type failure before DDL;
  - positive `batchsize`, exact chunk/tail behavior, and `numPartitions`;
  - empty partitions, per-partition transactions, rollback, committed-task
    retry, cleanup, and at-least-once behavior;
  - PostgreSQL 63-byte UTF-8 identifier limits and collision-safe staging names;
  - PostgreSQL `TRUNCATE ONLY`, inheritance, FK cascade, indexes, serial, and
    identity behavior;
  - MySQL/SQL Server truncate, identity/autoincrement, timeout, isolation, and
    connection-property particulars;
  - option case folding, precedence, credential redaction, and PostgreSQL
    connection-property mapping.
- Native Spark differential coverage for PostgreSQL, MySQL, and SQL Server:
  save modes, schema state, rows, column resolution, create types and metadata,
  truncate behavior, identity state, partition failure state, and supported
  write options.
- Full automatic-create type matrices now evaluate identical Spark SQL
  expressions on both engines. The isolated native oracle pins the Spark
  session and JVM process to UTC so timestamp comparisons do not inherit the
  machine timezone.
- Live PostgreSQL differentials for `SERIALIZABLE`, server-side timeout
  cancellation, and `ApplicationName` propagation.
- A dedicated JDBC integration workflow so the live dialect matrix is not only
  local coverage.

Latest full validation on the current source: 83 JDBC unit tests passed; the
complete live PostgreSQL/MySQL/SQL Server suite passed 115 tests in 303.61 s
with one CI-only oracle sentinel skipped locally, and that sentinel passed
separately with the CI guard enabled. The `sail-data-source` suite passed all
62 tests. Ruff, formatting, strict Clippy, and native-extension loading also
passed.

### Locally observed performance

As a bounded end-to-end check, native Spark 4.1.2 and Sail wrote through the
same network service to PostgreSQL 16. Both used four input partitions and
`batchsize=1000`; the container was limited to 1 CPU and 2 GiB. Two passes:

| Case | Native Spark 4.1.2 | Sail |
| --- | ---: | ---: |
| Create 10,000 rows | 3.507–3.707 s | 0.300–0.302 s |
| Append 10,000 rows | 0.550–0.571 s | 0.279–0.296 s |
| Create 100,000 rows | 1.399–1.630 s | 0.563–0.590 s |

Every run verified final row count and checksum. Sail was faster in this small
setup, particularly on PostgreSQL's ADBC/COPY path. These numbers are
diagnostic, not a general performance claim: both engines ran locally in one
container, native create includes JVM first-job cost, storage/network topology
was fixed, and two passes are not statistically strong enough for a release
benchmark.

The remaining intentional compatibility boundaries are custom JVM JDBC
drivers, Kerberos/JDBC connection providers, pgJDBC properties without an exact
native-client equivalent, and arbitrary destructive write-side `dbtable` SQL
expressions.

linhr:
Thanks for the updates. But unfortunately the most recent commits seem to expand the scope even more, with changes to CI workflows and Python data sources which may not be desirable. Given that the PR is around 6,000 lines, I'm afraid we're not able to accept it at the moment.

I appreciate your testing efforts and AI-assisted adversarial review rounds. But I'd prefer an approach where human is involved in the design early and the functionalities are implemented and reviewed progressively. The tests have to be designed carefully as well. With the amount of test code in the PR we are not able to assess the regular burden it may have in CI once the code is merged.

This does not mean that the code is bad. It's just an unfortunate reality that we do not have the bandwidth to review it and own it within our codebase. I feel the exploration is highly valuable and can serve as a good proof-of-concept once we get a chance to revisit this feature in the future.

ErikBPF:
Thanks for the feedback, @linhr. I understand and respect the decision not to accept #2167 in its current form. The PR grew to around 6,000 changed lines and expanded into generic Python data-source and CI changes, making it too large to review and own.

I will stop expanding #2167 and keep it as a proof of concept and reference. I will split the work into small, independently reviewable PRs, starting with [#2376](https://github.com/lakehq/sail/pull/2376). That PR is limited to explicit PostgreSQL append into an existing table. It excludes table creation, other save modes, other dialects, tuning options, and new CI workflows.

I will wait for feedback on each slice before preparing the next one.

ErikBPF:
Closing as discussed. The useful parts landed as smaller PRs: #2435 (identifiers), #2515 (query normalization), #2505 (CASE coercion). The Python datasource work isn't planned for now.

PR #2174, opened 2026-07-01
perf: various funcs StringBuilder byte-capacity 

davidlghellin:
No magic numbers — a fabricated `len * N` byte hint is noise; use 0 (unknown) or the input's get_buffer_memory_size() (correlated).


shehabgamin (approved):
Nice!!

PR #2175, opened 2026-07-01
perf: use arrow `binary` kernel for scalar-unit `time_diff`

davidlghellin:


davidlghellin:
Reverting in bc8bba9 to split this into its own PR — the change applies elsewhere too, so I'll
move it into a shared helper and land it standalone rather than dirty this branch's history.

shehabgamin (approved):
Nice!

PR #2181, opened 2026-07-01
feat: support Delta format MoR merge update operation

ctydw123:
## Overview

When the table uses delta format with deletion vector enabled, the current merge operation doesn't support update clause in either matched and unmatched case. This is pretty impactful, as merge update is commonly used feature with MoR. To be more specific, the query throws unsupported exception today,

```sql
MERGE INTO users_table AS target
USING incoming_batch AS source
ON target.id = source.id
WHEN MATCHED THEN
  UPDATE SET *
WHEN NOT MATCHED THEN
  INSERT (id, name) VALUES (source.id, source.name)
WHEN NOT MATCHED BY SOURCE THEN
  UPDATE SET target.name = 'Unknown'
```

##  Quick smoke test

```sql
CREATE TABLE IF NOT EXISTS users_table (
  id INT,
  name STRING
) 
USING delta
LOCATION '{path}'
TBLPROPERTIES ('delta.enableDeletionVectors' = 'true')

INSERT INTO users_table VALUES 
    (1, 'David'),
    (2, 'Eva')
```

```python
incoming_data = [
    (1, "Bobby"),  # Existing ID: This will trigger an UPDATE
    (3, "Frank")   # New ID: This will trigger an INSERT
]
incoming_df = spark.createDataFrame(incoming_data, schema="id INT, name STRING")
incoming_df.createOrReplaceTempView("incoming_batch")
```

```sql
MERGE INTO users_table AS target
USING incoming_batch AS source
ON target.id = source.id
WHEN MATCHED THEN
  UPDATE SET *
WHEN NOT MATCHED THEN
  INSERT (id, name) VALUES (source.id, source.name)
WHEN NOT MATCHED BY SOURCE THEN
  UPDATE SET target.name = 'Unknown'

SELECT * FROM users_table ORDER BY id
```

### Before
Unsupported error

### After

```
+---+-------+
| id|   name|
+---+-------+
|  1|  Bobby|
|  2|Unknown|
|  3|  Frank|
+---+-------+
```

### DIFF
--- a/crates/sail-delta-lake/src/logical/merge.rs
+++ b/crates/sail-delta-lake/src/logical/merge.rs
@@ -16,7 +16,7 @@ use crate::logical::table_source::DeltaTableSource;
 
 /// Expand MERGE information into a unified row-level write node for Delta.
 pub fn expand_merge_node(info: MergeInfo) -> Result<LogicalPlan> {
-    let row_index_column = (merge_has_delete_actions(&info)
+    let row_index_column = ((merge_has_delete_actions(&info) || merge_has_update_actions(&info))
         && merge_target_supports_deletion_vectors(info.target.as_ref())?)
     .then_some(MERGE_ROW_INDEX_COLUMN);
     let mut target_plan = ensure_merge_metadata_columns(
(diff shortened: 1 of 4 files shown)


lonless9:
Thanks for the PR! 

Since you have contributed to Sail before, CI should run automatically on new commits.

I’ve marked this as draft for now. I’ll start reviewing once the test regressions are fixed and the plan / delta_log snapshot changes are within the expected scope of this PR or clearly explained. You can use `hatch run pytest -k delta --snapshot-update`. 

Please keep the scope focused on DV/MoR MERGE UPDATE: COW/DataFrame MERGE UPDATE is already covered today, and the current changes appear to affect unchanged target rows there. Please also make sure the tests cover the example described in the PR and use meaningful scenarios that validate the intended behavior. It will be great to include the relevant execution plan and Delta log snapshot coverage. Since DataFrame `mergeInto` already has COW update coverage but no DV/MoR update coverage, please add a DataFrame test for the DV/MoR update path as well.

PR #2182, opened 2026-07-02
fix: preserve Spark function semantics across workers

lonless9:


crates/sail-plan/src/function/window.rs
```diff
@@ -429,18 +441,19 @@ fn count_if(input: WinFunctionInput) -> PlanResult<expr::Expr> {
...
                 fun: WindowFunctionDefinition::AggregateUDF(count::count_udaf()),
                 params: WindowFunctionParams {
-                    args: vec![lit(0)],
+                    args: vec![arg],
                     partition_by,
                     order_by,
                     window_frame,
-                    filter: Some(Box::new(filter)),
+                    filter: None,
                     null_treatment: get_null_treatment(ignore_nulls),
```
linhr, line 457: Is the `null_treatment` still correct?

linhr (approved):
This is great!

I just have one question.

crates/sail-plan/src/function/window.rs
lonless9: Added a test and it looks ok.

PR #2183, opened 2026-07-02
feat: push down projections through ExplicitRepartitionExec

nis12ram:
## Summary

Implements projection pushdown through `ExplicitRepartitionExec`, allowing the physical optimizer to push projections below explicit repartition nodes.

## Projection Pushdown

Pushes `ProjectionExec` below `ExplicitRepartitionExec` when all projection expressions are simple column references and the schema is narrowed.

### DIFF
--- a/crates/sail-physical-plan/src/repartition.rs
+++ b/crates/sail-physical-plan/src/repartition.rs
@@ -7,16 +7,19 @@ use datafusion::arrow::compute::take_arrays;
 use datafusion::arrow::datatypes::UInt32Type;
 use datafusion::arrow::record_batch::{RecordBatch, RecordBatchOptions};
 use datafusion::common::runtime::SpawnedTask;
+use datafusion::config::ConfigOptions;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
-use datafusion::physical_expr::Partitioning;
+use datafusion::physical_expr::{Partitioning, conjunction};
 use datafusion::physical_plan::execution_plan::{
     CardinalityEffect, EvaluationType, SchedulingType,
 };
+use datafusion::physical_plan::filter::FilterExec;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::{
-    DisplayAs, DisplayFormatType, ExecutionPlan, ExecutionPlanProperties, PlanProperties,
+    DisplayAs, DisplayFormatType, ExecutionPlan, ExecutionPlanProperties, PlanProperties, PhysicalExpr
 };
 use datafusion::physical_plan::projection::{ProjectionExec, all_columns, make_with_child, update_expr};
+use datafusion::physical_plan::filter_pushdown::{ChildPushdownResult, FilterDescription, FilterPushdownPhase, FilterPushdownPropagation, PushedDown};
 use datafusion_common::{internal_err, plan_err, Result, Statistics};
 use futures::{Stream, StreamExt};
 use sail_common_datafusion::extension::SessionExtensionAccessor;
@@ -501,7 +504,7 @@ impl ExecutionPlan for ExplicitRepartitionExec {
             return Ok(None);
         }
 
-        // Only push down simple column references — non-column expressions are potentially expensive
+        // Mostly no benefit if projection expressions are not simple column references
         if projection.benefits_from_input_partitioning()[0]
             || !all_columns(projection.expr())
         {
@@ -529,5 +532,48 @@ impl ExecutionPlan for ExplicitRepartitionExec {
             new_projection,
             new_partitioning
         ))))
+    }
+
+    fn gather_filters_for_pushdown(
+        &self,
+        _phase: FilterPushdownPhase,
+        parent_filters: Vec<Arc<dyn PhysicalExpr>>,
+        _config: &ConfigOptions,
+    ) -> Result<FilterDescription>
+    {
+        FilterDescription::from_children(parent_filters, &self.children())
+    }
+
+    fn handle_child_pushdown_result(
+        &self,
+        _phase: FilterPushdownPhase,
+        child_pushdown_result: ChildPushdownResult,
+        _config: &ConfigOptions,
+    ) -> Result<FilterPushdownPropagation<Arc<dyn ExecutionPlan>>>
+    {
+        // Collect parent filters that the child did NOT support
+        let unsupported_filters: Vec<Arc<dyn PhysicalExpr>> = child_pushdown_result
+            .parent_filters
+            .iter()
+            .filter(|&f| matches!(f.all(), PushedDown::No))
+            .map(|f| Arc::clone(&f.filter))
+            .collect();
+
+        if unsupported_filters.is_empty() {
+            return Ok(FilterPushdownPropagation::if_all(child_pushdown_result))
+        }
+
+        // Build a single conjunctive predicate from the unsupported filters
+        // and insert a FilterExec between this ExplicitRepartitionExec and its child.
+        let predicate = conjunction(unsupported_filters);
+        let new_child = Arc::new(FilterExec::try_new(predicate, Arc::clone(self.input()))?);
+        let new_explicit_repartition = Arc::new(ExplicitRepartitionExec::new(
+            new_child,
+            self.properties.partitioning.clone()
+        ));
+
+        Ok(FilterPushdownPropagation::with_parent_pushdown_result(
+            vec![PushedDown::Yes; child_pushdown_result.parent_filters.len()]
+        ).with_updated_node(new_explicit_repartition))
     }
 }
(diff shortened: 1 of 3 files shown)


crates/sail-physical-plan/src/repartition.rs
```diff
@@ -483,10 +493,84 @@ impl ExecutionPlan for ExplicitRepartitionExec {
...
+
+        if unsupported_filters.is_empty() {
+            return Ok(FilterPushdownPropagation::if_all(child_pushdown_result));
+        }
+
+        // Build a single conjunctive predicate from the unsupported filters
+        // and insert a FilterExec between this ExplicitRepartitionExec and its child.
+        let predicate = conjunction(unsupported_filters);
+        let new_child = Arc::new(FilterExec::try_new(predicate, Arc::clone(self.input()))?);
+        let new_explicit_repartition = Arc::new(ExplicitRepartitionExec::new(
```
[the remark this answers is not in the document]
nis12ram: DataFusion's physical optimizer already handles this. It updates `ExplicitRepartitionExec` with the modified child plan before calling `handle_child_pushdown_result`, so `self.input()` already returns the updated child.

Reference:
https://github.com/apache/datafusion/blob/main/datafusion/physical-optimizer/src/filter_pushdown.rs#L567-L573

`SortExec` follows the same pattern:
https://github.com/apache/datafusion/blob/main/datafusion/physical-plan/src/sorts/sort.rs#L1532

@shehabgamin @linhr Unless I'm missing something, I think Copilot may have misunderstood this point.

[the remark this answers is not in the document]
shehabgamin: @nis12ram Thanks for flagging this, lemme take a look over the next couple of days!

crates/sail-physical-plan/src/repartition.rs
```diff
@@ -483,10 +493,84 @@ impl ExecutionPlan for ExplicitRepartitionExec {
...
+
+        // Skip if projection expressions are not simple column references
+        if projection.benefits_from_input_partitioning()[0] || !all_columns(projection.expr()) {
+            return Ok(None);
+        }
+
+        let new_projection = make_with_child(projection, self.input())?;
+        // Handle all partitioning variants — 'ProjectionPushdown' runs before 'RewriteExplicitRepartition'
+        let new_partitioning = match &self.properties.partitioning {
+            Partitioning::Hash(partitions, size) => {
```
shehabgamin, line 506: Can we add test coverage in `test_partition.py` for the Hash remap path.

python/pysail/tests/spark/dataframe/test_repartition.py
```diff
@@ -93,6 +101,23 @@ def test_explicit_repartition_plan_shape_uses_expected_physical_nodes(spark):
     assert "RepartitionExec: partitioning=Hash([" in hash_plan
 
 
+def test_explicit_repartition_pushes_filter_down(spark):
```
shehabgamin: Can we also add assertions for the result to ensure correctness

python/pysail/tests/spark/dataframe/test_repartition.py
```diff
@@ -93,6 +101,23 @@ def test_explicit_repartition_plan_shape_uses_expected_physical_nodes(spark):
...
+    df = spark.range(6).repartition(5).filter(F.col("id") % 2 == 0)
+    plan = normalized_plan(df)
+
+    assert explicit_repartition_child_is(plan, "FilterExec:")
+
+
+def test_explicit_repartition_pushes_column_projection_down(spark):
```
shehabgamin: Can we also add assertions for the result to ensure correctness

python/pysail/tests/spark/dataframe/test_repartition.py
```diff
@@ -29,6 +29,14 @@ def normalized_plan(df):
     return normalize_plan_text(df._explain_string())  # noqa: SLF001
 
 
+def explicit_repartition_child_is(plan_text: str, child_name: str) -> bool:
+    lines = [line for line in plan_text.splitlines() if line.strip()]
+    for i, line in enumerate(lines[:-1]):
+        if "ExplicitRepartitionExec:" in line:
```
shehabgamin: I think we can remove the colon for greater matching

crates/sail-physical-plan/src/repartition.rs
```diff
@@ -483,10 +493,84 @@ impl ExecutionPlan for ExplicitRepartitionExec {
...
+        FilterDescription::from_children(parent_filters, &self.children())
+    }
+
+    fn handle_child_pushdown_result(
+        &self,
+        _phase: FilterPushdownPhase,
+        child_pushdown_result: ChildPushdownResult,
+        _config: &ConfigOptions,
+    ) -> Result<FilterPushdownPropagation<Arc<dyn ExecutionPlan>>> {
+        // Collect parent filters that the child did not support
```
shehabgamin: Right now we ignore phase. Should we gate the self-wrapping on `FilterPushdownPhase::Pre`?
```
if !matches!(phase, FilterPushdownPhase::Pre) {
    return Ok(FilterPushdownPropagation::if_all(child_pushdown_result));
}
```

shehabgamin (commented):
First round of review!

python/pysail/tests/spark/dataframe/test_repartition.py
linhr: For testing execution plans, we can use the `snapshot` fixture and commit the snapshot YAML file. In this way we won't need to write assertions on the plan text which is harder to maintain.

Here is an example:

<https://github.com/lakehq/sail/blob/671c5caf79343daabccd4732d06fcaa94e940cc3/python/pysail/tests/spark/test_clickbench.py#L156-L157>

crates/sail-physical-plan/src/repartition.rs
nis12ram: without remap 
Screenshot 2026-07-11 094119

with remap
Screenshot 2026-07-11 094824



added test `test_explicit_repartition_hash_partitioning_remaps_after_projection_pushdown_plan` to assert correctness

python/pysail/tests/spark/dataframe/test_repartition.py
nis12ram: added `test_explicit_repartition_pushes_filter_down_result`

python/pysail/tests/spark/dataframe/test_repartition.py
nis12ram: added `test_explicit_repartition_pushes_column_projection_down_result`

python/pysail/tests/spark/dataframe/test_repartition.py
nis12ram: @linhr Thanks, adopted the snapshot-based assertion.

crates/sail-physical-plan/src/repartition.rs
nis12ram: Yes @shehabgamin, that's why while experimenting I always encountered filter pushdown in the Pre phase. But just because of this thinking:

> Any physical optimization stage between `FilterPushdown(Pre)` and `FilterPushdown(Post)` restructures the plan and introduces a new `FilterExec`, then that `FilterExec` can be pushed down in the Post phase.

I did not encounter such an edge case yet. So i also think it is fine to limit it to the `Pre` phase.

If I'm missing something or have misunderstood anything, sorry about that.

nis12ram:
hope i addressed everything

linhr (commented):
The logic in `try_swapping_with_projection` mirrors what's present in `RepartitionExec`, which makes sense.

But for `handle_child_pushdown_result` it seems there are additional logic to inject `FilterExec`. Why is it needed?

Intuitively the optimizer behavior would be the same for `RepartitionExec` and `ExplicitRepartitionExec`. So I'm curious how the tests behave if we keep the logic exactly the same between the two.

linhr:
After thinking about it more, I feel we could even be more conservative to push down no filter. Imagine there's a filter containing a UDF which is super expensive to calculate. By having an explicit repartition, the user's intention is to split the data into more pieces so that the work can be better parallelized. But pushing down the filter ignores such intentions, even if the query result is correct.

But I'm open to discussions around this.

nis12ram:
@linhr Thanks for reviewing.

After thinking about it, I also feel the filter pushdown kind of ignores the user's intention of using explicit repartitioning.

So I think marking all parent filters of `ExplicitRepartitionExec` as unsupported for filter pushdown would be better overall (default).

Removed filter pushdown.

crates/sail-physical-plan/src/repartition.rs
```diff
@@ -483,10 +486,43 @@ impl ExecutionPlan for ExplicitRepartitionExec {
...
+        };
+        Ok(Some(Arc::new(ExplicitRepartitionExec::new(
+            new_projection,
+            new_partitioning,
+        ))))
+    }
+
+    // Do not implement Filter pushdown, as its evaluation can be potentially
+    // expensive, and the presence of explicit repartitioning indicates that
+    // the user wants to evaluate the filters after repartitioning.
```
linhr, line 527: Thanks for this note! Could you add back the test case and snapshot so that we can lock this behavior in the tests?

linhr (approved):
Looks good to me!

I just have one minor comment but otherwise I feel this is ready for merge.

crates/sail-physical-plan/src/repartition.rs
nis12ram: @linhr added test `test_explicit_repartition_does_not_push_filter_down_plan`

linhr (approved):
Thanks!

PR #2187, opened 2026-07-02
perf: build arrays with typed builders instead of Vec intermediates `base64`

davidlghellin:
`base64`/`unbase64` array paths built a `Vec<Option<String>>` / `Vec<Option<Vec<u8>>>` and then re-dispatched on the output type to copy everything into the Arrow buffer. Push the offset-type decision into each match arm and write straight into a `Generic{String,Binary}Builder<O>`, dropping both redundant output-`match` blocks.

encode_spark_base64_array / decode_spark_base64_array are now generic over `OffsetSizeTrait` (`::<i32>` for Utf8/Binary, `::<i64>` for Large outputs). Encode reuses a single `String` buffer via `encode_string`; decode gives the builder `with_capacity`. Scalar arms, `decode_spark_base64` (invalid-byte fallback), signatures and `return_type` are untouched — output is identical.

Measured (arrow 58.3, criterion/divan, per-invoke over N rows):
- encode: 1.27x-1.47x faster (3.85->3.04 us @100 rows, 296->201 us @8192); allocations collapse from N+7 to a constant 7 (8199->7 @8192).
- decode: time unchanged; peak live memory drops (8195->6 live buffers).

Behavior-preserving: 5 new unit tests over the array paths, additive JVM-validated base64.feature scenarios, full gate green, no gold-data drift.

### Allocations (divan) -- O(N) per-row Vec  ->  O(1) builder

| N (rows) | encode_old | encode_new |
|---------:|-----------:|-----------:|
| 100      | 107        | 7          |
| 1,024    | 1,031      | 7          |
| 8,192    | 8,199      | 7          |

The old path allocates one Vec<u8> per row (~N+7 allocs, ~N live at peak);
the typed builder is a constant 7 allocations regardless of row count.
At DataFusion's default 8192-row batch: 8,199 allocs -> 7 (~1170x fewer).

### Time (criterion) -- the effect

| N (rows) | encode_old | encode_new | Speedup |
|---------:|-----------:|-----------:|:--------|
| 1,024    | 37.9 µs    | 26.2 µs    | 1.45x   |
| 8,192    | 297.5 µs   | 202.0 µs   | 1.47x   |

### DIFF
--- a/crates/sail-function/src/scalar/string/spark_base64.rs
+++ b/crates/sail-function/src/scalar/string/spark_base64.rs
@@ -4,8 +4,9 @@ use base64::engine::general_purpose::{GeneralPurpose, GeneralPurposeConfig, STAN
 use base64::engine::DecodePaddingMode;
 use base64::{alphabet, Engine as _};
 use datafusion::arrow::array::{
-    Array, BinaryArray, BinaryBuilder, BinaryViewArray, FixedSizeBinaryArray, LargeBinaryArray,
-    LargeBinaryBuilder, LargeStringArray, StringArray, StringViewArray,
+    Array, BinaryArray, BinaryViewArray, FixedSizeBinaryArray, GenericBinaryArray,
+    GenericBinaryBuilder, GenericStringArray, GenericStringBuilder, LargeBinaryArray,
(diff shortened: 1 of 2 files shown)


shehabgamin (approved):
Well done!

PR #2189, opened 2026-07-02
perf: vectorize `spark_bround` with arrow::compute::binary

davidlghellin:
Replace the per-element `iter().zip().map(match).collect()` in the four vectorized branches of `spark_bround` (Float64/Float32/Int32/Int64 × Int32) with `arrow::compute::binary`. The kernel derives the output null buffer as the block-AND intersection of both input bitmaps and applies the closure only over valid values — identical semantics to the old `(Some, Some) => op, _ => None`, but without the per-element `Option` branch, enabling better autovectorization. Closure bodies copied verbatim; the rounding math is unchanged.

Behavior-preserving: existing unit tests stay green untouched, plus a new `test_bround_null_propagation` locks NULL-in-value and NULL-in-scale propagation. Also migrate the old `test_bround.txt` doctest into a new `bround.feature` with additive, JVM-validated scenarios (tie-to-even, negative scale, NULL/NaN/±Infinity multi-row mixes). Two FLOAT scenarios are tagged `@sail-bug` for a pre-existing `return_type` Float32→Float64 mismatch (unrelated to this refactor, fixed in a separate PR).


shehabgamin (approved):
Well done!

PR #2190, opened 2026-07-02
fix: count_min_sketch param types + optimize sketch aggregates

davidlghellin:
Two changes to the DataSketches aggregate family:

Fix (behavior): Spark's count_min_sketch requires `eps` and `confidence` to be DOUBLE literals and rejects DECIMAL/FLOAT/INT, but Sail's planner builder blindly cast both to Float64, silently accepting types Spark errors on. Drop the cast in count_min_sketch_args and tighten the aggregate's type validation to Float64 only for eps/confidence (seed stays INT/BIGINT), so Sail matches Spark. Add @datasketches-tagged scenarios for the type rules and fix the pre-existing scenarios that asserted Sail-specific error text or used decimal parameters Spark rejects.

Perf: count_min_sketch, hll_sketch_agg and theta_sketch_agg re-ran `match data_type()` + a dynamic downcast on every row, paying O(N) re-dispatch per batch. Hoist the match above the row loop and downcast once via the AsArray trait, iterating with `iter().flatten()` to skip nulls without a per-row branch. Also drop per-element allocations: inline count_min's binary hashing and reuse a scratch buffer for theta's array-element path. Behavior-neutral (same hashing, order and serialized bytes).


crates/sail-function/src/aggregate/count_min_sketch.rs
```diff
@@ -400,18 +351,22 @@ fn validate_count_min_sketch_types(arg_types: &[DataType]) -> Result<()> {
...
+            ));
         }
     }
-    if !matches!(arg_types[1], DataType::Float32 | DataType::Float64) {
+    if !matches!(arg_types[1], DataType::Float64) {
         return exec_err!(
-            "count_min_sketch requires a floating point eps argument, got {}",
+            "count_min_sketch requires a double eps argument, got {}",
             arg_types[1]
         );
```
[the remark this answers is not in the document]
davidlghellin: resolved [c81cd99](https://github.com/lakehq/sail/pull/2190/commits/c81cd996499403fdad5d84e3c718e8d0e37cdad7)

python/pysail/tests/spark/function/features/theta_sketch.feature
```diff
@@ -1,3 +1,4 @@
+@datasketches
```
lonless9, line 1: Why do we need this?

python/pysail/tests/spark/function/features/theta_sketch.feature
```diff
@@ -156,14 +157,14 @@ Feature: Theta sketch functions
         """
         SELECT theta_sketch_estimate(theta_intersection_agg(NULL)) AS result
         """
-      Then query error theta_intersection_agg cannot produce a result without any non-null input sketches
+      Then query error .*
```
lonless9: What do you think about explicit assert error messages? I just noticed that we have a lot of BDD tests using `.*`, but that can be addressed in the future.

python/pysail/tests/spark/function/features/datasketches.feature
```diff
@@ -1,3 +1,4 @@
+@datasketches
```
lonless9, line 1: Here as well.

lonless9 (approved):
Great work!

python/pysail/tests/spark/function/features/theta_sketch.feature
davidlghellin: It groups the whole DataSketches family (hll_*, theta_*, count_min_sketch) — which is split across `datasketches.feature` and `theta_sketch.feature` — so the family can be run in one command while working on it:

    `pytest ... -m datasketches`

Same idea as the existing `@lambda_hof` / `@aggregate` tags. It doesn't change test behavior, it's purely a selector

lonless9: Oh get it, great to help with dev!

python/pysail/tests/spark/function/features/theta_sketch.feature
davidlghellin: Good point — let me take a look and see if we can assert proper messages here and leave these tests clean.

davidlghellin: updated

lonless9 (approved):

PR #2188, opened 2026-07-02
fix: propagate NULL per-element in `make_timestamp_ntz` 6-arg branch

davidlghellin:

PR #2196, opened 2026-07-03
fix: honor NULL and overflow semantics in `make_ym_interval`

davidlghellin:
make_ym_interval diverged from Spark's MakeYMInterval:
- a NULL in either argument returned a value instead of NULL (null-intolerant semantics, same pattern as #2188 for make_timestamp_ntz)
- years * 12 + months could overflow silently instead of erroring
- the 0-argument form was rejected, and arity handling was split between the plan builder and the UDF


shehabgamin (approved):
Well done!

PR #2200, opened 2026-07-04
fix: support scalar subqueries in distributed execution

lonless9:


crates/sail-plan/src/explain.rs
```diff
@@ -413,6 +437,7 @@ async fn explain_from_collected(
         collected.logical_string(&collected.optimized_logical, PlanType::FinalLogicalPlan);
 
     let mut physical = PhysicalStrings::default();
+    let distributed_plan = distributed_plan_string(ctx, &collected.physical_plan);
 
     let output = match options.kind {
```
linhr: I feel we don't need `render_with_distributed_plan` to handle distributed plan specially. The distributed plan is just a normal section. We change change this `match` statement to return `Vec<String>`. And then we inline the logic from `render_with_distributed_plan` to update the sections and call `sections.join("\n\n")`.

python/pysail/tests/spark/execution/features/explain.feature
```diff
@@ -0,0 +1,34 @@
...
+      """
+    Then query plan matches snapshot
+
+  Scenario: EXPLAIN includes the scalar subquery distributed plan in local-cluster mode
+    When query
+      """
+      EXPLAIN
+      SELECT k, SUM(v) AS total
+      FROM VALUES (1, 2), (1, 3), (2, 10), (3, 1) AS t(k, v)
+      WHERE v > (SELECT MAX(x) FROM VALUES (1), (2) AS s(x))
```
linhr: I'm curious if it works when the subquery contains outer column reference. Could you add some test cases?

crates/sail-plan/src/explain.rs
lonless9: Done!

python/pysail/tests/spark/execution/features/explain.feature
lonless9: Done!

python/pysail/testing/spark/steps/sql.py
```diff
@@ -200,6 +200,27 @@ def query_result(datatable, ordered, query, spark):
...
+def _format_collected_value(value) -> str:
+    if value is None:
+        return "NULL"
+    if isinstance(value, bool):
+        return str(value).lower()
+    return str(value)
+
+
+@then(parsers.re("query result collected(?P<ordered>( ordered)?)"))
+def query_result_collected(datatable, ordered, query, spark):
```
linhr, line 212: Why do we need this?

linhr (approved):
Great work!

python/pysail/testing/spark/steps/sql.py
lonless9: This arose when I changed the test to BDD, and I think it can be considered as a follow-up work.

PR #2206, opened 2026-07-06
[DO NOT MERGE] keep catalog command in a single driver stage

linhr:
If the physical plan is only a catalog command, we should have a single driver stage to avoid the roundtrip between driver and worker.

But this PR has its own issue. It somehow causes tests to hang (e.g., `env SAIL_MODE=local-cluster hatch run pytest -k "test_delta_io.py"`). The root cause is unknown.

I feel we can have a proper implementation after certain refactoring on the job graph planning logic which has been getting complex over time.

### DIFF
--- a/crates/sail-execution/src/job_graph/planner.rs
+++ b/crates/sail-execution/src/job_graph/planner.rs
@@ -268,7 +268,13 @@ fn plan_job_graph_stages(
     }
 
     if let Some(barrier) = plan.downcast_ref::<BarrierExec>() {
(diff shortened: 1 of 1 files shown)


linhr:
I realized this issue when I was troubleshooting the failed tests mentioned in #2207. If it's a single driver stage, the bug mentioned in #2207 wouldn't have surfaced since Arrow Flight won't be involved for driver-only execution.

(Edit: I incorrectly mentioned #2077 in this comment. This has been updated.)

PR #2208, opened 2026-07-06
chore: migrate Rust edition to 2024

lonless9:


crates/sail-common-datafusion/src/system/types.rs
```diff
@@ -28,7 +28,7 @@ trait NumericType {
 }
 
 macro_rules! impl_numeric_type {
-    ($t:ty, $dt:expr) => {
+    ($t:ty, $dt:expr_2021) => {
```
linhr, line 31: This needs attention. (Same for other usage of `expr_2021`.)

linhr (approved):
🚀

PR #2220, opened 2026-07-08
fix: align +, -, *, / and try_* with Spark ANSI arithmetic semantics

### LINKED ISSUE #2198: try_divide function does not accept floating point numbers only integers
Function `try_divide` does not accept floating point numbers:
```python
spark.sql("select try_divide(10.0, 2.2)").show()
```

**ERROR:**
```
AnalysisException: Execution error: Function 'try_divide' user-defined coercion failed with: Execution error: Unsupported Data Type: Spark `try_divide` function expects Int32, Int64 o Interval(YearMonth) / Int32, got Decimal128(3, 1), Decimal128(2, 1). No function matches the given name and argument types 'try_divide(Decimal128(3, 1), Decimal128(2, 1))'. You might need to add explicit type casts.
        Candidate functions:
        try_divide(UserDefined)
```

davidlghellin:
Fixes #2198.

Sail's arithmetic operators (`+`, `-`, `*`, `/`) and their `try_*` variants had
several divergences from Spark: overflow was not ANSI-aware, `try_*` rejected
FLOAT/DECIMAL, decimal precision/scale did not follow Spark's rule, and a few
type-coercion cases were wrong. This aligns them with Spark 4.1.1 (all
divergences JVM-validated).

## What changed

The regular operators and their `try_*` variants are unified into per-operation
UDFs `SparkAdd` / `SparkSubtract` / `SparkMultiply` / `SparkDivide { ansi_mode,
safe }`, following the existing `SparkSum` consolidation pattern (one source of
truth per operation). Three modes, in priority order:

- `safe` (`try_*`)      → overflow / div-by-zero → NULL, ANSI-invariant
- `!safe && ansi`       → overflow / div-by-zero → ERROR
- `!safe && !ansi`      → integers wrap · decimals NULL · div-by-zero NULL

The old `SparkTryAdd/Subtract/Mult/Div` structs are retired. Regular ANSI-off
integral arithmetic stays on the native operator (native wrapping already
matches Spark, and it preserves `BinaryExpr` for simplification); only ANSI-on
or decimal operands route through the UDF.

## Divergences fixed (JVM-validated)

- Integer/decimal overflow now honors `spark.sql.ansi.enabled` (raises
  `ARITHMETIC_OVERFLOW` under ANSI on, wraps/NULLs under ANSI off), including
  narrow types (`TINYINT`/`SMALLINT`) which keep their type.
- `try_add/subtract/multiply/divide` now accept FLOAT and DECIMAL.
- Decimal `*` and `/` follow Spark's precision/scale rule (`adjustPrecisionScale`,
  HALF_UP), e.g. `decimal(10,2) / decimal(10,2)` → `decimal(23,13)`.
- Integer-literal narrowing: `decimal(10,2) * 3` → `decimal(12,2)`,
  `decimal(10,2) + 3` → `decimal(11,2)`.
- `DATE - DATE` → day-time interval (was `bigint`).
- String operand promotion: `'5' + 3` → numeric (was a coercion error).
- `float * decimal` → `double` (was `float`).
- Interval division by zero matches Spark exactly across the three interval
  representations: year-month and day-time raise `INTERVAL_DIVIDED_BY_ZERO` in
  **both** ANSI modes; `make_interval` (CalendarInterval) raises under ANSI on
  and returns NULL under ANSI off; `try_divide` always returns NULL.

## Known limitations (tracked as `@sail-bug`)

- **ANSI overflow with a scalar-subquery operand.** Wrapping arithmetic in the
  UDF breaks the DataFusion planner's count-bug check when a subquery is nested
  inside a correlated subquery body (fails to plan with "does not support
  logical expression ScalarSubquery"). As a workaround, `+`/`-`/`*` fall back to
  the native operator when an operand contains a scalar subquery
  (`has_scalar_subquery` in `math.rs`). Consequence: `(SELECT ...) + x` wraps on
  overflow under ANSI instead of raising. The root fix belongs in the fork's
  `evaluates_to_null` (relates to apache/datafusion#21240 and the
  `ScalarSubqueryExec` work in #22566 / #22530); the guard can be dropped once
  that lands.
- **`DATE - DATE` rendering.** Value is correct but renders as `INTERVAL … DAY TO
  SECOND` instead of Spark's `… DAY` granularity (Sail stores day-time intervals
  as `Duration(µs)`).
- **`'5' + 3` under ANSI on** yields `INT` where Spark yields `BIGINT` (value
  correct).
- **`try_divide(day-time interval, 0)`** has no `Duration` branch in
  `SparkDivide` yet (pre-existing gap, separate from this change).

## Testing

- New/updated feature files: `arithmetic_overflow`, `arithmetic_coercion`,
  `divide_by_zero`, `try_{add,subtract,multiply,divide}`, plus plan-snapshot
  (`@sail-only`) and `@sail-bug` scenarios for the subquery guard. All new
  scenarios JVM-validated against Spark 4.1.1.
- proto + codec updated for distributed execution of the four UDFs.
- Plan snapshots (tpcds/tpch/clickbench/delta) and gold data regenerated.


shehabgamin:
@davidlghellin it may be better to start with an analyzer rule. Using a UDF will lead to a large performance penalty.

Perhaps you want to pick up my work here?
https://github.com/lakehq/sail/pull/2137

davidlghellin:
> @davidlghellin it may be better to start with an analyzer rule. Using a UDF will lead to a large performance penalty.
> 
> Perhaps you want to pick up my work here? #2137

Thanks @shehabgamin — good point on the UDF perf cost. I'll review #2137 and the analyzer-rule approach properly. Thanks for the pointer!

Issue #1903, opened 2026-05-15
Use `return_field_from_args` instead of `return_type` for all UDFs


linhr:
Reference: <https://github.com/apache/datafusion/issues/19144>

This would ensure nullability and metadata are preserved in UDF outputs.

PR #2225, opened 2026-07-09
feat: add Spark-compatible arithmetic type coercion in plan builders

davidlghellin:
## What

Spark-compatible **arithmetic type coercion** for `+ - * / %` and string operands, done in the
plan builders (`crates/sail-plan/src/function/scalar/math.rs`). No UDFs, no custom `PhysicalExpr`
— pure logical `Expr` construction, so the logical plan is valid by construction (no
`ExprSchemable::get_type` / analyzer timing gap). Supersedes the #2223 POC.

**Operators (type + value validated against Spark 4.2.0):**
- **`/` decimal division** — Spark's scale rule `max(6, s1+p2+1)` + `adjustPrecisionScale` cap at
  38, computed with **HALF_UP** (widen to Decimal256, one guard digit, `round`, narrow back).
  Covers dec/dec, dec/int-literal (narrowing) and dec/int-column (`DecimalType.forType`),
  float×decimal→double, NULL, IEEE specials, signs, exact, high-scale, multi-row.
- **`%` modulo** — Spark's remainder rule `min(p1-s1, p2-s2) + max(s1,s2)`, sign of the dividend.
- **`+ - *`** — integer-literal narrowing to minimal decimal, float×decimal→double, and Spark's
  `adjustPrecisionScale` precision-38 capping (product/sum widened to Decimal256, then narrowed) so
  a capped result that overflows yields **NULL under ANSI off** (Spark's `CheckOverflow`) rather
  than raising — consistent across `+`, `-`, `*`.
- **operand-type rejection** — non-numeric operand pairs (`bool`, `date`, `timestamp`, `interval`,
  `binary`, …) are rejected at **plan time** with a `cannot resolve` error, matching Spark's
  analysis-time `DATATYPE_MISMATCH` decision. 0 over-accept / 0 over-reject vs the JVM over the full
  15×15 operand matrix, both ANSI modes, across all five operators.
- **string operand** promotion (ANSI off → double; ANSI on → the numeric operand's type). The
  ANSI-on integer *width* (`'5'+3` → int vs Spark's bigint) is a known gap below.

## Validation

- Every result **type and value** validated against **Spark 4.2.0 (JVM)** and asserted via BDD
  scenarios in `arithmetic_coercion.feature` plus the full operand-type matrix in
  `math/arithmetic_operand_matrix.feature` (no Rust unit tests — behavior tracked by BDD).
- **TPC-H / TPC-DS** plan + result snapshots regenerated; the division change propagates correctly
  (HALF_UP + Spark scale). The plan churn is decimal-width expression text, **not** a join reorder
  (same joins, same order); `.result.yaml` is unchanged. Verified the logical/physical optimizers
  still apply: CSE, filter evaluation and multi-stage aggregation all run over the new division
  construction, and the `round(division)` scenarios confirm `SimplifyExpressions` does not collapse
  the inner round.
- Full `nextest` + gold + `pytest` green.

## Performance

`+ - * %`, integer and float division are **unchanged** (coercion is plan-time cast insertion).
**Decimal/decimal division is heavier** (i256 widening + a HALF_UP `round` pass) — the inherent
cost of Spark's decimal-division semantics (Spark computes it in `BigDecimal`); it stays fully
vectorized (native Arrow kernels, **no UDF dispatch**). Division already keeps the intermediate in
**i128 when it provably cannot overflow** (an i128/i256 gate on the intermediate digit count),
widening to i256 only when needed; the `+ - *` overflow retype widens to i256 only when the exact
precision exceeds 38.

## Known gaps → follow-ups (tagged `@sail-bug`)

- **Nullability** — narrow decimal `+ - *` (precision ≤ 38) report `nullable=false` where Spark
  marks it `nullable=true`; the wide / overflow path is already correct (`try_cast`). Needs the
  custom `PhysicalExpr` follow-up (also covers `try_*` consolidation).
- **ANSI overflow** — integer `+ - *` wrap under ANSI off and don't raise under ANSI on; decimal
  overflow under ANSI **on** raises Sail's Arrow error rather than Spark's `NUMERIC_VALUE_OUT_OF_RANGE`
  class. (Decimal overflow under ANSI **off** → NULL is handled.)
- **ANSI `'5'+3` → bigint width** (Sail casts the string to int).
- **`round(decimal, n)` result type** keeps the input precision instead of Spark's
  `decimal(p-s+1+n, n)` — a pre-existing `round` divergence surfaced here (the `round_decimal_base`
  helper exists but isn't wired); unrelated to this change.


davidlghellin:
Opening for an early pass — added Copilot as a reviewer to get a first review

shehabgamin:
@davidlghellin There doesn't appear to be any difference in the Spark test suite. Can we perhaps isolate a couple tests from the Spark test suite that we expect should be impacted?

davidlghellin:
> @davidlghellin There doesn't appear to be any difference in the Spark test suite. Can we perhaps isolate a couple tests from the Spark test suite that we expect should be impacted?

Sure — I'll add BDD scenarios

davidlghellin:
> @davidlghellin There doesn't appear to be any difference in the Spark test suite. Can we perhaps isolate a couple tests from the Spark test suite that we expect should be impacted?

Good question — I isolated the impact two ways:

**1. BDD scenarios (feature/arithmetic_coercion.feature, tag `@arithmetic_coercion`).**
All 53 scenarios pass against real Spark JVM 4.1.1 (so the expected values are
Spark's, not hand-picked). Running the *same* scenarios against a `main` build
(no fix) gives **30 failures**, e.g.:

- `typeof(CAST(2.5 AS DECIMAL(10,2)) * 3)` → main: `decimal(21,2)`, Spark & this PR: `decimal(12,2)`
- `typeof(CAST(1.5 AS FLOAT) * CAST(2.0 AS DECIMAL(10,2)))` → main: not promoted, Spark & this PR: `double`

The 14 that already pass on main are cases DataFusion coerced the same way; the
9 xfails are the known decimal-nullability gap (needs the custom PhysicalExpr
follow-up).

**2. Existing Spark test suite (TPC-H / TPC-DS).**
The result snapshots already reflect the change — e.g. TPC-H Q14 `promo_revenue`
goes `14.5856344411` → `14.585634441128` (Spark's decimal scale via
multiply+divide capping), plus HALF_UP last-digit corrections across several
TPC-DS queries. Those are the "existing suite" tests that are impacted.

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -306,10 +580,40 @@ fn spark_divide(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
+        // widens to i256 and adds a HALF_UP `round` pass). That is the inherent cost
+        // of Spark's decimal-division semantics — Spark itself computes it in
+        // `BigDecimal`, and no pure-`Expr` alternative is both correct and cheaper.
+        // It stays fully vectorized (native Arrow kernels, no UDF dispatch), and only
+        // decimal/decimal division pays it; `+ - * %`, integer and float division are
+        // unchanged. A future optimization could keep the intermediate in i128 when it
+        // provably cannot overflow, instead of always widening to i256.
+        (Ok(DataType::Decimal128(p1, s1)), Ok(DataType::Decimal128(p2, s2))) => {
+            let (result_precision, result_scale) = spark_decimal_divide_type(*p1, *s1, *p2, *s2);
+            let dividend_scale = (*s1).max(result_scale - 3);
```
shehabgamin: I'm not sure that widening prevents intermediate overflow. Does Arrow rescale the numerator before division?

Can you add a test case for the following?
```
>>> spark.sql("SELECT (CAST('0.99999999999999999999999999999999999999' AS DECIMAL(38,38)) / CAST('0.99999999999999999999999999999999999999' AS DECIMAL(38,38))) AS result").show()
+--------+
|  result|
+--------+
|1.000000|
+--------+
```

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -176,12 +201,240 @@ fn spark_multiply(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
+                DataType::Decimal128(_, _) | DataType::Decimal256(_, _)
+            );
+            // ANSI off (any pairing) and ANSI-on string+decimal -> both to DOUBLE.
+            if !ansi_mode || has_decimal {
+                return (
+                    cast(left, DataType::Float64),
+                    cast(right, DataType::Float64),
+                );
+            }
+            // ANSI on, non-decimal: cast the string to the other numeric operand's type.
```
shehabgamin: I'm not sure whether or not this is true, but figured i'd flag it. Here is a review from Codex:
```
High — ANSI string promotion uses the peer’s exact numeric type
produces INT for string-plus-INT and FLOAT for string-plus-FLOAT.
Spark promotes integral operands to BIGINT and fractional operands to DOUBLE.
This changes both result types and valid ranges.
For example, a numeric string outside INT range can succeed in Spark but fail here.
The PR acknowledges only the integer case as @sail-bug, but the advertised string
feature remains non-parity.
```

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -176,12 +201,240 @@ fn spark_multiply(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
+    fn is_decimal(dt: &DataType) -> bool {
+        matches!(dt, DataType::Decimal128(_, _) | DataType::Decimal256(_, _))
+    }
+    (a.is_floating() && is_decimal(b)) || (is_decimal(a) && b.is_floating())
+}
+
+/// When `expr` is an integer literal and `other_type` is a decimal, returns the
+/// literal's minimal-precision decimal (`Decimal(digit_count, 0)`), matching Spark's
+/// `DecimalType.fromLiteral`. DataFusion would instead widen it to `Decimal(10, 0)`.
+fn spark_decimal_literal_datatype(expr: &Expr, other_type: &DataType) -> Option<DataType> {
```
shehabgamin: I'm not sure whether or not this is true, but figured i'd flag it. Here is a review from Codex:
```
High — Literal narrowing is applied outside Spark’s rule.

- Int8/TINYINT literals are narrowed by digit count, while Spark narrows only Short, Int, and Long literals.
typeof(DECIMAL(10,2) * 3Y) is decimal(14,2) in Spark but decimal(12,2) here

- Narrowing is unconditional, ignoring the exposed spark.sql.legacy.literal.pickMinimumPrecision=false.
Under that setting, Spark returns decimal(21,2) for DECIMAL(10,2) * 3; this code still returns decimal(12,2).
```

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -293,6 +546,27 @@ fn spark_divide(input: ScalarFunctionInput) -> PlanResult<Expr> {
     }
 
     let ansi_mode = function_context.plan_config.ansi_mode;
+
+    // Coerce operands the same way `*` does (narrow an integer literal combined with
+    // a decimal, promote float×decimal to double) before deriving the division type,
+    // because Spark's divide scale depends on the divisor precision.
+    let (dividend, divisor) = match (
```
shehabgamin, line 1286: Can you add these as test cases?
```
spark-sql (default)> SELECT typeof(CAST(3 AS DECIMAL(10,2)) / NULL);
decimal(23,13)
Time taken: 0.067 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT typeof(NULL / CAST(3 AS DECIMAL(10,2)));
double
Time taken: 0.042 seconds, Fetched 1 row(s)
spark-sql (default)>
```

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -176,12 +201,240 @@ fn spark_multiply(input: ScalarFunctionInput) -> PlanResult<Expr> {
...
+    //   ANSI off -> a string paired with a string or numeric promotes BOTH to DOUBLE.
+    //   ANSI on  -> a string paired with a non-decimal numeric is cast to that
+    //               numeric's exact type (strict cast); string+DECIMAL promotes both
+    //               to DOUBLE; string+string is left as-is (Spark rejects it).
+    // https://github.com/apache/spark/blob/v4.1.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala (PromoteStrings)
+    // https://github.com/apache/spark/blob/v4.1.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/AnsiTypeCoercion.scala
+    let left_string = is_string_type(left_type);
+    let right_string = is_string_type(right_type);
+    if left_string || right_string {
+        let left_ok = left_string || left_type.is_numeric();
```
shehabgamin: Here are some test cases to add:
```
spark-sql (default)> SET spark.sql.ansi.enabled = false;
spark.sql.ansi.enabled	false
Time taken: 0.473 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT 'not-a-number' + 1;
NULL
Time taken: 0.673 seconds, Fetched 1 row(s)
spark-sql (default)> SELECT '5' + NULL;
NULL
Time taken: 0.048 seconds, Fetched 1 row(s)
spark-sql (default)>
```

I think this branch needs null-on-failure casts and Spark’s NullType handling.

crates/sail-function/src/scalar/math/utils/decimal.rs
```diff
@@ -31,3 +31,44 @@ pub fn round_decimal_base(
...
+
+/// Spark's `DecimalType.MINIMUM_ADJUSTED_SCALE`: the minimum number of fractional
+/// digits preserved when a wide decimal result is capped at precision 38.
+pub const SPARK_MINIMUM_ADJUSTED_SCALE: i32 = 6;
+
+/// Spark's `adjustPrecisionScale`: when a computed decimal precision exceeds 38,
+/// cap it at 38 and reduce the scale, keeping at least
+/// `min(scale, SPARK_MINIMUM_ADJUSTED_SCALE)` fractional digits. DataFusion's
+/// coercion instead caps the scale at 38, which diverges from Spark for wide results.
+pub fn adjust_precision_scale(precision: i32, scale: i32) -> (u8, i8) {
```
shehabgamin: `allowPrecisionLoss=false` seems to be ignored.

shehabgamin (commented):
@davidlghellin Well done!!

I have some comments and I also posted a couple of comments that Codex flagged as well

crates/sail-plan/src/function/scalar/math.rs
davidlghellin: You were right to flag this — the widening does NOT make it overflow-proof. Arrow's
decimal `div` rescales the numerator by `10^(result_scale - s1 + s2)` (= `10^(4 + s2)`
here) before dividing, so the intermediate carries `(p1 - s1) + dividend_scale + 4 + s2`
digits — 80 for `decimal(38,38) / decimal(38,38)`, past i256's ~76. The overflow is
value-dependent, so the Spark-scale path is still taken (rejecting the whole type would
send every row to the native i128 divide, which overflows sooner). I fixed the code
comment (it claimed the opposite) and added your `1.000000` case as a scenario — it's
`@sail-bug`, because matching Spark exactly needs BigDecimal (`1e38 * 1e39 = 1e77`
doesn't fit i256 either), which is the custom-`PhysicalExpr` follow-up we discussed.

crates/sail-plan/src/function/scalar/math.rs
davidlghellin: Confirmed against Spark 4.1.1 — Codex is right. ANSI-on now promotes a string against an
integral operand to BIGINT and against a fractional one (float/decimal) to DOUBLE, not
the peer's exact type. This also fixes the range issue: `'3000000000' + CAST(1 AS INT)`
now succeeds. Covered per type (TINYINT/SMALLINT/INT/BIGINT/FLOAT/DOUBLE/DECIMAL), both
ANSI modes, literal and column, in arithmetic_coercion.feature.

crates/sail-plan/src/function/scalar/math.rs
davidlghellin: Confirmed and fixed (the flagged line is now `Outdated`). Both sub-points:
- Byte literals are no longer narrowed — `DataTypeUtils.fromLiteral` only matches
  Short/Int/Long, so a Byte falls through to `forType(ByteType)` = decimal(3,0):
  `typeof(decimal(10,2) * 3Y)` is now `decimal(14,2)`.
- `spark.sql.legacy.literal.pickMinimumPrecision` is now honoured; with it false,
  `typeof(decimal(10,2) * 3)` is `decimal(21,2)`.
Both are covered as scenarios, validated against Spark 4.1.1.

crates/sail-plan/src/function/scalar/math.rs
davidlghellin: Added both (as literal and column scenarios), and they were wrong before — fixed. The
division NULL coercion is asymmetric, matching your output exactly:
`typeof(dec / NULL)` is `decimal(23,13)` (the NULL takes the decimal's type, so the divide
rule applies) and `typeof(NULL / dec)` is `double` (that side falls back to DataFusion's
default). Verified against Spark 4.1.1.

crates/sail-plan/src/function/scalar/math.rs
davidlghellin: Done — both of your cases match now (verified vs Spark 4.1.1). Under ANSI off, a NULL
operand rides along as DOUBLE and a string that doesn't parse becomes NULL via a
null-on-failure cast, so `'not-a-number' + 1` and `'5' + NULL` are both NULL
(`typeof('5' + NULL)` is `double`). Under ANSI on, `'5' + NULL` and `'5' + '6'` are
rejected like Spark. The null-on-failure cast + NullType handling you called out is now
shared across all three string→numeric paths (CAST, the FLOAT()/DOUBLE() constructors,
and the operator coercion), and covered as scenarios in both ANSI modes, literal and
column.

crates/sail-function/src/scalar/math/utils/decimal.rs
davidlghellin: Fixed — the config is now honoured. `spark.sql.decimalOperations.allowPrecisionLoss`
switches between `adjustPrecisionScale` (the existing path) and `bounded`, so with it false
`typeof(decimal(38,10) * decimal(38,10))` is `decimal(38,20)` instead of `decimal(38,6)`.
Divide is handled too, and it does NOT simply swap in `bounded` — Spark keeps Hive's older
sizing there (`typeof(decimal(38,10) / decimal(38,10))` is `decimal(38,18)`). Both configs
default to true. Covered as scenarios.

shehabgamin:
@davidlghellin I had Codex 5.6 ultra do another pass. Let me know if any of this is relevant or not. If it is, please re-ping me when ready for re-review!
```
  ## Findings

  1. [P1] ANSI-off NaN now breaks heterogeneous VALUES.
     /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/common.rs:70 now creates TryCast, but /private/tmp/sail-pr2225-review/crates/sail-plan/src/resolver/query/values.rs:54 recognizes NaN only inside
     Expr::Cast.

     SELECT * FROM VALUES
       (0, FLOAT('NaN'), NULL),
       (1, NULL, 2.0),
       (2, 2.1, 3.5) t(a,b,c)

     With ANSI disabled, Sail infers b as decimal(14,7) and fails casting NaN. Spark infers DOUBLE and returns the rows. This accounts for seven new failures already visible in the PR’s Spark 3.5 comparison
     artifact.

  2. [P1] round(BIGINT) silently loses data.
     /private/tmp/sail-pr2225-review/crates/sail-function/src/scalar/math/spark_round.rs:85 delegates to DataFusion’s signature, which coerces integral inputs to Float64.

     round(CAST(9007199254740993 AS BIGINT), 0)

     Sail returns DOUBLE 9007199254740992.0; Spark returns the exact BIGINT 9007199254740993. Spark explicitly preserves non-decimal input types in opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/
     catalyst/expressions/mathExpressions.scala:1503. Casting back afterward cannot recover the lost bit.

  3. [P1] A capped decimal sum can overflow before reaching a valid Spark result.
     /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/scalar/math.rs:100 performs native Decimal128 addition before the scale-reducing round at /private/tmp/sail-pr2225-review/crates/sail-plan/src/
     function/scalar/math.rs:342.

     DECIMAL(38,2) '1000000000000000000000000000000'
       + DECIMAL(38,10) '0'

     Sail declares the correct decimal(38,6) type but errors while multiplying the first coefficient by 100000000. Spark returns 1000000000000000000000000000000.000000, using BigDecimal before overflow checking
     (opt/spark/sql/api/src/main/scala/org/apache/spark/sql/types/Decimal.scala:504).

  4. [P2] Decimal capping misses integer columns and NULL operands.
     The +/- gates at /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/scalar/math.rs:96 only run when both operands are already decimal; unlike multiply/divide, they never apply type-based integer
     conversion.
      - DECIMAL(38,18) +/- INT column: Sail returns decimal(38,18); Spark returns decimal(38,17) with correspondingly rounded values.
      - DECIMAL(38,18) * NULL: Sail returns decimal(38,36); Spark returns decimal(38,6).

     Spark converts integral expressions before applying decimal formulas (opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/DecimalPrecisionTypeCoercion.scala:176).

  5. [P2] pmod’s generic numeric coercion destroys Spark operand typing.
     Although /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/scalar/math.rs:1168 attempts to derive Spark’s remainder type, the call at /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/
     scalar/math.rs:1186 first forces both operands to DataFusion’s common numeric type.

     Confirmed differences include:
      - pmod(decimal(3,2), INT column): Sail decimal(12,2), Spark decimal(3,2).
      - pmod('5.5', decimal(10,2)): Sail decimal(30,15), Spark DOUBLE.
      - pmod('Infinity', decimal): Sail errors; Spark returns DOUBLE NaN.
      - ANSI-off pmod(NULL, '3'): Sail fails analysis; Spark returns DOUBLE NULL.

  6. [P2] round misses non-literal foldable scales.
     /private/tmp/sail-pr2225-review/crates/sail-function/src/scalar/math/spark_round.rs:65 only sees direct scalar literals.

     round(CAST(3.14159 AS DECIMAL(10,5)), 1 + 1)

     Sail types this as decimal(11,5); Spark evaluates the foldable scale and returns decimal(8,2).

  ## Known blocker

  The Decimal256 division path at /private/tmp/sail-pr2225-review/crates/sail-plan/src/function/scalar/math.rs:809 still overflows for valid inputs such as decimal(38,38) / itself; Spark returns decimal(38,6)
  1.000000. The PR documents this with an @sail-bug, but it remains incompatible with the repository’s exact-parity requirement.
```

davidlghellin:
> @davidlghellin I had Codex 5.6 ultra do another pass. Let me know if any of this is relevant or not. If it is, please re-ping me when ready for re-review!
> 

Thanks for running this @shehabgamin  — all six were legit. Went through each against Spark 4.1.1. **Fixed 1, 4, 5, 6.** 2 is half done, 3 is the overflow follow-up we already agreed to defer.

**1 — fixed.** `values.rs` only looked at `Expr::Cast`, so the `TryCast` the ANSI-off path emits slipped through and the column got inferred as `decimal(14,7)`. Matches both cast forms now. Your query returns `nan / NULL / 2.1`, and I added a heterogeneous-VALUES scenario.

**2 — half fixed.** `Round.dataType` keeps the input type, and rounding an int at a non-negative scale can't change it, so I return the argument as-is instead of pushing it through Float64. `round(9007199254740993L, 0)` stays exact and `bigint` now. Negative scale is the other half — it actually rounds and can overflow (Spark raises under ANSI, wraps otherwise: `round(127Y, -1)` → -126). Doing that right needs the ANSI flag inside the UDF plus a new proto message + kernel, so I'd rather split it into a small follow-up right after this lands than put proto changes in a coercion PR. Pinned with three xfails.

**3 — deferred (agreed).** The BigDecimal overflow case for the custom `PhysicalExpr` follow-up: the native add runs at the wider scale and overflows i128 before the round reduces it. Note the add that overflows is the same one `main` already does — this PR only re-types on top, and the cast can only widen, so it's not a new failure path. Pinned with Spark's value.

**4 — fixed.** The gate only fired when both sides were already decimal. Shared across all four operators now, and a NULL operand gets the peer's decimal type. `decimal(38,17)` for the int column, `decimal(38,6)` for `* NULL`.

**5 — fixed.** Two causes, both in shared code. The string branch was casting only the string and leaving the peer alone — fine for `+ - * /` (DataFusion re-widens the pair) but a UDF takes what it's handed, hence `pmod('5.5', decimal(10,2))` → `decimal(30,15)` and `pmod(NULL, '3')` failing to plan. And pmod wasn't doing the int→decimal conversion, so the remainder rule never saw two decimals. `pmod('Infinity', decimal)` gives NaN now, ANSI-off `pmod(NULL, '3')` returns NULL.

**6 — fixed.** Scale is folded at plan-build time now, so `round(x, 1 + 1)` is `decimal(8,2)`. While there I noticed ceil/floor share the rule and were worse — `ceil(x, 1 + 1)` errored where Spark returns a value; routed through the same helper. bround's still off (types decimals as double, errors on a decimal column, scalar path rounds half-up not half-even) — that's a #1903 thing, left pinned.

**Blocker (`decimal(38,38)`) — I don't think it blocks.** Not a regression: it's just as broken on `main`, where decimal division uses the native i128 path — this PR only surfaces it, dropping the test buys silence not parity. Mapped the shape with a grid (precision/scale × literal/column × both ANSI modes): only breaks at precision 38 with both scales ≥37. `decimal(20,20)`, `decimal(30,30)`, and `decimal(38,38) / decimal(38,10)` all divide fine — you can't get those two failing types without an explicit CAST. Matching Spark needs control over the intermediate scale, which an `Expr` can't express (Arrow derives it from the declared types); pre-lowering the scales trades overflow for divide-by-zero on a tiny divisor. So it's the `PhysicalExpr` follow-up — but if you'd rather not ship a known-failing test, say so and I'll pull it in.

shehabgamin:
@davidlghellin I converted this to a draft pr. Feel free to tag me when it's ready for re-review!

crates/sail-plan/src/function/scalar/math.rs
```diff
@@ -579,31 +1517,117 @@ fn double2(func: impl Fn(Expr, Expr) -> Expr) -> impl Fn(Expr, Expr) -> Expr {
...
-            )));
-        } else {
-            return Ok(Expr::Literal(ScalarValue::Null, None));
+    // Spark's `%` (`Remainder`, `inputType = NumericType`) rejects a non-numeric operand on
+    // either side at analysis with DATATYPE_MISMATCH; DataFusion would instead reinterpret a
+    // boolean/date/timestamp as its raw integer or an interval as its raw nanos and compute a
+    // meaningless remainder. Reject those pairs at plan time, matching the sibling `/` (its
+    // divisor reject set — numeric-only, intervals included — is exactly `%`'s on both sides).
+    // Strings are coerced to a numeric type upstream under ANSI off; under ANSI on they arrive
+    // as `Utf8` and Spark rejects string arithmetic, so a string operand is rejected there too.
```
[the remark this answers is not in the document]
davidlghellin: changed

PR #2228, opened 2026-07-10
fix: match Spark ordering semantics for agg and window functions

shehabgamin:


crates/sail-plan/src/resolver/query/aggregate.rs
```diff
@@ -279,6 +281,56 @@ impl PlanResolver<'_> {
...
+            .map(|x| Self::attach_ordering_to_aggregates(x, ordering))
+            .transpose()?;
+        Ok((projections, having))
+    }
+
+    fn attach_ordering_to_aggregates(expr: Expr, ordering: &[SortExpr]) -> PlanResult<Expr> {
+        Ok(expr
+            .transform_down(|e| match e {
+                Expr::AggregateFunction(mut function)
+                    if Self::ORDER_SENSITIVE_AGGREGATES.contains(&function.func.name())
```
linhr: Is it possible to use `is` to compare by type instead of comparing by name?

linhr (approved):
Nice!

crates/sail-plan/src/resolver/query/aggregate.rs
shehabgamin: Done!

PR #2230, opened 2026-07-10
feat: add hms support for spark datasource tables

### LINKED ISSUE #2210: Add Spark datasource path support for HMS catalog
Sail should add Spark datasource path support for HMS catalog tables as the next step toward DSV2 interop.

Spark datasource tables encode location through datasource `path` metadata, so Sail should understand that shape when reading and writing HMS tables.

This should be validated through HMS interop tests where JVM Spark and Sail read each other’s tables through the same metastore and warehouse.

To avoid drift, interop tests should discover cross-engine metadata behavior, unit tests should pin the conversion rules, and docs should state the supported surface.

The supported datasource provider set should be explicit rather than inferred from broad format names.

The HMS docs should also be updated because they currently drift from implementation details around read support, write support, metadata-only detection, and unsupported table-format behavior.

santosh-d3vpl3x:
## Summary

Closes https://github.com/lakehq/sail/issues/2210.

This PR adds Spark datasource path support for HMS catalog tables and expands the HMS/Spark interop coverage so the behavior is pinned against a real Hive Metastore + MinIO + JVM Spark setup.

It also updates the HMS contributor docs to describe the supported datasource/metadata contract and adjusts fork CI behavior so Codecov uploads and docs deployment do not fail fork pushes that lack upstream-only credentials.

## Changes

- Resolve Spark datasource table locations from datasource path metadata when present.
- Keep the supported Spark datasource provider set explicit.
- Preserve useful metadata behavior for unsupported datasource providers instead of collapsing everything to unknown.
- Add HMS/Spark DSV2 interop tests covering Spark-written and Sail-written tables across supported formats.
- Document HMS compatibility, test ownership, and current unsupported surfaces.
- Make fork CI tolerate missing Codecov credentials while preserving strict uploads for `lakehq/sail`.
- Skip docs deploy outside `lakehq/sail`, where AWS deployment variables are not available.

## Verification

Latest fork CI on `santosh/main` is green:

- Gold Data
- Python Lint
- Rust Lint
- Build, including Rust tests, Python matrix, Ibis, catalog integration, Spark tests, and docs build

Run: https://github.com/santosh-d3vpl3x/sail/actions/runs/29075964796


crates/sail-catalog-hms/src/convert.rs
```diff
@@ -210,10 +201,10 @@ pub(crate) fn build_generic_table(
     let mut parameters = vec_to_map(properties);
     insert_comment(&mut parameters, comment);
     let table_parameters = parameters.get_or_insert_with(AHashMap::new);
-    if format.logical_format.eq_ignore_ascii_case("delta") {
+    if let Some(provider) = spark_datasource_provider(format.logical_format) {
```
shehabgamin, line 204: Spark records the provider and path for every datasource table, are we sure we want to gate on `spark_datasource_provider`?

crates/sail-catalog-hms/src/convert.rs
```diff
@@ -229,14 +220,18 @@ pub(crate) fn build_generic_table(
...
     );
-    let serde_parameters = location.as_ref().map(|location| {
-        let mut parameters = AHashMap::new();
-        parameters.insert(
-            FastStr::from_static_str("path"),
-            FastStr::from_string(location.clone()),
-        );
-        parameters
-    });
+    let serde_parameters = if is_spark_datasource_format(format.logical_format) {
```
shehabgamin, line 223: Spark records the provider and path for every datasource table, are we sure we want to gate on `is_spark_datasource_format`?

crates/sail-catalog-hms/src/convert.rs
```diff
@@ -470,13 +467,72 @@ pub(crate) fn extract_property(
...
-        "deltalake" => "delta".to_string(),
-        _ => provider,
-    })
+    match provider.as_str() {
+        "delta" | "deltalake" => Some("delta".to_string()),
+        "parquet" => Some("parquet".to_string()),
+        "csv" => Some("csv".to_string()),
+        "json" => Some("json".to_string()),
+        "avro" => Some("avro".to_string()),
+        _ => None,
```
shehabgamin: Why fallback to None?

crates/sail-catalog-hms/src/convert.rs
```diff
@@ -470,13 +467,72 @@ pub(crate) fn extract_property(
...
+        "json" => Some("json"),
+        "avro" => Some("avro"),
+        _ => None,
+    }
+}
+
+pub(crate) fn table_location(table: &Table) -> Option<String> {
+    let storage = table.sd.as_ref()?;
+    if extract_property(table.parameters.as_ref(), SPARK_DATASOURCE_PROVIDER_KEY).is_some()
+        && let Some(path) = storage
```
shehabgamin: I think empty is actually a valid value in Spark. May want to do something like this instead:
```
pub(crate) fn table_location(table: &Table) -> Option<String> {
      let storage = table.sd.as_ref()?;
      if extract_property(table.parameters.as_ref(), SPARK_DATASOURCE_PROVIDER_KEY).is_some() {
          return storage
              .serde_info
              .as_ref()
              .and_then(|serde| {
                  get_case_insensitive(serde.parameters.as_ref(), SPARK_DATASOURCE_PATH_KEY)
              })
              .map(ToString::to_string);
      }
      storage.location.as_ref().map(ToString::to_string)
}
```

shehabgamin (commented):
@santosh-d3vpl3x Thanks for this PR! I just have some minor comments

crates/sail-catalog-hms/src/convert.rs
santosh-d3vpl3x: Agreed.

crates/sail-catalog-hms/src/convert.rs
santosh-d3vpl3x: This was intentionally conservative, but I think we can now generalize HMS classification (given I trust interop tests to highlight issues well now):

- `table_type=iceberg`: Iceberg; `metadata_location` is its metadata pointer but not a discriminator by itself.
- `spark.sql.sources.provider` present: Spark datasource table; preserve the provider and use SerDe `path`.
- Otherwise: Hive SerDe table; detect the format from SerDe/InputFormat/OutputFormat and use `sd.location`.

Unsupported providers remain listable/describable and fail during planning only when `TableFormatRegistry` cannot resolve them. Known aliases may still be canonicalized, such as `deltalake -> delta`, because the registry only registers `delta`.

This removes the execution-based HMS gates and addresses the other provider-related comments.

santosh-d3vpl3x: Implemented in 4376756be. HMS classification is now metadata-driven: an explicit table_type=iceberg marker selects Iceberg; otherwise a present spark.sql.sources.provider selects Spark datasource representation and its SerDe path; otherwise Hive SerDe detection and sd.location apply. Unsupported providers remain listable and describable, and fail later only if planning needs an unregistered TableFormat. The deltalake alias is still canonicalized to delta because the registry registers delta.

crates/sail-catalog-hms/src/convert.rs
santosh-d3vpl3x: Removed this fallback behavior in 4376756be. Any present datasource provider is trimmed, lowercased, and preserved; only the explicit deltalake alias is canonicalized to delta. SerDe format detection now runs only when the datasource provider marker is absent.

crates/sail-catalog-hms/src/convert.rs
santosh-d3vpl3x: Removed the is_spark_datasource_format execution-support gate in 4376756be. Generic HMS writes now emit the provider and SerDe path for datasource representation independently of current execution support; textfile and Iceberg retain their explicit Hive SerDe and Iceberg metadata representations.

crates/sail-catalog-hms/src/convert.rs
santosh-d3vpl3x: Implemented in 4376756be. Datasource path lookup is case-insensitive, and a present value is authoritative even when it is empty. If a datasource provider is present but path is absent, Sail now returns no location rather than falling back to sd.location.

santosh-d3vpl3x:
Marking as draft until I address couple of more things.

crates/sail-catalog/src/hive_format.rs
```diff
@@ -80,7 +80,7 @@ pub struct HiveCatalogFormat {
 impl HiveCatalogFormat {
     pub fn from_format(format: &str) -> CatalogResult<Self> {
         match format.trim().to_ascii_lowercase().as_str() {
-            "delta" | "deltalake" => Ok(Self {
+            "delta" => Ok(Self {
```
shehabgamin, line 83: I think `"deltalake"` may be valid

shehabgamin (approved):
@santosh-d3vpl3x Well done!

I just have a comment on 1 line. This is good to merge in after that's addressed

crates/sail-catalog/src/hive_format.rs
santosh-d3vpl3x: I couldn't find any references in spark or various varients of delta that may update HMS. To reduce risk of creating special case, I removed it.

Any place specific we should check?

shehabgamin: @santosh-d3vpl3x I saw it referenced elsewhere in the Sail codebase, which led me to make that assumption.

I just realized the only other place `"deltalake"` is used is in the `sail-catalog-glue` crate, so I think you're right. Nice catch!

Could you do a quick cleanup of `sail-catalog-glue` to remove the remaining uses of `"deltalake"` from the codebase?

santosh-d3vpl3x: Done!

linhr:
Thanks for the CI changes! This seems helpful for development on the fork for the community.

shehabgamin (approved):

PR #2238, opened 2026-07-13
fix: measure BINARY in bytes for length, octet_length and bit_length

davidlghellin:
Spark measures the character length of string data and the byte length of binary data. Sail routed `length`/`len`/`char_length`/`character_length` to DataFusion's `CharacterLengthFunc` and `octet_length`/`bit_length` through an explicit UTF-8 validation, both of which accept string data only. Binary input was therefore coerced to UTF-8 first, which silently counted characters instead of bytes (`length(CAST('josé' AS BINARY))` returned 4, Spark returns 5) and rejected outright any binary that is not valid UTF-8 (`length(X'DEADBEEF')`), including binary columns and aggregates over them.

Dispatch on the argument type in the plan builder instead: binary reaches the function untouched, collections are rejected as Spark rejects them (they were silently stringified — `length(array(1, 2, 3))` returned 9), and any other type is measured as its string form through `SparkToUtf8`. Casting through Arrow, as before, renders a timestamp with a time zone suffix, so `length` measured 25 characters where Spark measures 19, even though Sail's own `CAST(ts AS STRING)` is correct.

`length` and its aliases now use datafusion-spark's `SparkLengthFunc`, which measures binary and always returns INT. DataFusion has no binary-aware `octet_length`/`bit_length`, so add both, delegating to the Arrow kernels, which measure from the offset buffer and already wrap on overflow the way Spark does (the bit length of more than 2^28 bytes wraps to i32::MIN in both engines). The kernels return BIGINT for large offsets, so normalize to INT, truncating rather than nullifying to preserve that wrap.

Every expected value is validated against Spark JVM. The feature also pins the output schema, which a values-only test cannot see: Spark returns INT for every input type, whatever the width of the input's offsets.

The two remaining divergences are pinned as `@sail-bug`: the string form Sail renders for a double in scientific notation and for an interval differs from Spark, so `length` measures the right thing about the wrong string. The root cause is the CAST-to-string formatter, not this family; the strict xfail will XPASS and force the markers off once that is fixed.

### DIFF
--- a/crates/sail-execution/src/proto/codec.rs
+++ b/crates/sail-execution/src/proto/codec.rs
@@ -73,6 +73,7 @@ use datafusion_spark::function::math::hex::SparkHex;
 use datafusion_spark::function::math::width_bucket::SparkWidthBucket;
 use datafusion_spark::function::string::elt::SparkElt;
 use datafusion_spark::function::string::format_string::FormatStringFunc;
+use datafusion_spark::function::string::length::SparkLengthFunc;
 use datafusion_spark::function::string::luhn_check::SparkLuhnCheck;
 use datafusion_spark::function::url::try_url_decode::TryUrlDecode;
 use datafusion_spark::function::url::url_decode::UrlDecode;
@@ -213,6 +214,7 @@ use sail_function::scalar::string::soundex::Soundex;
 use sail_function::scalar::string::spark_base64::{SparkBase64, SparkUnbase64};
 use sail_function::scalar::string::spark_concat_ws::SparkConcatWs;
 use sail_function::scalar::string::spark_encode_decode::{SparkDecode, SparkEncode};
+use sail_function::scalar::string::spark_length::{SparkBitLength, SparkOctetLength};
 use sail_function::scalar::string::spark_mask::SparkMask;
 use sail_function::scalar::string::spark_quote::SparkQuote;
 use sail_function::scalar::string::spark_regexp_extract_all::{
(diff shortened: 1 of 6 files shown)


shehabgamin:
@davidlghellin It appears that this modifies the planning of some clickbench queries.

Can you do a local benchmark and see if the affected queries have any difference in performance?

davidlghellin:
> @davidlghellin It appears that this modifies the planning of some clickbench queries.
> 
> Can you do a local benchmark and see if the affected queries have any difference in performance?

Sure — I'll benchmark the affected ClickBench queries locally (main vs this
branch) and post the before/after timings here.

davidlghellin:
> @davidlghellin It appears that this modifies the planning of some clickbench queries.
> 
> Can you do a local benchmark and see if the affected queries have any difference in performance?

Done — benchmarked locally (main vs this branch), 1M rows, on the exact path this PR changes:
`length(BINARY)` moving from `character_length` (which forces an implicit `CAST(binary→utf8)`) to
`SparkLengthFunc` (byte length off the offset buffer). Q28/Q29 (`AVG(length(URL))` /
`AVG(length(Referer))`, BINARY columns) hit this path.

**Time (criterion, 1M rows)**

| case | time |
|---|---|
| BINARY — old (`character_length` + implicit cast) | 3.56 ms |
| BINARY — cast alone (`cast_binary_to_utf8`) | 1.54 ms |
| **BINARY — new (`SparkLengthFunc`)** | **0.32 ms** |
| STRING — old (`character_length`) | 2.01 ms |
| STRING — new (`SparkLengthFunc`) | 2.13 ms |

image


 https://github.com/davidlghellin/sail-benchmarks/pull/1

shehabgamin (approved):
Nice!!

PR #2239, opened 2026-07-14
feat: support the DataFrame inputFiles operation

### LINKED ISSUE #481: PySpark DataFrame Doctest: inputFiles
`pyspark/sql/dataframe.py::pyspark.sql.dataframe.DataFrame.inputFiles`

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.inputFiles.html

fangchenli:
Closes #481

Implements DataFrame.inputFiles(), whose Spark Connect analyze handler was an unimplemented stub that made every call fail. Enumeration lives in sail-session, which inspects the optimized plan and unions the files of each constituent listing relation (Parquet/JSON/CSV/binary/…), honoring format-level filters like the binary pathGlobFilter; hidden files (_SUCCESS, .crc, and anything under a hidden directory) are excluded via a shared predicate in list_all_files, so listing scans and inputFiles agree. Paths are returned as deduplicated, percent-encoded URIs that match Spark's semantics, so the upstream PySpark doctest passes. Delta support is deferred to a follow-up under #171


crates/sail-data-source/src/listing/utils.rs
```diff
@@ -225,16 +226,80 @@ pub async fn list_all_files<'a>(
...
+/// The per-name rule mirrors `InMemoryFileIndex.shouldFilterOut` in Spark: a name is
+/// hidden when it starts with `_` (unless it contains `=`, i.e. a partition
+/// directory) or with `.`, except for the Parquet summary files `_metadata` and
+/// `_common_metadata`.
+pub fn has_hidden_path_component(url: &ListingTableUrl, location: &Path) -> bool {
+    let is_hidden = |name: &str| {
+        let exclude = (name.starts_with('_') && !name.contains('=')) || name.starts_with('.');
+        let keep = name.starts_with("_common_metadata") || name.starts_with("_metadata");
+        exclude && !keep
+    };
```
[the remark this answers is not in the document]
fangchenli: My impl matches Spark. See https://github.com/apache/spark/blob/v4.1.1/core/src/main/scala/org/apache/spark/util/HadoopFSUtils.scala#L346-L358

crates/sail-data-source/src/listing/input_files.rs
linhr, line 1: The `sail-session` crate is for general session logic, so I feel this is more related to `sail-data-source`.

Could you move this to e.g., `crates/sail-data-source/src/listing/utils.rs`? I think there will be no circular dependency among crates, but let me know if I missed anything.

crates/sail-data-source/src/listing/input_files.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let mut files: Vec<String> = vec![];
+    // List each (path, filter) once so a repeated scan (e.g. `df.union(df)`) isn't re-listed.
+    let mut listed: HashSet<(String, Option<String>)> = HashSet::new();
+
+    for source in &listing_sources {
+        // Apply the format-level name filter (e.g. binary `pathGlobFilter`).
+        let raw_glob = source.config().read_format.input_file_name_glob();
+        let name_glob = raw_glob
+            .map(Pattern::new)
```
linhr, line 42: We have our own glob logic in `GlobUrl` to support patterns not understood by `glob::Pattern`. Could you add a test case around alternation (`{a,b}`)?

crates/sail-data-source/src/listing/input_files.rs
```diff
@@ -0,0 +1,82 @@
...
+            .map_err(|e| DataFusionError::Plan(format!("invalid path glob filter: {e}")))?;
+        for table_path in &source.config().table_paths {
+            if !listed.insert((
+                table_path.as_str().to_string(),
+                raw_glob.map(str::to_string),
+            )) {
+                continue;
+            }
+            let store = ctx.runtime_env().object_store(table_path)?;
+            let base = Url::parse(table_path.object_store().as_str())
```
linhr, line 53: Does this work with absolute path on Windows?

You could try the `multi-platform-build.yml` workflow to run the tests on Windows for your fork.

linhr (commented):
Thanks for the contribution!

I have a few thoughts, but this overall makes sense to me!

linhr:
@fangchenli Let me know if you have time to wrap up this PR, or if there is anything that we can help with!

fangchenli:
> @fangchenli Let me know if you have time to wrap up this PR, or if there is anything that we can help with!

Thanks for checking. I was on vacation for the past few days and plan to pick this up tomorrow.

crates/sail-data-source/src/listing/input_files.rs
fangchenli: I feel like this deserves its own module instead of being inside `utils.rs`. But I could move it to `utils.rs` if it's preferred.

crates/sail-data-source/src/listing/input_files.rs
fangchenli: It seems to work on Windows. https://github.com/fangchenli/sail/actions/runs/29878010629

crates/sail-data-source/src/listing/input_files.rs
linhr: Yeah a dedicated file within `sail-data-source` sounds good to me!

crates/sail-data-source/src/listing/input_files.rs
linhr: I realized our existing implementation of `pathGlobFilter` does not use `GlobUrl` but uses `glob::Pattern` so alternation is indeed unsupported. I am glad that you added the test case to reflect this behavior. 

I created an issue #2275 to track improvements around this data source option.

linhr (approved):
This looks good to me! Thanks again for having this!

PR #2241, opened 2026-07-14
test: cover behaviour-governing arguments passed as a column

davidlghellin:
Our Spark-semantics fixes live in the plan builder and match on `Expr::Literal`. When the same value arrives as a column, that code never runs and the query falls through to the raw DataFusion function, which usually demands a scalar. Literals are also constant-folded, so a scenario written with a literal often never exercises the columnar kernel at all -- the path production actually takes.

The scenarios are derived from Spark's own `DESCRIBE FUNCTION EXTENDED` examples, with one literal argument at a time lifted into a two-row column. Every expected value was captured on Spark JVM 4.1.1, and every divergence was re-checked with the argument inlined as a literal, so that a plain value divergence is not misreported as a column bug.

Adds 129 scenarios over 37 feature files, all tagged `@column_args`. Each function also gets an untagged literal control that is green on both engines, so the file itself shows the gap: the literal passes, the column does not. The 88 divergent scenarios are tagged `@sail-bug` and carry a comment stating what Sail does today: 8 apply the first row's argument to every row (wrong data, no error), 3 return NULL for every row, 18 reject the column outright, and 14 accept a column where Spark requires a foldable argument.

No production code is changed.

### DIFF
--- a/python/pysail/tests/spark/function/features/aes_decrypt.feature
+++ b/python/pysail/tests/spark/function/features/aes_decrypt.feature
@@ -1,0 +1,52 @@
+Feature: aes_decrypt with an argument coming from a column
+  # A behaviour-governing argument given as a literal is constant-folded, so the literal
+  # scenarios never exercise the columnar kernel. These scenarios pass the same argument
+  # through a column. All expected values were captured on Spark JVM 4.1.1.
+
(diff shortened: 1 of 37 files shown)


shehabgamin (approved):
Nice!

PR #2244, opened 2026-07-14
perf: optimize queries with linear sums

lonless9:
## Summary

Enable [DataFusion's linear aggregate optimization](https://github.com/apache/datafusion/issues/15524) for Spark `SUM` expressions used by ClickBench Q29.

We initially produces `SUM(CAST(x + c AS BIGINT))`, whose outer cast prevents DataFusion from recognizing the `SUM(x + c)` pattern. This PR rewrites the argument as `SUM(CAST(x AS BIGINT) + c)` when the original `Int32` addition cannot overflow.

DataFusion can then rewrite:

`SUM(x + c) → SUM(x) + c * COUNT(x)`

This reduces Q29 from 90 aggregate accumulators to one `SUM` and one `COUNT`, with the remaining expressions evaluated in the final projection.

The ClickBench plan snapshot is updated accordingly.


linhr (approved):
This is great!

PR #2254, opened 2026-07-16
fix: resolve `next_day` day names per row and stop trimming them

davidlghellin:
next_day handled (scalar,scalar), (array,scalar) and (array,array) but not (scalar,array), so a literal date with the day name in a column fell through to a catch-all error leaking Rust Debug output. Share the array path between both cases, which also makes the match exhaustive and removes the catch-all.

parse_day_of_week trimmed the day name; Spark does not, so 'MON ' must be invalid like any unknown name. Matching stays case-insensitive.


shehabgamin (approved):
Looks great!

PR #2255, opened 2026-07-16
fix: validate CSV expression function options like Spark

davidlghellin:
`from_csv`, `to_csv` and `schema_of_csv` silently ignored every option they did not themselves use, so a bad value was accepted where Spark rejects the call: `from_csv('1', 'a INT', map('header', 'garbage'))` returns `{1}` in Sail and fails in Spark.

Each `from_map` was an allowlist — it pulled out the keys it knew and dropped the rest. Spark instead builds `CSVOptions` eagerly, so every option is parsed as soon as the map is read, even by a function that never uses the resulting value: `from_csv` rejects a bad `quoteAll` and `to_csv` rejects a bad `header`. Validation now lives in one shared place that every function routes through.

Each behaviour below was measured against Spark 4.1.1 rather than inferred:

- boolean options, including `multiLine`, which Spark reads with Scala's `String.toBoolean` and therefore reports differently to the other seven
- single-character options, sized with Java's UTF-16 `String.length`: an emoji is two units and is rejected, `ñ` is one and is accepted
- an empty `sep`, and an empty `lineSep`, which only the writer rejects
- numeric and enum options, and `DROPMALFORMED`, which only `from_csv` rejects
- NULL option values, which Spark rejects even under a key it does not know
- option keys, matched case-insensitively and reported by their canonical name

The three copies of the map lookup had each drifted apart: one compared keys exactly, one guarded a NULL value and another did not, one built its default eagerly. They are now a single `find_option`, which makes the NULL bug unrepresentable rather than merely guarded.

Ten scenarios are tagged `@sail-bug` for divergences this does not fix: the parse mode (Sail reads no `mode`, so it always aborts where Spark's default PERMISSIVE nulls the failed field and keeps the good ones), the extra `.trim()` on every token, and `schema_of_csv` ignoring `dateFormat` and a time-less `timestampFormat`. They pass against the JVM and fail against Sail, so CI reports an XPASS the day any of them is fixed.


davidlghellin:
This is the CSV slice of a systemic gap: from_json/to_json and from_xml/to_xml build their options the same allowlist way and drop bad values just as silently. They are deliberately left out — each family is its own PR, with its own JVM sweep, since the option sets and error messages differ.

crates/sail-function/src/scalar/csv/options.rs
```diff
@@ -0,0 +1,199 @@
...
+        .column_by_name(SAIL_MAP_KEY_FIELD_NAME)?
+        .as_any()
+        .downcast_ref::<StringArray>()?;
+    let values = map
+        .entries()
+        .column_by_name(SAIL_MAP_VALUE_FIELD_NAME)?
+        .as_any()
+        .downcast_ref::<StringArray>()?;
+    Some((keys, values))
+}
```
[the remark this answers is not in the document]
davidlghellin: Good catch — confirmed: make_scalar_function chains Hint::Pad, so the literal map is expanded to one copy per row and validate_options was walking the same options once per row.

shehabgamin:
```
The added cases are useful, but the implementation is not yet Spark-equivalent and introduces a null-handling regression.

  ## Findings

  1. P1 — The validator omits options Spark validates eagerly.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:29 and /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:140 use a partial allowlist. Missing
     options include preferDate, columnPruning, enableDateTimeParsingFallback, ignoreCorruptFiles, ignoreMissingFiles, inputBufferSize, encoding/charset, compression/codec, timeZone, and extension.

     For example, Spark rejects:

     SELECT from_csv('1', 'a INT', map('preferDate', 'garbage'))

     with a Boolean-option error, while this implementation does not validate preferDate and proceeds. Similar discrepancies occur for invalid encodings, compression codecs, time zones, and buffer sizes. Spark
     parses these in opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVOptions.scala:107 and inherited file options.

  2. P1 — Eager validation breaks Spark’s null propagation.
     Options are validated before inspecting the input at /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/spark_from_csv.rs:233 and /private/tmp/sail-pr2255-review/crates/sail-function/src/
     scalar/csv/spark_to_csv.rs:264.

     Spark returns NULL for both:

     SELECT from_csv(
       CAST(NULL AS STRING),
       'a INT',
       map('header', 'garbage')
     );

     SELECT to_csv(
       CAST(NULL AS STRUCT<a: INT>),
       map('lineSep', '')
     );

     The PR raises an option error before null propagation. Spark creates the parser/writer lazily inside the non-null evaluation path; see opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/
     expressions/csv/CsvExpressionEvalUtils.scala:52. Validation of option values should therefore wait until a non-null input is encountered. Structural errors such as null option-map entries must remain eager
     because Spark rejects those during map conversion.

  3. P1 — Duplicate keys, casing, and aliases do not follow Spark’s effective-map semantics.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:96 finds the first case-insensitive match, while /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/
     options.rs:121 validates every raw entry. Spark first creates a CaseInsensitiveMap, where the last case-variant wins, and then reads canonical options in a fixed order.

     Confirmed differences include:
      - map('header','garbage','HEADER','true'): Spark succeeds; the PR rejects the shadowed first value.
      - map('quoteAll','false','QUOTEALL','true'): Spark uses true; the PR retrieves false.
      - map('sep','|','delimiter',''): Spark uses sep and ignores the shadowed invalid alias; the PR rejects delimiter.
      - /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/schema_of_csv.rs:195 manually overwrites sep with a later delimiter, whereas Spark gives sep canonical priority.

     The cleanest fix—and the main simplification identified by the ponytail review—is one normalized, case-insensitive effective option map reused by all three expressions, with Spark’s alias precedence encoded
     once.

  4. P2 — Delimiter parsing does not implement Spark’s escape grammar.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:154 checks only character count and callers retain the raw string. Spark runs separators through CSVExprUtils.toDelimiterStr.

     Consequently:
      - \q is rejected by Spark as an unsupported special character but accepted by this validator.
      - \t means an actual tab in Spark but remains an unconverted two-character value here.

     See opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVExprUtils.scala:120.

  5. P2 — samplingRatio uses Rust’s float grammar instead of Java’s.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:162 uses Rust f64 parsing. Spark uses Java/Scala Double.parseDouble semantics.

     Spark accepts values such as " 1 ", "1d", and "0x1.0p0" that this implementation rejects. Conversely, Rust accepts "inf", which Spark rejects.

  6. P2 — lineSep length validation is missing for to_csv.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:178 checks only for an empty value. Spark’s writer rejects separators longer than one Unicode character—implemented as more than
     two UTF-16 code units—so lineSep='abc' must fail. Reader expressions intentionally override/ignore this option, meaning the validation must be expression-specific. Spark’s behavior is in opt/spark/sql/
     catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/CSVOptions.scala:264.

  7. P2 — Nonliteral or non-string option maps still bypass Spark analysis rules.
     /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:65 reads only row zero and silently treats a failed StringArray downcast as no options. The expression coercion accepts any map
     type.

     Spark requires a foldable map<string,string> expression and rejects option columns, map_from_arrays, and non-string maps during analysis. Sail can instead apply one row’s options globally or silently fall
     back to defaults. Although some acceptance predates this PR, the new common validator relies on that incorrect invariant and therefore cannot provide the claimed Spark validation parity.

  8. P3 — ASCII-only case folding differs from Spark for Unicode values.
     The eq_ignore_ascii_case checks at /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:143 and /private/tmp/sail-pr2255-review/crates/sail-function/src/scalar/csv/options.rs:211
     differ from Java/Scala Unicode case handling. For example, Spark accepts the long-s forms multiLine='falſe' and unescapedQuoteHandling='ſkip_value'; the PR rejects them. This is obscure but observable, and
     this project requires exact parity.
```

davidlghellin:
> @davidlghellin I had Codex 5.6 ultra do a first pass. Let me know if any of this is relevant or not. If it is, please re-ping me when ready for re-review!


Thanks for the thorough review — and for running Codex on it. I went through all 8 findings against Spark JVM, measuring each one rather than trusting the diff. Answering in your order:

**1. (P1) Validator omits options Spark validates eagerly**
→ **Partially fixed.** Added the boolean/int ones (`preferDate`, `columnPruning`, `inputBufferSize`, `enableDateTimeParsingFallback`, `ignoreCorruptFiles`, `ignoreMissingFiles`). The domain ones (`encoding`/`compression`/`timeZone`) are deferred with a `@sail-bug` — they need charset/zone/codec validation.

**2. (P1) Eager validation breaks Spark's null propagation** *(the regression)*
→ **Fixed.** Value validation is now lazy (inside the non-null path); the structural NULL-entry check stays eager, matching `CsvExpressionEvalUtils`' lazy parser and `convertToMapData`'s eager NPE.

**3. (P1) Duplicate keys / casing / aliases don't follow the effective-map semantics**
→ **Fixed.** Read-driven `CaseInsensitiveMap`: last case-variant wins, `sep` takes precedence over `delimiter`, shadowed values are never validated. This was the main simplification you flagged, and it made the code shorter.

**4. (P2) Delimiter parsing doesn't implement Spark's escape grammar** *(`\t` ≠ tab, `\q`)*
→ **Deferred** (`@sail-bug`). Needs a full `CSVExprUtils.toDelimiterStr` port — validation and unescaping are coupled.

**5. (P2) samplingRatio uses Rust's float grammar instead of Java's**
→ **Deferred** (`@sail-bug`). I measured Java `Double.parseDouble` vs Rust `f64`: 8 divergent cases (whitespace, `d`/`f` suffix, case-sensitive `Infinity`/`NaN`, hex floats). A partial parser would introduce new divergences, so it deserves its own change.

**6. (P2) lineSep length validation missing for to_csv**
→ **Fixed.** `encode_utf16().count() > 2`, so `ab` is accepted and `abc` rejected.

**7. (P2) Non-literal / non-string option maps bypass Spark's analysis rules**
→ **Deferred** (`@sail-bug`). This is analysis-time rejection, which lives in `sail-plan`, not the expression.

**8. (P3) ASCII-only case folding differs from Spark for Unicode**
→ **Fixed.** `getBool` folds ASCII-only (`toLowerCase(ROOT)`); `toBoolean`/enum options fold Unicode (`equalsIgnoreCase`).

---

**Summary: 2, 3, 6, 8 fully fixed · 1 partially · 4, 5, 7 deferred** — each deferred item is tracked with a red `@sail-bug` (passes on JVM, fails on Sail), so nothing is lost. Happy to fold any of them into this PR if you'd prefer — just say the word. Re-ping when ready for re-review.

crates/sail-function/src/scalar/csv/options.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+    // Spark reads `sep`, falling back to `delimiter`; only the value it would read is validated.
+    if let Some(value) = find_option(map, "sep").or_else(|| find_option(map, "delimiter"))
+        && value.is_empty()
+    {
+        return exec_err!("Delimiter cannot be empty");
+    }
+    for option in INT_OPTIONS {
+        if let Some(value) = find_option(map, option)
+            && value.parse::<i32>().is_err()
```
shehabgamin, line 231: Is it possible to use Java/Scala numeric parsing here?

crates/sail-function/src/scalar/csv/options.rs
```diff
@@ -0,0 +1,270 @@
...
+/// from the case-insensitive map (last case-variant wins, via [`find_option`]), NOT by scanning raw
+/// entries. A shadowed value — an earlier case-variant, or the losing side of an alias such as
+/// `delimiter` when `sep` is present — is never read and therefore never validated.
+///
+/// Spark builds `CSVOptions` eagerly relative to other options — so `from_csv` rejects a bad
+/// `quoteAll` and `to_csv` rejects a bad `header` — but LAZILY relative to the input: a bad value is
+/// not seen when every input row is NULL. Callers therefore invoke this only when the input has at
+/// least one non-null row; the structural NULL-entry check ([`reject_null_entries`]) runs
+/// unconditionally instead. Unknown keys are ignored, also matching Spark.
+pub(super) fn validate_options(map: &MapArray, function: CsvFunction) -> Result<()> {
```
shehabgamin, line 199: I think Spark also validates extension, charset, time zone, compression codec, etc. But this can be followup work.

shehabgamin (commented):
Well done! I only have a couple of comments

crates/sail-function/src/scalar/csv/options.rs
davidlghellin: Good catch — I measured it, and there is a real divergence, though a narrow one:

| value | Java | Rust `parse::<i32>()` |
|---|---|---|
| `٥` (Arabic-Indic), `१२` (Devanagari) | accepts | rejects |
| `" 5 "`, `"5 "` | rejects | rejects ✓ |
| `+5` | accepts | accepts ✓ |
| `2147483648` (overflow) | rejects | rejects ✓ |

Whitespace, sign and overflow already match. The only gap is **Unicode decimal digits**: Java's `Integer.parseInt` goes through `Character.digit` (Unicode Nd), while Rust's parse is ASCII-only. Matching it isn't a one-liner — Rust's `char::to_digit` is ASCII-only too, so it needs an Nd table or a new dependency.

**Added a test for it** (`@sail-bug`, so CI tracks it):

```gherkin
@sail-bug
Scenario: from_csv accepts a Unicode decimal digit in an integer option
  SELECT from_csv('1', 'a INT', map('maxColumns', '٥')) AS result
  → | {1} |

File: crates/sail-function/src/scalar/csv/options.rs (hunk 3 above)
Comment (davidlghellin) [reply crates/sail-function/src/scalar/csv/options.rs:199]: Agreed — `encoding`/`charset`, `timeZone` and `compression` are tracked with `@sail-bug` scenarios so CI keeps an eye on them, and they're on the followup list.

One caveat on `extension` specifically: I measured it and it looks like a Spark quirk rather than validation we should mirror. It's a file-writer option (added by [SPARK-50616](https://issues.apache.org/jira/browse/SPARK-50616) for the CSV DataSource writer), not an expression option. Passing it to `to_csv` accepts `ab`/`abc`/`abcd` but throws `INTERNAL_ERROR` on a non-letter value like `a1` — and `INTERNAL_ERROR` is Spark's "this shouldn't happen" class, not a user-facing validation. So I'd lean toward leaving Sail ignoring it as an unknown option on the expression path, rather than replicating it. Happy to revisit if you'd rather match it.

File: crates/sail-function/src/scalar/csv/options.rs [hunk 4]
```diff
@@ -0,0 +1,316 @@
...
+        && let Some(value) = find_option(map, MODE_OPTION)
+        && fold_eq(value, DROP_MALFORMED_MODE)
+    {
+        return exec_err!(
+            "The function `from_csv` doesn't support the {DROP_MALFORMED_MODE} mode. Acceptable modes are PERMISSIVE and FAILFAST."
+        );
+    }
+    // Spark's `CSVOptions` validates the charset against a fixed set, matched case-insensitively
+    // (Spark lower-cases with `Locale.ROOT`), and rejects everything else — narrower than Java's.
+    for option in CHARSET_OPTIONS {
```
shehabgamin: I think we want to do something like this:
```
let charset = find_option(map, "encoding").or_else(|| find_option(map, "charset"));
```

I believe that Spark treats `charset` as a fallback alias for `encoding`. It doesn't validate both values when both are present:
```
spark-sql (default)> SELECT from_csv('1', 'a INT', map('encoding', 'utf-8', 'charset', 'utf-99'));
{"a":1}
Time taken: 1.155 seconds, Fetched 1 row(s)
spark-sql (default)>
```

shehabgamin (commented):
@davidlghellin Just 1 small change and then this is good to go!

crates/sail-function/src/scalar/csv/options.rs
davidlghellin: Done — good catch! Switched to the effective-value fallback exactly as you suggested (wrapped in a small find_option_with_alias helper, so it reads find_option(map, "encoding").or_else(|| find_option(map, "charset"))). Spark aliases compression/codec the same way (CSVOptions.scala:118, 155), so I applied the same fix there too, and added scenarios for the shadowing case — e.g. map('encoding', 'utf-8', 'charset', 'utf-99') → accepted, map('compression', 'gzip', 'codec', 'garbage') → accepted.

shehabgamin (approved):
Nice!!

PR #2256, opened 2026-07-17
test: fix Delta location for Windows

linhr:
It seems that for BDD tests we need `LOCATION {{ location.sql }}` for Delta tables but `LOCATION {{ location.uri }}` for Iceberg tables.


lonless9 (approved):
Nice!

PR #2259, opened 2026-07-17
fix: spark-compatible xpath — foldable path, first-node semantics, XPath 1.0 numbers

davidlghellin:
The xpath family diverged from Spark in five ways. All expected values were captured on Spark JVM 4.1.1, and an 81-probe sweep of the family now agrees with Spark on 80 of them (the remaining one is namespaces, see below).

1. The path must be foldable. Spark compiles the XPath once, when it analyzes the query, so it requires `path` to be a foldable expression and rejects a per-row one with DATATYPE_MISMATCH.NON_FOLDABLE_INPUT. Sail accepted a column and evaluated it row by row. Note that foldable is not the same as literal: Spark's rule is "no column references, and deterministic", so `concat('a', '/b')` and `substring('zza/b', 3)` are both valid. Gating on `Expr::Literal` -- or on "can I evaluate it right now", which fails whenever the argument types still need coercion -- would reject queries Spark accepts. Only the path is gated; the XML is an ordinary column.

2. A path matching several nodes now takes the first one. `xpath_string('<a><b>b1</b><b>b2</b></a>', 'a/b')` errored with XPTY0004: XPath 2.0 rejects a sequence of more than one item, while Spark evaluates with XPath 1.0 and takes the first node. Same for xpath_int / xpath_double / xpath_float / xpath_number.

3. xpath_short wraps instead of saturating. Spark narrows the double the Java way, `(short) (int) value`: double -> int saturates, and int -> short then truncates the bits, so 99999 comes out as -31073. Rust's `as i16` was saturating at 32767. xpath_int and xpath_long already agreed (both saturate).

4. Numbers follow XPath 1.0's lexical space. XPath 1.0's number grammar is only [-]digits[.digits]: no exponent, no INF. Anything else is NaN. Sail read `INF` as Infinity and `1.5e2` as 150.0, because xee implements XPath 2.0. The integer variants inherit it: in Spark `xpath_int('<a><b>INF</b></a>', 'a/b')` is 0, because the value is NaN and Java's (int) NaN is 0. A node is now converted through its string value and parsed by the 1.0 rules, while an atomic result (`sum()`, `count()`) stays a number and keeps its precision.

5. A pre-existing test asserted Sail's own error wording ("Invalid XML") and so failed against Spark JVM. It now accepts either engine's wording, while still requiring the error to be about the document.

Tests: 40 new scenarios across the family, all validated against Spark JVM 4.1.1 -- 78 pass there, and on Sail 77 pass with 3 xfail. Plus two Sail-only plan snapshots pinning the invariant the gate creates: the path reaches the function as a literal, whether the user wrote a literal or a constant expression.

Known divergence, left as @sail-bug with its own follow-up: Spark parses the document without namespace awareness, so an unprefixed path still matches a node in a default namespace, while Sail (XPath 2.0) honours it and comes back empty. Every real-world document carrying an `xmlns` -- SOAP, RSS, SVG, Atom -- is affected, and none of the ways out is free (rewriting the path to `*:a/*:b` means parsing the XPath expression; stripping the xmlns breaks prefixed documents), so it needs a decision of its own.

### DIFF
--- a/crates/sail-function/src/scalar/xml/xpath_typed.rs
+++ b/crates/sail-function/src/scalar/xml/xpath_typed.rs
@@ -8,7 +8,7 @@ use datafusion::arrow::datatypes::DataType;
 use datafusion_common::utils::take_function_args;
 use datafusion_common::{DataFusionError, Result, plan_err};
 use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility};
-use xee_xpath::{Documents, Queries, Query};
+use xee_xpath::{Documents, Item, Queries, Query};
 
 use crate::functions_utils::make_scalar_function;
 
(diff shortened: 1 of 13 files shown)


shehabgamin:
```
  ### Findings

  1. P1 — Atomic strings and booleans regress to NaN.
     xpath_typed.rs:201 (https://github.com/lakehq/sail/blob/aadf240c08cd81438c23d6d97e6368567c2544de/crates/sail-function/src/scalar/xml/xpath_typed.rs#L199-L202) uses try_into_value::<f64>(), which xee supports
     only for numeric atomics. Consequently:
      - xpath_double(..., 'a/b = 10'): Spark 1.0, PR NaN
      - xpath_int(..., 'a/b = 10'): Spark 1, PR 0
      - xpath_double(..., 'string(a/b)'): Spark 10.0, PR NaN

     This is a direct regression from the previous number(path) conversion.

  2. P1 — XPath 1.0 conversion is applied too late.
     xpath_typed.rs:191-202 (https://github.com/lakehq/sail/blob/aadf240c08cd81438c23d6d97e6368567c2544de/crates/sail-function/src/scalar/xml/xpath_typed.rs#L191-L202) only applies xpath1_number when the final xee
     result remains a node. Internal XPath functions still use xee’s XPath 2 rules:
      - number(a/b) and sum(a/b) over 1.5e2: Spark NaN, PR 150.0
      - number(a/b) over two nodes: Spark converts the first node; PR raises XPTY0004

     Thus the advertised number and first-node parity does not hold inside ordinary XPath expressions.

  3. P1 — The foldability check admits non-foldable, potentially per-row paths.
     xml.rs:27 (https://github.com/lakehq/sail/blob/aadf240c08cd81438c23d6d97e6368567c2544de/crates/sail-plan/src/function/scalar/xml.rs#L26-L40) equates “no Expr::Column and nonvolatile” with Spark foldability.
     It admits OuterReferenceColumn, aggregates, windows, scalar subqueries, and deterministic UDFs; line 40 then preserves them when literal evaluation fails. Spark rejects all of these with NON_FOLDABLE_INPUT. A
     correlated outer reference can therefore make Sail’s path vary per row, defeating the central invariant of this PR.

  4. P2 — The [1] wrapper accepts XPath 2 syntax Spark rejects.
     xpath_typed.rs:191,218 (https://github.com/lakehq/sail/blob/aadf240c08cd81438c23d6d97e6368567c2544de/crates/sail-function/src/scalar/xml/xpath_typed.rs#L191-L218) turns path 1, 2 into (1, 2)[1]. Spark rejects
     the comma sequence as invalid XPath; the PR returns 1.0 or "1". The pre-PR wrappers errored.

  5. P2 — Unicode whitespace is incorrectly accepted around numbers.
     xpath_typed.rs:171 (https://github.com/lakehq/sail/blob/aadf240c08cd81438c23d6d97e6368567c2544de/crates/sail-function/src/scalar/xml/xpath_typed.rs#L170-L179) uses Rust str::trim, which removes Unicode
     whitespace. XPath 1.0 only trims space, tab, CR, and LF. Spark returns NaN for NBSP, U+2003, or NEL around 42; the PR returns 42.
```

PR #2260, opened 2026-07-18
fix: honor Spark's lambda short-circuit in `exists` and `forall` under ANSI

davidlghellin:
Spark's `ArrayExists`/`ArrayForAll` walk the elements in order and stop at the
first `true`/`false`, so an element past the stopping point is never evaluated
and can never raise. Sail evaluates the whole flattened batch in one vectorized
pass, so it raises on elements Spark would have skipped:

```
SELECT exists(array(1, 0, 2), x -> 10 / x > 4)   -- ANSI on
-- Spark: true (stops at the first element, never divides by the 0)
-- Sail:  ERROR Division by zero
```

Add `short_circuit_boolean_reduce` to `lambda_utils`, shared by both functions
and parameterized by the stopping value, and fall back to it only when the
vectorized evaluation returns an error. The happy path is unchanged, and the
recovery path reproduces Spark's evaluation order exactly, one element at a
time, per row.

The fallback needs no ANSI plumbing and no proto change: with
`spark.sql.ansi.enabled = false` the predicate yields NULL instead of raising,
so the vectorized pass succeeds and the fallback is never reached. The error
itself is the gate, and it carries no state that would have to reach a worker —
the recovery lives entirely inside `invoke_with_args`.

Sail's errors were a superset of Spark's and the values never differed — the
reduce only reads the prefix up to the stopping point — so recovering from the
error is a complete fix rather than a partial one.

Cover both functions with column-borne arrays, which neither feature file
exercised before (every scenario used an array literal, so the multi-row
columnar kernel was never hit), and pin the short-circuit order with ANSI
on/off pairs asserting opposite outcomes for the same query. Five of the new
scenarios assert that the query still fails with Spark's own message, so a
fallback that swallowed legitimate errors would not pass.

exists: 35 -> 54 scenarios, forall: 33 -> 50, all verified against Spark JVM.

### DIFF
--- a/crates/sail-function/src/scalar/array/lambda_utils.rs
+++ b/crates/sail-function/src/scalar/array/lambda_utils.rs
@@ -10,12 +10,17 @@
 /// (`index_array`/`offsets_to_indices`).
 use std::sync::Arc;
 
-use datafusion::arrow::array::{Array, ArrayRef, AsArray, Int32Array, OffsetSizeTrait};
+use datafusion::arrow::array::{
+    Array, ArrayRef, AsArray, BooleanArray, BooleanBuilder, Int32Array, OffsetSizeTrait,
+};
 use datafusion::arrow::buffer::OffsetBuffer;
+use datafusion::arrow::compute::take_arrays;
 use datafusion::arrow::datatypes::DataType;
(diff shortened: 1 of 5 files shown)


shehabgamin (approved):
Nice!

PR #2262, opened 2026-07-19
feat: accept a non-lambda expression in a higher-order function's lambda position

davidlghellin:
Spark accepts a plain expression wherever a higher-order function expects a lambda, wrapping it in a lambda whose parameters go unused. Sail rejected all six:

    exists(array(1,2), true)     -- Spark: true    | Sail: ERROR
    transform(array(1,2), 9)     -- Spark: [9, 9]  | Sail: ERROR
    aggregate(array(1,2), 0, 9)  -- Spark: 9       | Sail: ERROR

DataFusion is what rejects them, so the wrapping happens before it, in the resolver. The body is resolved outside of a lambda scope, so nothing in it can bind to the parameters declared around it — and those parameters use opaque hidden names that cannot shadow an outer lambda variable. The body's type is still checked at analysis, so `exists(array(1,2), 1)` remains an error.

An untyped `NULL` body carries the `Null` type; Sail rejected it where Spark coerces it, so `exists(array(1,2), x -> NULL)` failed with a real lambda too — now fixed.

**Parity fixes from review:**
- A user-registered UDF named like a built-in HOF (e.g. `transform`) now   shadows the built-in instead of being intercepted by the HOF rewrite.
- filter/exists/forall validate the predicate return type (Boolean or untyped NULL) during analysis, independent of the empty/null fast path.
- Subquery expressions and Python UDFs inside a HOF lambda are rejected during analysis, matching Spark's CheckAnalysis.
- array_sort skips the comparator for NullType arrays.
- Wrapping a non-lambda body declares a single element parameter (not the optional index), avoiding an unused per-element column.
- exists/forall output nullability is derived from the array and predicate; the feature tests now assert HOF output schemas.

The remaining `@sail-bug` scenarios document pre-existing executor gaps this PR surfaces but does not fix — non-deterministic per-element / short-circuit state (`rand(seed)` in exists/forall/aggregate) and ANSI overflow inside lambda bodies. They predate this PR for explicit lambdas and will be addressed
separately.

80 new scenarios across the six feature files, all verified against Spark JVM, plus Python UDF tests for the shadowing and rejection behavior.


crates/sail-plan/src/function/scalar/lambda.rs
```diff
@@ -64,6 +64,23 @@ pub(crate) fn is_higher_order_function(name: &str) -> bool {
...
+/// an arity that matches no form yields no positions so that the non-lambda
+/// overloads (e.g. `array_sort(a)`) keep resolving as ordinary functions.
+pub(crate) fn lambda_argument_positions(function_name: &str, arity: usize) -> Vec<usize> {
+    match (function_name.trim().to_lowercase().as_str(), arity) {
+        ("aggregate" | "reduce", 3) => vec![2],
+        ("aggregate" | "reduce", 4) => vec![2, 3],
+        ("filter" | "transform" | "exists" | "forall" | "array_sort", 2) => vec![1],
+        _ => vec![],
+    }
+}
```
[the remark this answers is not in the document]
davidlghellin: Thanks — I went a different route than the suggested rewrite (the guard chain
reads worse than the pattern-alternation match), but the point pushed me to a
better fix.

There were three separate name matches for the same HOF set. I consolidated them
into one `higher_order_udf` lookup: `is_higher_order_function` is now
`.is_some()`, and `lambda_argument_positions` derives the lambda slots from the
UDF's own signature instead of a hand-written table. So the name is matched in
one place (unavoidable — this runs before the catalog lookup), and the
`to_lowercase()` is confined to that single resolution-time helper.

shehabgamin:

```
I found several Spark-parity correctness failures, including two direct regressions that silently change behavior.

  ## Blocking findings

  1. High — Synthetic lambda parameters can capture outer variables and return wrong data.

     lambda.rs:126 (https://github.com/lakehq/sail/blob/994680ce669e5fbdeabad265ea733f6a5b85c3fc/crates/sail-plan/src/resolver/expression/lambda.rs#L126-L132) creates ordinary, user-spellable names such as
     __sail_unused_lambda_param_0. DataFusion later resolves lambda variables by name, so an inner generated parameter can shadow an outer user parameter:

     SELECT transform(
       array(1),
       __sail_unused_lambda_param_0 ->
         transform(array(2), __sail_unused_lambda_param_0)
     );

     Sail returns [[2]]; Spark returns [[1]].

     Spark explicitly creates these lambdas with hidden = true to prevent this collision in opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/higherOrderFunctions.scala:98. Sail needs
     equivalent hidden/identity-based semantics, not predictable textual names.

  2. High — Catalog UDFs named like HOF built-ins are intercepted before UDF resolution.

     function.rs:99 (https://github.com/lakehq/sail/blob/994680ce669e5fbdeabad265ea733f6a5b85c3fc/crates/sail-plan/src/resolver/expression/function.rs#L99-L124) chooses the HOF argument path solely from the
     function name, before the catalog lookup at line 124.

     With a registered transform(INT, INT) UDF:

     SELECT transform(1, 2);

     Spark invokes the UDF and returns 3; Sail now errors with transform expected a list, got Int32. A shadowing exists UDF similarly fails in lambda evaluation instead of being called.

     The HOF rewrite must be gated on the catalog lookup missing, preserving Spark’s temporary-UDF precedence.

  3. High — Predicate validity is data-dependent instead of analysis-time checked.

     spark_array_filter.rs:118 (https://github.com/lakehq/sail/blob/994680ce669e5fbdeabad265ea733f6a5b85c3fc/crates/sail-function/src/scalar/array/spark_array_filter.rs#L118-L133) does not validate the lambda
     return field, and its empty/null fast path returns before the execution-time Boolean downcast.

     Confirmed:

     SELECT filter(array(), 1);                    -- Sail: []
     SELECT filter(CAST(NULL AS ARRAY<INT>), 1);   -- Sail: NULL
     SELECT IF(false, exists(array(1), 1), false); -- Sail: false

     Spark rejects all three during analysis, regardless of whether execution would evaluate the predicate. filter, exists, and forall need Boolean return-type validation during planning, while retaining Spark’s
     untyped-NULL coercion.

  4. High — Spark’s HOF analyzer restrictions are bypassed.

     The wrapper at lambda.rs:123 (https://github.com/lakehq/sail/blob/994680ce669e5fbdeabad265ea733f6a5b85c3fc/crates/sail-plan/src/resolver/expression/lambda.rs#L123-L132) accepts bodies that Spark rejects:
      - filter(array(1,2), (SELECT true)) returns [1,2] in Sail; Spark rejects HOF subqueries by default. Sail already exposes the corresponding
        spark.sql.analyzer.allowSubqueryExpressionsInLambdasOrHigherOrderFunctions configuration with default false, but does not enforce it here.

      - With a Python plus_one UDF, transform(array(1,2), plus_one(3)) returns [4,4] in Sail; Spark rejects Python UDFs inside HOF functions via opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/
        analysis/CheckAnalysis.scala:374.

     The PR’s subquery scenarios are marked @sail-bug, so they document rather than close this parity gap.

  5. High — Stateful expressions do not preserve Spark’s evaluation order.

     Bare volatile expressions are now accepted, but the existing vectorized HOF evaluators do not implement Spark’s per-element, short-circuit state semantics:

     SELECT exists(c, rand(42) > 0.6)
     FROM VALUES (array(1,2)), (array(3)) t(c);
     -- Spark: [true, false]
     -- Sail:  [true, true]

     SELECT forall(c, rand(42) < 0.6)
     FROM VALUES (array(1,2)), (array(3)) t(c);
     -- Spark: [false, true]
     -- Sail:  [false, false]

     aggregate(array(1,2), 0D, rand(42)) also returns 0.619189370225301 in Sail versus Spark’s 0.5096018842446481, because Sail recreates the seeded generator for each fold depth.

     These executor gaps predate this PR for explicit lambdas, but the PR newly exposes them through the advertised bare-expression syntax and can now produce deterministic wrong answers.

  ## Additional parity gaps

  - Medium — array_sort evaluates comparators for NullType arrays.
    array_sort(array(NULL,NULL), CAST(raise_error('boom') AS INT)) returns [NULL,NULL] in Spark, which skips sorting for NullType; Sail raises boom.

  - Low — Bare filter/transform expressions declare both element and index parameters.
    Spark creates only one hidden element parameter. Sail’s two generated parameters cause DataFusion to materialize a full unused index array—avoidable O(n) work.

  - Coverage — exists/forall nullability remains different from Spark.
    Sail hardcodes nullable Boolean output; Spark reports non-nullable output for calls such as exists(array(1,2), true). The new result tests do not assert schemas.
```

davidlghellin:
Thanks for the thorough pass, @shehabgamin — all of it was relevant.

**1 — Synthetic lambda params capturing outer variables** ✅
Non-lambda bodies are now wrapped with opaque, plan-unique hidden parameter names that can't be spelled as user lambda variables, so the `__sail_unused_lambda_param_0` collision is gone.

**2 — Catalog UDFs shadowing HOF built-ins** ✅
The HOF rewrite is now gated on the catalog lookup: a registered `transform(INT, INT)` UDF shadows the built-in and `transform(1, 2)` invokes the UDF. Covered by a new UDF test.

**3 — Predicate validity is now analysis-time** ✅
filter/exists/forall validate the lambda return type (Boolean or untyped NULL) in `return_field_from_args`, so `filter(array(), 1)`, `filter(CAST(NULL AS ARRAY<INT>), 1)` and `IF(false, exists(array(1), 1), false)` are all rejected during analysis, matching Spark.

**4 — HOF analyzer restrictions** ✅
Both subquery expressions and Python UDFs inside a HOF lambda are now rejected during analysis (mirroring CheckAnalysis). The `@sail-bug` subquery scenarios now pass and were untagged; there's a Python-UDF rejection test.

**5 — Stateful expressions / evaluation order** ⏳ follow-up
This is the pre-existing executor gap you noted (seeded `rand`, short-circuit state, `aggregate` fold-depth reseeding): the vectorized HOF evaluators don't implement Spark's per-element, short-circuit state semantics. It predates this
PR for explicit lambdas — the bare-expression syntax only surfaces it — so I've kept those scenarios as `@sail-bug` and will fix the per-element short-circuit evaluator in a dedicated follow-up PR rather than widen this one.


**array_sort on NullType** ✅ short-circuits before the comparator.
**Bare filter/transform param count** ✅ now a single element parameter.
**exists/forall nullability** ✅ derived from array || predicate, with schema
assertions added to the feature tests.

crates/sail-plan/src/resolver/expression/lambda.rs
```diff
@@ -102,10 +154,70 @@ impl PlanResolver<'_> {
...
+                    // only to satisfy the arity the function expects and to give
+                    // the evaluation batch its row count.
+                    let body = self
+                        .resolve_named_expression(expression, schema, state)
+                        .await?;
+                    // Opaque, plan-unique parameter names. DataFusion resolves
+                    // lambda variables by name at evaluation time, so a plain name
+                    // like `x` could shadow a captured outer variable of the same
+                    // name (Spark avoids this with `hidden = true`). The generated
+                    // `#N` ids cannot be spelled as user lambda variables.
```
shehabgamin: What about the following example?
```
SELECT transform(array(1), `#0` -> transform(array(2), `#0`));
```

I believe that Spark's hidden lambda variables use expression identity and do not participate in name resolution.

crates/sail-plan/src/resolver/expression/lambda.rs
```diff
@@ -102,10 +154,70 @@ impl PlanResolver<'_> {
...
+                    NamedExpr::new(
+                        vec![name],
+                        expr::Expr::Lambda(Lambda::new(params, body.expr)),
+                    )
+                }
             };
             names.push(name.one()?);
             exprs.push(expr);
         }
+        for expr in &exprs {
```
shehabgamin:
```
spark-sql (default)> SELECT transform((SELECT array(1)), 9);
[UNSUPPORTED_SUBQUERY_EXPRESSION_CATEGORY.HIGHER_ORDER_FUNCTION] Unsupported subquery expression: Subquery expressions are not supported within higher-order functions. Please remove all subquery expressions from higher-order functions and then try the query again. SQLSTATE: 0A000
spark-sql (default)>
````

`SPARK-47509` was a critical correctness fix that prohibits the use of subquery expressions within lambda and higher-order functions:
https://issues.apache.org/jira/browse/SPARK-47509

File: crates/sail-plan/src/resolver/expression/lambda.rs [hunk 4]
```diff
@@ -52,17 +96,25 @@ impl PlanResolver<'_> {
...
-        for argument in arguments {
+        for (position, argument) in arguments.into_iter().enumerate() {
             if is_spec_lambda_argument(&argument) {
                 let Some((function, arguments)) = take_spec_lambda_argument(argument) else {
                     return Err(PlanError::internal(
                         "lambda argument predicate and extraction disagreed",
                     ));
                 };
                 slots.push(Slot::Lambda(function, arguments));
+            } else if lambda_positions.contains(&position) {
```
shehabgamin, line 126: Does Sail break with this example?
```
spark-sql (default)> SELECT exists(c, rand(42) > 0.6) FROM VALUES (array(1, 2)), (array(3)) AS t(c);
true
false
Time taken: 0.084 seconds, Fetched 2 row(s)
spark-sql (default)>
```

crates/sail-plan/src/resolver/expression/lambda.rs
```diff
@@ -102,10 +154,70 @@ impl PlanResolver<'_> {
...
             names.push(name.one()?);
             exprs.push(expr);
         }
+        for expr in &exprs {
+            let expr::Expr::Lambda(lambda) = expr else {
+                continue;
+            };
+            // Spark rejects subquery expressions inside a higher-order function
+            // (SPARK-47509 is a correctness bug); the guard is on by default
+            // (`spark.sql.analyzer.allowSubqueryExpressionsInLambdasOrHigherOrderFunctions`
```
shehabgamin: Shouldn't we check for the config value and adjust behavior accordingly?

crates/sail-plan/src/resolver/expression/lambda.rs
shehabgamin: Another example:
```
spark-sql (default)> SELECT aggregate(array(1, 2), CAST(0.0 AS DOUBLE), rand(42));
0.5096018842446481
Time taken: 0.068 seconds, Fetched 1 row(s)
spark-sql (default)>
```

shehabgamin (commented):
Next round of review!

crates/sail-plan/src/resolver/expression/lambda.rs
davidlghellin: You're right — Spark binds hidden lambda vars by `exprId` (identity), not name. 
DataFusion resolves by name, so `#N` was still backtick-spellable and shadowed the captured variable (`[[2]]` instead of `[[1]]`). Fixed: the placeholder now avoids every name the body captures. Verified against Spark, with scenarios.

crates/sail-plan/src/resolver/expression/lambda.rs
davidlghellin: Good catch — the check only looked at lambda bodies. It now rejects a subquery in any argument (value or lambda), matching SPARK-47509. Added scenarios for the value-position case.

crates/sail-plan/src/resolver/expression/lambda.rs
davidlghellin: Yes, Sail diverges here — seeded `rand(42)` advances per call and Sail's RNG + eager (non-short-circuit) evaluation don't match Spark's. It's a pre-existing executor gap (not the resolver this PR touches); keeping it `@sail-bug` and fixing it in a dedicated follow-up PR.

crates/sail-plan/src/resolver/expression/lambda.rs
davidlghellin: Done — gated the rejection on the config (default `false`). With it set to `true`, subqueries are allowed again (`transform((SELECT array(1,2)), x -> x+1)` → `[2, 3]` on both engines).

davidlghellin:
> Next round of review!

Thanks for the thorough review, @shehabgamin

shehabgamin:
@davidlghellin I converted this to a draft pr. Feel free to tag me when it's ready for re-review!

PR #2264, opened 2026-07-20
feat: add UPDATE support for delta-lake

nis12ram:


nis12ram:
@lonless9  I'd be grateful if you could take a look at it and share your thoughts when you have some time.

lonless9:
@nis12ram Thanks for putting this together! I’m currently focused on basic `MERGE` support for Iceberg and several Delta Lake/Iceberg fixes. There is also ongoing redesign around DML and catalog operations which is needed to support more features cleanly. Due to the rapidly evolving internals, we decided to pause `UPDATE` support (#1462 and #1716) and advanced features such as Delta Lake row tracking (#1773).
Given these commitments, I’m afraid we won’t be able accept PRs around major Delta or Iceberg features at the moment. Thank you for your understanding.

PR #2266, opened 2026-07-20
fix: delta nested column mapping fix

zemin-piao:
When writing to Delta tables with column mapping enabled (delta.columnMapping.mode = name/id), nested struct fields were matched only by logical name. This breaks when the physical parquet field names differ from the logical schema names — nested fields could be misaligned, dropped, or have stale parquet field-ID metadata leak into the committed table schema. This PR makes nested field resolution column-mapping-aware across the read (schema evolution) and write paths.

### DIFF
--- a/crates/sail-common-datafusion/src/schema_evolution.rs
+++ b/crates/sail-common-datafusion/src/schema_evolution.rs
@@ -10,6 +10,7 @@ use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::common::tree_node::{Transformed, TransformedResult, TreeNode};
 use datafusion::common::{DataFusionError, Result, ScalarValue, exec_err};
 use datafusion::functions::core::getfield::GetFieldFunc;
+use datafusion::parquet::arrow::PARQUET_FIELD_ID_META_KEY;
 use datafusion::physical_expr::expressions::{self, Column, Literal};
 use datafusion::physical_expr::{PhysicalExpr, ScalarFunctionExpr};
 use datafusion::physical_expr_adapter::{PhysicalExprAdapter, PhysicalExprAdapterFactory};
@@ -17,7 +18,16 @@ use datafusion::physical_plan::ColumnarValue;
 use datafusion_common::format::DEFAULT_CAST_OPTIONS;
 use parquet_variant_compute::{VariantArray, unshred_variant};
 
+use crate::array::record_batch::cast_array_recursively;
 use crate::variant::{is_binary_variant_field, is_variant_arrow_field, is_variant_storage_type};
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash)]
+pub enum StructFieldMatching {
+    #[default]
+    Name,
+    PhysicalName,
+    FieldId,
+}
 
 #[derive(Debug)]
 pub struct SchemaEvolutionPhysicalExprAdapterFactory {}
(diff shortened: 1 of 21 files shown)


crates/sail-delta-lake/src/schema/converter.rs
```diff
@@ -336,3 +337,49 @@ pub fn enrich_arrow_with_parquet_field_ids(
...
+        let physical = get_physical_arrow_schema(&logical, ColumnMappingMode::Name);
+        let details = physical
+            .field_with_name("col-details")
+            .expect("physical details field");
+        let ArrowDataType::Struct(children) = details.data_type() else {
+            panic!("details must be a struct");
+        };
+        let rate = children.first().expect("nested rate field");
+
+        assert_eq!(rate.name(), "rate");
```
lonless9: As every field, including nested fields, must be written using `delta.columnMapping.physicalName`, I think this should remain `col-rate` rather than `rate` and we need to preserve nested physical names.

crates/sail-common-datafusion/src/schema_evolution.rs
```diff
@@ -338,15 +336,57 @@ fn can_cast_field_with_schema_evolution(source: &Field, target: &Field) -> Resul
...
+        return Some(matched);
+    }
+
+    if let Some(target_id) = struct_field_id(target_field)
+        && let Some(matched) = source_fields
+            .iter()
+            .enumerate()
+            .find(|(_, source)| struct_field_id(source) == Some(target_id))
+    {
+        return Some(matched);
```
lonless9: Delta name mode resolves data columns only by `delta.columnMapping.physicalName`; column IDs are explicitly not used. So we need to make resolution mode-aware and reserve ID matching for id mode.

python/pysail/tests/spark/delta/test_delta_column_mapping_interop.py
```diff
@@ -0,0 +1,124 @@
...
+    CREATE TABLE delta.`...` (
+        id INT, label STRING, details STRUCT<amount: INT, active: BOOLEAN>
+    ) USING delta TBLPROPERTIES ('delta.columnMapping.mode' = 'name',
+        'delta.minReaderVersion' = '2', 'delta.minWriterVersion' = '5');
+    INSERT INTO ... VALUES (1, 'alpha', named_struct('amount', 10, 'active', true)),
+                           (2, 'beta', named_struct('amount', 20, 'active', false));
+    ALTER TABLE ... RENAME COLUMN label TO code;
+    ALTER TABLE ... RENAME COLUMN details.amount TO total;
+    INSERT INTO ... VALUES (3, 'gamma', named_struct('total', 30, 'active', true));
+"""
```
lonless9: I think both Parquet files retain the same field IDs and physical names across the renames. Rename changes only logical/display names, and name mode does not use IDs for resolution. 
The external fixture remains useful as a positive interoperability case, but I think we can cover writer output in the existing nested column-mapping BDD by recursively asserting that Parquet field names and IDs match the Delta metadata, and cover forbidden name-mode ID/logical-name fallbacks with negative unit tests. So we don't need to include the binary fixture finally.

lonless9 (commented):
Looks great! Just a couple of minor points. BTW we can treat `RENAME COLUMN` support as a follow-up work.

linhr:
Thanks for the PR! One thing I noticed is that the test data is committed, which I'm afraid is not something we would like to do. Binary files such as Parquet data cannot be viewed easily in diff and are not auditable, so we'd like to remove them even if they are small.

We can keep the test as an integration test that includes the data generation steps, or if data generation is too slow or requires too many extra dependencies, we can skip them entirely via marker so that we track the test case in the codebase (and run it manually when needed) without incurring CI overhead.

crates/sail-delta-lake/src/schema/converter.rs
zemin-piao: Agree! Adjusted to use `delta.columnMapping.physicalName` and the test expects `col-rate` now.

crates/sail-common-datafusion/src/schema_evolution.rs
zemin-piao: Make sense, I made the resolution mode-aware (based on whether they are `name`, `id` or pure parquet / iceberg)

python/pysail/tests/spark/delta/test_delta_column_mapping_interop.py
zemin-piao: I removed the binary fixtures now. I replaced by a test doing sail data generation ++ verification. I'll look into how to do an integration test with delta-spark etc.

zemin-piao:
> Thanks for the PR! One thing I noticed is that the test data is committed, which I'm afraid is not something we would like to do. Binary files such as Parquet data cannot be viewed easily in diff and are not auditable, so we'd like to remove them even if they are small.
> 
> We can keep the test as an integration test that includes the data generation steps, or if data generation is too slow or requires too many extra dependencies, we can skip them entirely via marker so that we track the test case in the codebase (and run it manually when needed) without incurring CI overhead.

Thank you for your suggestion. I removed the binary files ++ created an integration test that indeed includes a step for data generation, with the integration marker

lonless9:
Love it!
Since we just added Spark 4.2 support, and our catalog tests run against the latest Spark version, these tests are currently skipped in CI because `delta-spark` only supports Spark <= 4.1.x. We can still run them locally with:

```bash
hatch run test.spark-4.1.1:pytest -m integration \
  python/pysail/tests/spark/delta/test_delta_column_mapping_delta_spark.py
```
Hopefully once `delta-spark` catches up with Spark 4.2, they’ll just work in CI as well.
cc @linhr

PR #2270, opened 2026-07-21
feat: add basic checkpoint support

lonless9:


crates/sail-spark-connect/src/service/plan_executor.rs
```diff
@@ -515,19 +518,69 @@ pub(crate) async fn handle_execute_streaming_query_listener_bus_command(
...
+}
+
+pub(crate) async fn handle_execute_remove_cached_remote_relation_command(
+    ctx: &SessionContext,
+    _command: RemoveCachedRemoteRelationCommand,
+    metadata: ExecutorMetadata,
+) -> SparkResult<ExecutePlanResponseStream> {
+    let spark = ctx.extension::<SparkSession>()?;
+    // Checkpoints have session lifetime so plans can safely share their immutable relation ID.
+    // The registry and object-store namespace are cleared together after the job runner stops.
```
linhr, line 574: I feel we can support removing cached relations if it's an intention from the client. It seems PySpark handles best-effort removal during garbage collection in `CachedRemoteRelation.__del__()`.

This can be out of scope for this PR.

crates/sail-common/src/config/application.yaml
```diff
@@ -317,6 +317,17 @@
     This setting can only be configured at startup and cannot be changed at runtime.
   experimental: true
 
+- key: checkpoint.root
```
linhr: Let's change this to `execution.checkpoint.path`.

crates/sail-common/src/config/application.yaml
```diff
@@ -317,6 +317,17 @@
...
+  type: string
+  default: ""
+  description: |
+    Server-authorized object-store root for DataFrame checkpoints. The root must be readable and
+    writable by the driver and every worker. Checkpoint operations are rejected when this value is
+    empty; Sail does not fall back to driver memory or local worker storage. Sail removes each
+    session namespace on shutdown; an object-store lifecycle policy can clean up data left by an
+    ungraceful process exit.
```
linhr: How about this description?

```
The object storage base path for DataFrame checkpoints managed by the server. The path must be readable and
writable by the driver and every worker. Checkpoint operations are rejected when this value is
empty. Sail does not fall back to driver memory or local worker storage. Sail removes each
session namespace on shutdown. An object storage lifecycle policy can clean up data left by an
ungraceful process exit.
```

crates/sail-logical-plan/src/remote_checkpoint.rs
```diff
@@ -0,0 +1,67 @@
...
+pub struct RemoteCheckpointRelationNode {
+    relation_id: String,
+    schema: DFSchemaRef,
+}
+
+impl PartialOrd for RemoteCheckpointRelationNode {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.relation_id.partial_cmp(&other.relation_id)
+    }
+}
```
linhr: Let's use `educe`.

crates/sail-object-store/src/path.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn store(&self) -> &Arc<dyn ObjectStore> {
+        &self.store
+    }
+
+    pub async fn put_bytes(&self, location: &Path, bytes: Vec<u8>) -> Result<PutResult> {
+        self.store
+            .put(location, PutPayload::from(bytes))
+            .await
+            .map_err(|error| DataFusionError::ObjectStore(Box::new(error)))
+    }
```
linhr: I feel this method may be misused for large objects so we should probably just remove it. The caller should use multipart upload instead.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+            self.mode,
+            ExecutionMode::LocalCluster | ExecutionMode::KubernetesCluster
+        ) && (object_store_url.starts_with("file:") || object_store_url.starts_with("memory:"))
+        {
+            return Err(DataFusionError::Plan(format!(
+                "checkpoint.root must use shared object storage in cluster mode, not {object_store_url}"
+            )));
+        }
+        Ok(())
+    }
```
linhr: We don't need to validate the URL so that we don't need to depend on `ExecutionMode` in the checkpoint service.

A `file://` URL is totally fine in the cluster if the file system is a mounted network file system shared by the driver and workers. And in the future we may even support `memory://` if certain memory space is shared in distributed settings.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+
+    fn session_prefix(&self, root: &Path) -> Path {
+        // `v1` versions the object layout only; checkpoint discovery remains session-local.
+        root.clone()
+            .join("_sail")
+            .join("checkpoints")
+            .join("v1")
```
linhr: We can remove these constant levels because (1) the user is aware that the root directory is dedicated to checkpoints so that we don't need subdirectories to isolate the intent and (2) the content format is specific to each Sail release so we don't need a general versioning mechanism.

crates/sail-spark-connect/src/executor.rs
linhr, line 1: The keep-alive fix in this file is super valuable, so I feel it deserves it's on PR so that we can highlight it in the changelog. Could you maybe remove the change for now and I can follow up separately? I plan to play with it a bit and see if I can come up with some elegant solution to cover both query and command heartbeats.

python/pysail/testing/spark/steps/plan.py
```diff
@@ -74,6 +74,20 @@ def normalize_path(path: str) -> str:
         text,
     )
     text = pytest_tmp_prefix.sub(lambda m: f"{m.group(1)}<tmp>/", text)
+    # Preserve checkpoint partition layout while removing generated identifiers.
+    checkpoint_uuid = r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
+    text = re.sub(
+        rf"(_sail/checkpoints/v1/){checkpoint_uuid}/{checkpoint_uuid}",
```
linhr: Since I suggested removing `_sail/checkpoints/v1/` in the Rust code, for testing we can assume that the checkpoint root path is always something like `<tmp>/__checkpoint_testing__` when defining the `remote` fixture, so that we can match it by regex here.

python/pysail/tests/spark/dataframe/test_checkpoint.py
```diff
@@ -0,0 +1,174 @@
...
+    aggregate_plan = normalize_plan_text(checkpointed.groupBy("key").count()._explain_string())  # noqa: SLF001
+    ordered_plan = normalize_plan_text(checkpointed.sortWithinPartitions("key", "id")._explain_string())  # noqa: SLF001
+    consumer_plans = "\n".join(
+        [
+            f"== Grouped Aggregate From Checkpoint ==\n{aggregate_plan}",
+            f"== Sort Within Partitions From Checkpoint ==\n{ordered_plan}",
+        ]
+    )
+
+    assert consumer_plans == snapshot
```
linhr: I feel for simplicity we don't need to concatenate the plans here. We could just assert `== snapshot` twice and let Syrupy manage multiple snapshots within the test.

python/pysail/tests/spark/dataframe/test_checkpoint.py
```diff
@@ -0,0 +1,174 @@
...
+@pytest.mark.yamlsnapshot(group="plan")
+def test_checkpoint_supports_iterative_weakly_connected_components(spark, snapshot):
+    nodes = spark.range(7)
+    pairs = spark.createDataFrame(
+        [(0, 1), (1, 2), (2, 3), (3, 4), (4, 5)],
+        "a LONG, b LONG",
+    )
+
+    expected = {frozenset(range(6)), frozenset({6})}
+    baseline = _weakly_connected_components(nodes, pairs, checkpoint=False)
```
linhr: I wonder if we should just remove `baseline` since we have precomputed `expected` result anyway. I'm worried WCC without checkpoint causes repeated computation and generates a long plan which might cause issues if we later adjust the test to have larger input datasets.

crates/sail-physical-plan/src/remote_checkpoint.rs
```diff
@@ -0,0 +1,757 @@
...
+
+pub const PARTITION_COLUMN: &str = "partition";
+pub const LOCATION_COLUMN: &str = "location";
+pub const SIZE_COLUMN: &str = "size";
+pub const ROW_COUNT_COLUMN: &str = "row_count";
+pub const ROW_MARKER_COLUMN: &str = "__sail_checkpoint_row_marker";
+
+/// Pass-through scan boundary that reports properties captured at checkpoint materialization.
+#[derive(Debug)]
+pub struct RemoteCheckpointScanExec {
```
linhr: I feel we don't need `RemoteCheckpointScanExec`. We should look at how `FileScanConfig` is created for checkpoint data and make sure the properties are derived correctly or fix the root cause if that's not the case.

crates/sail-physical-plan/src/remote_checkpoint.rs
```diff
@@ -0,0 +1,757 @@
...
+use parquet::file::properties::WriterPropertiesBuilder;
+use sail_common_datafusion::array::record_batch::record_batch_with_schema;
+use tokio::io::AsyncWrite;
+use uuid::Uuid;
+
+pub const PARTITION_COLUMN: &str = "partition";
+pub const LOCATION_COLUMN: &str = "location";
+pub const SIZE_COLUMN: &str = "size";
+pub const ROW_COUNT_COLUMN: &str = "row_count";
+pub const ROW_MARKER_COLUMN: &str = "__sail_checkpoint_row_marker";
```
linhr: Instead of using Arrow record batches to represent commit metadata and the special column for row marker, I wonder if we should have a single JSON string so that it makes the code much simpler to read/write the information.

crates/sail-physical-plan/src/remote_checkpoint.rs
```diff
@@ -0,0 +1,757 @@
...
+    ) -> Result<SendableRecordBatchStream> {
+        let input = self.input.execute(partition, Arc::clone(&context))?;
+        let store = context.runtime_env().object_store(&self.object_store_url)?;
+        // Attempt-unique immutable objects make retries safe without rename or copy.
+        let location = self
+            .prefix
+            .clone()
+            .join("attempts")
+            .join(format!("partition-{partition:020}"))
+            .join(format!("{}.parquet", Uuid::new_v4()));
```
linhr: I'd just make it `part-{partition:05}-{uuid}.parquet` so that it's closer to Spark's or Hadoop's convention for partitioned data. "attempts" is confusing since we may think it's the attempt in cluster mode data processing, which is not.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+            .join(self.session_namespace.as_str())
+    }
+
+    fn new_relation_id(&self, registry: &RemoteCheckpointRegistry) -> Result<String> {
+        for _ in 0..8 {
+            let relation_id = Uuid::new_v4().to_string();
+            if registry.get(&relation_id)?.is_none() {
+                return Ok(relation_id);
+            }
+        }
```
linhr: We don't need retry. Just fail on the first failure since UUID duplication is extremely rare.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+    let fields = logical_schema
+        .fields()
+        .iter()
+        .enumerate()
+        .map(|(index, field)| {
+            Arc::new(
+                field
+                    .as_ref()
+                    .clone()
+                    .with_name(format!("__sail_checkpoint_col_{index:05}")),
```
linhr: For simplicity just use `_c{index}`. But glad that you identified the potential problem for duplicated column names.

crates/sail-spark-connect/src/service/plan_executor.rs
```diff
@@ -515,19 +518,69 @@ pub(crate) async fn handle_execute_streaming_query_listener_bus_command(
...
+            .materialize(&operation_ctx, plan, logical_schema)
+            .await?;
+        Ok(ExecutorBatch::CheckpointCommandResult(Box::new(
+            CheckpointCommandResult {
+                relation: Some(CachedRemoteRelation {
+                    relation_id: descriptor.relation_id.clone(),
+                }),
+            },
+        )))
+    });
```
linhr: I feel we don't need `resolve_and_execute_query` and the changes in `crates/sail-plan/src/lib.rs`. We should define a command in the spec to materialize checkpoint so that it flows through plan resolver and logical planning while remaining transparent to the plan executor for Spark Connect.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+};
+use sail_physical_plan::remote_checkpoint::{
+    LOCATION_COLUMN, PARTITION_COLUMN, ROW_COUNT_COLUMN, ROW_MARKER_COLUMN,
+    RemoteCheckpointWriteExec, SIZE_COLUMN,
+};
+use tokio::sync::RwLock;
+use uuid::Uuid;
+
+#[derive(Debug)]
+pub struct RemoteCheckpointService {
```
linhr: I don't think we need `RemoteCheckpointService`. If we define the command in the spec according to the other command I have, we can move all the logic to a physical plan node and access the checkpoint registry during physical execution.

crates/sail-session/src/checkpoint.rs
```diff
@@ -0,0 +1,518 @@
...
+                        "checkpoint failed: {error}; additionally failed to clean {relation_prefix}: {cleanup_error}"
+                    )),
+                }
+            }
+        };
+        drop(lifecycle);
+        output
+    }
+
+    pub async fn cleanup_session(
```
linhr: The cleanup logic should just be moved to the registry.

jordantgh:
@lonless9 I've been interested in this for a while, esp for iterative processing. I've been using a manual in memory write/read cycle till now to deal with that. Would you find it useful at this stage for me to run any kind of "stress" testing on some jobs I have where checkpointing has been a challenge, or are you still iterating on this PR?

linhr:
@jordantgh Thanks for reaching out! The testing would be super valuable. We plan to include this PR in the upcoming release next week, and you can try that out and see how it goes. We'd love to hear from you. Your feedback would help us a lot when we iterate on the design once the initial version is out.

lonless9:
@jordantgh I'm wrapping up this PR. If the release build takes too long, the debug mode should also show meaningful improvements. Thanks for being involved!

jordantgh:
> @jordantgh I'm wrapping up this PR. If the release build takes too long, the debug mode should also show meaningful improvements. Thanks for being involved!

Cheers, I did check out the branch and do some testing but it will be cleaner on the release and can report back anything in an issue then, if that sounds sensible.

lonless9:
@jordantgh Sounds great!

crates/sail-object-store/src/path.rs
```diff
@@ -0,0 +1,144 @@
...
+        ));
+    }
+    if url.get_glob().is_some()
+        || url.get_url().query().is_some()
+        || url.get_url().fragment().is_some()
+    {
+        return Err(DataFusionError::Plan(
+            "object store path cannot contain a query, fragment, or glob".to_string(),
+        ));
+    }
```
linhr: I'm not sure if we want these validations. It's up to the object store implementation to decide whether these components are valid.

crates/sail-session/src/planner.rs
```diff
@@ -99,8 +123,195 @@ impl ExtensionPlanner for ExtensionPhysicalPlanner {
...
+                .map(|ordering| checkpoint_schema_ordering(ordering, &storage_schema))
+                .transpose()?;
+            let writer: Arc<dyn ExecutionPlan> = Arc::new(RemoteCheckpointWriteExec::try_new(
+                Arc::clone(input),
+                object_store_url.clone(),
+                prefix.clone(),
+                Arc::clone(&storage_schema),
+            )?);
+            let commit_input: Arc<dyn ExecutionPlan> =
+                Arc::new(CoalescePartitionsExec::new(writer));
```
linhr: This may be added automatically by the physical optimizer.

linhr (approved):
This is great! 🚀

I just have a few minor comments but overall it's exciting to see this in a good shape!

PR #2272, opened 2026-07-21
perf: build arrays_zip result with a single take instead of per-row concat

davidlghellin:
## Benchmark

Self-contained harness (public `arrow` only, pinned): davidlghellin/sail-benchmarks#2 →
[`sail-2272-arrays-zip`](https://github.com/davidlghellin/sail-benchmarks/tree/master/sail-2272-arrays-zip).
`old` = the previous per-row `StructArray` + `concat(N)` build; `new` = compute the output offsets once
and gather each struct column with a single `arrow::compute::take`.

**Time — criterion, 200k rows**

| case | old (per-row struct + concat) | new (flatten + take) | speedup |
|---|---|---|---|
| 2 args | 72.0 ms | 5.56 ms | **~13×** |
| 4 args | 130.0 ms | 11.76 ms | **~11×** |

**Memory — divan, 200k rows × 4 args**

| metric | old | new | improvement |
|---|---|---|---|
| time | 116.4 ms | 11.19 ms | ~10× |
| bytes allocated | 273.6 MB | 26.4 MB | **~10×** |
| **allocation count** | **3,000,046** | **56** | **~54,000×** |
| peak memory | 264 MB | 37.6 MB | ~7× |

The old path builds one `StructArray` per row (200k of them) and then `concat`s them — ~3M tiny
allocations. The new path computes offsets once and does a single `take` per struct column: 56
allocations total. So the win is both CPU and, above all, allocator pressure.

**Correctness:** `cargo test -p sail-2272-arrays-zip` — `old` and `new` produce byte-identical
`ArrayData` on uniform, ragged (null-padded), null-row, null-element and empty/mixed inputs (5 tests).

_Numbers on Apple M-series; exact values vary by CPU / Rust / arrow version — treat as illustrative._

benchmarks https://github.com/davidlghellin/sail-benchmarks/pull/2

### DIFF
--- a/crates/sail-function/src/scalar/array/arrays_zip.rs
+++ b/crates/sail-function/src/scalar/array/arrays_zip.rs
@@ -346,6 +346,10 @@ fn arrays_zip_generic<O: OffsetSizeTrait>(
     offsets.push(zero_offset);
     let mut last_offset = zero_offset;
 
+    // Reused across rows to avoid a per-row allocation of the two scratch vectors.
+    let mut arrays_one_row: Vec<ArrayRef> = Vec::with_capacity(lists.len());
+    let mut lens_one_row: Vec<O> = Vec::with_capacity(lists.len());
+
     for row_idx in 0..num_rows {
         if validity_mask_opt
             .as_ref()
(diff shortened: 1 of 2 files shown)


shehabgamin (approved):
Well done!

PR #2283, opened 2026-07-23
feat: support Aliyun OSS object storage

weixiuli:


linhr:
Thanks for the PR! I will review it shortly.

It seems that this PR uses the S3 protocol to support the `oss://` URL scheme. This seems a good first step to improve user experience. Later I can swap the implementation to support the actually OSS protocol to avoid limitations in Aliyun's S3-compatible endpoints. (I researched the OSS protocol a while ago and I feel I can possibly get inspiration from the OpenDAL implementation.)

linhr (approved):
This is great! Thanks for the contribution!

PR #2285, opened 2026-07-24
feat: support explicit Glue catalog IDs

pomykalakyle:
Adds optional catalog_id configuration to the Glue catalog provider and applies it consistently to database, table, and view requests.

Omitting catalog_id preserves the existing AWS account-default behavior. The Glue guide and catalog integration fixtures now cover explicit catalog IDs.


linhr (approved):
Nice!

PR #2288, opened 2026-07-24
feat: support a bearer token file for the Iceberg REST catalog

### LINKED ISSUE #2287: Iceberg REST catalog: support a bearer token file that is re-read per request
### Problem

The Iceberg REST catalog takes a bearer token from config
(`catalog.list[].bearer_access_token`, env `SAIL_CATALOG__LIST`). The value is
read once at startup and held for the process lifetime. In Kubernetes the common
way to authenticate to a REST catalog is a projected ServiceAccount token, which
kubelet rotates on a schedule (by default once 80% of the token lifetime has
elapsed). Because Sail never re-reads the token, the rotated token is ignored and
requests start failing with 401 once the original token expires, until the server
is restarted.

### Proposal

Add an optional `bearer_access_token_file` to the Iceberg REST catalog config: a
path to a file holding the bearer token. When set, the token is read from that
file and refreshed when the file changes, so a rotated projected token is picked
up without a restart. It takes precedence over `bearer_access_token`.

`main` already resolves credentials per request through
`CatalogCredentials::retrieve()`, so this is a small addition: a new
`FileCatalogCredentials` implementation plus the config field and wiring.

### Alternatives considered

- Static `bearer_access_token`: cannot rotate, the problem above.
- OAuth token exchange: heavier, needs an auth server and does not match the
  projected-token pattern where kubelet already handles rotation on disk.

Robin-Everaars:
## Summary

Adds an optional `bearer_access_token_file` to the Iceberg REST catalog config.
When the path is set, the bearer token is read from that file and re-read on
rotation, so a kubelet-projected ServiceAccount token is picked up without
restarting Sail. It falls back to the existing static `bearer_access_token` when
the file is not set.

Closes #2287.

## Changes

- `sail-catalog`: new `FileCatalogCredentials` implementing `CatalogCredentials`.
  `retrieve()` reads the token file with `tokio::fs`, trims it and caches it in
  memory. It re-reads when the file modification time changes or a short refresh
  interval (60s) elapses, so a rotated token is picked up promptly while avoiding
  a disk read on every catalog request.
- `sail-common`: new optional `bearer_access_token_file: Option<String>` on the
  `IcebergRest` catalog config. The path is not a secret so it is a plain string.
- `sail-session`: build `FileCatalogCredentials` when the path is set, otherwise
  the existing static or empty credentials. The file takes precedence.

Because the Iceberg provider already builds its client and calls
`credentials.retrieve()` per request, no provider changes are needed.

## Behavior

- No config change for existing users: without `bearer_access_token_file`,
  behavior is unchanged.
- The token file is expected to hold the raw bearer token, optionally with
  trailing whitespace or a newline (trimmed). This matches a projected
  ServiceAccount token file.
- Refresh is driven by file mtime with a 60s TTL fallback. The interval is a
  constant for now and can be made configurable if there is demand.

## Testing

New unit tests on `FileCatalogCredentials` in
`crates/sail-catalog/src/credentials.rs`. They add `tempfile` as a
dev-dependency and drive the refresh through the file mtime (via
`std::fs::FileTimes`), so no test sleeps for the refresh interval:

- `retrieve_returns_token_from_file`: a file holding a token returns that token.
- `retrieve_trims_surrounding_whitespace`: leading and trailing whitespace or
  newlines are trimmed.
- `retrieve_rereads_when_modification_time_changes`: after rewriting the file and
  bumping its mtime, `retrieve()` returns the new token within the refresh
  interval, which exercises the mtime-driven re-read.
- `retrieve_reports_error_for_missing_file`: a missing file yields
  `CatalogError::External`.

Commands run locally (cargo/rustc 1.95.0 from nixpkgs, CI is authoritative):

- `cargo test -p sail-catalog`: 16 passed, 0 failed (the four new tests plus the
  crate's existing tests).
- `cargo clippy -p sail-catalog -p sail-common -p sail-session --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt --check` on the touched files: clean.

## Notes for reviewers

- Blocking vs async IO: `retrieve()` is async and uses `tokio::fs`. The file is
  small and local, so the cost is negligible.
- The same feature on the pre-refactor 0.6.x line needs a different shape (the
  provider held a single cached client there). This PR targets the current
  per-request credential resolution on `main`.


Robin-Everaars:
Addressed both review comments and the patch coverage gap in c59ec9f:

- An empty or whitespace-only token file is now an error instead of a cached empty credential, so no bare `Authorization: Bearer` header can go out. The empty read is not cached, which means the next `retrieve()` recovers as soon as the file holds a token again (test covers the recovery).
- Dropped the clone on the token file path. Small note: the config is matched by reference there, so the binding is `&String` rather than a moved value; the change is a simplification, the byte copy into `PathBuf` happens either way.
- Codecov flagged the credentials wiring in `sail-session`. I extracted it into `iceberg_rest_credentials` and added behavioral tests for the four paths (file precedence over a static token, static bearer, OAuth fallback and no token), so the previously untested lines are now exercised.

On the Ibis report delta: the one additional unique failure is in the SQL planner categories (`TRUNCATE` parsing, `approx_percentile_cont` signatures) that this change does not touch, so it looks pre-existing.

`cargo test -p sail-catalog -p sail-session` (17 pass in each), clippy with `-D warnings` and `fmt --check` are green on the updated branch.

crates/sail-session/src/catalog.rs
```diff
@@ -317,6 +314,28 @@ fn wrap_catalog_provider(
     Ok(Arc::new(provider))
 }
 
+/// Credentials for an Iceberg REST catalog. A token file takes precedence over
```
shehabgamin, line 317: I'm not sure if this is true for every outbound REST request. If i'm tracing the code correctly, `IcebergRestCatalogProvider::configuration()` seems to resolve the credential once and then copy it into the `ApiClient` header map. Then, `drop_database(cascade)` reuses one client across all `list`/`drop` requests, so a token that rotates and expires mid-cascade continues using the old header. Because the per-object drop errors are ignored, this can partially apply before the final namespace drop fails:
- https://github.com/lakehq/sail/blob/main/crates/sail-catalog-iceberg/src/provider.rs#L935-L942
- https://github.com/lakehq/sail/blob/main/crates/sail-catalog-iceberg/src/provider.rs#L950-L952

One approach is to ensure that a rotated credential can be observed or recovered between outbound REST requests while retaining the shared `reqwest::Client`. For example, the request layer could use the cached token initially, then force-reload the credential file and retry that request once on a 401 response.

And then we would want to add a regression test where the server rejects the old token between cascade requests.

crates/sail-common/src/config/application.rs
```diff
@@ -551,6 +551,14 @@ pub enum CatalogType {
...
         )]
         bearer_access_token: Option<SecretString>,
+        /// Path to a file holding the bearer token. When set, the token is
+        /// re-read from this file per request (with a short refresh interval)
+        /// so a rotated token (for example a kubelet-projected service account
+        /// token) is picked up without restarting the server. Takes precedence
+        /// over `bearer_access_token`. The path is not a secret, so it is kept
+        /// as a plain string.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        bearer_access_token_file: Option<String>,
```
shehabgamin, line 645: Can you please document this new option in `docs/guide/catalog/iceberg-rest.md`

crates/sail-catalog/src/credentials.rs
```diff
@@ -34,3 +37,197 @@ impl CatalogCredentials for StaticCatalogCredentials {
...
+            path: path.into(),
+            cached: Mutex::new(None),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl CatalogCredentials for FileCatalogCredentials {
+    async fn retrieve(&self) -> CatalogResult<Option<String>> {
+        let modified = tokio::fs::metadata(&self.path)
```
shehabgamin: Non-blocking: `retrieve()` still performs asynchronous filesystem metadata I/O on every call, so the cache does not provide a filesystem-free fast path while adding mutex, mtime, and TTL state.

shehabgamin (commented):
@Robin-Everaars This approach makes sense to me. Just have a couple of comments.

Thanks so much for this contribution!

Robin-Everaars:
Thanks. I moved the retry to the request layer: a `with_auth_retry` helper rebuilds the client from a freshly read credential and retries once on a 401, wrapped around every catalog operation including each request in the `drop_database` cascade, so a mid-cascade rotation recovers per request (the one-time startup config fetch is the exception, since a 401 there is a hard startup failure). Added `drop_database_cascade_recovers_when_token_rotates_midway`, a mock-server test that flips the token file to B mid-cascade and asserts the cascade completes authorized, plus unit tests asserting the helper retries exactly once on a 401 and leaves other errors untouched.

Per your other note I dropped the mtime, TTL and mutex cache so `retrieve()` is a fresh read each call, and documented `bearer_access_token_file` in `docs/guide/catalog/iceberg-rest.md`. `cargo test`, `clippy --all-targets --all-features -D warnings` and `fmt --check` are green for the four crates.

Robin-Everaars:
> @Robin-Everaars This approach makes sense to me. Just have a couple of comments.
> 
> Thanks so much for this contribution!

And you're very welcome! Hope to see you next month in NL :)

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -188,6 +188,27 @@ impl IcebergRestCatalogProvider {
...
     }
 
+    /// Run a single outbound REST request, retrying it once if the server
+    /// answers `401 Unauthorized`. Each attempt builds an [`ApiClient`] from a
+    /// freshly resolved credential, so a projected service account token that
+    /// rotated mid-operation is picked up on the retry. The credential is
+    /// re-read per request, so every request in a `drop_database` cascade sees
+    /// the current token. The shared `reqwest::Client` and its connection pool
+    /// are reused across attempts.
+    async fn with_auth_retry<T, E, F, Fut>(&self, call: F) -> CatalogResult<Result<T, ApiError<E>>>
```
shehabgamin, line 198: `GET /v1/config` at lines 217–226 is the only outbound REST request that bypasses this helper.

I know you said the following:
> the one-time startup config fetch is the exception, since a 401 there is a hard startup failure

But there is 1 edge case scenario:
If the credential rotates between `bootstrap_client()` reading token `A` and the server rejecting it, the first catalog operation fails instead of retrying with token `B`. I think we can add a non-recursive bootstrap retry using `bootstrap_client()`.

docs/guide/catalog/iceberg-rest.md
```diff
@@ -20,6 +20,10 @@ An Iceberg REST catalog can be configured using the following options:
   If unset or empty, Sail uses the Iceberg REST default unit separator.
 - `oauth_access_token` (optional): The OAuth 2.0 access token.
 - `bearer_access_token` (optional): The bearer token for authentication.
+- `bearer_access_token_file` (optional): Path to a file that holds the bearer token.
```
shehabgamin, line 23: Can we please document the complete token-file contract: the file must contain the raw token without a `Bearer ` prefix, surrounding whitespace is trimmed, empty or unreadable files produce an error rather than falling back. And then the actual precedence is `bearer_access_token_file` → `bearer_access_token` → `oauth_access_token`. Currently only file-versus-bearer precedence is described.

shehabgamin (approved):
@Robin-Everaars This looks great! I have very minor comments so i went ahead and approved.

Btw we have an upcoming release and would love to include this :)

Robin-Everaars:
@copilot please fix the merge conflicts in this pull request.

Robin-Everaars:
@shehabgamin great! Happy to read that! I'm off the next couple days, but I'll try to get it merged in time!

docs/guide/catalog/iceberg-rest.md
linhr: #2346

PR #2297, opened 2026-07-27
fix: propagate swallowed cascade errors in Iceberg drop_database

Robin-Everaars:
## What

Rework `IcebergRestCatalogProvider::drop_database` so the CASCADE path no longer discards per-object and listing failures. Every `list_tables`, `list_views`, `drop_table` and `drop_view` call now propagates a genuine failure as an error, and only specific benign statuses are tolerated.

## Why

The previous code used `if let Ok(...)` around the table and view listings and `let _ =` around each per-object drop. Any error was silently swallowed, so a DROP DATABASE ... CASCADE could report success while tables or views were left behind. A transient 500 on a table drop, or a permission error on a listing, produced a half applied cascade that still deleted (or claimed to delete) the namespace. The failure was invisible to the caller.

## Tolerance rationale

- NOT_FOUND on a per-object drop means a concurrent drop already removed the object, which is an acceptable outcome for a cascade, so we keep going.
- NOT_FOUND on a listing means the namespace itself is already gone. We fall through to `drop_namespace`, which applies the canonical `if_exists` handling (a 404 there succeeds when `if_exists` is set and errors otherwise).
- 405 METHOD_NOT_ALLOWED and 501 NOT_IMPLEMENTED on `list_views` mean the catalog has no views endpoint, which is optional in the Iceberg REST spec. The pre-fix `if let Ok(views)` tolerated any views-list failure, so a view-less catalog must still succeed. These two statuses stay tolerated for `list_views` only. The tables endpoint is mandatory, so a 405 or 501 on `list_tables` is surfaced as a genuine failure.

## Tests

Added regression tests against the existing wiremock harness:

- `test_drop_database_cascade_propagates_table_drop_failure`: a per-object table drop 500 makes `drop_database` return an error and the namespace drop is never sent (asserted with `expect(0)`).
- `test_drop_database_cascade_tolerates_missing_table`: a per-object table drop 404 is tolerated and the cascade proceeds to drop the namespace.
- `test_drop_database_cascade_propagates_list_tables_failure`: a `list_tables` 500 propagates and the namespace drop is never sent.
- `test_drop_database_cascade_propagates_list_views_failure`: a `list_views` 500 propagates and the namespace drop is never sent.
- `test_drop_database_cascade_tolerates_missing_views_endpoint`: a `list_views` 405 is tolerated and the cascade still drops the namespace.
- `test_drop_database_cascade_tolerates_unimplemented_views_endpoint`: a `list_views` 501 is tolerated the same way.
- `test_drop_database_cascade_tolerates_missing_namespace`: listings return 404 and, with `if_exists` set, the trailing namespace 404 succeeds.

## Verification

- `cargo test -p sail-catalog-iceberg`: all pass.
- `cargo clippy -p sail-catalog-iceberg --all-targets --all-features -- -D warnings`: exit 0.
- `cargo fmt --check`: clean.

## Note

This addresses the swallowed-error observation raised in the review of the open bearer-token-file PR (#2288). It touches the same `drop_database` method, so a mechanical rebase is expected once one of the two merges.

### DIFF
--- a/crates/sail-catalog-iceberg/src/provider.rs
+++ b/crates/sail-catalog-iceberg/src/provider.rs
@@ -975,42 +975,99 @@ impl CatalogProvider for IcebergRestCatalogProvider {
             // For CASCADE, first drop all tables and views in the namespace before dropping the namespace.
             // Each request re-reads the credential and retries once on a 401, so a token that rotates
             // partway through the cascade is recovered per request instead of leaving a partial drop.
-            let tables_result = self
+            match self
                 .with_auth_retry(|client| {
                     let prefix = prefix.clone();
                     let ns_string = ns_string.clone();
                     async move { client.list_tables(prefix, ns_string, None, None).await }
                 })
-                .await?;
-            if let Ok(tables) = tables_result {
-                for identifier in tables.inner.identifiers.unwrap_or_default() {
-                    let _ = self
-                        .with_auth_retry(|client| {
-                            let prefix = prefix.clone();
-                            let ns_string = ns_string.clone();
-                            let name = identifier.name.clone();
-                            async move { client.drop_table(prefix, ns_string, name, Some(true)).await }
-                        })
-                        .await?;
+                .await?
+            {
+                Ok(tables) => {
+                    for identifier in tables.inner.identifiers.unwrap_or_default() {
+                        match self
+                            .with_auth_retry(|client| {
+                                let prefix = prefix.clone();
+                                let ns_string = ns_string.clone();
+                                let name = identifier.name.clone();
+                                async move {
+                                    client.drop_table(prefix, ns_string, name, Some(true)).await
+                                }
+                            })
+                            .await?
+                        {
+                            Ok(_) => {}
+                            // The table was already removed (a concurrent drop), which is
+                            // an acceptable outcome for a cascade, so keep going.
+                            Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                            Err(e) => {
+                                return Err(CatalogError::External(format!(
+                                    "Failed to drop table '{}' while cascading namespace drop: {e}",
+                                    identifier.name
+                                )));
+                            }
+                        }
+                    }
+                }
+                // The namespace itself is already gone; fall through to drop_namespace,
+                // which applies the canonical if_exists handling below.
+                Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                Err(e) => {
+                    return Err(CatalogError::External(format!(
+                        "Failed to list tables while cascading namespace drop: {e}"
+                    )));
                 }
             }
-            let views_result = self
+            match self
                 .with_auth_retry(|client| {
                     let prefix = prefix.clone();
                     let ns_string = ns_string.clone();
                     async move { client.list_views(prefix, ns_string, None, None).await }
                 })
-                .await?;
-            if let Ok(views) = views_result {
-                for identifier in views.inner.identifiers.unwrap_or_default() {
-                    let _ = self
-                        .with_auth_retry(|client| {
-                            let prefix = prefix.clone();
-                            let ns_string = ns_string.clone();
-                            let name = identifier.name.clone();
-                            async move { client.drop_view(prefix, ns_string, name).await }
-                        })
-                        .await?;
+                .await?
+            {
+                Ok(views) => {
+                    for identifier in views.inner.identifiers.unwrap_or_default() {
+                        match self
+                            .with_auth_retry(|client| {
+                                let prefix = prefix.clone();
+                                let ns_string = ns_string.clone();
+                                let name = identifier.name.clone();
+                                async move { client.drop_view(prefix, ns_string, name).await }
+                            })
+                            .await?
+                        {
+                            Ok(_) => {}
+                            // The view was already removed (a concurrent drop), which is
+                            // an acceptable outcome for a cascade, so keep going.
+                            Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                            Err(e) => {
+                                return Err(CatalogError::External(format!(
+                                    "Failed to drop view '{}' while cascading namespace drop: {e}",
+                                    identifier.name
+                                )));
+                            }
+                        }
+                    }
+                }
+                // The namespace itself is already gone; fall through to drop_namespace,
+                // which applies the canonical if_exists handling below.
+                Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                // The views endpoint is optional in the Iceberg REST spec, so a catalog
+                // that does not implement it answers 405 or 501. There are then no views
+                // to cascade, so tolerate it and continue to the namespace drop. The tables
+                // endpoint is mandatory, so the list_tables arm above does not tolerate these
+                // statuses and a 405 or 501 there is surfaced as a genuine failure.
+                Err(e)
+                    if matches!(
+                        e.status(),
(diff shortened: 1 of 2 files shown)


crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -927,29 +927,84 @@ impl CatalogProvider for IcebergRestCatalogProvider {
...
+                                )));
+                            }
+                        }
+                    }
+                }
+                // The namespace itself is already gone; fall through to drop_namespace,
+                // which applies the canonical if_exists handling below.
+                Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                Err(e) => {
+                    return Err(CatalogError::External(format!(
```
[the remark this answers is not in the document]
Robin-Everaars: Fixed in 8aee0cfa. A 404 from the mandatory list_tables endpoint now skips the optional list_views request and proceeds directly to drop_namespace, preserving the existing if_exists handling. The missing-namespace regression makes the views endpoint return 500 with expect(0), so it proves that no views request is sent.

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -927,29 +927,84 @@ impl CatalogProvider for IcebergRestCatalogProvider {
...
-                        .drop_view(prefix.clone(), ns_string.clone(), identifier.name)
-                        .await;
+                .await
+            {
+                Ok(views) => {
+                    for identifier in views.inner.identifiers.unwrap_or_default() {
+                        match client
+                            .drop_view(prefix.clone(), ns_string.clone(), identifier.name.clone())
+                            .await
+                        {
```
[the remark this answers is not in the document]
Robin-Everaars: Covered in 8aee0cfa. The new per-view regression returns 500 from DELETE /views/view1, asserts that drop_database returns an error and sets expect(0) on the namespace delete.

crates/sail-catalog-iceberg/src/provider.rs
```diff
@@ -922,51 +922,111 @@ impl CatalogProvider for IcebergRestCatalogProvider {
...
+                    }
+                }
+                // The namespace itself is already gone; fall through to drop_namespace,
+                // which applies the canonical if_exists handling below.
+                Err(e) if e.status() == Some(reqwest::StatusCode::NOT_FOUND) => {}
+                // The views endpoint is optional in the Iceberg REST spec, so a catalog
+                // that does not implement it answers 405 or 501. There are then no views
+                // to cascade, so tolerate it and continue to the namespace drop. The tables
+                // endpoint is mandatory, so the list_tables arm above does not tolerate these
+                // statuses and a 405 or 501 there is surfaced as a genuine failure.
```
lonless9, line 1077: When the `/v1/config` response advertises a non-empty endpoints list that does not contain `GET /v1/{prefix}/namespaces/{namespace}/tables`

lonless9 (commented):
@Robin-Everaars Thanks for this fix, and also for the other Iceberg-related PRs you have opened recently! We appreciate the effort here. We would like to share some context about how we prefer to validate Iceberg changes and how we plan to coordinate the upcoming work.

For Iceberg changes, we generally prefer integration tests as the primary correctness coverage. Many important behaviors depend on the interaction between Sail, the various catalogs, object storage, and Iceberg metadata, so mocked REST tests alone are often insufficient. Wiremock tests are still useful for request-shape validation and cases that require injecting specific HTTP responses, but user-visible catalog behavior should normally also be exercised against a real catalog service.

Our current test infrastructure is roughly divided as follows:
- Visible behavior should normally be covered by the BDD integration tests under `python/pysail/tests/spark/catalog/iceberg_rest/`. For this PR, a good test would go through the SQL-facing interface, and verify that the objects and namespace are actually gone. The scenario can be added to `features/namespace.feature`, reusing the existing shared steps and fixtures.

- More general Iceberg format behavior belongs under `python/pysail/tests/spark/iceberg/features/`.

- Low-level provider interoperability can be tested with the Testcontainers-based Rust harness in crates/sail-catalog-iceberg/tests/rest_integration_test.rs. But we plan to move all of these to pytests. 

The focused catalog tests can be run with:
```
hatch run maturin develop
hatch run pytest -m integration \
  python/pysail/tests/spark/catalog/iceberg_rest/test_namespace_features.py \
  -k cascade -q
 ```
The ignored Rust catalog integration tests can be run with:
```
cargo nextest run -p sail-catalog-iceberg \
  --test rest_integration_test \
  --run-ignored ignored-only
```
On a PR, the catalog integration jobs can also be enabled by adding the run catalog tests label or posting a `[catalog test]` commit message. The CI should auto-run after your first PR gets merged. 

There is also some broader coordination context. We currently have a larger Iceberg MOR feature PR in progress. It may overlap or conflict with some of the other Iceberg PRs, but it is also expected to resolve several correctness issues and provide reusable infrastructure for subsequent Iceberg work. Starting next week, we plan to begin a broader set of Delta Lake and Iceberg changes.

Until that work settles, we will likely focus our review bandwidth on small, self-contained bug-fix PRs such as this one and defer broader or more cross-cutting changes. For larger follow-up PRs, we can coordinate like opening an issue first and avoid some rebasing or adaptation work.

File: crates/sail-catalog-iceberg/src/provider.rs (hunk 3 above)
Comment (Robin-Everaars) [reply crates/sail-catalog-iceberg/src/provider.rs:1077]: Could you clarify the intended behavior here? The comment appears to end after the endpoint path. The current change treats list-tables as mandatory and propagates 405/501 responses, while tolerating those statuses only for the optional list-views endpoint. Should `drop_database(..., cascade: true)` instead inspect the advertised endpoint list before making either request?

**Comment by Robin-Everaars** [comment]
You're welcome. The integration-testing guidance makes sense, especially for catalog behavior that users exercise through SQL. I'm excited about Sail's future and happy to keep contributing.

I've rebased this PR onto the current `main` branch and added the requested SQL-facing BDD scenario in `python/pysail/tests/spark/catalog/iceberg_rest/features/namespace.feature`. It creates a namespace containing an Iceberg table and a view, drops the namespace with `CASCADE`, verifies that the namespace disappears, recreates it and verifies that neither object remains. The Wiremock coverage remains focused on failures that require injected HTTP responses, including listing and per-object drop failures.

The BDD commit includes `[catalog test]` so the catalog integration job runs. I'll keep the broader COW and scoped-overwrite work paused as agreed in #2313 until #2222 lands. That work will be consolidated and validated end to end rather than advanced as stacked PRs.

File: crates/sail-catalog-iceberg/src/provider.rs (hunk 3 above)
Comment (lonless9) [reply crates/sail-catalog-iceberg/src/provider.rs:1077]: Thanks for the clarification. I did some additional probing against Apache Iceberg, Lakekeeper, Nessie, and Unity Catalog. All four advertise `ConfigResponse.endpoints`, but Unity advertises `LOAD_VIEW` without `LIST_VIEWS`; calling the unadvertised route returns `404`, while other unsupported routes return `405`. The spec may also use `406`, so relying on `405/501` is not portable, and missing `LIST_VIEWS` does not necessarily mean there are no views.
The JVM client uses `endpoints` for capability discovery before issuing requests. I think we should similarly preserve this field and use it to preflight the operations required by CASCADE before mutation. For this PR, a focused implementation and tests for CASCADE should be sufficient; broader capability handling can remain a follow-up.

Comment (lonless9) [reply crates/sail-catalog-iceberg/src/provider.rs:1077]: For now here we can say: 
```
// Preserve compatibility with catalogs that signal an unsupported views
// listing with 405 or 501. Proper capability gating should use
// `/v1/config.endpoints`.
```

lonless9 (approved):
Everything looks good! Thanks for the PR! 
Let me think about how to handle the endpoints later.

Robin-Everaars:
> Everything looks good! Thanks for the PR! Let me think about how to handle the endpoints later.

Good to read! You're very welcome!

PR #2299, opened 2026-07-27
test: migrate function doctests to feature files and consolidate with Scenario Outlines

davidlghellin:
## What

Migrates every `test_*.txt` Spark doctest under `python/pysail/tests/spark/function/` into pytest-bdd `.feature` scenarios, deletes the migrated `.txt` files, and reorganizes the result into `Scenario Outline` + `Examples` tables in the style of
`features/math/abs.feature`.

## Why

- The `.txt` doctests were a second, parallel test format. Folding them into the   `.feature` suite gives us one source of truth that runs (and is JVM-validated via `-m` tags) the same way as everything else.
- Consolidating repetitive per-case scenarios into `Scenario Outline` + `Examples` makes the suites far shorter and easier to read/extend, and surfaced a few exact duplicates and over-broad assertions.

## Scope

- **~700 doctests migrated** across math, string, datetime, json, url, csv, map, conditional, array and aggregate. `function/` no longer contains any `test_*.txt`.
- **Grouping into Scenario Outlines** for the migrated blocks *and* the value-heavy math features (`bin`, `bround`, `negative`, `pmod`, `uniform`, `ceil_floor`, …).  Schema-only nullability trios and `@column_args` features are left as individual scenarios (different dimension), matching `abs.feature`.


davidlghellin:
3.5.7 CI fix: the 10 failures were PySpark version differences in schema rendering (3.5 prints null/interval, 4.x prints void/interval day to second; make_interval can't even convert calendar_interval on 3.5) — not Sail behaviour (same type/value underneath). Tagged the affected @spark_null output-schema rules with @spark-4 (skipif pyspark < 4), same as the existing variant/to_time precedent; nullability stays covered on 4.x and values are still tested on 3.5. Verified against the 3.5.7 client: full function suite → 4007 passed, 384 skipped, 0 failed.

[f0925d9](https://github.com/lakehq/sail/pull/2299/commits/f0925d9b860b3282ca2cd9a197864f7e34178245)

linhr (approved):
@davidlghellin It's great to see this!

Since this PR is huge, I'm going to merge it as is so that we can iterate on it in follow-up PRs. I may have some thoughts around tags/markers and test organization and I'm going to explore the ideas after this PR is merged.

davidlghellin:
> @davidlghellin It's great to see this!
> 
> Since this PR is huge, I'm going to merge it as is so that we can iterate on it in follow-up PRs. I may have some thoughts around tags/markers and test organization and I'm going to explore the ideas after this PR is merged.

Thanks @linhr ! 🙏 Tried hard not to drop any case (everything re-validated against a running Sail server; the non-groupable ones kept as individual scenarios), and added the output-schema nullability coverage too.

PR #2303, opened 2026-07-27
perf: optimize regexp_replace for anchored capture extraction

lonless9:
This PR optimizes `regexp_replace` for anchored single-capture extraction by normalizing `$1` to `${1}` and omitting the global flag, enabling DataFusion’s extraction fast path. Other patterns retain existing global-replacement behavior.


linhr (approved):
Nice!

PR #2308, opened 2026-07-28
fix: match Spark's output type for datediff, bitmap helpers and regexp_count/instr

davidlghellin:


shehabgamin:
@davidlghellin Here's an initial review from Codex!
```
  1. crates/sail-plan/src/function/scalar/datetime.rs, new line 363

  > This strict narrowing cast changes Spark's overflow behavior. DateDiff subtracts its internal Int day values directly (end.asInstanceOf[Int] - start.asInstanceOf[Int]), so differences across the full valid
  > DateType range wrap even in ANSI mode. In bundled Spark 4.1.1, SELECT datediff(date_from_unix_date(CAST(2147483647 AS INT)), date_from_unix_date(CAST(-2147483648 AS INT))) returns -1; this path instead
  > computes 4294967295 as Int64, and DataFusion's strict Int64 -> Int32 cast errors. Please implement Spark-compatible wrapping 32-bit subtraction and add this boundary case rather than using a strict narrowing
  > cast.

  2. crates/sail-plan/src/function/scalar/misc.rs, new line 113
     This also covers the identical cast at line 127.

  > This raw DataFusion cast is broader than Spark's implicit LONG coercion. Arrow permits BOOLEAN, DATE, and TIMESTAMP to be cast to Int64, so this PR newly makes calls such as bitmap_bit_position(true) and
  > bitmap_bit_position(DATE '2024-01-02') return 0 and 19723; Spark 4.1.1 rejects both with DATATYPE_MISMATCH.UNEXPECTED_INPUT_TYPE. The pre-PR expression also rejected these during arithmetic coercion, so this
  > is a regression introduced here. DataFusion's explicit cast is additionally strict in non-ANSI mode, whereas Spark's malformed STRING-to-LONG coercion yields NULL. Please validate the source type using
  > Spark's implicit-cast rules and honor ANSI cast semantics before converting to Int64; the same issue applies to bitmap_bucket_number below.

  3. crates/sail-plan/src/function/scalar/misc.rs, new line 119

  > Negating the widened value is not safe for Long.MIN_VALUE. DataFusion wraps array negation but uses checked ScalarValue::arithmetic_negate() for scalar constants, so
  > bitmap_bit_position(CAST(-9223372036854775808 AS DECIMAL(19,0))) now fails with an arithmetic-overflow error after the new Int64 cast; before this hunk it remained decimal and returned 0. Spark's Java
  > arithmetic wraps here and its BitmapExpressionUtilsSuite explicitly expects bitmapBitPosition(Long.MinValue) == 0. Please reduce modulo before negating, or otherwise preserve Spark's wrapping semantics, and
  > add Long.MIN_VALUE scalar coverage rather than stopping at Int.MIN_VALUE.

  I found no additional issues in the regexp changes, nullability changes, aliases, or cluster codec behavior. Current CI is fully green, but it does not cover these boundaries. I did not repeat the three
  return-type test comments already posted by Copilot.
  ```

PR #2309, opened 2026-07-28
fix: validate Iceberg mutation snapshot conflicts

Robin-Everaars:
## Summary

- bind Iceberg mutation commits to the snapshot observed during physical planning
- preserve the tri-state expected snapshot through optimizer child rewrites and remote-execution protobuf round trips
- validate both initially empty tables and concrete snapshots before writing mutation metadata

## Stack

Stacked on #2300. The S4-only change is commit `92ea472f`; review this PR after the selective-rewrite base.

## Verification

- `cargo test -p sail-execution iceberg_commit_ --lib`
- `cargo test -p sail-iceberg --lib`
- `cargo clippy -p sail-execution -p sail-iceberg --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7856,6 +7856,7 @@ dependencies = [
  "serde",
  "serde_arrow",
  "serde_json",
+ "tokio",
  "url",
  "uuid",
 ]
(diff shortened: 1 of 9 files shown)


Robin-Everaars:
Superseded by #2473.

PR #2310, opened 2026-07-28
fix: derive scalar UDF output nullability from `return_field_from_args`

davidlghellin:
DataFusion's default `return_field_from_args` hardcodes `nullable = true` (datafusion-expr 54, udf.rs:685), so every scalar UDF that only implemented `return_type` reported a nullable output regardless of its inputs.

Move the output type logic into `return_field_from_args` across all 124 scalar UDF impls and turn `return_type` into a guard, matching the DataFusion idiom (21 of the 22 upstream impls that override the hook do the same).

The nullability rule for each function was read from the Spark source (`Expression.nullable`, resolved through the class hierarchy) rather than inferred from the function name:

- null-intolerant expressions derive from the argument fields
- `override def nullable = false` becomes a fixed `false`
- `override def nullable = true` stays `true`, with the source cited inline

25 functions now report the output nullability Spark declares, clearing 40 `@sail-bug` tags from the `@spark_null` suite: abs, array_compact, array_position, arrays_zip, base64, bin, hash, is_variant_null, last_day, levenshtein, make_ym_interval, negative, now, regexp_extract, regexp_extract_all, reverse, schema_of_csv, soundex, to_char, to_unix_timestamp, to_varchar, unbase64, unix_timestamp, version and year. Each tag was verified individually against a running server before being removed.

`convert_tz`, `spark_date_format` and `make_valid_utf8` deliberately keep `nullable = true`. Each backs a composed function (`to_utc_timestamp`, `from_utc_timestamp`, `from_unixtime`) whose Spark counterpart has a nullable implicit cast as its child, which Sail constant-folds away. Deriving from the arguments there would report `false` where Spark reports `true`, letting the optimizer drop null checks.

Also fixes the four callers of `ScalarUDFImpl::return_type` that the guard would otherwise have broken at runtime rather than at compile time: the explode resolver in sail-plan, `spark_try_parse_url`, and the unit tests in `array_intersect` and `spark_base64`.

Refs #1903


shehabgamin:
@davidlghellin Here's a first pass from Codex!
```
  ## Introduced regressions

  1. spark_sequence.rs:82 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/array/spark_sequence.rs#L82)

  > Spark's implicit coercion can make a Sequence child nullable even when the source expression is not. With ANSI enabled, Spark 4.1.1 reports sequence(DATE '2020-01-01', '2020-01-03') as nullable because the
  > string is implicitly cast to Date, and Cast.forceNullable(StringType, DateType) is true. Sail's coerce_types likewise maps Utf8 to Date32, but DataFusion preserves the source field's nullable = false, so this
  > line reports false. Please preserve the cast-induced nullability here and add this mixed DATE/string schema case.

  2. spark_interval.rs:67 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_interval.rs#L67)

  > This loses Spark's forced nullability for string-to-interval casts. These UDFs implement CAST(string AS INTERVAL), but Spark's Cast.nullable is child.nullable || Cast.forceNullable(...), and forceNullable
  > returns true for every String-to-non-string cast. Spark therefore declares all three interval cast forms nullable, whereas this returns false for non-null literals. Please model Spark's cast nullability here
  > rather than deriving it from the kernel's error behavior.

  3. spark_last_day.rs:62 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_last_day.rs#L62)

  > This misses the nullable implicit cast that Spark inserts for string inputs. DataFusion's UDF coercion changes a non-null Utf8 field to Date32 while preserving nullable = false, but Spark's Cast(StringType,
  > DateType) has forceNullable = true; consequently Spark reports last_day('2024-01-15') as nullable. This new rule reports it non-nullable, so please preserve the string-to-date cast's nullability rather than
  > copying the coerced field bit.

  4. spark_make_ym_interval.rs:58 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_make_ym_interval.rs#L58)

  > This misses the nullable implicit casts for the string arguments accepted by coerce_types. DataFusion replaces their field types with Int32 but keeps a non-null string field's nullable = false; Spark's implicit
  > String-to-Integer casts are force-nullable, so make_ym_interval('1', '2') has nullable = true in Spark while this declares false. Please account for the coercion-induced nullability here.

  5. spark_time_diff.rs:71 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_time_diff.rs#L71)

  > TimeDiff is RuntimeReplaceable, so Spark does not use TernaryExpression's nullability here. RuntimeReplaceable.nullable delegates to the replacement, and this replacement is a StaticInvoke with the default
  > returnNullable = true. Spark therefore reports time_diff as always nullable. Please keep this field nullable.

  6. spark_time_trunc.rs:65 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_time_trunc.rs#L65)

  > TimeTrunc is also RuntimeReplaceable; its nullability comes from the replacement, not BinaryExpression. The replacement is a StaticInvoke with the default returnNullable = true, so Spark always reports this
  > result nullable. Please keep this field nullable.

  7. spark_unix_timestamp.rs:60 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_unix_timestamp.rs#L60)

  > This is only Spark's ANSI-on rule. ToUnixTimestamp and UnixTimestamp set failOnError = SQLConf.ansiEnabled, and ToTimestamp.nullable is if (failOnError) children.exists(_.nullable) else true. With ANSI
  > disabled, a non-null malformed string may return NULL, so advertising this field as non-nullable is unsafe. Please carry ansi_mode into this UDF and its codec, or conservatively keep the result nullable.

  8. spark_year.rs:62 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/datetime/spark_year.rs#L62)

  > This misses the nullable implicit cast that Spark inserts for string inputs. DataFusion's UDF coercion changes a non-null Utf8 field to Date32 while preserving nullable = false, but Spark's Cast(StringType,
  > DateType) has forceNullable = true; consequently Spark reports year('2024-01-15') as nullable. This new rule reports it non-nullable, so please preserve the string-to-date cast's nullability rather than copying
  > the coerced field bit.

  9. spark_abs.rs:91 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/math/spark_abs.rs#L91)

  > abs cannot blindly inherit the coerced argument field's nullability here. Sail's UDF coercion turns STRING into a regular DataFusion Cast(..., Float64), whose field preserves the non-null source flag. Spark's
  > implicit STRING-to-DOUBLE cast is instead always nullable, so Spark reports abs('13') as nullable = true in both ANSI modes. Before this PR DataFusion's default UDF field was true; this change makes it false.
  > Please carry nullable metadata through the STRING coercion path and add a schema regression case.

  10. spark_bin.rs:64 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/math/spark_bin.rs#L64)

  > bin cannot always inherit the argument field's nullability here. Spark casts its input to LONG, and both STRING-to-LONG and fractional-to-integral casts are force-nullable. Sail instead keeps STRING/FLOAT
  > inputs and performs the fallible conversion inside this UDF. Thus this PR makes calls such as bin('13'), bin('ab'), and bin(13.3) non-nullable, while Spark reports them nullable (bin(13) remains false). Please
  > carry Spark-equivalent cast nullability through these coercion paths and add schema regression cases.

  11. spark_negative.rs:88 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/math/spark_negative.rs#L88)

  > Under ANSI mode, the planner wraps STRING input in a regular Cast(..., Float64) before this UDF, and DataFusion preserves the non-null string field's nullability on that cast. Spark's STRING-to-DOUBLE cast is
  > force-nullable even in ANSI mode, so Spark reports negative('13') as nullable = true; this PR changes it from the previous default true to false. Please make the ANSI STRING-cast path carry nullable metadata
  > into this hook and cover it with a schema test. The non-ANSI path already uses TryCast and remains nullable.

  12. spark_to_char.rs:82 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/string/spark_to_char.rs#L82)

  > to_char cannot always inherit the raw argument fields' nullability here. Spark first casts the value to DecimalType; STRING-to-DECIMAL and FLOAT/DOUBLE-to-DECIMAL are force-nullable casts, so Spark reports
  > to_char('12', '99') and to_char(CAST(12 AS DOUBLE), '99') as nullable, while an integral input is not. Sail keeps those inputs as UTF-8/float and performs the fallible conversion inside this UDF, so this hook
  > marks non-null STRING/float calls false. Please account for these coercion paths and add schema regression cases.

  ## Incorrect newly added rules

  These do not worsen the base behavior, but the new hooks still encode rules that contradict Spark.

  13. spark_to_binary.rs:57 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/string/spark_to_binary.rs#L57)

  > ToBinary is format-dependent, but this applies the default/hex rule to every format. Spark resolves to_binary(..., 'base64') to UnBase64, which is null-intolerant with no nullable override, so to_binary('YWJj',
  > 'base64') is non-nullable. Please inspect scalar_arguments and derive the base64 branch from the input field rather than hard-coding true for every format.

  14. update_struct_field.rs:194 (https://github.com/lakehq/sail/blob/7382be650f126c9766e32c41190434708ba9dd19/crates/sail-function/src/scalar/update_struct_field.rs#L194)

  > Spark does not derive UpdateFields nullability from all children. complexTypeCreator.scala:755 explicitly defines override def nullable: Boolean = structExpr.nullable; a nullable replacement value changes the
  > nested field's nullability but cannot make a non-null input struct null. Please derive this top-level field only from args.arg_fields[0].
```

davidlghellin:
To wrap this up, I'll split it into a few PRs, grouping related changes together.

PR #2311, opened 2026-07-28
feat: support copy-on-write Iceberg delete

Robin-Everaars:
## Summary

- plan Iceberg `DELETE` as a strict-snapshot, manifest-pruned copy-on-write mutation
- scan and rewrite only candidate data files, preserving SQL `NULL` semantics with `IS NOT TRUE` and supporting conditionless delete-all
- reject tables with active merge-on-read delete files to prevent row resurrection
- carry removed paths and `Operation::Delete` through optimizer rewrites and the distributed physical-plan codec
- write conservative Parquet-derived Iceberg bounds and preserve them through action batches and Avro manifest decoding so sparse predicates can safely reuse unaffected manifests

## Stack

- depends on #2309 (strict Iceberg snapshot conflicts)
- #2309 depends on #2300 (selective Iceberg rewrite)

## Validation

- `cargo test -p sail-iceberg --lib` — 105 passed
- `cargo test -p sail-execution iceberg_commit_rewrite_round_trip_preserves_paths_and_operation --lib` — passed
- `cargo clippy -p sail-iceberg -p sail-execution --lib --tests -- -D warnings` — passed
- current-wheel Iceberg DELETE BDD — 4 passed (selective manifest reuse, SQL NULL retention, delete-all, validated no-candidate no-op)
- `cargo fmt --all -- --check` — passed
- Ruff check and format check for the changed Python step module — passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7856,6 +7856,7 @@ dependencies = [
  "serde",
  "serde_arrow",
  "serde_json",
+ "tokio",
  "url",
  "uuid",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7852,6 +7852,7 @@ dependencies = [
  "sail-common",
  "sail-common-datafusion",
  "sail-data-source",
+ "sail-logical-plan",
  "sail-sql-analyzer",
  "serde",
  "serde_arrow",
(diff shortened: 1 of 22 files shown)


Robin-Everaars:
Follow-up `09de996e` corrects the Iceberg `DELETE` result count. Copy-on-write still rewrites survivors, but the command now returns predicate-matched affected rows by subtracting the committed survivor count from the candidate files' manifest row total.

This preserves zero for validation-only and conservatively pruned false-positive candidates. `IS NOT TRUE` survivor semantics also keep SQL `NULL` handling unchanged.

Verified with:

- all 109 `sail-iceberg` tests
- Clippy with warnings denied
- an exact rebuilt-wheel test that deletes two of three rows, returns `count = 2` and retains the one survivor

Robin-Everaars:
Superseded by #2473.

PR #2312, opened 2026-07-28
refactor: consolidate schema type merging logic

### LINKED ISSUE #1646: Consolidate type merging logic for `schema_of_` functions
Right now, for example, for the `schema_of_variant` function, we merge the types within an array by parsing the type string again. This is unnecessary if we have some internal representation for the type during schema inference. We then convert such internal representation to the string at the end after the recursion.

Such logic can be reused for other formats such as `schema_of_json`.

rohankumardubey:
## Summary

- Introduce a shared internal `InferredType` representation and recursive type-merging logic.
- Reuse it across `schema_of_json`, `schema_of_variant`, and `schema_of_variant_agg`.
- Preserve the format-specific merging and rendering behavior for JSON and Variant schemas.
- Avoid rendering and reparsing type strings while recursively merging Variant arrays.
- Add regression coverage for merging object fields containing decimal types.

Previously, types such as `DECIMAL(3,2)` could be incorrectly parsed while merging rendered object schemas because the type string contains a comma.

Closes #1646

## Testing

- `cargo test -p sail-function`
- `cargo clippy --all-targets --all-features -- -D warnings`
- Focused `schema_of_json`, `schema_of_variant`, and `schema_of_variant_agg` PySpark feature tests

### DIFF
--- a/crates/sail-function/src/aggregate/schema_of_variant_agg.rs
+++ b/crates/sail-function/src/aggregate/schema_of_variant_agg.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::sync::Arc;
 
 use arrow::array::{Array, ArrayRef};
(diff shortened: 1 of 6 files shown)


linhr (approved):
This looks great! Thanks for the contribution!

Issue #1972, opened 2026-05-24
Full support for Java datetime format


linhr:
As I work on #1925 I realized there are still gaps for supporting Java datetime format.

Right now, we are relying on translating the Java formatting strings to the Rust `chrono` formatting strings using rules defined by regex, but the two libraries have many differences. For example, I don't see `chrono` supporting optional fields, and specifying literals in the formatting string requires actual work beyond the regex-based translation rules.

I feel we should implement all the parsing and formatting logic from scratch by translating the Java code into Rust. I found a library as a proof of concept: <https://github.com/palantir/date-time-formatter-rs>. This is a literal translation of Java code into Rust with large scale generated test cases. However, I think we can still partially rely on the `chrono` libraries and design the code following Rust convention and practices.

Here is a high-level design based on my analysis:

1. Define a `DateTimeFormat` struct which contains a `Vec` of `DateTimeItem`s. Each `DateTimeItem` is one element in the formatting string. This "format" will be used to instruct how we parse and format strings.
2. Use Chumsky to write a parser that parses Java formatting strings into this `DateTimeFormat` representation.
3. Implement the formatter and parser for `DateTimeFormat`:
    1. For the formatter, we can design an input struct similar to `DelayedFormat` in `chrono`. We then translate the Java code to format each `DateTimeItem`. We don't need to port `java.time.temporal.TemporalAccessor` since we will access `chrono` date and time structs directly.
    2. For the parser, we can directly return `Parsed` struct from `chrono` instead of translating Java's equivalent of `java.time.format.Parsed`. This allows us to customize the parser while reusing the resolution logic from `chrono`. (In the chrono documentation, it specifically mentions the `Parsed` struct can be used for the case of custom parsers.)

The related Java code is `java.time.format.DateTimeFormatter` and `java.time.format.DateTimeFormatterBuilder`. I guess AI coding might help us with the translation, and since this is a well-defined problem with easily generated test cases, the correctness can be verified easily.

To begin with, we support only the US locale (since locale support is unstable in `chrono`) and Gregorian calendar.

linhr:
As the next step, we would like to figure out the following:

1. The detailed design of `DateTimeFormat` struct and `DateTimeItem` enum.
2. The design of formatter input, similar to `DelayedFormat`.
3. An analysis of whether the `Parsed` struct (parser output) covers all corner cases of the Java datetime formatting logic.
[...]

PR #2322, opened 2026-07-30
fix: improve Spark datetime compatibility

lonless9:
- Align Java datetime pattern validation and parsing/formatting semantics with Spark.
- Enforce strict input consumption, clock fields, fractions, eras, optional sections, and timezone offsets.
- Preserve wall-clock values when parsing formatted `TIMESTAMP_NTZ` inputs.
- Fix `to_date`, `unix_timestamp`, and `to_unix_timestamp` handling for dynamic/NULL formats, typed inputs, ANSI mode, nullability, lazy evaluation, and session timezones.
- Propagate Unix timestamp ANSI state through the driver/worker codec.
- Migrate TIME, CSV, JSON, and XML consumers to the shared datetime engine and remove the legacy Spark-to-Chrono format converter.


lonless9:
Part of https://github.com/lakehq/sail/issues/1972, but since the conversion layer has already been removed, we also need to assess whether the performance optimization brought by https://github.com/lakehq/sail/pull/1925 has been lost or not, which would be a follow-up work.

PR #2329, opened 2026-07-31
refactor: specify components for actor creation

linhr:


crates/sail-session/src/session_manager/actor/handler.rs
```diff
@@ -256,6 +256,7 @@ impl SessionManagerActor {
                 if let Some(driver_id) = driver_id.take() {
                     self.drivers.remove(driver_id);
                 }
+                session.deleted_at.get_or_insert_with(Utc::now);
```
linhr, line 259: This reverts the code removal in this [commit](https://github.com/lakehq/sail/pull/2327/changes/1710b10cc9c49e4880fff3b2679e8f3f05f62008) in #2327.

PR #2331, opened 2026-07-31
[WIP]  add Web UI

weixiuli:
This Web UI is provided only as an example and is implemented with reference to the Spark Web UI.
1. Physical Operator DAG 
740fc93c63cf6c03386429cbfd797292

2. Detailed View of the Physical Operator DAG
f0608279defd3490cb4bf9a97904576a


weixiuli:
@linhr

shehabgamin:
@weixiuli This is epic!!

linhr:
Thanks for sharing this demo!

I skimmed through the code and now I get an understanding around potential work involved to support UI. It's also good to see this end-to-end working sample so that it's clear what the UI needs to support to provide useful insights around query execution performance.

I'm going to move this PR to draft for now. I'll refer to the implementation here as I improve observability data collection to support UI development.

Issue #2333, opened 2026-07-31
Binary_view is not supported in conversion to Arrow.


weixiuli:
This issue did not occur in previous versions, but it appears when using the latest main branch.

The last commit where the page layout worked correctly was:https://github.com/lakehq/sail/commit/93afaab0f3622554804dde1aa18ee25dfb3bd653

```
pyspark.errors.exceptions.connect.PythonException:
  An exception was thrown from the Python worker. Please see the stack trace below.
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/pyspark/sql/pandas/types.py", line 465, in from_arrow_type
    raise PySparkTypeError(
pyspark.errors.exceptions.base.PySparkTypeError: [UNSUPPORTED_DATA_TYPE_FOR_ARROW_CONVERSION] binary_view is not supported in conversion to Arrow.
```

shehabgamin:
I asked Codex to diagnose, it appears that bumping to Spark 4.2 is the culprit. cc @linhr

---
## Diagnosis

The responsible commit is [568910654a58af180b0f19d4202f2affcac87a20](https://github.com/lakehq/sail/commit/568910654a58af180b0f19d4202f2affcac87a20), `chore: support Spark 4.2 (#2289)`.

I confirmed this with an exact historical A/B:

- `93afaab` with PySpark 4.1.1: an ordinary binary identity UDF over Parquet passes.
- Current main with PySpark 4.2.0: the same UDF fails at `pyspark/sql/pandas/types.py:465` with `binary_view is not supported`.

### Exact breaking change

The commit made two jointly necessary changes:

1. It adopted Spark 4.2’s default:

```diff
spark.sql.execution.pythonUDF.arrow.enabled:
- false
+ true
```

Compare [Spark 4.1 config](/Users/r/Desktop/lakesail/sail/crates/sail-spark-connect/data/config/spark-4.1.json:1203) with [Spark 4.2 config](/Users/r/Desktop/lakesail/sail/crates/sail-spark-connect/data/config/spark-4.2.json:1274). Consequently, an ordinary `udf(...)` now uses the Arrow-batched path automatically.

2. It added this PySpark 4.2 fast path at [spark.py:703](/Users/r/Desktop/lakesail/sail/crates/sail-python-udf/src/python/spark.py:703):

```python
return _output_column(
    self._udf(None, iter([_record_batch(args, num_rows=num_rows)]))
)
```

[...]

PR #2334, opened 2026-07-31
fix: cast view-type inputs in Arrow-optimized Python UDFs

shehabgamin:
https://github.com/lakehq/sail/issues/2333

### DIFF
--- a/crates/sail-python-udf/src/cereal/mod.rs
+++ b/crates/sail-python-udf/src/cereal/mod.rs
@@ -162,7 +162,12 @@ fn build_input_types_json(input_types: &[DataType]) -> PyUdfResult<String> {
             .iter()
             .enumerate()
             .map(|(i, dt)| -> PyResult<_> {
-                let arrow_type = dt.to_pyarrow(py)?;
+                // Arrow view types are internal representations, not Spark SQL types.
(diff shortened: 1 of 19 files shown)


crates/sail-python-udf/src/cereal/mod.rs
```diff
@@ -162,7 +163,16 @@ fn build_input_types_json(input_types: &[DataType]) -> PyUdfResult<String> {
...
             .map(|(i, dt)| -> PyResult<_> {
-                let arrow_type = dt.to_pyarrow(py)?;
+                // Arrow view types are internal representations, not Spark SQL types.
+                let arrow_type = match dt {
+                    DataType::BinaryView if large_var_types => {
+                        DataType::LargeBinary.to_pyarrow(py)?
+                    }
+                    DataType::BinaryView => DataType::Binary.to_pyarrow(py)?,
+                    DataType::Utf8View if large_var_types => DataType::LargeUtf8.to_pyarrow(py)?,
+                    DataType::Utf8View => DataType::Utf8.to_pyarrow(py)?,
```
linhr: I'm not sure if we need this. I think it can also be taken care of in `normalize_data_type`?

linhr (approved):
Nice!

PR #2336, opened 2026-08-01
fix: require cascade for non-empty memory databases

ShiroKSH:
## Summary

- reject non-CASCADE drops of non-empty in-memory databases
- make the precondition check and removal atomic, so a concurrent table or view create cannot be erased after the check
- add regression coverage for tables and views surviving the rejected drop and CASCADE removal succeeding

## Tests

- `cargo fmt --check`
- `cargo test -p sail-catalog-memory`
- `cargo clippy -p sail-catalog-memory --all-targets --all-features -- -D warnings`

### DIFF
--- a/crates/sail-catalog-memory/src/provider.rs
+++ b/crates/sail-catalog-memory/src/provider.rs
@@ -149,19 +149,29 @@ impl CatalogProvider for MemoryCatalogProvider {
     ) -> CatalogResult<()> {
         let DropDatabaseOptions {
             if_exists,
-            cascade: _,
+            cascade,
         } = options;
-        if self.databases.remove(database).is_none() {
-            if if_exists {
+        match self.databases.entry(database.clone()) {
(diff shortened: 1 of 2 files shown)


crates/sail-catalog-memory/src/provider.rs
```diff
@@ -521,3 +526,76 @@ impl CatalogProvider for MemoryCatalogProvider {
         }
     }
 }
+
+#[cfg(test)]
```
linhr: Could you remove Rust unit tests but add BDD tests in `.feature` files?

linhr (commented):
Thanks for the contribution!

The implementation looks good to me. I have one comment regarding testing.

ShiroKSH:
I updated the branch with requested BDD migration + atomic non-empty check+removal behavior in in-memory catalog drop. @linhr please review again.

linhr (approved):
This looks great! Thanks again!

PR #2337, opened 2026-08-01
fix: do not round-robin repartition streaming plans

calvinchengx:
## Problem

A streaming query starts, reports `isActive`, raises no exception, and delivers
nothing to its sink for a very long time.

```python
spark.readStream.format("rate").option("rowsPerSecond", 5).load() \
     .writeStream.format("console").start()
```

On a multi-core machine this prints nothing. The query is not stuck and no data
is lost — it is buffered.

## Cause

The physical planner inserts a round-robin repartition to use the available
parallelism:

```
ConsoleSinkExec
  RepartitionExec: partitioning=RoundRobinBatch(14), input_partitions=1
    RateSourceExec: rows_per_second=5, num_partitions=1
```

`RepartitionExec` coalesces on the producer side. `OutputChannel::coalesce`
pushes into a `LimitedBatchCoalescer` and drains only batches that have reached
`target_batch_size` (`execution.batch_size`, default 8192). The residual is
flushed in `finalize()`, which runs when the last input sender finishes.

An unbounded source never finishes, so the residual is never flushed. The rate
source emits one row per batch, so the first output appears only once 8192 rows
have accumulated — about 27 minutes at 5 rows/s, and never at all for a source
that is idle.

This is confirmed by varying only `batch_size`, with the repartition present in
both cases:

| `default_parallelism` | `batch_size` | rows generated | rows delivered |
|---|---|---|---|
| 4 | 1 | 24 | 24 |
| 4 | 8192 (default) | 24 | 0 |

## Fix

Round-robin repartitioning trades latency for parallelism, which is the wrong
trade for an unbounded input — and streaming plans are not meant to be
repartitioned in the first place: the streaming rewriter already rejects a
logical `Repartition` with `not_impl_err!("streaming repartition")`. This keeps
the physical planner from reintroducing one, by disabling
`enable_round_robin_repartition` for streaming plans only.

A streaming source that is itself multi-partition keeps its partitions, so real
parallelism is unaffected. Batch queries are untouched.

## Tests

`test_streaming_repartition.py` pins the plan shape both ways: a streaming plan
has no `RoundRobinBatch`, and a batch plan still does.

Asserting delivery directly would be better, but no sink is observable from a
Connect client — `console` writes to the server's stdout and `memory` is not
implemented. That gap is why the existing streaming tests pass while nothing
flows: they assert `isActive`, `status`, and `exception() is None`, all of which
hold during a total stall.

## Note for upstream DataFusion

Reported as https://github.com/apache/datafusion/issues/24044, with a standalone
repro that does not involve Sail.

Avoiding the repartition side-steps the buffering here, but any
`RepartitionExec` over an unbounded stream has the same unbounded-latency
behaviour. The change in this PR stands on its own regardless of what DataFusion
decides: the streaming rewriter already rejects a logical `Repartition`, and
artificial round-robin fan-out over an unbounded source buys nothing.

### DIFF
--- a/crates/sail-plan/src/lib.rs
+++ b/crates/sail-plan/src/lib.rs
@@ -41,30 +41,13 @@ pub async fn resolve_and_execute_plan(
     let NamedPlan { plan, fields } = resolver.resolve_named_plan(plan).await?;
     info.push(plan.to_stringified(PlanType::InitialLogicalPlan));
     let df = execute_logical_plan(ctx, plan).await?;
-    let (mut session_state, plan) = df.into_parts();
+    let (session_state, plan) = df.into_parts();
     let plan = session_state.optimize(&plan)?;
-    let streaming = is_streaming_plan(&plan)?;
-    let plan = if streaming {
+    let plan = if is_streaming_plan(&plan)? {
         rewrite_streaming_plan(plan)?
     } else {
         plan
     };
-    if streaming {
-        // Round-robin repartitioning exists to buy parallelism, and for an
-        // unbounded input it costs unbounded latency instead: `RepartitionExec`
-        // coalesces on the producer side, holding rows until `batch_size` of
-        // them accumulate, and only flushes the remainder when the input ends.
-        // A stream does not end, so at the default batch size a slow source
-        // delivers nothing for a very long time. Streaming plans are not meant
-        // to be repartitioned anyway — the streaming rewriter rejects a logical
-        // `Repartition` outright — so keep the physical planner from
-        // reintroducing one.
-        session_state
-            .config_mut()
-            .options_mut()
-            .optimizer
-            .enable_round_robin_repartition = false;
-    }
     info.push(plan.to_stringified(PlanType::FinalLogicalPlan));
     let plan = session_state
         .query_planner()
--- a/crates/sail-plan/src/lib.rs
+++ b/crates/sail-plan/src/lib.rs
@@ -41,13 +41,30 @@ pub async fn resolve_and_execute_plan(
     let NamedPlan { plan, fields } = resolver.resolve_named_plan(plan).await?;
     info.push(plan.to_stringified(PlanType::InitialLogicalPlan));
     let df = execute_logical_plan(ctx, plan).await?;
-    let (session_state, plan) = df.into_parts();
+    let (mut session_state, plan) = df.into_parts();
     let plan = session_state.optimize(&plan)?;
-    let plan = if is_streaming_plan(&plan)? {
+    let streaming = is_streaming_plan(&plan)?;
+    let plan = if streaming {
         rewrite_streaming_plan(plan)?
     } else {
         plan
     };
+    if streaming {
+        // Round-robin repartitioning exists to buy parallelism, and for an
+        // unbounded input it costs unbounded latency instead: `RepartitionExec`
+        // coalesces on the producer side, holding rows until `batch_size` of
+        // them accumulate, and only flushes the remainder when the input ends.
+        // A stream does not end, so at the default batch size a slow source
+        // delivers nothing for a very long time. Streaming plans are not meant
+        // to be repartitioned anyway — the streaming rewriter rejects a logical
+        // `Repartition` outright — so keep the physical planner from
+        // reintroducing one.
+        session_state
+            .config_mut()
+            .options_mut()
+            .optimizer
+            .enable_round_robin_repartition = false;
+    }
     info.push(plan.to_stringified(PlanType::FinalLogicalPlan));
     let plan = session_state
         .query_planner()
--- a/python/pysail/tests/spark/streaming/test_streaming_repartition.py
+++ b/python/pysail/tests/spark/streaming/test_streaming_repartition.py
@@ -8,36 +8,38 @@ size a slow source delivers nothing for a very long time — the query reports
 active the whole while.
 
 There is no sink a Connect client can assert delivery against (`console` writes
-to the server's stdout, and `memory` is not implemented), so this pins the plan
-shape instead.
+to the server's stdout, and `memory` is not implemented), so these tests pin the
+plan shape instead: the streaming plan snapshot must not contain a
+`RepartitionExec`, and the batch plan snapshot must keep its `RoundRobinBatch`
+so the streaming rule is known not to have leaked into batch planning.
 """
 
 from __future__ import annotations
 
 import pytest
 
+from pysail.testing.spark.steps.plan import normalize_plan_text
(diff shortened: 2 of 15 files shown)


python/pysail/tests/spark/streaming/test_streaming_repartition.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pysail.testing.spark.utils.common import is_jvm_spark
+
+if is_jvm_spark():
+    pytest.skip("Sail streaming tests", allow_module_level=True)
+
+
+def test_streaming_plan_has_no_round_robin_repartition(spark, capsys):
+    query = spark.readStream.format("rate").option("rowsPerSecond", 5).load().writeStream.format("console").start()
+    try:
```
linhr, line 32: Let's use the `snapshot` fixture to capture the full physical plan.

For batch query, use the `_explain_string` method to get the string. There are existing Python tests that you can refer to for snapshot testing on `EXPLAIN` outputs.

For streaming query, here is the implementation in the spark source code. You can implement a utility function in `python/pysail/testing/spark/utils/sql.py` to return the explain string using the similar idea. Then you can having snapshot testing logic.

```python
    def explain(self, extended: bool = False) -> None:
        cmd = pb2.StreamingQueryCommand()
        cmd.explain.extended = extended
        result = self._execute_streaming_query_cmd(cmd).explain.result
        print(result)
```

crates/sail-plan/src/lib.rs
```diff
@@ -41,13 +41,30 @@ pub async fn resolve_and_execute_plan(
...
+        // delivers nothing for a very long time. Streaming plans are not meant
+        // to be repartitioned anyway — the streaming rewriter rejects a logical
+        // `Repartition` outright — so keep the physical planner from
+        // reintroducing one.
+        session_state
+            .config_mut()
+            .options_mut()
+            .optimizer
+            .enable_round_robin_repartition = false;
+    }
```
linhr: It is not desirable to modify session config. The plan resolver may be run concurrently for multiple queries received by the server, so it's hard to reason about potential race condition here.

linhr (changes requested):
Thanks for looking into this.

I think the issue makes sense, but the bug fix needs more thoughts. Let me know if you have questions!

crates/sail-plan/src/lib.rs
calvinchengx: Agreed — thanks, that's a fair objection and broader than a race: even if `into_parts()` returns an owned clone today, relying on that is fragile.

Moved the fix into a physical optimizer rule, `RemoveStreamingRoundRobinRepartition`, alongside the existing custom rules in `sail-physical-optimizer`. No session state is touched — `crates/sail-plan/src/lib.rs` is back to unchanged and has dropped out of this PR entirely.

Two things worth flagging about the rule:

- It removes a `RepartitionExec` only when the partitioning is `RoundRobinBatch` **and** the input carries the flow event schema. A hash repartition changes semantics rather than merely adding parallelism, so if one ever appears in a streaming plan it is left alone rather than silently dropped. There is a unit test for each of those three cases.
- It is placed immediately **before** `EnsureCooperative`, not right after `EnforceDistribution`. That ordering is load-bearing: once the repartition is gone, the unbounded source needs its own `CooperativeExec` yield point, and `EnsureCooperative` has to see the final shape of the plan to insert one.

python/pysail/tests/spark/streaming/test_streaming_repartition.py
calvinchengx: Done — both tests now use the `snapshot` fixture.

The batch one uses `_explain_string`, and for the streaming one I added `streaming_explain_string()` to `python/pysail/testing/spark/utils/sql.py`, following the implementation you linked. I checked `_execute_streaming_query_cmd` and the `StreamingQueryCommand` proto exist on PySpark 3.5.9 as well as 4.2.0, so it should hold across the test matrix.

Two notes on the resulting diff:

- Adding the rule changed **28 codegen-mode snapshots**, each by a single `physical_plan after RemoveStreamingRoundRobinRepartition:` / `SAME TEXT AS ABOVE` line. No plan changed.
- Regenerating snapshots across several test directories in one `pytest` session also produced spurious `physical_plan after JoinReorder:` lines. Running the `execution` directory on its own did not. I have not chased it down, but `optimizer/test_features.py` sets `SAIL_OPTIMIZER__ENABLE_JOIN_REORDER=true` for its own server, so it looks like that setting reaching servers created later in the same process. I regenerated per directory to avoid it — mentioning it in case it bites someone else.

One limitation I should be upfront about: these snapshots pin the plan *shape*, not delivery. There is no sink a Connect client can assert delivery against — `console` writes to the server's stdout and `memory` is not implemented — so a future change that broke delivery without changing the plan would not be caught here. That gap is why this bug went unnoticed: the existing streaming tests assert `isActive`, `status` and `exception() is None`, all of which hold during a total stall. Happy to add a client-observable sink separately if that would be useful.

calvinchengx:
Both comments are addressed. I have also included CI evidence, since the runs on this PR are still waiting on maintainer approval and there is nothing visible from your side.

**Session config.** Gone. The fix is now a physical optimizer rule, `RemoveStreamingRoundRobinRepartition`, in `sail-physical-optimizer` alongside `RewriteExplicitRepartition` and the other custom rules, so no shared state is touched. `crates/sail-plan/src/lib.rs` is byte identical to `main` and no longer appears in the diff.

Two details are load bearing. The rule removes only `RoundRobinBatch` partitioning over a flow event input, leaving hash repartitions alone since those change semantics rather than merely adding parallelism. And it runs immediately before `EnsureCooperative`, because once the repartition is gone the unbounded source needs its own `CooperativeExec` yield point.

**Snapshot tests.** Both plan assertions now use the `snapshot` fixture: the batch one via `_explain_string`, the streaming one via a new `streaming_explain_string()` helper in `python/pysail/testing/spark/utils/sql.py`, built from the implementation you linked. I checked the underlying proto exists on PySpark 3.5.9 as well as 4.2.0.

On diff size: 28 snapshot files change, each by a single `physical_plan after RemoveStreamingRoundRobinRepartition:` / `SAME TEXT AS ABOVE` line. No plan changed.

CI is green on the same pipeline for this exact commit (`9a75962c`), run on my public fork:

- [Build](https://github.com/calvinchengx/sail/actions/runs/30739366050): Rust Build, Rust Tests, Python Tests on Spark 3.5.9, 4.0.4, 4.1.3 and 4.2.0, Docs Build
- [Rust Lint](https://github.com/calvinchengx/sail/actions/runs/30739365961)
- [Python Lint](https://github.com/calvinchengx/sail/actions/runs/30739365984)

No rush from my side, just making the state visible.

PR #2348, opened 2026-08-03
fix: retry Iceberg REST bootstrap authentication

Robin-Everaars:
## Summary

- Retry the initial `GET /v1/config` request once after a `401 Unauthorized` response.
- Rebuild the bootstrap client with freshly loaded credentials without recursively resolving the catalog configuration.
- Cover token rotation between the first bootstrap request and its retry.

## Context

This follows up on #2288. Regular Iceberg REST requests reload credentials and retry once after a 401, but catalog initialization made its configuration request before that retry path was available. A projected token rotating during the bootstrap request could therefore prevent the catalog from initializing.

The token-file input, error and precedence contract is already documented on the current `main` branch.

## Testing

- `cargo test -p sail-catalog -p sail-catalog-iceberg`
- `cargo clippy -p sail-catalog -p sail-catalog-iceberg --all-targets --all-features -- -D warnings`
- `cargo +nightly fmt -- --check`

### DIFF
--- a/crates/sail-catalog-iceberg/src/provider.rs
+++ b/crates/sail-catalog-iceberg/src/provider.rs
@@ -188,6 +188,28 @@ impl IcebergRestCatalogProvider {
         .await
     }
 
+    /// Retry the bootstrap configuration request once on `401 Unauthorized`.
+    /// This uses [`Self::bootstrap_client`] rather than [`Self::client`]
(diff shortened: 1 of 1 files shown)


Robin-Everaars:
@shehabgamin here are the last changes of what I was working on yesterday!

shehabgamin (approved):
Looks great!

PR #2351, opened 2026-08-03
fix: validate explicit Parquet read schemas

### LINKED ISSUE #2298: Reading Parquet with an explicit schema silently narrows values instead of failing like Spark
## Description

When reading Parquet with an explicit `StructType`, Spark only allows a small set of safe upcasts between the declared type and the type stored in the file. Any other conversion is rejected at read time with `FAILED_READ_FILE.PARQUET_COLUMN_DATA_TYPE_MISMATCH`.

Sail performs the cast instead, and returns silently converted values. For narrowing conversions this is silent data loss: `9.5` is read back as `9`, and the integers `1, 2, 3` are read back as `true, true, true`.

This matters most in the setup where it was found: no catalog at all, reading files directly with a declared schema. There the `StructType` is the only type contract there is, so a cast that Spark treats as a hard error goes completely unnoticed.

## Steps to reproduce

Parquet file written with `id INT, score DOUBLE`, values `(1, 9.5), (2, 7.25), (3, NULL)`.

```python
import os
import tempfile

from pyspark.sql.types import DoubleType, IntegerType, StructField, StructType

remote = os.environ.get("SPARK_REMOTE", "sc://localhost:50051")
if remote == "local":
    # PySpark's builder treats SPARK_REMOTE as a Connect URL, so it has to go
    # before we can ask for a local master.
    os.environ.pop("SPARK_REMOTE", None)
    spark = SparkSession.builder.master("local[2]").getOrCreate()
    engine = "SPARK JVM"
else:
    spark = SparkSession.builder.remote(remote).getOrCreate()
    engine = "SAIL"

path = os.path.join(tempfile.mkdtemp(), "t.parquet")
spark.createDataFrame(
    [(1, 9.5), (2, 7.25)],
    StructType([StructField("id", IntegerType()), StructField("score", DoubleType())]),
).write.parquet(path)

print(f"===== {engine} =====")
print("file schema:    ", spark.read.parquet(path).schema.simpleString())

# The file stores score as DOUBLE. Declare it as INT on purpose.
narrowed = StructType([StructField("score", IntegerType())])
print("declared schema:", narrowed.simpleString())

try:
    rows = spark.read.schema(narrowed).parquet(path).collect()
    print(f"RESULT: {rows}   <-- 9.5 and 7.25 were truncated, no error raised")
except Exception as e:  # noqa: BLE001
    first = " ".join(str(e).split())[:200]
    print(f"RAISED: {type(e).__name__}: {first}")

```

Run against Sail (server on port 50051):

```bash
SPARK_REMOTE="sc://localhost:50051" hatch run python repro_parquet_schema_cast.py
```

Run against Spark JVM:

```bash
source "$HOME/.sdkman/bin/sdkman-init.sh" && sdk use java 17.0.9-tem SPARK_REMOTE="local" hatch run python  repro_parquet_schema_cast.py
```

```
===== SAIL =====
file schema:     struct<id:int,score:double>
declared schema: struct<score:int>
RESULT: [Row(score=9), Row(score=7)]   <-- 9.5 and 7.25 were truncated, no error raised

===== SPARK JVM =====
file schema:     struct<id:int,score:double>
declared schema: struct<score:int>
.....
RAISED: Py4JJavaError: An error occurred while calling o54.collectToPython. : org.apache.spark.SparkException: [FAILED_READ_FILE.PARQUET_COLUMN_DATA_TYPE_MISMATCH] Encountered error while reading file file:///var/folders/p4

```

rohankumardubey:
## Summary

- validate explicit Parquet schemas against physical file schemas
- reject incompatible type conversions with a Spark-style error
- support compatible widening and recursively validate nested types
- preserve missing fields, partition columns, and case-insensitive matching
- add Sail-only read/write round-trip tests without JVM Spark

Closes #2298

## Testing

- `cargo test -p sail-data-source --lib`
- `cargo clippy -p sail-data-source --all-targets -- -D warnings`
- `hatch run python -m pytest python/pysail/tests/spark/datasource/test_parquet.py -q`

### DIFF
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -226,8 +226,7 @@ fn validate_parquet_field(path: &str, requested: &Field, physical: &Field) -> Re
         }
         (physical_type, requested_type) => plan_err!(
             "[FAILED_READ_FILE.PARQUET_COLUMN_DATA_TYPE_MISMATCH] Data type mismatches when reading \
-             Parquet column `{path}`. Expected Spark type {requested_type}, actual Parquet type \
-             {physical_type}."
+             Parquet column `{path}`. Expected {requested_type}, actual Parquet type {physical_type}."
         ),
     }
 }
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use datafusion::arrow::datatypes::{Fields, Schema, SchemaRef, TimeUnit};
+use datafusion::arrow::datatypes::{DataType, Field, Fields, Schema, SchemaRef, TimeUnit};
 use datafusion::catalog::Session;
 use datafusion::datasource::physical_plan::ParquetSource;
 use datafusion::datasource::physical_plan::parquet::CachedParquetFileReaderFactory;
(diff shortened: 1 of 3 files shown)


linhr (approved):
This looks great! Thanks for having this!

PR #2353, opened 2026-08-04
test: initialize the global state when reusing an existing server

davidlghellin:
When it starts a server in process, `spark_connect_server` creates a `SparkConnectServer`, which initializes the process-global state as a side effect. That state holds the snapshot of environment variables that static configuration warnings are compared against, and it is captured lazily, the first time the global state is initialized.

When `SPARK_REMOTE` points to an existing server the helper returns early and never creates one, so the global state stays uninitialized. The first test in `test_config.py` then captures the snapshot itself, after its own `monkeypatch`, so `warn_if_changed` finds no difference and emits no warning. Whichever of the three tests runs first fails, and the other two pass only because the first one left the global state initialized.

Creating the server here, without starting it, makes both branches behave the same. The constructor does not bind a listener; `start` does.

```
SPARK_REMOTE="sc://localhost:50051" hatch run pytest python/pysail/tests/spark/test_config.py -k deletion
```
fail:
image

```
SPARK_REMOTE="sc://localhost:50051" hatch run pytest python/pysail/tests/spark/test_config.py -k "deletion or modification"
```
With two tests, the first one fails.

image

ok:
image

### DIFF
--- a/python/pysail/tests/spark/test_config.py
+++ b/python/pysail/tests/spark/test_config.py
@@ -8,7 +8,7 @@ from pysail.spark import SparkConnectServer
 @pytest.fixture(scope="module", autouse=True)
 def server():
     # The tests below need the global state to be initialized while the environment variables
-    # are still pristine, which does not happen when running against a server from `SPARK_REMOTE`.
+    # are still pristine. This fixture ensures this even when running the tests against a server from `SPARK_REMOTE`.
     _ = SparkConnectServer()
 
 
--- a/python/pysail/tests/spark/test_config.py
+++ b/python/pysail/tests/spark/test_config.py
@@ -5,6 +5,13 @@ import pytest
 from pysail.spark import SparkConnectServer
 
 
+@pytest.fixture(scope="module", autouse=True)
+def server():
+    # The tests below need the global state to be initialized while the environment variables
+    # are still pristine, which does not happen when running against a server from `SPARK_REMOTE`.
+    _ = SparkConnectServer()
+
+
 def test_warning_on_static_config_addition(monkeypatch):
     key = "SAIL_TELEMETRY__EXPORTER__OTLP__ENDPOINT"
     assert key not in os.environ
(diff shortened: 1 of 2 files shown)


linhr (commented):
Thanks for finding out this edge case due to selected test runs!

I'd prefer to avoid modifying the common settings. How about adding the following to `test_config.py`? And you can add a comment explaining why this is needed.

```python
@pytest.fixture(scope="module", autouse=True)
def server():
    _ = SparkConnectServer()
```

davidlghellin:
> Thanks for finding out this edge case due to selected test runs!
> 
> I'd prefer to avoid modifying the common settings. How about adding the following to `test_config.py`? And you can add a comment explaining why this is needed.
> 
> ```python
> @pytest.fixture(scope="module", autouse=True)
> def server():
>     _ = SparkConnectServer()
> ```

Sure. Reverted the change to the common settings and moved it into `test_config.py` with a comment explaining why it is needed.

davidlghellin:
> ## Pull request overview
> This PR fixes order-dependent/flaky behavior in `python/pysail/tests/spark/test_config.py` when running tests against an existing Spark Connect server via `SPARK_REMOTE`, by ensuring Sail’s process-global state (including the environment snapshot used for static-config change warnings) is initialized before any test-level `monkeypatch` modifications occur.
> 
> **Changes:**
> 
> * Add a module-scoped, autouse pytest fixture that instantiates `SparkConnectServer` once at module setup time to initialize the process-global state while environment variables are still pristine.
> * Ensure consistent warning behavior regardless of whether tests reuse an external server (`SPARK_REMOTE`) or start an in-process server.

That summary is from the bot — I just enabled auto review and it fired right away 😄

python/pysail/tests/spark/test_config.py
```diff
@@ -5,6 +5,13 @@
 from pysail.spark import SparkConnectServer
 
 
+@pytest.fixture(scope="module", autouse=True)
+def server():
+    # The tests below need the global state to be initialized while the environment variables
+    # are still pristine, which does not happen when running against a server from `SPARK_REMOTE`.
```
linhr:
```suggestion
    # The tests below need the global state to be initialized while the environment variables
    # are still pristine. This fixture ensures this even when running the tests against a server from `SPARK_REMOTE`.
```

linhr (approved):
Nice! Thanks!

PR #2361, opened 2026-08-05
feat: add zip_with and map_zip_with functions

twsl:
Part of #229

- Add support for `zip_with` and `map_zip_with`


crates/sail-function/src/scalar/array/lambda_utils.rs
```diff
@@ -229,3 +296,1130 @@ fn offsets_to_indices<O: OffsetSizeTrait>(
     }
     Ok(Arc::new(Int32Array::from(out)) as ArrayRef)
 }
+
+#[cfg(test)]
+mod tests {
```
linhr, line 301: Why do we have three copies of duplicated tests?

Please review the 6 `mod tests` in this file and make sure they are intentional. Ideally we should just have one `mod tests` with `#[cfg(test)]`.

crates/sail-plan/src/explain.rs
```diff
@@ -271,15 +271,23 @@ async fn collect_plan_with(
...
-                            explain_config.show_statistics,
-                            explain_config.show_schema,
-                        ),
-                    ));
+                    if current != previous || optimizer_name != "wrap_higher_order_functions" {
+                        stringified.push(StringifiedPlan::new(
+                            PlanType::OptimizedPhysicalPlan { optimizer_name },
+                            current,
+                        ));
+                    }
```
linhr, line 290: Let's avoid changes to this file. If the added optimizer changes any physical plan snapshots, please commit them.

crates/sail-spark-connect/tests/gold_data/function/lambda.json
```diff
@@ -774,8 +774,8 @@
         }
       },
       "output": {
-        "failure": "not implemented: function: zip_with"
+        "success": "ok"
       }
     }
   ]
-}
+}
```
linhr, line 781: This end of file change is unintentional.

linhr (commented):
Thanks for the contribution!

I haven't reviewed this PR, but here are just a few thoughts on the overall structure.

linhr:
Here is a review from Codex. I haven't verified the accuracy, but I just wanted to share it here for your reference.

> - **P1: Preserve aggregate ordering when rewriting expressions**  
>   `crates/sail-physical-optimizer/src/wrap_higher_order.rs:173`
> 
>   `wrap_aggregate` calls `with_new_expressions(new_args, vec![])`, dropping `ORDER BY` expressions. For order-sensitive aggregates, rewriting fails and leaves the higher-order expression unwrapped, which then fails distributed serialization.
> 
> - **P1: Preserve grouping-set metadata when rebuilding aggregates**  
>   `crates/sail-physical-optimizer/src/wrap_higher_order.rs:200`
> 
>   Rebuilding `PhysicalGroupBy` with `new_single` discards grouping-set masks, null expressions, and the grouping-set flag. Queries using `GROUPING SETS`, `ROLLUP`, or `CUBE` can return different results when a higher-order function appears in a grouping key.
> 
> - **P1: Preserve join configuration when wrapping join filters**  
>   `crates/sail-physical-optimizer/src/wrap_higher_order.rs:382`
> 
>   Reconstructing `HashJoinExec` resets its projection, partition mode, and `null_aware` flag. This can change the output schema or join semantics. The equivalent nested-loop-join rewrite also drops its projection. Rebuild from the existing join configuration and replace only the filter.
> 
> - **P1: Wrap higher-order repartition keys**  
>   `crates/sail-physical-optimizer/src/wrap_higher_order.rs:503`
> 
>   The wrapping rule does not handle hash partitioning expressions in repartition nodes. A higher-order function used as a repartition key remains unwrapped, then reaches distributed serialization and is rejected by the codec.
> 
> - **P1: Fix `FixedSizeList` normalization**  
>   `crates/sail-function/src/scalar/array/lambda_utils.rs:104`
> 
>   The `FixedSizeList` conversion constructs the resulting list child field with `List<item>` type but supplies `item` values. Arrow rejects that type mismatch, causing a panic for `zip_with` inputs backed by `FixedSizeList`.

It seems that the expression wrapper may have many issues. I can see that this is a workaround for the missing codec in DataFusion. Would it make sense to have the proper fix upstream? It might also be useful for a few other projects that rely on distributed processing for DataFusion. What do you think?

twsl:
this is definitely a pr with a scope that got bigger than anticipated. i agree that the optimizer workaround is too broad and serialization should be fixed in datafusion without physical plan rewriting. not sure if or when i can find the time. maybe treat this as inspiration 😄

PR #2372, opened 2026-08-06
fix: correct SQL sequence syntax descriptor cardinality

lonless9:
## Summary

Describe `Sequence<T, S>` as `T (S T)*` instead of `T (S T)+`

The runtime parser already accepts one head item followed by zero or more
separator-item pairs. This change aligns the generated syntax descriptor with
that behavior.


crates/sail-sql-parser/tests/syntax.rs
```diff
@@ -1,6 +1,26 @@
...
+        sequence_syntax.node,
+        SyntaxNode::Sequence(vec![
+            SyntaxNode::NonTerminal(TypeId::of::<Ident>()),
+            SyntaxNode::ZeroOrMore(Box::new(SyntaxNode::Sequence(vec![
+                SyntaxNode::NonTerminal(TypeId::of::<Comma>()),
+                SyntaxNode::NonTerminal(TypeId::of::<Ident>()),
+            ]))),
+        ])
+    );
+}
```
linhr: We can remove this test.

linhr (approved):
Nice catch!

PR #2374, opened 2026-08-06
fix: resolve Delta MERGE constraints by visible names

alexy:
## Summary

Resolve Delta MERGE constraints from visible source names so the generic Sail correction used by Grust cognition can land upstream.

## Verification

- Built the exact local Sail source revision.
- Ran the bounded Grust Sail integration harness in source mode: 26 Sail executor tests, two live backend tests, and two live cognition parity/secrecy tests passed.


lonless9 (approved):
LGTM! Thanks for having this!

PR #2375, opened 2026-08-06
test: align function tests with Spark values

davidlghellin:


python/pysail/tests/spark/function/test_format_string.py
```diff
@@ -88,43 +90,45 @@ def test_format_string_grouping_separator(spark):
...
 
 def test_format_string_scientific_uppercase(spark):
-    actual = spark.sql("SELECT format_string('Scientific: %E', 1234.5) AS result").toPandas()
+    actual = spark.sql("SELECT format_string('Scientific: %E', 1234.5D) AS result").toPandas()
     expected = pd.DataFrame({"result": ["Scientific: 1.234500E+03"]})
     pd.testing.assert_frame_equal(actual, expected)
 
 
+# Java's %g keeps 6 significant digits (`1234.50`); Sail trims to `1234.5`.
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 117: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

python/pysail/tests/spark/function/test_format_string.py
```diff
@@ -304,31 +343,37 @@ def test_format_string_float_negative_infinity(spark):
...
     actual = spark.sql(
-        "SELECT format_string('String: %s, Integer: %d, Float: %.2f', 'test', 42, 3.14159) AS result"
+        "SELECT format_string('String: %s, Integer: %d, Float: %.2f', 'test', 42, 3.14159D) AS result"
     ).toPandas()
     expected = pd.DataFrame({"result": ["String: test, Integer: 42, Float: 3.14"]})
     pd.testing.assert_frame_equal(actual, expected)
 
 
+# Spark surfaces Java's MissingFormatArgumentException verbatim; Sail words it differently.
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 353: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

python/pysail/tests/spark/function/test_format_string.py
```diff
@@ -304,31 +343,37 @@ def test_format_string_float_negative_infinity(spark):
...
-        AnalysisException,
-        match=r"Argument index .* out of bounds",
+        IllegalArgumentException,
+        match=r"Format specifier '%d'",
     ):
         spark.sql("SELECT format_string('Value: %d')").collect()
 
 
+# Same: Spark reports `Format specifier '%s'`, Sail its own message.
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 363: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

python/pysail/tests/spark/function/test_format_string.py
```diff
@@ -304,31 +343,37 @@ def test_format_string_float_negative_infinity(spark):
...
-        AnalysisException,
-        match="Argument index 2 is out of bounds",
+        IllegalArgumentException,
+        match=r"Format specifier '%s'",
     ):
         spark.sql("SELECT format_string('Values: %d %s', 42)").collect()
 
 
+# Spark reports Java's IllegalFormatConversionException; Sail its own message.
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 373: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

python/pysail/tests/spark/function/test_product.py
```diff
@@ -0,0 +1,190 @@
...
+    rows = (
+        spark.sql("SELECT * FROM VALUES ('a', 2), ('a', 3), ('a', 4) AS t(g, v)")
+        .select("g", "v", F.product("v").over(window).alias("result"))
+        .orderBy("v")
+        .collect()
+    )
+    assert [r["result"] for r in rows] == [2.0, 6.0, 24.0]
+
+
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 184: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

python/pysail/tests/spark/function/test_quote.py
```diff
@@ -3,11 +3,16 @@
 import pandas as pd
 import pytest
 
-from pysail.testing.spark.utils.common import pyspark_version
+from pysail.testing.spark.utils.common import is_jvm_spark, pyspark_version
 
 
+@pytest.mark.xfail(not is_jvm_spark(), reason="Known Sail bug", strict=True)
```
shehabgamin, line 9: We should maybe add a `TODO` comment or `FIXME` comment coupled with an issue so this doesn't get lost?

shehabgamin (approved):
Looks great, just left comments in places where adding a `TODO`/`FIXME` coupled with a GitHub issue could help with visibility

linhr:
I feel maybe an epic issue to track all `xfail` test cases would be enough. We can easily filter tests by markers so that we can examine those cases and plan the fix. There are also a ton of `@sail-bug` in BDD tests (which translates to the `xfail` marker during test setup) so adding TODO with GitHub links would be tedious. Also if the test case is fixed, `strict=True` ensures that "unexpected success" triggers a test failure which would remind us to remove the marker.

@shehabgamin I feel the marker itself is good enough for tracking the work, without individual issues for each case. What do you think?

PR #2377, opened 2026-08-06
fix: promote date interval addition to timestamp

### LINKED ISSUE #2370: DATE + sub-day INTERVAL returns DATE (drops time), Spark returns TIMESTAMP
## Summary

Adding a **sub-day `INTERVAL`** (minutes/hours/seconds) to a `DATE` keeps the result type as `DATE` and silently drops the time-of-day component. Spark promotes the result to a timestamp and carries the time. As a result, `date + 5*INTERVAL 1 MINUTE` and `date + 10*INTERVAL 1 MINUTE` are indistinguishable on Sail, which also causes downstream row loss through `dropDuplicates`.

## Environment

- Sail server **4.2.0** (pysail v0.7.0), Spark Connect
- pyspark **3.5.7** client
- Session timezone UTC
- Reproduced on a fresh cluster with fully synthetic data.

## Reproduction

```python

r = spark.sql("""
  SELECT to_date('20260730','yyyyMMdd')                        AS d,
         to_date('20260730','yyyyMMdd') + 5*INTERVAL 1 MINUTE  AS plus5,
         to_date('20260730','yyyyMMdd') + 10*INTERVAL 1 MINUTE AS plus10
""")
print(r.dtypes)
r.show(truncate=False)
```

### Actual (Sail 4.2.0)

```
dtypes: [('d', 'date'), ('plus5', 'date'), ('plus10', 'date')]
+----------+----------+----------+
|d         |plus5     |plus10    |
+----------+----------+----------+
|2026-07-30|2026-07-30|2026-07-30|
+----------+----------+----------+
```

### Expected (Spark 3.5)

```
dtypes: [('d','date'), ('plus5','timestamp_ntz'), ('plus10','timestamp_ntz')]
+----------+-------------------+-------------------+
|d         |plus5              |plus10             |
+----------+-------------------+-------------------+
|2026-07-30|2026-07-30 00:05:00|2026-07-30 00:10:00|
+----------+-------------------+-------------------+
```

## Downstream impact: silent row loss via dropDuplicates

Because the sub-day component is discarded, rows that differ only in the added minutes become byte-identical and are collapsed by `dropDuplicates`:

```python
df = spark.createDataFrame([("A", 1), ("A", 2)], ["id", "n"])
df2 = df.selectExpr(
    "id",
    "to_date('20260730','yyyyMMdd') + (n*5)*INTERVAL 1 MINUTE AS ts",
)
print("before dedup:", df2.count(), "after:", df2.dropDuplicates().count())
# Sail  -> before dedup: 2 after: 1   (ts is date 2026-07-30 for both rows)
# Spark -> before dedup: 2 after: 2   (ts is 00:05:00 vs 00:10:00)
```

A common ETL pattern derives a per-record timestamp as `to_date(<date_string>) + (<sequence_column> * k) * INTERVAL 1 MINUTE` to give records within the same day distinct times. On Spark those rows are distinct and survive a subsequent `dropDuplicates`; on Sail they collapse to the same bare `date`, so rows are silently lost and the output column comes out as `date` where Spark produces `timestamp_ntz`.

## Notes

- Timestamp base already works: `timestamp + n*INTERVAL 1 MINUTE` is correct on Sail. The bug is specific to a **`DATE`** left operand with a sub-day interval not being promoted to timestamp.
- Related but distinct: #2237 (`date_trunc` timestamp return type), #2350 (interval leading/trailing fields). Neither covers `DATE + sub-day INTERVAL` type promotion.

rohankumardubey:
## Summary

- promote `DATE` operands to `TIMESTAMP_NTZ` when adding sub-day intervals
- preserve minute, hour, and second components instead of truncating them
- keep existing date-plus-number behavior unchanged
- add parameterized BDD coverage for values, result types, nulls, reversed operands, and column-derived intervals

Closes #2370

## Testing

- `cargo test -p sail-plan --lib`
- `cargo clippy -p sail-plan --all-targets -- -D warnings`
- focused BDD tests for date and interval arithmetic

### DIFF
--- a/crates/sail-plan/src/function/scalar/math.rs
+++ b/crates/sail-plan/src/function/scalar/math.rs
@@ -64,10 +64,12 @@ fn spark_plus(input: ScalarFunctionInput) -> PlanResult<Expr> {
         );
         Ok(match (left_type, right_type) {
             (Ok(DataType::Date32), Ok(DataType::Duration(TimeUnit::Microsecond))) => {
-                left + cast(right, DataType::Interval(IntervalUnit::MonthDayNano))
+                cast(left, DataType::Timestamp(TimeUnit::Microsecond, None))
+                    + cast(right, DataType::Interval(IntervalUnit::MonthDayNano))
             }
             (Ok(DataType::Duration(TimeUnit::Microsecond)), Ok(DataType::Date32)) => {
-                cast(left, DataType::Interval(IntervalUnit::MonthDayNano)) + right
+                cast(left, DataType::Interval(IntervalUnit::MonthDayNano))
+                    + cast(right, DataType::Timestamp(TimeUnit::Microsecond, None))
             }
             (Ok(left_type), Ok(DataType::Date32)) if left_type.is_numeric() => {
                 cast(left + cast(right, DataType::Int32), DataType::Date32)
--- a/python/pysail/tests/spark/function/features/datetime/date_interval_arithmetic.feature
+++ b/python/pysail/tests/spark/function/features/datetime/date_interval_arithmetic.feature
@@ -27,7 +27,7 @@ Feature: Date and interval arithmetic
         FROM (
           SELECT n, DATE '2026-07-30' + (n * 5) * INTERVAL 1 MINUTE AS result
           FROM VALUES (1), (2) AS t(n)
-        )
+        ) AS q
         ORDER BY n
         """
       Then query result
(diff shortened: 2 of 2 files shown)


shehabgamin:
@rohankumardubey Looks like you may need to update the snapshots:
https://docs.lakesail.com/sail/latest/development/build/python.html#updating-test-snapshots

rohankumardubey:
> @rohankumardubey Looks like you may need to update the snapshots: https://docs.lakesail.com/sail/latest/development/build/python.html#updating-test-snapshots

@shehabgamin Thanks! The snapshot differences revealed that the current change also promotes day-only intervals, which should continue returning `DATE`. Updating the snapshots would hide that regression. I’ll wait for #2350, then rebase and use its interval-field metadata to promote only intervals containing sub-day fields.

PR #2401, opened 2026-08-10
fix: lay the correctness groundwork for the DataFusion upgrade

lonless9:
## Summary

This PR fixes several correctness and Spark parity issues on DataFusion 54:

- Preserve `fetch`, null-aware semantics, and dynamic-filter state when rewriting `CollectLeft` hash joins.
- Disable dynamic-filter pushdown in cluster driver and worker sessions until cross-stage routing is supported. Local execution remains unchanged.
- Mark physical projections containing scalar subqueries as nullable, since an empty scalar subquery returns `NULL`.
- Preserve arbitrary binary bytes in `concat` instead of casting them through UTF-8 strings.
- Implement Spark-compatible `spark.sql.mapKeyDedupPolicy`:
  - default: `EXCEPTION`
  - optional: `LAST_WIN`
  - preserve the policy through remote UDF codecs
- Add remote physical-plan codec support for `PartitionedTopKExec`.

## PartitionedTopK

`WindowTopN` continues to use DataFusion's `enable_window_topn` option, which defaults to `false`. DataFusion keeps it opt-in because `PartitionedTopKExec` may regress memory usage and runtime for high-cardinality partition keys.

## Dynamic filters

Dynamic-filter pushdown is disabled for local-cluster and Kubernetes-cluster execution at both planning and worker-session boundaries.

## Scalar subquery nullability

The physical planner now makes expressions containing scalar subqueries nullable. The accompanying `EXPLAIN CODEGEN` snapshot verifies that:

- the outer and subquery `RangeExec` columns remain non-nullable;
- the ordinary output column remains `Int64`;
- the expression containing the scalar subquery becomes `Int64;N`;
- the distributed plan preserves the scalar-subquery stage and broadcast input.

This change corrects the physical execution schema. Logical `AnalyzePlan.Schema` behavior remains outside the scope of this workaround.


crates/sail-execution/src/proto/codec.rs
```diff
@@ -2893,7 +2963,7 @@ impl PhysicalExtensionCodec for RemoteExecutionCodec {
             "spark_width_bucket" | "width_bucket" => {
                 Ok(Arc::new(ScalarUDF::from(SparkWidthBucket::new())))
             }
-            "str_to_map" => Ok(Arc::new(ScalarUDF::from(StrToMap::new()))),
+            "str_to_map" => Ok(Arc::new(ScalarUDF::from(StrToMap::default()))),
```
linhr: Should we just remove this?

crates/sail-function/src/scalar/map/str_to_map.rs
```diff
@@ -22,16 +22,17 @@ use crate::scalar::string::spark_split::{SparkSplit, parse_regex, split_to_array
...
     signature: Signature,
+    last_value_wins: bool,
 }
 
 impl Default for StrToMap {
     fn default() -> Self {
-        Self::new()
+        Self::new(false)
     }
 }
```
linhr: I feel we should just remove this to prevent misuse.

crates/sail-session/src/session_factory/server.rs
```diff
@@ -230,3 +231,19 @@ impl ServerSessionFactory {
...
+    mode: &ExecutionMode,
+) -> Result<()> {
+    if matches!(
+        mode,
+        ExecutionMode::LocalCluster | ExecutionMode::KubernetesCluster
+    ) {
+        config.options_mut().set(
+            "datafusion.optimizer.enable_dynamic_filter_pushdown",
+            "false",
+        )?;
```
linhr: I'd prefer to set the field directly instead of using string keys, so that issues can be caught at compile time. Also in this way we won't need to return `Result<()>` since setting the field would always succeed.

Also let's use `match` to match all variants, so that we will notice in case we add another cluster mode in the future.

crates/sail-session/src/session_factory/worker.rs
```diff
@@ -33,11 +33,15 @@ impl SessionFactory<()> for WorkerSessionFactory {
...
-        let config = SessionConfig::default()
+        let mut config = SessionConfig::default()
             .with_extension(Arc::new(DeltaTableCache::default()))
             .with_extension(Arc::new(RepartitionBufferConfig::new(
                 self.repartition_buffer_size,
             )));
+        config.options_mut().set(
+            "datafusion.optimizer.enable_dynamic_filter_pushdown",
+            "false",
+        )?;
```
linhr: We won't need to worry about worker since the optimizer is not run there.

linhr (approved):
This is great!

PR #2411, opened 2026-08-11
fix: align USING and NATURAL join output-key ownership with Spark

### LINKED ISSUE #2387: Join on columns with different types (DATE vs STRING) fails with coalesce coercion error instead of implicit cast
## Summary

Joining two DataFrames `on=` a column that is typed `DATE` on one side and `STRING` on the other fails on Sail 4.2.0 with an internal `coalesce` type-coercion error, whereas Spark implicitly casts one side and performs the join. The error message references `coalesce`, suggesting Sail's join-key type unification is implemented via a `coalesce`-style coercion path that doesn't have a case for `Date32`/`Utf8`.

## Environment

## Reproduction

```python

left = spark.createDataFrame([("p1", "2026-01-31")], ["pid", "period_end"]) \
    .withColumn("period_end", F.expr("last_day(period_end)"))  # -> DATE
right = spark.createDataFrame([("p1", "2026-01-31")], ["pid", "period_end"])  # stays STRING

print(left.dtypes)   # [('pid', 'string'), ('period_end', 'date')]
print(right.dtypes)  # [('pid', 'string'), ('period_end', 'string')]

joined = left.join(right, on=["pid", "period_end"], how="inner")
joined.show()
```

Output:

```
pyspark.errors.exceptions.connect.AnalysisException:
Execution error: Function 'coalesce' user-defined coercion failed with: Execution error:
Fail to find the coerced type, errors: Execution error: Expect to get struct but got Date32.
No function matches the given name and argument types 'coalesce(Date32, Utf8)'.
You might need to add explicit type casts.
	Candidate functions:
	coalesce(UserDefined)
```

## Expected vs actual

- **Expected (Spark):** the join implicitly casts the STRING side to DATE (or vice versa) for key comparison and returns the matched row.
- **Actual (Sail 4.2.0):** the join-key type-unification path calls `coalesce(Date32, Utf8)` internally, which has no matching signature, and the whole join fails with an `AnalysisException` before any row comparison happens.

## Impact

Any join where one side's join key was derived via a date function (`last_day`, `to_date`, etc.) upstream and the other side's matching key is still a raw string column — a common shape when two sources are prepared independently before being joined on a shared key like a period-end date — fails outright on Sail, even though the values are otherwise identical and Spark handles the coercion transparently.

lonless9:
## Summary

Align visible output-key ownership for `USING` and `NATURAL` joins with Spark.

Inner and left joins retain the left key, right joins retain the right key, and only full outer joins synthesize a coalesced key using Spark-compatible coercion. Join-predicate type coercion remains a separate existing path, and hidden child-key aliases remain available for qualified DataFrame column resolution.

Closes #2387

### DIFF
--- a/crates/sail-plan/src/resolver/query/join.rs
+++ b/crates/sail-plan/src/resolver/query/join.rs
@@ -3,11 +3,12 @@ use std::sync::Arc;
 use datafusion_common::{Column, JoinType, NullEquality};
 use datafusion_expr::utils::split_conjunction;
 use datafusion_expr::{Expr, LogicalPlan, LogicalPlanBuilder, build_join_schema};
-use datafusion_functions::expr_fn::coalesce;
 use sail_common::spec;
 use sail_python_udf::udf::pyspark_udf::PySparkUDF;
 
 use crate::error::{PlanError, PlanResult};
+use crate::function::common::{FunctionContextInput, ScalarFunctionInput};
+use crate::function::get_built_in_function;
 use crate::resolver::PlanResolver;
 use crate::resolver::state::PlanResolverState;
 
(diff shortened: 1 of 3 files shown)


linhr (approved):
Looks good to me!

PR #2412, opened 2026-08-11
fix: replace malformed UTF-8 in CSV reads

### LINKED ISSUE #2388: CSV reader errors on invalid UTF-8 bytes instead of reading leniently like Spark 3.5
## Summary

Sail's CSV reader aborts the entire read with a hard `Csv error: Encountered invalid UTF-8 data ...` when a source file contains any byte that is not valid UTF-8 (e.g. a lone `0xFF`, a bare continuation byte, or a Latin-1 byte such as `0xE9` for "é"). Spark 3.5's CSV reader tolerates such bytes — it decodes leniently, substituting the Unicode replacement character (U+FFFD) for malformed input, and reads the row rather than failing the whole file.

This makes any CSV that carries legacy / Latin-1 / mixed-encoding bytes unreadable on Sail, where the same file reads fine on Spark 3.5.

## Environment

- Sail server **4.2.0** / pysail **0.7.0**
- Client: **pyspark 3.5.7** over Spark Connect (`sc://<host>:50051`)

## Minimal repro

Self-contained: writes a tiny CSV whose second data row carries bytes that are not valid UTF-8 — a Latin-1 `0xE9` ("é") followed by a lone `0xFF` — then reads it back. Adjust the path to any location both the client and Sail server can see.

```python

path = "/tmp/bad_utf8.csv"

# Exact bytes written (27 bytes). 0xE9 and 0xFF are both invalid as UTF-8.
csv_bytes = b"id,label\n1,ok\n2,caf\xe9_bad_\xff\n"
with open(path, "wb") as f:
    f.write(csv_bytes)

df = spark.read.option("header", "true").csv(path)
df.show()
```

The file is:

```
id,label
1,ok
2,caf<0xE9>_bad_<0xFF>
```

## Expected (Spark 3.5.7)

The read succeeds. Spark's CSV reader decodes bytes using the configured charset (default UTF-8) with a lenient decoder that replaces malformed/undecodable input with the Unicode replacement character (U+FFFD) instead of raising. All rows are returned; the offending field simply contains replacement characters. The whole-file read does not fail on a single bad byte.

## Actual (Sail 4.2.0)

The entire read fails with:

```
SparkRuntimeException: Csv error: Encountered invalid UTF-8 data for line 3 and field 2
```

(Confirmed live on Sail 4.2.0 / pysail 0.7.0 with the exact bytes above.) No rows are returned — the byte-level UTF-8 validation rejects the whole file.

## Impact

Any CSV source containing legacy-encoded, Latin-1, or otherwise non-UTF-8 bytes is completely unreadable on Sail, whereas Spark 3.5 reads it (lossily). A single stray byte anywhere in the file fails the entire read, so this blocks migration of any pipeline whose upstream CSVs are not guaranteed clean UTF-8. Ideally Sail should match Spark's tolerant/lossy decode (replacement-char substitution), and/or honor an `encoding` / charset read option and a permissive parse mode.

## Scope note

Distinct from #2118 (CSV reader fails on gzipped files with quoted embedded commas — an *incorrect-number-of-fields* / field-parsing bug). This report is a **byte-level encoding-validation** divergence: the read fails on invalid UTF-8 bytes regardless of gzip, quoting, or field count. Different mechanism, different code path.

lonless9:
## Summary

Make UTF-8 CSV reads tolerant of malformed and incomplete input by replacing invalid units with `U+FFFD` using JDK-compatible replacement boundaries instead of failing the entire read.

The same decoding behavior is applied during schema inference and data scanning, after decompression and before CSV parsing, and is preserved across worker execution. This change covers Sail's currently supported UTF-8 CSV reads; configured `encoding` or `charset` handling, CSV writes, and row-level corrupt-record modes are unchanged.

Closes #2388

### DIFF
--- a/crates/sail-data-source/src/formats/csv/decoder.rs
+++ b/crates/sail-data-source/src/formats/csv/decoder.rs
@@ -1,0 +1,222 @@
+use std::io::{self, Read};
+
+use bytes::Bytes;
+use datafusion_common::Result;
+use futures::stream::{self, BoxStream};
+use futures::{StreamExt, TryStreamExt};
+
+const READ_BUFFER_SIZE: usize = 8 * 1024;
+const REPLACEMENT_CHARACTER: &[u8] = "\u{FFFD}".as_bytes();
+
+#[derive(Debug, Default)]
+struct LossyUtf8Decoder {
+    incomplete: Vec<u8>,
+}
+
+impl LossyUtf8Decoder {
+    fn decode_bytes(&mut self, input: Bytes) -> Bytes {
+        if self.incomplete.is_empty() && std::str::from_utf8(&input).is_ok() {
+            return input;
+        }
+        self.decode(&input)
+    }
+
+    fn decode(&mut self, input: &[u8]) -> Bytes {
+        let mut bytes = std::mem::take(&mut self.incomplete);
+        bytes.extend_from_slice(input);
+        self.repair(bytes, false)
+    }
+
+    fn finish(&mut self) -> Bytes {
+        let bytes = std::mem::take(&mut self.incomplete);
+        self.repair(bytes, true)
+    }
+
+    fn repair(&mut self, bytes: Vec<u8>, end_of_input: bool) -> Bytes {
+        let mut output = Vec::with_capacity(bytes.len());
+        let mut offset = 0;
+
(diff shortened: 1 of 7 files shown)


linhr (approved):
Looks good to me!

PR #2413, opened 2026-08-11
fix: preserve lambda bindings across projection pushdown

### LINKED ISSUE #2395: FILTER/TRANSFORM HOF lambda variable fails physical binding ("x" != "#N") when the HOF column survives to a self-join
## Summary

A higher-order-function column built with `FILTER(array(...), x -> ...)` fails at materialization on Sail 4.2.0 when that column survives to the left side of a self-join, with:

```
AnalysisException: Field of physical LambdaVariable with index 0 doesn't match batch field during evaluation
Field { "x": nullable Utf8 } != Field { "#47": nullable Utf8, metadata: {"SPARK::metadata::json": "{}"} }
```

Sail binds the lambda's bound variable by *name* (`x`) but the physical batch has re-numbered that field to an internal id (`#47`), so the physical `LambdaVariable` lookup fails. The plain expression evaluates fine on its own — the mismatch only appears once enough upstream operators (a join plus a handful of derived columns) have bumped the internal attribute ids, and the HOF column is then re-referenced through a self-join. Spark evaluates the identical plan without error.

## Environment

## Reproduction

```python

base = spark.createDataFrame([("h1", "T_A"), ("h2", "T_B")], ["entity_id", "col_1"])
aux = spark.createDataFrame([("h1", "x"), ("h2", "y")], ["entity_id", "seg"])
base = base.join(aux, "entity_id", "left")

# a handful of derived columns to bump the internal attribute ids
for i in range(4):
    base = base.withColumn(f"d{i}", F.expr(f"concat(entity_id, '_{i}')"))

# HOF column: FILTER over an array, lambda-bound variable named x
base = base.withColumn("arr", F.expr("FILTER(array(coalesce(col_1, NULL)), x -> x IS NOT NULL)"))

# explode the HOF column, then self-join the result back to the base
# (base still carries the FILTER column on its left side)
child = base.withColumn("t", F.explode("arr")).select("entity_id", "t") \
    .withColumnRenamed("entity_id", "eid2")
out = base.join(child, base.entity_id == child.eid2, "left").drop("eid2")

out.count()  # <-- AnalysisException here
```

## Expected vs actual

- **Expected (Spark):** the query materializes; `out.count()` returns the joined row count and the `arr` column holds the filtered array.
- **Actual (Sail 4.2.0):** materialization fails with `Field of physical LambdaVariable with index 0 doesn't match batch field ... Field { "x" } != Field { "#47" }`. The lambda variable is resolved by name against a batch whose field has been renumbered to an internal id.

Notes on the trigger boundary (from bisecting):
- With 0 derived columns, or with 3 or fewer, the same plan passes — the failure appears at ~4 derived columns once the internal ids are high enough.
- Removing the upstream join makes it pass regardless of derived-column count.
- Materializing the `FILTER` column alone (no self-join) passes; exploding it alone passes. The failure requires the HOF column to still be present on the left side of a join whose right side is derived by exploding that same column.

## Impact

Any pipeline that builds an array column via a HOF (`FILTER`/`TRANSFORM` with a lambda), then explodes it to join against a lookup and re-attaches the aggregated result to the original frame via a self-join, fails on Sail once the surrounding plan has moderate depth — a common ETL "explode → join → group → attach back" shape. Spark handles it transparently. The bound-variable resolution needs to track the physical field id rather than matching on the lambda variable's source name.

lonless9:
## Summary

Preserve physical lambda-variable bindings when DataFusion projection pushdown rewrites higher-order expressions across schema boundaries.

Temporary optimizer boundaries protect lambda-bearing projections from unsafe projection unification and protect lambda-bearing nested-loop join filters from join-filter extraction, while ordinary child projections remain eligible for optimization. All boundaries are removed before the optimized plan is returned. This addresses both higher-order self-join expressions and lambdas that capture columns across join sides.

Closes #2395


python/pysail/tests/spark/dataframe/test_higher_order_self_join.py
```diff
@@ -0,0 +1,68 @@
...
+
+    result = left.join(
+        right,
+        F.expr("exists(l.arr, x -> x IS NOT NULL) = (l.id > r.k)"),
+        "inner",
+    ).select("l.id", "r.k")
+
+    plan = result._explain_string()  # noqa: SLF001
+    assert "NestedLoopJoinExec" in plan
+    assert "join_proj_push_down" not in plan
```
linhr: Can we have a snapshot for this?

linhr (approved):
This workaround makes sense to me!

Would you like to create an issue with a minimum reproducible example (such as a unit test) so that the DataFusion community is aware of the root cause?

python/pysail/tests/spark/dataframe/test_higher_order_self_join.py
lonless9: Done!

lonless9:
@linhr Thanks for the review! I'll create an issue with an MRE upstream.

lonless9:
Upstream issue created https://github.com/apache/datafusion/issues/24372
cc @linhr

PR #2416, opened 2026-08-12
test: port HMS provider integration tests to Python

lonless9:
## Summary

- Port the HMS provider integration suite from ignored Rust tests to pytest.
- Switch integration tests from Spark 4.2.0 to 4.1.3.
- Fix compatibility issues exposed by the new catalog integration suite:
  - use Delta Lake 4.3.1 for Spark 4.1;
  - allow explicit Parquet schemas to select either Spark timestamp family, matching Spark behavior.


python/pysail/testing/containers/hms.py
```diff
@@ -0,0 +1,307 @@
+"""Hive Metastore container fixtures."""
```
linhr, line 1: As the next step we will also remove the similar setup in `python/pysail/tests/spark/catalog/hms/conftest.py`. This can be a follow-up.

crates/sail-python/src/catalog/mod.rs
linhr, line 1: Instead of `hms`, we should have `catalog/hms.rs` for the catalog provider and `catalog/status.rs` for status, so that we can add other catalog providers later with shared utilities.

crates/sail-data-source/src/formats/parquet/read.rs
```diff
@@ -343,3 +341,23 @@ fn parse_coerce_int96_string(setting: &str) -> Result<TimeUnit> {
         ))),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
linhr: Remove this test.

python/pysail/tests/hms/__init__.py
```diff
@@ -0,0 +1 @@
+"""Hive Metastore integration tests."""
```
linhr: Let's move this to `python/pysail/tests/catalog/hms/__init__.py`.

crates/sail-python/src/catalog/hms.rs
```diff
@@ -0,0 +1,397 @@
...
+use crate::hms::status::{PyDatabaseStatus, PyTableStatus};
+use crate::hms::to_py_error;
+
+type PyColumnDefinition = (String, String, bool);
+
+#[pyclass(name = "HmsCatalog")]
+pub(super) struct PyHmsCatalog {
+    provider: HmsCatalogProvider,
+    runtime: RuntimeHandle,
+}
```
linhr, line 22: Before porting tests for other catalog providers, let's think about how we can reduce code duplication.

If we can have a general `PyCatalogProvider` that wraps `Arc<dyn CatalogProvider>` so that we can reuse the wrapper for trait methods. And we can consider having e.g., `PyHmsCatalogProvider` (or maybe some factory) that wraps only the `new` method.

Or may be we keep the current `PyHmsCatalog` struct but implement a macro to define all trait methods from `CatalogProvider` and reuse it for other catalogs.

linhr (approved):
Exciting!

I just have a few thoughts for code organization and extensibility.

Also it seems the code coverage is pretty low for this PR. Do we know why?

lonless9:
@linhr Thanks for the review! I checked the update history of the code coverage report, and it seems that some uploader failures were recorded while others succeeded, so I'm inclined to think it's an issue with Codecov.

crates/sail-python/src/catalog/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    module.add(
+        "TableNotFoundError",
+        parent.py().get_type::<TableNotFoundError>(),
+    )?;
+    module.add(
+        "ViewNotFoundError",
+        parent.py().get_type::<ViewNotFoundError>(),
+    )?;
+    parent.add_submodule(&module)?;
+    Ok(())
```
linhr, line 32: The modules could be `_catalog` and `_catalog._hms`. But this can be done when we migrate tests for other catalog providers.

linhr (approved):
Nice!

PR #2421, opened 2026-08-12
feat: implement Iceberg copy-on-write row operations

lonless9:
## Summary

Connect Iceberg DELETE, UPDATE, and MERGE to the shared row-level plans and targeted copy-on-write commit path.

- use copy-on-write by default and honor explicit `write.delete.mode`, `write.update.mode`, and `write.merge.mode` properties
- expand DELETE, UPDATE, and MERGE into write-row and touched-file effects
- restrict rewrites to rows from affected data files while retaining inserts
- project row-level operation metadata out before writing replacement data files
- support conditionless DELETE and UPDATE
- support aliases and path-based targets
- preserve SQL false and unknown predicate results
- support updates that move rows between partitions
- avoid advancing the snapshot for no-op operations
- commit insert-only MERGE operations as fast appends
- retain existing delete manifests when COW follows a merge-on-read delete
- support Iceberg table format versions 1, 2, and 3

Existing merge-on-read DELETE and MERGE paths remain available. UPDATE with `write.update.mode=merge-on-read` remains explicitly unsupported; this PR adds UPDATE through copy-on-write.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/crates/sail-iceberg/src/lake_source.rs
+++ b/crates/sail-iceberg/src/lake_source.rs
@@ -16,11 +16,9 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use datafusion::arrow::datatypes::{Field as ArrowField, Schema as ArrowSchema};
 use datafusion::catalog::{Session, TableProvider};
-use datafusion::common::{
-    DataFusionError, Result, TableReference, ToDFSchema, not_impl_err, plan_err,
-};
+use datafusion::common::{DataFusionError, Result, not_impl_err, plan_err};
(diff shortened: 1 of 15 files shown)


lonless9:
I'm trying stacked PRs and finding it a bit tricky to handle sync when the stack gets too deep. I'll close this PR and continue in a new one.

PR #2429, opened 2026-08-12
fix: support three-argument dateadd

### LINKED ISSUE #2385: DATEADD(unit, n, date) parses the unit keyword as a column reference instead of a date-part literal
## Summary

`DATEADD(<unit>, <n>, <date_col>)` — the 3-argument T-SQL-style form, valid Spark SQL syntax — fails on Sail 4.2.0 because the date-part unit keyword (e.g. `week`) is parsed as a column/attribute reference rather than a literal, raising an `AnalysisException` instead of shifting the date. The 2-argument native form `DATE_ADD(date, days)` works correctly, so the parser path for the 3-arg unit form specifically is affected.

## Environment

## Reproduction

```python

df = spark.createDataFrame([("2026-02-01",)], ["d"]).withColumn("d", F.to_date("d"))

# FAILS — 3-arg unit form
df.selectExpr("DATEADD(week, -4, d) as x").show()
# pyspark.errors.exceptions.connect.AnalysisException:
#   attribute ObjectName([Identifier("week")]) is missing from the schema: cannot resolve attribute

# Also fails uppercase
df.selectExpr("DATEADD(WEEK, -4, d) as x").show()
# attribute ObjectName([Identifier("WEEK")]) is missing from the schema: cannot resolve attribute

# WORKS — native 2-arg day form
df.selectExpr("DATE_ADD(d, -28) as x").show()
# +----------+
# |         x|
# +----------+
# |2026-01-04|
# +----------+
```

## Isolation

| Call | Result |
|---|---|
| `DATE_ADD(d, -28)` (2-arg, days) | ✅ works |
| `DATEADD(week, -4, d)` (3-arg, lowercase unit) | ❌ `AnalysisException` |
| `DATEADD(WEEK, -4, d)` (3-arg, uppercase unit) | ❌ `AnalysisException` |

## Expected vs actual

- **Expected (Spark):** `DATEADD(week, -4, d)` shifts `d` back 4 weeks, same as `DATE_ADD(d, -28)`.
- **Actual (Sail 4.2.0):** the planner tries to resolve `week` as an attribute in the input schema and fails with `AnalysisException`, before ever reaching date arithmetic.

## Impact

Any workflow using the 3-argument `DATEADD(unit, n, col)` form — a common pattern for "N periods ago" comparisons in CASE expressions — fails outright on Sail, even though it's valid Spark SQL and the 2-argument day-granularity form works fine.

rohankumardubey:
## What changes

- Support three-argument `DATEADD(unit, quantity, value)` using existing `TIMESTAMPADD` semantics.
- Resolve date-part keywords such as `WEEK` correctly.
- Preserve the existing two-argument `DATEADD(date, days)` behavior.

Closes #2385

### DIFF
--- a/crates/sail-plan/src/formatter.rs
+++ b/crates/sail-plan/src/formatter.rs
@@ -515,8 +515,13 @@ impl PlanFormatter for SparkPlanFormatter {
                 name, arguments, 2, "12",
             )),
             "dateadd" => {
+                let name = if arguments.len() == 3 {
+                    "timestampadd"
+                } else {
+                    "date_add"
+                };
                 let arguments = arguments.join(", ");
-                Ok(format!("date_add({arguments})"))
+                Ok(format!("{name}({arguments})"))
             }
             "sum" => {
                 let mut args = arguments.join(", ");
--- a/crates/sail-plan/src/formatter.rs
+++ b/crates/sail-plan/src/formatter.rs
@@ -514,11 +514,11 @@ impl PlanFormatter for SparkPlanFormatter {
             "theta_union" => Ok(format_function_with_default_argument(
                 name, arguments, 2, "12",
             )),
-            "dateadd" => {
-                let name = if arguments.len() == 3 {
-                    "timestampadd"
-                } else {
-                    "date_add"
+            "date_add" | "dateadd" => {
+                let name = match arguments.len() {
+                    2 => "date_add",
+                    3 => "timestampadd",
+                    _ => name,
                 };
                 let arguments = arguments.join(", ");
                 Ok(format!("{name}({arguments})"))
--- a/crates/sail-plan/src/function/scalar/datetime.rs
+++ b/crates/sail-plan/src/function/scalar/datetime.rs
@@ -156,7 +156,7 @@ fn format_interval(interval: Expr, unit: &str) -> Expr {
     })
 }
 
-fn timestampadd_interval(unit: &str, quantity: Expr) -> PlanResult<Expr> {
+fn timestampadd_interval(function_name: &str, unit: &str, quantity: Expr) -> PlanResult<Expr> {
     let zero_i32 = || lit(0_i32);
     let zero_f64 = || lit(0_f64);
     let quantity_i32 = || cast(quantity.clone(), DataType::Int32);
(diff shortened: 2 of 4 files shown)


rohankumardubey:
@linhr , this is ready for review. I’ve added support for the three-argument `DATEADD` form while preserving the existing two-argument behavior. Thanks!

crates/sail-plan/src/resolver/expression/function.rs
```diff
@@ -453,6 +453,7 @@ impl PlanResolver<'_> {
         mut arguments: Vec<spec::Expr>,
     ) -> Vec<spec::Expr> {
         const DATE_PART_FUNCTIONS: &[&str] = &[
+            "dateadd",
```
shehabgamin, line 457: Following the pattern of the others, i think we can add `date_add` as well

crates/sail-plan/src/function/scalar/datetime.rs
```diff
@@ -1182,10 +1190,7 @@ pub(super) fn list_built_in_datetime_functions() -> Vec<(&'static str, ScalarFun
             F::custom(|input| interval_arithmetic(input, "days", Operator::Minus)),
         ),
         ("date_trunc", F::custom(date_trunc)),
-        (
-            "dateadd",
-            F::custom(|input| interval_arithmetic(input, "days", Operator::Plus)),
-        ),
+        ("dateadd", F::custom(dateadd)),
```
shehabgamin: There is also `date_add` in this same file

crates/sail-plan/src/formatter.rs
```diff
@@ -515,8 +515,13 @@ impl PlanFormatter for SparkPlanFormatter {
                 name, arguments, 2, "12",
             )),
             "dateadd" => {
+                let name = if arguments.len() == 3 {
+                    "timestampadd"
```
shehabgamin: Are we sure this is correct? Can you test with JVM spark?
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.date_add.html

shehabgamin (commented):
Largely looks good! Just a couple comments :)

crates/sail-plan/src/resolver/expression/function.rs
rohankumardubey: I’ve added `date_add` here as well. Thanks!

crates/sail-plan/src/function/scalar/datetime.rs
rohankumardubey: will Update `date_add` to use the same dispatch, so it now supports both the two- and three-argument forms.

crates/sail-plan/src/formatter.rs
rohankumardubey: Confirmed with JVM Spark doc. Both `dateadd` and `date_add` are formatted as `date_add()` with two arguments and `timestampadd()` with three arguments. I will  update the formatter to match this behavior.

shehabgamin (approved):
LGTM!

PR #2430, opened 2026-08-12
fix: improve sequence, sampling, and convert_tz parity

### LINKED ISSUE #2368: sequence(start, col) hangs indefinitely when the stop argument is a column (literal bounds work)
## Summary

`sequence(start, stop)` hangs indefinitely (never returns) when the `stop` argument is a **column**. The identical call with **literal** bounds returns instantly. On the same cluster in the same session, `range`, `createDataFrame`, and `sequence(1, 3)` (literals) all return in under a second — so the cluster is healthy and the hang is specific to a non-literal `sequence` bound.

## Environment

- Sail server 4.2.0 (pysail v0.7.0)
- pyspark 3.5.7 Spark Connect client
- Reproduced on a fresh 2-node `xs` cluster via Spark Connect, with fully synthetic data.

## Reproduction

```python

# WORKS — literal bounds, ~0.1s
spark.sql("SELECT sequence(1, 3) AS s").collect()          # -> [1, 2, 3]

# HANGS — column upper bound, never returns (killed at >120s) on a 3-row DataFrame
df = spark.createDataFrame([(1,), (3,), (12,)], ["n"])
df.withColumn("s", F.expr("sequence(1, n)")).collect()     # <-- hang
```

## Isolation

| Test | Result |
|---|---|
| `range(0, 1000).count()` | ✅ 0.7s |
| `createDataFrame([(1,), (2,)]).collect()` | ✅ 0.1s |
| `sequence(1, 3)` literal (SQL) | ✅ 0.1s → `[1, 2, 3]` |
| `sequence(1, n)` column bound, 3-row DF | ❌ **hang >120s** |

## Expected vs actual

- **Expected (Spark):** `[[1], [1, 2, 3], [1..12]]` returned in milliseconds.
- **Actual (Sail 4.2.0):** the query never returns.

## Impact

This blocks any event-expansion pattern that builds a per-row range from a count column and then `explode`s it (e.g. `F.expr("sequence(0, count_col - 1)")` followed by `explode`), which is a common way to disaggregate counts into individual rows.

shehabgamin:
Closes https://github.com/lakehq/sail/issues/2368

Future work needs to rewrite Sail's sampling implementation:
A scalar UDF is invoked once per Arrow `RecordBatch`, so `Random::invoke_with_args` and `RandPoisson::invoke_with_args` construct a fresh RNG for every batch and neither UDF receives the physical partition index. Spark's `SampleExec` owns one mutable sampler for the entire partition.


crates/sail-execution/src/driver/task_assigner/state.rs
```diff
@@ -81,7 +81,7 @@ impl DriverResource {
         let count = self.external_streams.len();
         self.external_streams
             .retain(|(stream_job_id, stream_stage)| {
-                *stream_job_id != job_id || !stage.is_none_or(|stage| stage == *stream_stage)
+                *stream_job_id != job_id || stage.is_some_and(|stage| stage != *stream_stage)
```
shehabgamin, line 84: Clippy made this change

PR #2433, opened 2026-08-13
feat: bring max_by/min_by to Spark 4.2 parity with top-k, sliding windows and clause checks

davidlghellin:
Aligns `max_by`/`min_by` with Spark 4.2 across type checking, ordering semantics, the top-k form, window frames and the clause surface. Every behaviour below was checked against the Spark source at `opt/spark` (v4.2.0) and measured on both engines.

## `max_by` / `min_by`

**Type checking and arity**
- The ordering argument must be orderable, mirroring `OrderUtils.isOrderable` arm by arm: MAP, VARIANT and calendar intervals are rejected at any nesting depth. The exhaustive match has no wildcard, so a new Arrow variant fails the build instead of being silently accepted.
- The check runs on `DataType` in `coerce_types` and again on the argument fields in `return_field`. Only the fields carry the metadata that identifies GEOMETRY/GEOGRAPHY and extension-only VARIANT.
- Arity is 2 or 3 (`MaxByBuilder`/`MinByBuilder`). Before, `max_by()` panicked and killed the RPC, and `max_by(x, y, k)` silently dropped `k`.

**Top-k form (`MaxMinByK`)**
- Returns `ARRAY<value>`, or NULL when no ordering is non-null, and honours DISTINCT.
- `k` is cast to INT only from numeric, string or NULL, must be foldable and within `[1, 100000]`. It is checked at analysis when it is a literal, so an empty grouped result still fails as in Spark, and again when the accumulator is built.
- The accumulator is bounded to `2k` rows. Result and partial-state lists are built with the declared element field, so value columns that carry field metadata (for example from `createDataFrame`) keep a matching schema.

**Ordering semantics**
- The window accumulator compares keys in Arrow row format with nulls first, instead of `ScalarValue::partial_cmp`. That gives nested NULLs as smallest, `-0.0 == 0.0` and NaN as largest, and a tie takes the newer row as in `If(old > new, old, new)`.
- This fixes a NULL struct field comparing `Equal` to anything, and a NULL list element ranking greatest (the Postgres rule).
- The aggregate path still rewrites to `last_value` (`max_by`) or `first_value` (`min_by`), both `ASC NULLS FIRST`, so a nested NULL ranks smallest.
- A key containing floats is wrapped in an internal `spark_ordering_key` UDF (`normalize_floats_for_ordering`), registered in the codec.
- A VOID ordering key is treated as NULL through `logical_nulls()`.

**Windows**
- Frames that do not start at `UNBOUNDED PRECEDING` get a retractable accumulator through `create_sliding_accumulator`: a sorted map plus an arrival queue. It also serves the top-k form.
- `FILTER (WHERE p) … OVER w` is rewritten to `max_by(x, CASE WHEN p THEN y END) OVER w`. A NULL ordering drops the row, and `datafusion-proto` does not serialize window filters, so passing the filter through would be lost in cluster mode.

**Clauses and errors**
- `IGNORE NULLS` and `WITHIN GROUP` are rejected (`INVALID_SQL_SYNTAX.FUNCTION_WITH_UNSUPPORTED_SYNTAX`).
- Rejections carry Spark's error classes: `INVALID_ORDERING_TYPE`, `VALUE_OUT_OF_RANGE`, `NON_FOLDABLE_INPUT`, `UNEXPECTED_INPUT_TYPE` and `WRONG_NUM_ARGS`.
- The result field keeps the value field's metadata, so GEOMETRY/GEOGRAPHY values stay typed.

## Shared changes these required
- A DISTINCT aggregate used as a window function is rejected for every function (`DISTINCT_WINDOW_FUNCTION_UNSUPPORTED`, `WindowResolution.checkWindowFunction`).
- Named arguments are rejected for built-ins without a `functionSignature` in Spark 4.2 (`NAMED_PARAMETERS_NOT_SUPPORTED`). The allowlist is `mask`, `count_min_sketch`, the explode family and `tuple_*`. No Sail built-in reads argument names today, and unknown functions still report as unknown.
- An unaliased aggregate with FILTER is named `<call> FILTER (WHERE <condition>)`, as in Spark.
- `array()` keeps the arguments' field metadata when all non-NULL arguments agree. `named_struct()` copies each value's field metadata, as `struct()` already did, through a `SparkNamedStruct` wrapper registered in the codec. Without this, a nested GEOMETRY was invisible to the orderability check.
- The gold data entries for `max_by(x, y, 2)` / `min_by(x, y, 2)` are back to `"success": "ok"`.

## Tests
- `max_by.feature` and `min_by.feature` mirror each other scenario for scenario. They cover orderability, arity, top-k, ordering semantics, running and sliding windows, GROUP BY, ROLLUP, CUBE, HAVING, UNION, RANGE, PIVOT, the clause surface, output schema, error classes and shuffled input (`REPARTITION(8)`).
- `test_max_min_by_dataframe.py` covers the DataFrame API: `groupBy().agg`, `rollup`, `rowsBetween`, top-k with `rangeBetween`, `pivot`, and default column names.
- `test_geometry_output.py`, `pivot.feature` and `array_coercion.feature` cover the geo and pivot surfaces touched above.

## Known gaps, recorded as `@sail-bug`
- A GEOMETRY ordering key produced by `CASE`: DataFusion builds the CASE field without metadata.
- Top-k inside PIVOT: on its general path Spark wraps `k` in `IF(...)`, so it is rejected as non-foldable, while Sail's FILTER-based pivot answers.
- `PIVOT … FOR col IN (…)` without parentheses fails in Sail's SQL parser.
- Collecting GEOMETRY/GEOGRAPHY to the client (`collect`/`toArrow`): Sail sends WKB `binary` where Spark sends `struct<srid, wkb>`.
- `array(geometry, binary)` is accepted, where Spark raises `DATA_DIFF_TYPES`.
- Year-month interval `*` and `/` typing, and `try_multiply` on day-time intervals, both unrelated to these functions.

## Verification
- `max_by`/`min_by`, pivot and DataFrame tests: Sail 340 passed / 7 xfailed; Spark 4.2 JVM 347 passed.
- The same selection passes with the 3.5.9, 4.0.x and 4.1.x clients.
- Local mode and `SAIL_MODE=local-cluster` agree on shuffled input; values match the JVM.
- The full `function/`, `dataframe/`, `analyzer/` and `execution/` Python suites pass with no regressions.


shehabgamin:
@davidlghellin Here is an initial review from Codex!

---
## Review comments

### 1. P1 — Field metadata bypasses orderability validation

Place on [max_min_by.rs:148](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/crates/sail-function/src/aggregate/max_min_by.rs#L148):

```text
[P1] `coerce_types` cannot enforce Spark’s logical orderability for types whose identity is carried by Arrow `Field` metadata. Supported Iceberg/Delta VARIANT columns may have only the outer `arrow.parquet.variant` extension (no Sail child marker), and GEOMETRY/GEOGRAPHY are plain Binary plus `geoarrow.wkb`; all three therefore pass this `DataType` check. Nested extension-only VARIANT also passes because the struct/list recursion discards the child `Field`. Spark 4.2’s `OrderUtils` rejects VariantType, GeometryType, and GeographyType before the AtomicType arm. Please add a recursive field-aware predicate and invoke it from `AggregateUDFImpl::return_field`, which receives the real argument fields.
```

Exact production fix:

```diff
diff --git a/crates/sail-common-datafusion/src/ordering.rs b/crates/sail-common-datafusion/src/ordering.rs
--- a/crates/sail-common-datafusion/src/ordering.rs
+++ b/crates/sail-common-datafusion/src/ordering.rs
@@
-use datafusion::arrow::datatypes::{DataType, IntervalUnit};
+use datafusion::arrow::datatypes::{DataType, Field, IntervalUnit};
+use sail_common::geoarrow::extension::GeoArrowWkbType;

-use crate::variant::is_marked_variant_storage_type;
+use crate::variant::{is_marked_variant_storage_type, is_variant_storage_field};
@@
-/// Everything atomic is orderable, `STRUCT` and the list family recurse, and anything else is
-/// not. Spark fails closed on the types it does not name (`case _ => false`); this port has no
-/// wildcard arm at all, so a new *Arrow* variant fails the build and has to be classified
-/// deliberately. That guarantee does not extend to a new *Spark* type carried inside an existing
-/// Arrow variant — such a type is silently treated as orderable, which is exactly the GEOMETRY
-/// case described below.
+/// `NULL` and atomic types other than `VARIANT`, `GEOMETRY`, and `GEOGRAPHY` are orderable;
+/// `STRUCT` and the list family recurse, and anything else is not. Spark fails closed on the
+/// types it does not name (`case _ => false`); this port has no wildcard arm at all, so a new
+/// *Arrow* variant fails the build and has to be classified deliberately. Logical types carried
+/// in Arrow field metadata are handled by [`is_orderable_field`].
@@
-/// `GEOMETRY` and `GEOGRAPHY` are unorderable in Spark too, but in Sail they are lowered to plain
-/// `Binary` and their identity lives in the field metadata, which `coerce_types` never receives.
-/// Catching them would mean moving the check to a hook that takes `FieldRef`s.
+/// `GEOMETRY` and `GEOGRAPHY` are lowered to `Binary`; [`is_orderable_field`] rejects their
+/// `geoarrow.wkb` extension before inspecting that storage type.
 pub fn is_orderable(data_type: &DataType) -> bool {
@@
-        DataType::Struct(fields) => fields.iter().all(|field| is_orderable(field.data_type())),
+        DataType::Struct(fields) => fields.iter().all(|field| is_orderable_field(field)),
         DataType::List(field)
         | DataType::LargeList(field)
         | DataType::FixedSizeList(field, _)
         | DataType::ListView(field)
         | DataType::LargeListView(field)
-        | DataType::RunEndEncoded(_, field) => is_orderable(field.data_type()),
+        | DataType::RunEndEncoded(_, field) => is_orderable_field(field),
         DataType::Dictionary(_, value_type) => is_orderable(value_type),
@@
     }
 }
+
+/// Spark orderability for an Arrow field, including logical extension metadata.
+pub fn is_orderable_field(field: &Field) -> bool {
+    if field.extension_type_name() == Some(GeoArrowWkbType::NAME)
+        || is_variant_storage_field(field)
+    {
+        false
+    } else {
+        is_orderable(field.data_type())
+    }
+}
diff --git a/crates/sail-function/src/aggregate/max_min_by.rs b/crates/sail-function/src/aggregate/max_min_by.rs
--- a/crates/sail-function/src/aggregate/max_min_by.rs
+++ b/crates/sail-function/src/aggregate/max_min_by.rs
@@
-use sail_common_datafusion::ordering::is_orderable;
+use sail_common_datafusion::ordering::{is_orderable, is_orderable_field};
@@
 fn get_min_max_by_result_type(
@@
     Ok(coerced)
 }
+
+fn validate_min_max_by_fields(
+    function_name: &str,
+    input_fields: &[FieldRef],
+) -> Result<(), DataFusionError> {
+    let (_, ordering_field) = max_min_by_args(function_name, input_fields)?;
+    if !is_orderable_field(ordering_field.as_ref()) {
+        return Err(generic_exec_err(
+            function_name,
+            &format!(
+                "does not support ordering on type {}",
+                ordering_field.data_type()
+            ),
+        ));
+    }
+    Ok(())
+}
@@
     fn return_type(&self, arg_types: &[DataType]) -> Result<DataType, DataFusionError> {
         let (value_type, _) = max_min_by_args(self.name(), arg_types)?;
         Ok(value_type.to_owned())
     }
+
+    fn return_field(&self, arg_fields: &[FieldRef]) -> Result<FieldRef, DataFusionError> {
+        validate_min_max_by_fields(self.name(), arg_fields)?;
+        datafusion::logical_expr::udaf_default_return_field(self, arg_fields)
+    }
@@
     fn return_type(&self, arg_types: &[DataType]) -> Result<DataType, DataFusionError> {
         let (value_type, _) = max_min_by_args(self.name(), arg_types)?;
         Ok(value_type.to_owned())
     }
+
+    fn return_field(&self, arg_fields: &[FieldRef]) -> Result<FieldRef, DataFusionError> {
+        validate_min_max_by_fields(self.name(), arg_fields)?;
+        datafusion::logical_expr::udaf_default_return_field(self, arg_fields)
+    }
```

Exact tests:

```diff
diff --git a/crates/sail-function/src/aggregate/max_min_by.rs b/crates/sail-function/src/aggregate/max_min_by.rs
--- a/crates/sail-function/src/aggregate/max_min_by.rs
+++ b/crates/sail-function/src/aggregate/max_min_by.rs
@@
 impl AggregateUDFImpl for MinByFunction {
@@
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use datafusion::arrow::datatypes::{DataType, Field, FieldRef};
+    use datafusion::logical_expr::AggregateUDFImpl;
+    use parquet_variant_compute::VariantType;
+    use sail_common::geoarrow::extension::{GeoArrowMetadata, GeoArrowWkbType};
+
+    use super::{MaxByFunction, MinByFunction};
+
+    fn extension_only_variant_field(name: &str) -> Field {
+        Field::new(
+            name,
+            DataType::Struct(
+                vec![
+                    Field::new("metadata", DataType::Binary, false),
+                    Field::new("value", DataType::Binary, false),
+                ]
+                .into(),
+            ),
+            true,
+        )
+        .with_extension_type(VariantType)
+    }
+
+    fn geo_field(name: &str) -> Field {
+        Field::new(name, DataType::Binary, true).with_extension_type(GeoArrowWkbType {
+            metadata: GeoArrowMetadata::default(),
+        })
+    }
+
+    #[test]
+    fn rejects_extension_metadata_ordering_fields() {
+        let value: FieldRef = Arc::new(Field::new("value", DataType::Int32, true));
+        let ordering_fields: [FieldRef; 3] = [
+            Arc::new(extension_only_variant_field("ordering")),
+            Arc::new(Field::new(
+                "ordering",
+                DataType::Struct(vec![extension_only_variant_field("nested")].into()),
+                true,
+            )),
+            Arc::new(geo_field("ordering")),
+        ];
+
+        for ordering in ordering_fields {
+            let fields = [Arc::clone(&value), ordering];
+            assert!(MaxByFunction::new().return_field(&fields).is_err());
+            assert!(MinByFunction::new().return_field(&fields).is_err());
+        }
+    }
+}
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    @sail-bug @spark-4.2
+    @spark-4.2
     Scenario: max_by rejects a GEOMETRY ordering column
diff --git a/python/pysail/tests/spark/function/features/aggregate/min_by.feature b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/min_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
@@
-    @sail-bug @spark-4.2
+    @spark-4.2
     Scenario: min_by rejects a GEOMETRY ordering column
```

### 2. P1 — Windowed ties still retain the first row

Place on [max_min_by.rs:233](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/crates/sail-function/src/aggregate/max_min_by.rs#L233):

```text
[P1] This only fixes constant ordering keys for aggregate expressions. `AggregateUDFImpl::simplify` is not applied when this UDAF is used as a window function, so the window form still runs `MaxMinByAccumulator`, whose `Ordering::Equal` arm declines to update. With a unique `ORDER BY i`, Spark’s running `max_by/min_by(x, 1)` window returns a,b,c while Sail returns a,a,a. Please update the shared accumulator on `Equal` and add a running-window constant-key case for both functions.
```

Exact production fix:

```diff
diff --git a/crates/sail-function/src/aggregate/max_min_by.rs b/crates/sail-function/src/aggregate/max_min_by.rs
--- a/crates/sail-function/src/aggregate/max_min_by.rs
+++ b/crates/sail-function/src/aggregate/max_min_by.rs
@@
                 match ordering_val.partial_cmp(&self.ordering) {
                     Some(Ordering::Greater) => self.is_max,
                     Some(Ordering::Less) => !self.is_max,
-                    _ => false,
+                    Some(Ordering::Equal) => true,
+                    None => false,
                 }
```

Exact tests:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
       Examples:
         | case            | ordering         |
         | literal         | 1                |
         | foldable sum    | 1 + 1            |
         | foldable concat | concat('a', 'b') |

+    # The unique window ordering makes tie replacement deterministic within every running frame.
+    Scenario: max_by keeps the current value on ties in a running window
+      When query
+        """
+        SELECT i,
+               max_by(x, 1) OVER (
+                 ORDER BY i ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+               ) AS result
+        FROM VALUES (1, 'a'), (2, 'b'), (3, 'c') AS t(i, x)
+        ORDER BY i
+        """
+      Then query result ordered
+        | i | result |
+        | 1 | a      |
+        | 2 | b      |
+        | 3 | c      |
+
     # NullType is orderable in Spark, which returns NULL here.
diff --git a/python/pysail/tests/spark/function/features/aggregate/min_by.feature b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/min_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
@@
       Examples:
         | case            | ordering         |
         | literal         | 1                |
         | foldable sum    | 1 + 1            |
         | foldable concat | concat('a', 'b') |

+    # The unique window ordering makes tie replacement deterministic within every running frame.
+    Scenario: min_by keeps the current value on ties in a running window
+      When query
+        """
+        SELECT i,
+               min_by(x, 1) OVER (
+                 ORDER BY i ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW
+               ) AS result
+        FROM VALUES (1, 'a'), (2, 'b'), (3, 'c') AS t(i, x)
+        ORDER BY i
+        """
+      Then query result ordered
+        | i | result |
+        | 1 | a      |
+        | 2 | b      |
+        | 3 | c      |
+
     # NullType is orderable in Spark, which returns NULL here.
```

### 3. P2 — The global tied-key tests assert a nondeterministic winner

Place on [max_by.feature:309](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L309):

```text
[P2] Spark chooses the newer buffer value in one update/merge step, but max_by/min_by are explicitly documented as nondeterministic when several input rows share the extremal ordering key because partial-state merge order is not guaranteed. This expected `c` therefore encodes a stronger global guarantee than Spark provides and can become flaky under different partitioning. Please assert only that the result is one of the associated values in both global constant-key outlines; the uniquely ordered running-window cases should cover directional tie replacement.
```

This is observable: Spark returned different tied winners when the same input was repartitioned to 1, 2, 4, 8, and 16 partitions.

Production fix: none.

Exact test fix:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    # A constant ordering key makes every row tie. Spark keeps the newer row on a tie
-    # (MaxByAndMinBy.scala: `If(old > new, old, new)` is false when they are equal), so the
-    # answer is the last one. Foldable arguments are constant-folded to the same literal.
+    # A constant ordering key makes every row tie. Spark does not specify which tied value is
+    # returned, so only assert that it is one of the associated values. Foldable arguments are
+    # constant-folded to the same literal.
@@
-        SELECT max_by(x, <ordering>) AS result
+        SELECT max_by(x, <ordering>) IN ('a', 'b', 'c') AS result
@@
-        | c      |
+        | true   |
diff --git a/python/pysail/tests/spark/function/features/aggregate/min_by.feature b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/min_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
@@
-    # A constant ordering key makes every row tie. Spark keeps the newer row on a tie
-    # (MaxByAndMinBy.scala: `If(old < new, old, new)` is false when they are equal), so the
-    # answer is the last one, the same as for max_by. Foldable arguments are constant-folded
-    # to the same literal.
+    # A constant ordering key makes every row tie. Spark does not specify which tied value is
+    # returned, so only assert that it is one of the associated values. Foldable arguments are
+    # constant-folded to the same literal.
@@
-        SELECT min_by(x, <ordering>) AS result
+        SELECT min_by(x, <ordering>) IN ('a', 'b', 'c') AS result
@@
-        | c      |
+        | true   |
```

### 4. P2 — VARIANT scenarios run against Spark 3.5

Place on [max_by.feature:80](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L80):

```text
[P2] This and the other new VARIANT examples call `parse_json`, which was introduced in Spark 4.0. Without `@spark-4` they fail when this suite is used as a Spark 3.5 oracle—the configured matrix includes 3.5.9. Please tag the standalone VARIANT scenarios and split VARIANT rows out of the mixed outlines, preserving the MAP coverage on 3.5; the same issue occurs in min_by.
```

Production fix: none.

Exact test fix:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    Scenario: max_by rejects a VARIANT ordering column
+    @spark-4
+    Scenario: max_by rejects a VARIANT ordering column
@@
         | STRUCT<MAP>        | struct(map('k', y))        |
         | ARRAY<ARRAY<MAP>>  | array(array(map('k', y)))  |
         | STRUCT<ARRAY<MAP>> | struct(array(map('k', y))) |
-        | ARRAY<VARIANT>     | array(parse_json(x))       |
-        | STRUCT<VARIANT>    | struct(parse_json(x))      |
+
+    @spark-4
+    Scenario Outline: max_by rejects a nested Spark 4 <case> ordering column
+      When query
+        """
+        SELECT max_by(x, <ordering>) AS result
+        FROM VALUES ('{"v":1}', 1), ('{"v":2}', 2) AS t(x, y)
+        """
+      Then query error (?s)max_by.*does not support ordering on type
+
+      Examples:
+        | case            | ordering              |
+        | ARRAY<VARIANT>  | array(parse_json(x))  |
+        | STRUCT<VARIANT> | struct(parse_json(x)) |
@@
       Examples:
         | case        | value                                      | result  |
         | MAP         | max_by(map('a', y), y)['a']                | 3       |
-        | VARIANT     | to_json(max_by(parse_json(j), y))          | {"v":3} |
         | ARRAY<MAP>  | max_by(array(map('k', y)), y)[0]['k']      | 3       |
         | STRUCT<MAP> | max_by(struct(map('k', y) AS m), y).m['k'] | 3       |
+
+    @spark-4
+    Scenario: max_by accepts a VARIANT value argument
+      When query
+        """
+        SELECT to_json(max_by(parse_json(j), y)) AS result
+        FROM VALUES ('{"v":1}', 1), ('{"v":3}', 3), ('{"v":2}', 2) AS t(j, y)
+        """
+      Then query result
+        | result  |
+        | {"v":3} |
diff --git a/python/pysail/tests/spark/function/features/aggregate/min_by.feature b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/min_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
@@
-    Scenario: min_by rejects a VARIANT ordering column
+    @spark-4
+    Scenario: min_by rejects a VARIANT ordering column
@@
       Examples:
-        | case            | ordering              |
-        | STRUCT<MAP>     | struct(map('k', y))   |
-        | STRUCT<VARIANT> | struct(parse_json(x)) |
+        | case        | ordering            |
+        | STRUCT<MAP> | struct(map('k', y)) |
+
+    @spark-4
+    Scenario: min_by rejects a nested STRUCT<VARIANT> ordering column
+      When query
+        """
+        SELECT min_by(x, struct(parse_json(x))) AS result
+        FROM VALUES ('{"v":1}', 1), ('{"v":2}', 2) AS t(x, y)
+        """
+      Then query error (?s)min_by.*does not support ordering on type
@@
       Examples:
         | case        | value                                      | result  |
         | MAP         | min_by(map('a', y), y)['a']                | 1       |
-        | VARIANT     | to_json(min_by(parse_json(j), y))          | {"v":1} |
         | ARRAY<MAP>  | min_by(array(map('k', y)), y)[0]['k']      | 1       |
         | STRUCT<MAP> | min_by(struct(map('k', y) AS m), y).m['k'] | 1       |
+
+    @spark-4
+    Scenario: min_by accepts a VARIANT value argument
+      When query
+        """
+        SELECT to_json(min_by(parse_json(j), y)) AS result
+        FROM VALUES ('{"v":3}', 3), ('{"v":1}', 1), ('{"v":2}', 2) AS t(j, y)
+        """
+      Then query result
+        | result  |
+        | {"v":1} |
```

### 5. P2 — IGNORE NULLS expects a Spark-4-only error class

Place on [max_by.feature:220](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L220):

```text
[P2] This exact error class is Spark-4-only. Spark 3.5 reports the same semantic rejection as `_LEGACY_ERROR_TEMP_1023` without `INVALID_SQL_SYNTAX`, while Spark 4.0 moved `functionWithUnsupportedSyntaxError` to `INVALID_SQL_SYNTAX.FUNCTION_WITH_UNSUPPORTED_SYNTAX`. Please add `@spark-4` so the regex is valid across the configured oracle matrix.
```

Production fix: none.

Exact test fix:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    @sail-bug
+    @sail-bug @spark-4
     Scenario: max_by rejects the clause IGNORE NULLS
```

### 6. P2 — Generic WITHIN GROUP is unavailable in Spark 3.5

Place on [max_by.feature:231](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L231):

```text
[P2] Generic `WITHIN GROUP` syntax was added in Spark 4.0 (SPARK-46009). Spark 3.5 parses it only for the dedicated percentile productions, so this exact max_by query fails earlier with `PARSE_SYNTAX_ERROR` rather than the asserted `INVALID_SQL_SYNTAX` message. Please gate this oracle with `@spark-4` for the configured 3.5.9/4.2.0 matrix.
```

Production fix: none.

Exact test fix:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    @sail-bug
+    @sail-bug @spark-4
     Scenario: max_by rejects the clause WITHIN GROUP
```

### 7. P2 — FILTER on a window aggregate is Spark 4.2-only

Place on [max_by.feature:251](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L251):

```text
[P2] FILTER on a window aggregate is only supported starting in Spark 4.2; Spark 3.5 and 4.1 reject this exact shape with `Window aggregate function with filter predicate is not supported yet`. Because this feature suite is used with the supported 3.5.9 and 4.2.0 oracles, please add `@spark-4.2`, as done for the other 4.2-only max_by surfaces.
```

Production fix: none.

Exact test fix:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
-    @sail-bug
+    @sail-bug @spark-4.2
     Scenario: max_by supports the clause FILTER combined with OVER
```

### 8. P3 — The marked-versus-shaped VARIANT distinction is untested

Place on [max_by.feature:142](https://github.com/lakehq/sail/blob/836ef34a1fdde6fab52a0831ef2b1b27d7b13c36/python/pysail/tests/spark/function/features/aggregate/max_by.feature#L142):

```text
[P3] Please pin the distinction implemented by this PR between a marked VARIANT storage struct and an ordinary, unmarked struct that merely has VARIANT’s physical field names and types. The implementation was corrected from `is_variant_storage_type` to `is_marked_variant_storage_type`, but these accepted tables only exercise generic structs, so that exact regression can return unnoticed. Spark 4.2 accepts an unmarked `named_struct(metadata BINARY, value BINARY)` as an ordering key; add it to both max_by and min_by tables.
```

Production fix: none; the PR-head implementation currently handles this correctly.

Exact test addition:

```diff
diff --git a/python/pysail/tests/spark/function/features/aggregate/max_by.feature b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/max_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/max_by.feature
@@
         | STRUCT<INT>            | struct(y)                                                      |
         | STRUCT<STRUCT>         | struct(struct(y))                                              |
+        | Unmarked VARIANT shape | named_struct('metadata', CAST(CAST(y AS STRING) AS BINARY), 'value', CAST(CAST(y AS STRING) AS BINARY)) |
         | ARRAY<INT>             | array(y)                                                       |
diff --git a/python/pysail/tests/spark/function/features/aggregate/min_by.feature b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
--- a/python/pysail/tests/spark/function/features/aggregate/min_by.feature
+++ b/python/pysail/tests/spark/function/features/aggregate/min_by.feature
@@
         | STRUCT<INT>            | struct(y)                                                      |
         | STRUCT<STRUCT>         | struct(struct(y))                                              |
+        | Unmarked VARIANT shape | named_struct('metadata', CAST(CAST(y AS STRING) AS BINARY), 'value', CAST(CAST(y AS STRING) AS BINARY)) |
         | ARRAY<INT>             | array(y)                                                       |
```

I found no additional correctness issues in the PR’s added year-month interval scenarios, `try_multiply` scenarios, or the changed gold snapshots. I also did not treat the explicitly unsupported three-argument top-k overload as an in-domain defect.

---

PR #2434, opened 2026-08-14
test: port Iceberg REST catalog tests to Python

lonless9:
https://github.com/lakehq/sail/pull/2416#discussion_r3780974432
https://github.com/lakehq/sail/pull/2416#discussion_r3781105687


python/pysail/tests/catalog/iceberg_rest/test_catalog.py
```diff
@@ -0,0 +1,543 @@
...
+    _assert_partition_transform(iceberg_rest_catalog, "partition_bucket", "bucket_table", ("id", "bucket(16)"))
+
+
+def test_create_table_partition_truncate(iceberg_rest_catalog: IcebergRestCatalogProvider) -> None:
+    _assert_partition_transform(
+        iceberg_rest_catalog,
+        "partition_truncate",
+        "truncate_table",
+        ("name", "truncate(10)"),
+    )
```
linhr, line 543: These tests can now be written as parameterized tests, which was not straightforward in Rust.

(This is just a note. We can do this in follow-up PRs when we port other tests.)

linhr (approved):
This is great! The use of `PyClassInitializer` seems a very nice idea!

PR #2436, opened 2026-08-14
refactor: declare output nullability for scalar functions Spark always marks nullable

davidlghellin:
## Why

In DataFusion the output `Field` of a scalar function comes from `return_field_from_args`, whose default hardcodes `nullable = true`. Sail inherited that default in ~80 scalar UDFs, so the nullability Sail reports was not decided by Sail at all — it was a library default that had never been checked against Spark.

## What

This migrates the bucket where Spark's rule is the **constant `true`**: 31 scalar functions. For each one:

- `return_field_from_args` becomes the single source of truth and carries Spark's rule with a permalink to the exact line in `apache/spark@v4.2.0`.
- `return_type` becomes an explicit `internal_err!` guard, so a caller that still reaches it is a bug rather than a silent second answer.
- The output type is moved **verbatim** into the hook, never rewritten — including where it is known to be wrong (`bround` on `Float32`), which stays for its own PR.

Five functions that already had the hook (`try_to_number`, `variant_get`, `date_trunc`, `from_json`, `from_xml`) were audited against Spark and given the same guard.

## How the bucket was closed

Not by grepping for a literal `override def nullable = true` — that finds 80 expressions and misses whole families. Spark also reaches `true` by inheritance, three ways:

1. an abstract base (`RoundBase`, `XPathExtract`, `DivModLike`, `UnaryMathExpression`);
2. `RuntimeReplaceable` with an `override lazy val replacement` (`Encode`, `AesEncrypt`);
3. `RuntimeReplaceable` with `replacement` as a **constructor parameter** — which is how the entire `try_*` family qualifies, via `TryEval.nullable = true` and, on the numeric branch, `evalMode == EvalMode.TRY` in `BinaryArithmetic.nullable`.

The transitive closure over `opt/spark` was used as a candidate generator only; every candidate was confirmed by reading the class. That killed two false positives: `LastDay` declares `nullIntolerant`, a different mechanism, and `UnixTimestamp` inherits a config-dependent rule from `ToTimestamp`.

## Behavior

**No observable schema changes.** DataFusion's default already answered `true`, so this
bucket pins the rule rather than fixing a value. That is the point: it is the zero-risk
half of the split, and it establishes the contract that the riskier buckets build on. A
wrong `false` is the direction that corrupts results — codegen drops the null check
entirely — so it gets its own PR with a per-function proof that the kernel cannot emit
NULL.

## Verification

- 44 queries measured on **both** engines (Spark JVM 4.2.0 via Connect, and Sail): every
  nullability flag and output type identical.
- The pre-existing `@function(nullability)` scenarios stay green — that is the end-to-end
  proof that nothing moved.
- ~55 Rust tests pin the contract per function, each on the discriminating input: a
  **non-nullable** argument, the case where the arity default would say `false`.

## Out of scope

- The derived bucket (`children.exists(_.nullable)`) and the constant-`false` bucket.
- Aggregate and window functions: different trait, different PR.
- `to_number`, whose rule is derived rather than constant, stays tagged `@sail-bug`.


shehabgamin (approved):
Looks great!

PR #2437, opened 2026-08-15
fix: support Python data source filter pushdown on PySpark 4.0

maxvwa:
## Problem

`DataSourceReader.pushFilters` was added in PySpark 4.1, but `InProcessExecutor` invokes it unconditionally whenever a scan carries filters. On PySpark 4.0 the base class has no such method, so **every filtered scan against every Python data source** fails at planning:

```
AttributeError: '<Reader>' object has no attribute 'pushFilters'
```

This affects the built-in `jdbc` and `vortex` sources as well as any user-defined one.

## Fix

Probe with `hasattr` and skip the push when the reader does not implement it.

This is safe because `supports_filters_pushdown` reports `TableProviderFilterPushDown::Inexact`, so DataFusion keeps the predicate in the plan and applies it above the scan. That is the same outcome as a 4.1 reader that declines to push anything — no rows are lost either way.

The probe only runs when there are filters to push, so unfiltered scans keep their existing FFI cost.

## Testing

Adds `test_python_filter_pushdown_reader_without_push_filters`, using a duck-typed reader with no `pushFilters` attribute to stand in for the 4.0 situation. It also covers readers that are not `DataSourceReader` subclasses.

Verified load-bearing: against the pre-fix binary the test reproduces the exact error above; after the fix it passes and the filter is still applied correctly (`id = 3` returns exactly one row).

Full `python/pysail/tests/spark/datasource/` suite: 201 passed, 8 skipped.

---

Split out of #2267, where this was found. It is independent of the Kafka work and can merge on its own.

### DIFF
--- a/crates/sail-data-source/src/formats/python/executor.rs
+++ b/crates/sail-data-source/src/formats/python/executor.rs
@@ -10,6 +10,7 @@ use arrow_schema::SchemaRef;
 use async_trait::async_trait;
 use datafusion_common::Result;
 use futures::stream::BoxStream;
+use pyo3::exceptions::PyAttributeError;
 use pyo3::prelude::*;
 use pyo3::types::PyAnyMethods;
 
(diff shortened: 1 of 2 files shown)


maxvwa:
Pushed a fix for the `Lint` failure — `N806`, a function-local variable in the new test that I had named as a constant. Also restructured the guard into a plain nested `if`/`else` rather than two conditions each re-testing `!filters.is_empty()`; same behaviour, easier to read.

On the Codecov note (75% patch, 1 line missing): the uncovered line is the `map_err(|e| ctx.wrap_py_error(e))` closure on the `hasattr` call. `hasattr` only propagates an error when an object's `__getattr__` raises something other than `AttributeError`, which no reader does, so that arm isn't reachable from a test without a deliberately pathological data source. The two lines that matter — the probe itself and the skip branch — are both covered by `test_python_filter_pushdown_reader_without_push_filters`.

I kept `?` propagation rather than `unwrap_or(false)` there: silently treating a genuine Python error as "no pushFilters" would hide a real failure, and the branch costs one uncovered line.

maxvwa:
Thanks — both Copilot comments were right, and both are addressed in the latest push.

**Error context (`executor.rs:304`).** The probe was wrapped with `ctx`, so a failure there would have been reported against `partitions` rather than `pushFilters`. The `pushFilters` context is now created before the probe and used for it.

**Callability (`executor.rs:316`).** Correct, and it reproduces: a reader with `pushFilters = None` passes `hasattr` and then fails at the call site with exactly the error this PR exists to prevent. I confirmed it by reverting just the callability check and running the new test:

```
SparkRuntimeException: Python error: [non_callable_push_filters_test::pushFilters]
TypeError: 'NoneType' object is not callable
```

The probe is now `getattr` + `is_callable()` rather than `hasattr`, which also lets the three cases be distinguished properly:

- attribute absent → skip the push (the PySpark 4.0 case)
- attribute present but not callable → skip the push, since calling it would fail anyway
- `getattr` raising anything other than `AttributeError` → propagate, with the `pushFilters` context

Added `test_python_filter_pushdown_reader_with_non_callable_push_filters` alongside the existing absent-attribute test; both assert the filter is still applied above the scan (`id = 3` returns exactly one row), so skipping the push never changes results.

shehabgamin (approved):
Looks great, thanks for the contribution!

PR #2438, opened 2026-08-16
fix: propagate cluster task panics

### LINKED ISSUE #2390: Improve cluster mode execution to return errors early on panic
Context:
https://github.com/lakehq/sail/issues/2368#issuecomment-5199849631

rohankumardubey:
This prevents cluster-mode queries from hanging when a task panics.

Panics during plan execution or while polling the task stream are now reported as failed task statuses, allowing the error to propagate back to the client through the existing retry and scheduling flow.

Added tests for panic reporting, panic message extraction, and task cancellation.

Closes #2390

### DIFF
--- a/crates/sail-execution/src/task_runner/actor/handler.rs
+++ b/crates/sail-execution/src/task_runner/actor/handler.rs
@@ -1,3 +1,4 @@
+use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::sync::Arc;
 
 use datafusion::arrow::datatypes::Schema;
@@ -28,7 +29,7 @@ use crate::stream::reader::TaskStreamSource;
 use crate::stream::writer::{TaskStreamChannelSink, TaskStreamSink};
 use crate::task::definition::{TaskDefinition, TaskInput, TaskOutput};
 use crate::task_runner::monitor::TaskMonitor;
-use crate::task_runner::{TaskRunnerActor, TaskRunnerMessage, TaskRunnerPlacement};
+use crate::task_runner::{TaskRunnerActor, TaskRunnerMessage, TaskRunnerPlacement, panic_message};
 use crate::worker::{WorkerLocation, WorkerMessage};
 
 impl TaskRunnerActor {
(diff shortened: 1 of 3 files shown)


linhr (commented):
Thanks for the contribution. I think the approach makes sense.

I'm a bit concerned about the usage of `AssertUnwindSafe` though, since it requires certain safety assumption and there is no way to detect violation of such assumption as the codebase evolves.

I did some research with Codex and here is an alternative that seems interesting, which wraps the task with a "supervisor" that detects panic upon task join. I haven't tried the code myself, but let me know if you'd be interested in exploring this direction.

```rust
ctx.spawn(async move {
    let monitor = AbortOnDropHandle::new(tokio::spawn(
        TaskMonitor::new(handle, key.clone(), stream, rx).run(),
    ));

    match monitor.await {
        Ok(()) | Err(error) if error.is_cancelled() => {}
        Err(error) if error.is_panic() => {
            // report panic as task failure
        }
        Err(error) => {
            // report unexpected error as task failure
        }
    }
});
```

* We need the `tokio-util` crate for `AbortOnDropHandle`.
* I feel we can ignore the panic for `self.execute_plan` for now. Handling it properly requires some refactoring to move `execute_plan` to a Tokio task without depending on `self` (if we don't use `catch_unwind`), and we can revisit it in the future.

Feel free to create a new PR if you'd like to play with this alternative approach and we can have this PR as a reference.

rohankumardubey:
Thanks @linhr for looking into this. I agree that the supervisor approach is safer than relying on `AssertUnwindSafe`, especially as the task runner evolves.

I’ll try this approach in a new PR: run `TaskMonitor` in a supervised Tokio task, handle panics through `JoinError`, and report them through the existing failed-task path. I’ll leave `execute_plan` panics out of scope for now and verify the change with the original local-cluster reproducer and focused tests.

I’ll keep this PR as a reference. Thanks for the suggestion!

linhr:
Superseded by #2480.

PR #2450, opened 2026-08-19
fix: correct DV scans and log replay planning for Delta Lake

lonless9:
## Summary

- Separate eager snapshot-file scans from metadata-as-data log replay using explicit scan source types.
- Replay file actions by logical-file identity `(path, deletionVector.uniqueId)` instead of path alone.
- Preserve physical column identities throughout column-mapped Parquet scans, statistics, partition values, and replay.

## Background

Delta `stats.numRecords` is the number of physical rows in the Parquet file. For a logical file with a deletion vector, the visible row count is:

```text
logical rows = stats.numRecords - deletionVector.cardinality
```
When stats.tightBounds = false, min/max values remain safe for data skipping but are not exact logical statistics. Wide nullCount values are meaningful only in the protocol-defined all-null and no-null cases.

Deletion vectors address absolute physical row positions. Delta therefore applies DV filtering using a stable file-local row index and prevents file splitting or predicate pushdown from changing those positions unless the Parquet metadata row index is used.

Delta log replay treats `(file path, deletion-vector unique ID)` as the identity of a logical file action. Multiple logical versions of the same Parquet file may coexist in the log with different DVs, so replaying by path alone can remove or retain the wrong Add.

Column mapping similarly requires Parquet fields, partition values, and column statistics to be resolved by physical names, while logical names are used only for the query-facing schema.


crates/sail-plan/src/function/aggregate.rs
```diff
@@ -461,6 +462,14 @@ fn count(input: AggFunctionInput) -> PlanResult<expr::Expr> {
     // For COUNT(DISTINCT *), the resolver already expanded the wildcard to column references
     // (with hidden-column filtering). For COUNT(*), convert to COUNT(1).
     let args = transform_count_star_wildcard_expr(arguments);
+    // Spark counts every row for a non-null literal. Use DataFusion's canonical COUNT(*)
+    // expansion so its aggregate-statistics rule recognizes all literal types consistently.
+    let args = match args.as_slice() {
```
shehabgamin, line 467: @lonless9 For consistency should we also apply this in `crates/sail-plan/src/function/window.rs`? I am not aware of any impact it would have to window functions, but it doesn't hurt to apply and there might be future relevant optimizations in DF.

lonless9: We can treat this case as the first step, and consider window functions later.

lonless9: Moved to https://github.com/lakehq/sail/pull/2461

PR #2457, opened 2026-08-19
fix: keep Python thread states alive on worker threads for UDF and Python data source execution

### LINKED ISSUE #2456: Python UDF execution destroys the PyThreadState after every call, breaking libraries with threading.local-tied native state (use-after-free in pyproj)
### Summary

Sail invokes Python UDFs on its tokio worker threads via per-call `Python::attach` (`PyGILState_Ensure`/`Release`). The final release **destroys the transient `PyThreadState`**, so every UDF call on a worker thread gets a fresh thread state. Libraries that keep native resources keyed by `threading.local` — whose lifetime follows the *thread state*, not the OS thread — get their cleanup callbacks run between calls while the OS thread lives on. pyproj is a concrete casualty: its per-thread `PJ_CONTEXT*` lives in OS-thread TSS, but cleanup is a `threading.local` destructor calling `proj_context_destroy`. The thread state dies after call 1, the context is freed, the TSS keeps the dangling pointer, and **the second UDF call on the same worker thread hands libproj a freed context → SIGSEGV.**

### Native stack (macOS, 3/3 runs identical)

```
thread 'tokio-rt-worker' EXC_BAD_ACCESS  (freed-memory address)
  libproj  getDBcontext(pj_ctx*)      <- dereferencing a freed PJ_CONTEXT
  libproj  proj_create
  pyproj   _crs.CRS.__init__
  ... CPython eval (UDF body) ...
  sail     PySparkGroupMapper::call -> BatchAggregateAccumulator::evaluate
  DataFusion GroupedHashAggregateStream::emit   (inline on a core tokio worker)
```

Deterministic signature: with N logical CPUs, the crash lands at exactly N completed groups — the first wave populates every worker thread; the first *reuse* dies.

### Minimal repro (~25 lines, pysail + pyproj 3.7 + pyspark 3.5)

```python
import os; os.environ["SPARK_CONNECT_MODE_ENABLED"] = "1"
import pandas as pd
from pyspark.sql.connect.session import SparkSession as ConnectSession
from pyspark.sql.types import StructType, StructField, IntegerType, StringType

server = SparkConnectServer(ip="127.0.0.1", port=0); server.start(background=True)
h, p = server.listening_address
spark = ConnectSession.builder.remote(f"sc://{h}:{p}").create()

schema = StructType([StructField("g", IntegerType()), StructField("crs", StringType())])

def body(pdf):
    import pyproj
    crs = pyproj.CRS("EPSG:32633")   # thread-local ctx created on 1st call per thread,
    return pd.DataFrame({"g": pdf["g"].iloc[:1], "crs": [crs.srs]})  # freed ctx on 2nd

GROUPS = 64  # > tokio worker count => guarantees second-wave reuse
pdf = pd.DataFrame({"g": [i for i in range(GROUPS) for _ in range(3)], "x": [1.0]*(GROUPS*3)})
df = spark.createDataFrame(pdf)
for wave in range(4):
    print(wave, df.groupby("g").applyInPandas(body, schema=schema).count())
print("SURVIVED")
```

Crashes within the second wave (validated on 0.5.3; the UDF invocation pattern is unchanged on current `main`). With `GROUPS` <= worker count it survives.

### Root cause chain

1. pyo3's `Python::attach` maps to `PyGILState_Ensure`/`Release`; the outermost release deletes the thread state (pyo3 0.26 `internal/state.rs`).
2. `threading.local` contents are deallocated with the thread state, not the OS thread.
3. pyproj (`_context.pyx`) frees its `PJ_CONTEXT` from a `threading.local` destructor while the pointer remains in OS-thread TSS. (Arguably also a pyproj bug — being reported there separately — but the pattern is common: any `threading.local` cache of native handles behaves this way.)
4. Diagnostic marker: a `threading.local` probe inside a UDF reads FRESH on every call on current Sail; REUSED with the fix below.

### Proposed fix (patch available)

Keep one persistent `PyThreadState` per worker thread: on a thread's first UDF call, `PyGILState_Ensure` once and retain that reference for the thread's lifetime (then `PyEval_SaveThread` so the GIL is not held). ~70 lines plus wiring at the UDF entry points (`pyspark_udf.rs`, `pyspark_group_map_udf.rs`, `pyspark_udaf.rs`, `pyspark_cogroup_map_udf.rs`, `pyspark_map_iter_udf.rs`, `pyspark_udtf.rs`). Precedent already in-tree: `PyMapStream`'s dedicated thread holds a single long-lived attach. Cost: one retained thread state per pool thread. Side benefit: `threading.local`-keyed caches in user libraries (pyproj contexts among them) stop being rebuilt on every call.

A/B on the patch (against 0.5.3): the repro above crashes 100% on unpatched code and survives 4x64 groups with the fix; a 150-group grouped-map workload with heavy pyproj/geopandas/shapely usage went from 3/3 crashes to clean completion.

Happy to open a PR porting the patch to `main`.

james-willis:
Fixes #2456.

## What

Per-call `Python::attach` (`PyGILState_Ensure`/`Release`) destroys the `PyThreadState` after every Python invocation on a worker thread. That runs `threading.local` teardown between calls while the OS thread lives on — freeing native resources that some libraries key by thread. The concrete casualty in #2456 is pyproj: its per-thread `PJ_CONTEXT*` lives in OS-thread TSS while cleanup hangs off a `threading.local` destructor, so the second call on the same worker thread hands libproj a freed context (SIGSEGV, deterministic repro in the issue).

This PR keeps one persistent thread state per worker thread: on a thread's first Python call, `PyGILState_Ensure` once and retain that reference for the thread's lifetime, then `PyEval_SaveThread` so the GIL is not held between calls. Subsequent `Python::attach` calls reuse the registered thread state and can never drop the gilstate count to zero.

- New `sail-python-udf/src/threadstate.rs` (`attach_persistent()`, ~70 lines).
- Wired at the UDF entry points: `pyspark_udf.rs`, `pyspark_group_map_udf.rs`, `pyspark_udaf.rs`, `pyspark_cogroup_map_udf.rs`, `pyspark_map_iter_udf.rs`, `pyspark_udtf.rs`.
- Also wired at the 18 Python attach sites in the Python data source format (`sail-data-source/src/formats/python/`) — six of these run inside `spawn_blocking`, i.e. on the reaped blocking pool where transient thread states churn hardest. The helper lives in a new minimal `sail-python-runtime` crate (pyo3-only dependency) so both `sail-python-udf` and `sail-data-source` can use it without depending on each other; happy to move it if you'd prefer a different home.
- `PyMapStream`'s dedicated thread is unchanged — it already holds a single long-lived attach, which is the precedent this generalizes.

## Why not just fix the libraries

Any library that caches native handles in `threading.local` has this hazard; pyproj is just the one with a deterministic crash. Persistent thread states also mean such caches survive across calls (pyproj otherwise rebuilds its PROJ context on every UDF invocation) and the per-call thread-state setup cost disappears.

## Cost

One retained `PyThreadState` per worker thread for the process lifetime.

## Validation

- The mechanism was A/B-validated on a 0.5.3-based build: the #2456 repro crashes 100% unpatched and survives patched; a `threading.local` marker probe inside a UDF flips from fresh-per-call to reused-per-thread.
- On this branch (`main`): `cargo check` and `cargo clippy` clean on `sail-python-udf` and `sail-data-source` under the workspace lint denies. The crate currently has no unit tests; happy to add the #2456 repro as a Python regression test if you'd like it in CI — and a wheel built from this branch reproduces/validates the fix in seconds with the issue's script.


shehabgamin:
@james-willis Nice find! And thank you for the fix.

Yeah adding a regression test would be nice if it's not too much trouble to do so

crates/sail-python-runtime/src/threadstate.rs
```diff
@@ -0,0 +1,72 @@
...
+fn pin_current_thread() {
+    THREAD_STATE_PINNED.with(|pinned| {
+        if pinned.get() {
+            return;
+        }
+        // SAFETY: the interpreter is guaranteed to be initialized here (this
+        // code runs while executing a query inside an embedded interpreter or
+        // one started by the host process), and we are not in interpreter
+        // finalization (query execution has the server alive).
+        unsafe {
```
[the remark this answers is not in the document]
james-willis: fixed

crates/sail-python-runtime/src/threadstate.rs
```diff
@@ -0,0 +1,72 @@
...
+                // `PyGILState_Ensure` calls find and reuse it, and their
+                // matching releases can never drop the counter to zero, so
+                // the thread state is never destroyed.
+                let tstate = ffi::PyEval_SaveThread();
+                let _ = tstate;
+            }
+            // If the GIL was already held (`PyGILState_LOCKED`), an outer
+            // scope owns the thread state; the unreleased `Ensure` reference
+            // taken above still pins it for this thread's lifetime, and we
+            // must not touch the GIL that the caller is holding.
```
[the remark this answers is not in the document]
james-willis: fixed. same issue basically as the other comment

crates/sail-python-runtime/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "sail-python-runtime"
```
linhr: Let's call this crate `sail-common-python`.

crates/sail-python-runtime/src/thread_state.rs
```diff
@@ -0,0 +1,143 @@
...
+/// process. Retaining the state is the safer trade.
+pub fn attach_persistent<F, R>(f: F) -> R
+where
+    F: for<'py> FnOnce(Python<'py>) -> R,
+{
+    Python::attach(|py| {
+        pin_thread_state(py);
+        f(py)
+    })
+}
```
linhr: Can we remove this wrapper and just call `pin_thread_state(py);` on each site? This makes the intention clear, and if we need other common initialization in `Python::attach` in the future, the logic is easier to composite instead of defining another wrapper.

linhr (commented):
Thanks for the fix!

I'll need some more time to review the safety of this approach, but here are some initial comments for consideration.

crates/sail-python-runtime/src/thread_state.rs
james-willis: sure

PR #2458, opened 2026-08-20
feat: add basic OPTIMIZE compaction

ErikBPF:
## Summary

- add `OPTIMIZE <named_delta_table>` SQL support for Delta tables
- compact active files independently per partition through the existing Delta writer
- emit `OPTIMIZE` commits with add/remove `dataChange=false` and existing file metrics
- skip the commit when every partition already has at most one active file

## Scope

Intentionally excludes path targets, `WHERE`, `ZORDER BY`, `FULL`, size-aware bin selection, and the Delta SQL metrics result row. Those remain separate follow-up slices.

## Tests

- `cargo test --workspace --all-targets`: 1092 passed, 60 ignored
- targeted Clippy with `-D warnings`: clean
- full Delta BDD suite: 200 passed, 59 snapshots passed
- OPTIMIZE BDD scenarios: compaction/data integrity, no-data-change actions and metrics, repeat no-op, partition isolation


ErikBPF:
Closing to recreate the reviewed and simplified draft under the ErikBPF account.

PR #2459, opened 2026-08-20
feat: add basic OPTIMIZE compaction

ErikBPF:
## Summary

- add `OPTIMIZE <named_delta_table>` SQL support for Delta tables
- compact active files independently per partition through the existing Delta writer
- emit `OPTIMIZE` commits with add/remove `dataChange=false` and existing file metrics
- skip commits when no same-partition files fit within the target output size

## Compatibility

The slice follows the [Delta Lake bin-packing contract](https://docs.delta.io/optimizations-oss/): compaction is partition-local, file actions do not change table data, and repeating compaction without new candidate files is a no-op.

The parser deliberately rejects `FULL`, `WHERE`, and `ZORDER BY` until those behaviors are implemented and tested rather than silently accepting partial semantics.

## Scope

Intentionally excludes path targets, `WHERE`, `ZORDER BY`, `FULL`, candidate-bin-only rewrites, and the Delta SQL metrics result row. Those remain separate follow-up slices.

## Review and simplification

- pin write planning to the input Delta snapshot so concurrent appends are not removed
- make repeat OPTIMIZE a no-op for healthy multi-file partitions
- request catalog data-write access for maintenance operations
- use the existing DataFusion plan visitor for snapshot discovery
- update the affected logical-plan golden snapshot
- retain whole-table rewrite as the deliberate minimal implementation

## Tests

- `cargo test --workspace --all-targets`: 1094 passed, 60 ignored
- targeted Clippy with `-D warnings`: clean
- OPTIMIZE BDD scenarios: 2 passed
- affected distributed Delta commit snapshot: passed
- parser rejects deferred `FULL`, `WHERE`, and `ZORDER BY` modifiers
- OPTIMIZE coverage: compaction/data integrity, no-data-change actions and metrics, repeat no-op, partition isolation


linhr:
@ErikBPF As a heads up, we won't be able to accept feature contributions around Delta Lake and Iceberg. We have turned down a few PRs from other contributors in the past already.

The internals for the lakehouse format are evolving rapidly, and it is often faster for us to generate the code ourselves instead of reviewing contributions.

I appreciate your continued interest in contributing. At the moment, the most valuable pieces are bug reports and bug fixes. It's fine to share prototypes for new features, but the PR will likely be closed.

Thanks for your understanding.

PR #2462, opened 2026-08-20
feat: free-threaded (no-GIL) Python support

james-willis:
> [!NOTE]
> **Draft / WIP.** Opened early so the free-threading CI job on this PR can demonstrate the Linux cp314t wheel build and the test run under a free-threaded interpreter. See the checklist at the bottom for what is intentionally not done yet.

## What this PR does

Adds first-class support for running Sail on free-threaded (no-GIL) CPython builds (3.13t/3.14t):

- **Declares `pysail._native` free-threading-compatible** via `#[pymodule(gil_used = false)]` in `crates/sail-python/src/lib.rs`. Without this declaration, importing `pysail` on a free-threaded interpreter re-enables the GIL for the whole process, so nothing else matters until it is in place.
- **Adds regression tests** (`python/pysail/tests/test_free_threading.py`) that skip themselves unless the interpreter is free-threaded with the GIL actually disabled:
  - `import pysail` alone keeps `sys._is_gil_enabled() == False` (checked in a subprocess with `PYTHON_GIL` unset, the only condition under which the `gil_used` declaration is observable);
  - a Python UDF observes a disabled GIL and executes on multiple threads across 8 partitions inside the in-process Spark Connect server;
  - UDF results are exact and stable across repeated concurrent runs;
  - several distinct UDFs evaluated concurrently do not corrupt Sail's cached Python state (the per-UDF `PyOnceLock` wrapper caches and the shared helper-module cache), whose first initialization is raced across worker threads.
- **Adds a CI job** (`free-threading-tests.yml`, wired into `build.yml`) that builds a versioned **cp314t** wheel on Linux — same recipe as `rust-build.yml`, plus `-i python3.14t`; pyo3 drops the `abi3-py38` feature automatically for a free-threaded target — and runs the tests through a dedicated `test-freethreaded` Hatch environment (standard `hatch run install-pysail` + `hatch run pytest` pattern) with `PYTHON_GIL=0` set by the environment. The environment pins `pandas>=3`, `pyarrow>=25`, and `pillow>=11.3`, deliberately bumped relative to the regular `test` environment: those are the minimum versions that publish cp314t wheels. This PR's own checks demonstrate that the wheel builds and the tests pass on Linux.
- **The CI scope covers the Python execution surface under free-threading**: `pysail.tests.spark.udf` (all UDF types), `pysail.tests.spark.datasource` (including the Python data source), and the dedicated free-threading tests — 225 tests passing on a free-threaded interpreter. Exclusions, all dependency-related rather than free-threading findings: four csv/parquet tests are deselected because their comparison helpers are incompatible with pandas 3 (they fail identically on the same interpreter with the GIL *enabled*; the free-threaded environment cannot use pandas 2, which has no cp314t wheels), and the vortex datasource test skips itself because `vortex-data` ships only abi3 wheels, which free-threaded interpreters cannot use. Integration-marked tests are deselected by default as usual.
- **Ships cp314t wheels in releases**: `release.yml` gains a `build-wheel-freethreaded` job that builds free-threaded wheels for Linux x86_64/aarch64 and macOS x86_64/aarch64 alongside the existing abi3 wheels (which are unchanged). The artifacts use the `package-wheel-*` prefix, so the existing test/review/publish artifact patterns pick them up, and the review/publish jobs wait on the new job so a release cannot ship without them. pip then selects the cp314t wheel on free-threaded 3.14 interpreters (which cannot use abi3 wheels) and the abi3 wheel everywhere else. Two deliberate deviations from the abi3 matrix: the Linux aarch64 leg uses the `2_28` manylinux image (the `2_24` image used for the abi3 aarch64 build predates free-threaded CPython and has no `python3.14t`), and Windows is excluded for now (see below). Note the release workflow is tag-triggered, so this job is not exercised by this PR's CI; its interpreter setup and maturin invocation mirror the free-threading test job that PR CI does validate, plus the abi3 release recipe (protoc, targets, release flags), and the same `-i python3.14t` release build has been run locally on macOS.

Verified locally on CPython 3.14.6 (free-threading build): the cp314t wheel builds, the bare-import check passes, and all four tests pass with genuinely concurrent UDF execution (multiple OS threads, `sys._is_gil_enabled() == False` inside the UDF).

## Why the declaration is sound: audit summary

On a GIL build, attaching to the interpreter grants mutual exclusion, so extension code can accidentally rely on the GIL as a mutex around shared state; on a free-threaded build it does not, and any such reliance is a data race. The crates that touch Python (`sail-python`, `sail-python-udf`, `sail-pyarrow`, and the Python data source support in `sail-data-source`) were audited for this, and **no GIL-as-mutex assumptions were found — nothing needed fixing**:

- All shared Python object caches already use `pyo3::sync::PyOnceLock` (free-threading-safe by construction): the embedded helper-module caches in `sail-python-udf/src/python/spark.rs` and `sail-data-source/src/formats/python/discovery.rs`, the per-UDF wrapper caches (`LazyPyObject` in `sail-python-udf/src/lazy.rs`), the PySpark version cache in the `sail-python-udf` cereal module, the type caches in `sail-pyarrow`, and the global runtime state (`sail-python/src/globals.rs`).
- The remaining shared state is lock-based: the Python data source registry is a `DashMap` with atomic entry insertion, and `PythonDataSource` caches use `once_cell`/`Mutex`.
- Every `Python::attach` call site operates on per-call locals (arguments in, converted values out); `#[pyclass]` mutation goes through pyo3's runtime borrow checking, which is atomic-based; accumulators mutate only via `&mut self` under DataFusion's exclusivity guarantee.
- The embedded Python helper classes (e.g. `PySparkBatchUdf` in `spark.py`) set all instance attributes in `__init__` and never mutate `self` in `__call__`, so one shared wrapper instance can be called from many threads.

So `gil_used = false` is an assertion of properties the code already has, and the new tests exist to keep it that way.

## Relationship to #2457

This PR is independent of #2457 (persistent Python thread states) and applies directly to `main`: the free-threading tests pass without it, since none of them relies on thread-local Python state surviving across UDF calls. The two changes compose: #2457 fixes per-thread native-resource lifetimes for libraries like pyproj (on GIL and free-threaded builds alike — the thread-state lifecycle semantics of the `PyGILState_*` API are the same on both), while this PR makes the interpreter-level concurrency mode work.

## grpcio and real deployments

`grpcio` does not yet declare free-threading support (grpc/grpc#38762), so importing the Spark Connect *client* re-enables the GIL unless the process is started with `PYTHON_GIL=0`. The tests and the CI job set `PYTHON_GIL=0` only because they colocate the client and the in-process server in a single interpreter.

The split deployment needs **no override at all**, and this has been verified end to end rather than just argued: a standalone Sail server on free-threaded `python3.14t` with `PYTHON_GIL` *unset* keeps the GIL genuinely off — the server's gRPC is implemented in Rust (tonic), so grpcio is never imported in the server process — while a client on an ordinary GIL interpreter connects over `sc://` with grpcio as usual, and the UDFs execute server-side across multiple threads with the GIL off.

One constraint to be aware of in that split: the client and the server must run the same Python **minor** version. Sail's UDF version check compares `major.minor`, and free-threading is a build flavor rather than a version — so a GIL 3.14 client pairs fine with a 3.14t server (both are "3.14"), while e.g. a 3.12 client against a 3.14t server is rejected.

## Out of scope

Whether the native libraries used *inside* a given UDF (numpy, pandas, etc.) are thread-safe under free-threading is a property of the user workload, not of Sail — Sail already invoked UDFs from multiple threads on GIL builds; free-threading widens the exposure. Interpreters without free-threading are unaffected: they keep using the abi3 wheel, and the free-threaded wheel is additive.

## Remaining work (why this is a draft)

- [ ] Broader concurrency coverage: pandas/Arrow UDFs, UDTFs, aggregate/map/co-group UDFs, and Python data sources under sustained multi-threaded stress (the current tests target scalar UDFs and the shared caches).
- [ ] Decide whether a free-threaded interpreter joins the standard test matrix (its dependencies are now managed by the dedicated `test-freethreaded` Hatch environment, with `pandas>=3` / `pyarrow>=25` pinned as the minimum versions publishing cp314t wheels).
- [ ] Windows cp314t wheels (the free-threaded Windows toolchain is unvalidated for this project and the release job cannot be exercised by PR CI, so it is excluded rather than shipped untested); decide whether cp313t wheels are also worth shipping.
- [ ] A release-time smoke test of the free-threaded wheels (the release `test-wheel` job runs on a GIL interpreter and installs the abi3 wheel; the free-threaded wheels are validated by PR CI instead).
- [ ] Performance numbers for multi-threaded UDF execution (GIL vs. free-threaded) worth publishing in the docs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,6 +141,13 @@ jobs:
       - setup
       - rust-build
 
+  free-threading-tests:
+    name: Free-Threading Tests
+    if: ${{ needs.setup.outputs.build_package == 'true' }}
+    uses: ./.github/workflows/free-threading-tests.yml
+    needs:
+      - setup
+
   catalog-tests:
     name: Catalog Integration Tests
     if: ${{ needs.setup.outputs.build_package == 'true' && needs.setup.outputs.run_catalog_tests == 'true' }}
--- a/.github/workflows/free-threading-tests.yml
+++ b/.github/workflows/free-threading-tests.yml
@@ -24,10 +24,8 @@ jobs:
           cache-key: rust-build-free-threaded
           install-testing-tools: false
 
-      # Build a versioned cp314t wheel. pyo3 automatically drops the
-      # `abi3-py38` feature when building for a free-threaded interpreter,
-      # so no feature changes are needed; the build recipe is otherwise
-      # identical to the abi3 build in `rust-build.yml`.
+      # pyo3 drops the abi3 feature for free-threaded targets, so this emits
+      # a cp314t wheel.
       - name: Build Python Package
         uses: PyO3/maturin-action@v1
         with:
@@ -39,25 +37,10 @@ jobs:
       - name: Install Package
         run: hatch -e test-freethreaded run install-pysail
 
-      # The `test-freethreaded` Hatch environment sets PYTHON_GIL=0, which
-      # keeps the GIL disabled even though the Spark Connect client imports
-      # grpcio: grpcio does not yet declare free-threading support
-      # (grpc/grpc#38762) and would otherwise re-enable the GIL. This is only
-      # needed because the tests colocate the client and the in-process
-      # server in a single interpreter; in real deployments the server's
-      # gRPC is implemented in Rust (tonic), so grpcio never enters the
-      # free-threaded server process and no override is needed.
-      # The scope covers the Python execution surface, where free-threading
-      # actually matters: all UDF types, the Python data source, and the
-      # dedicated free-threading tests.
-      #
-      # The deselected tests are pandas>=3 test incompatibilities, not
-      # free-threading failures: their comparison helpers sort
-      # `to_dict(orient="records")` output whose missing-value handling
-      # changed in pandas 3, and they fail identically on this interpreter
-      # with the GIL enabled. The free-threaded environment cannot use
-      # pandas 2 (no cp314t wheels), so they are excluded until the tests
-      # support pandas 3.
+      # The deselected tests fail because of pandas>=3 (which this
+      # environment must use; pandas 2 has no cp314t wheels), not because of
+      # free-threading: they fail identically with the GIL enabled.
+      # Re-include them when the tests support pandas 3.
       - name: Run Free-Threading Tests
         run: >-
           hatch -e test-freethreaded run pytest --pyargs
--- a/.github/workflows/free-threading-tests.yml
+++ b/.github/workflows/free-threading-tests.yml
@@ -33,26 +33,19 @@ jobs:
         with:
           args: -i python3.14t
 
-      - name: Install Package and Test Dependencies
-        run: |
-          python -m venv .venv-ft
-          .venv-ft/bin/pip install pysail --no-index -f target/wheels
-          # `testcontainers` is imported unconditionally by the test suite's
-          # conftest (no containers are started by the free-threading tests).
-          # `pandas` and `pyarrow` are intentionally unpinned here: the
-          # versions pinned for the regular test matrix do not all publish
-          # cp314t wheels yet.
-          .venv-ft/bin/pip install "pyspark-client==4.2.0" "pytest>=9.1,<10" "testcontainers>=4,<5" pandas pyarrow
+      - name: Install Hatch
+        uses: pypa/hatch@install
 
-      # PYTHON_GIL=0 keeps the GIL disabled even though the Spark Connect
-      # client imports grpcio, which does not yet declare free-threading
-      # support (grpc/grpc#38762) and would otherwise re-enable the GIL.
-      # This is only needed because the tests colocate the client and the
-      # in-process server in a single interpreter; in real deployments the
-      # server's gRPC is implemented in Rust (tonic), so grpcio never enters
-      # the free-threaded server process and no override is needed.
+      - name: Install Package
+        run: hatch -e test-freethreaded run install-pysail
+
+      # The `test-freethreaded` Hatch environment sets PYTHON_GIL=0, which
+      # keeps the GIL disabled even though the Spark Connect client imports
+      # grpcio: grpcio does not yet declare free-threading support
+      # (grpc/grpc#38762) and would otherwise re-enable the GIL. This is only
+      # needed because the tests colocate the client and the in-process
+      # server in a single interpreter; in real deployments the server's
+      # gRPC is implemented in Rust (tonic), so grpcio never enters the
+      # free-threaded server process and no override is needed.
       - name: Run Free-Threading Tests
-        env:
-          PYTHON_GIL: "0"
-        run: |
-          .venv-ft/bin/python -m pytest --pyargs pysail.tests.test_free_threading -v
+        run: hatch -e test-freethreaded run pytest --pyargs pysail.tests.test_free_threading -v
--- a/.github/workflows/free-threading-tests.yml
+++ b/.github/workflows/free-threading-tests.yml
@@ -47,5 +47,24 @@ jobs:
       # server in a single interpreter; in real deployments the server's
       # gRPC is implemented in Rust (tonic), so grpcio never enters the
       # free-threaded server process and no override is needed.
+      # The scope covers the Python execution surface, where free-threading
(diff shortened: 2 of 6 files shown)


james-willis:
Something interesting I found is that the current behavior on main makes freethreaded python runs with the GIL enabled perform much much worse that regular abi3 python versions. There's more threadstate contents in the no-gil runtime apparently:

> On a stock build that teardown is nearly free — shared pymalloc arenas, zero samples. On a free-threaded build every thread state owns per-thread mimalloc heaps, so teardown runs PyThreadState_Clear → mi_heap_collect_ex → segment reclaim.

So freethreaded runtimes benefit in gil and no-gil cases from this fix, and both out perform the abi3 runtime.

Python UDF throughput in rows/s, best of 5, Apple M4 Pro (8P+4E), CPython 3.14.6/3.14.6t.
`before` = sail 0.7.0 + PR #2462; `after` = the same tree with persistent UDF thread states.

**`intmix`** — 64-bit masked integer arithmetic

| parts | gil | ft-gil-on | nogil | nogil ÷ gil |
|---|---|---|---|---|
| 1 | 55,072 → 54,339 (-1%) | **44,516 → 60,769 (+37%)** | **44,040 → 62,547 (+42%)** | 0.80× → **1.15×** |
| 2 | 53,699 → 55,304 (+3%) | **36,315 → 61,963 (+71%)** | 98,242 → 121,571 (+24%) | 1.83× → **2.20×** |
| 4 | 53,835 → 54,120 (+1%) | **29,430 → 62,103 (+111%)** | **153,275 → 200,567 (+31%)** | 2.85× → **3.71×** |
| 8 | 52,621 → 53,374 (+1%) | **25,974 → 61,988 (+139%)** | 252,588 → 309,429 (+23%) | 4.80× → **5.80×** |
| 12 | 53,080 → 52,947 (-0%) | **21,663 → 61,027 (+182%)** | 304,377 → 338,554 (+11%) | 5.73× → **6.39×** |

**`strwork`** — slice/concat churn, allocation-heavy

| parts | gil | ft-gil-on | nogil | nogil ÷ gil |
|---|---|---|---|---|
| 1 | 42,208 → 43,243 (+2%) | **32,179 → 43,291 (+35%)** | 32,920 → 42,367 (+29%) | 0.78× → **0.98×** |
| 2 | 42,410 → 43,002 (+1%) | **29,358 → 42,210 (+44%)** | 68,170 → 83,436 (+22%) | 1.61× → **1.94×** |
| 4 | 42,256 → 41,527 (-2%) | **24,922 → 42,641 (+71%)** | 120,681 → 144,542 (+20%) | 2.86× → **3.48×** |
| 8 | 41,600 → 42,099 (+1%) | **21,293 → 42,234 (+98%)** | 178,382 → 205,930 (+15%) | 4.29× → **4.89×** |
| 12 | 40,813 → 40,751 (-0%) | **18,870 → 41,443 (+120%)** | 195,108 → 231,073 (+18%) | 4.78× → **5.67×** |


And the breakdown in where the time goes. Native stack samples, free-threaded + GIL on, 12 partitions, 10s window:

| frame | GIL build | before | after |
|---|---|---|---|
| `take_gil` | 325 | 199 | 394 |
| `_PyThreadState_Attach` | 3 | 185 | 302 |
| `PyThreadState_Clear` | 0 | 244 | **0** |
| `mi_heap_collect_ex` | 0 | 349 | **0** |

linhr:
Thanks for this change and the experiments. Supporting free-threaded Python would be a very nice thing to have.

I feel what we'll likely do is to accept the Rust code change (assuming that it does not impact the existing GIL-enabled Python versions) so that the user can build from source for `3.13t` and `3.14t` by themselves, but we won't have the bandwidth to maintain official multi-platform release for free-threaded wheels. Building and testing free-threaded Python on every commit would also double the CI time, so we cannot do that.

I'm going to put this PR as draft for now, but there is no need to remove code from this PR. The end-to-end code, test, and CI changes give us a good picture of the complexity involved, so these are good reference. We just need to figure out a path forward, possibly by cherry-picking the required changes and some CI setup to support ad-hoc testing.

james-willis:
In terms of support, in Sedona db we just tell users that support is experimental for freethreading and make no promise it’ll work well.

PR #2470, opened 2026-08-20
fix: type CASE and IF from every branch, not just the first

davidlghellin:
`CASE`/`IF` took their type from the first non-null `THEN` branch and ignored the `ELSE`,
so any conditional whose widest branch was not written first was under-typed.

Spark unifies the branches **before** typing the expression: `CaseWhen` and `If` are
`ComplexTypeMergingExpression`s whose `inputTypesForMerging` covers every `THEN` value
*and* the `ELSE` value, and `CaseWhenCoercion` / `IfCoercion` cast all of them to
`findWiderCommonType` before analysis completes. That is why the declared type never
depends on branch order in Spark.

```sql
SELECT typeof(CASE WHEN true THEN CAST(1 AS INT) ELSE CAST(2 AS BIGINT) END);
-- Spark: bigint    Sail before: int
```

The values were already right — DataFusion's `TypeCoercion` analyzer coerces the branches
at execution time — so what was wrong was the schema Spark Connect *announces*. That still
breaks real consumers: anything that builds from the declared schema fails on a value the
narrow type cannot hold.

```
CASE WHEN c THEN CAST(1 AS INT) ELSE CAST(3000000000 AS BIGINT) END
  df.collect()  ->  3000000000            (the data is wide)
  df.toArrow()  ->  ArrowInvalid: Integer value 3000000000 not in range
```

The fix unifies the branches in the resolver. That placement is load-bearing: `analyze_schema`
serves the client schema off `plan.schema()` of the **resolved** plan, before any analyzer rule
runs, so a `TypeCoercion` analyzer rule could not fix the reported type.

Because the fix now inserts real casts, it also has to reproduce two Spark rules that
DataFusion answers differently — without them the casts themselves would lose data:

- **`FractionalType` ⊕ `DecimalType` → `DOUBLE`** (`findWiderTypeForDecimal`). DataFusion
  answers `Decimal128(30,15)`, which holds only 15 integral digits and overflows on
  magnitudes a `DOUBLE` carries fine.
- **Past 38 digits, cut the SCALE, not the integral digits**
  (`DecimalType.boundedPreferIntegralDigits`). DataFusion clamps precision and scale
  independently, so a `DECIMAL(38,0)` branch would be cast to `decimal(38,20)`.

Both also apply to the element type of an `array`/`map`/`struct` branch, since Spark's
`findTypeForComplex` recurses `findWiderTypeForTwo` into the element.

### Tests

`case_branch_type_coercion.feature` — 30 scenarios covering branch order, the two decimal
rules, `IF`, the `CASE <operand> WHEN` form, NULL branches, nested containers, the reported
output schema, both ANSI dialects, and the pairs with no common type.

`test_conditional_arrow_schema.py` — 13 cases reading the result through `toArrow()`, which
builds from the declared schema and is the only assertion that catches the schema and the
data disagreeing. It lives in Python because there is no Arrow step in the BDD vocabulary.

Every expected type and value was captured from Spark Connect 4.2.0 (JVM), and the whole
file passes against it. TPC-H and TPC-DS are unchanged — **zero snapshots moved**.

### Known gaps, left out on purpose

- `float` ⊕ integral answers `float` where ANSI Spark answers `double`. The root cause is
  DataFusion's `numerical_coercion`, which answers the same for `coalesce`, `greatest`,
  `least`, `array`, `UNION ALL`, `+` and `*` — fixing it only here would make the engine
  inconsistent with itself. Tagged `@sail-bug`, with the ANSI-off counterpart pinned as a
  passing scenario since Sail is right in that dialect.
- `CaseWhen` nullability: Spark's `TrueLiteral` rule is not reproduced, so a `CASE` with an
  `ELSE` reports `nullable = true` where Spark reports `false`. `IF` already matches. Separate
  axis, untouched here.
- `coalesce`/`nvl`/`ifnull` are unified by the same Spark function and still carry the
  original bug.

### Relationship to #2225

Same class of bug on a different surface: #2225 fixes Spark coercion for the `+ - * /`
operators, this one for `CASE`/`IF`. Both land in the resolver rather than in an
analyzer rule, which is the placement agreed for #2225 — and for conditionals it is
not just a preference: `analyze_schema` reads `plan.schema()` off the resolved plan,
before any AnalyzerRule runs, so an analyzer rule cannot reach the reported type at all.

The two are independent and can land in either order. The `float` ⊕ integral gap listed
below is shared between them and belongs in whichever shared numeric-coercion fix comes
later.

### DIFF
--- a/crates/sail-plan/src/function/scalar/conditional.rs
+++ b/crates/sail-plan/src/function/scalar/conditional.rs
@@ -1,9 +1,11 @@
 use std::sync::Arc;
 
-use arrow::datatypes::{DataType, TimeUnit};
+use arrow::datatypes::{DataType, Field, FieldRef, Fields, TimeUnit};
 use datafusion::functions::expr_fn;
 use datafusion_common::ScalarValue;
+use datafusion_expr::type_coercion::binary::type_union_coercion;
 use datafusion_expr::{ExprSchemable, ScalarUDF, cast, expr, lit};
+use sail_common::spec::ARROW_DECIMAL128_MAX_PRECISION;
 use sail_common_datafusion::utils::items::ItemTaker;
 use sail_function::scalar::datetime::spark_date::SparkDate;
 use sail_function::scalar::datetime::spark_timestamp::SparkTimestamp;
@@ -33,7 +35,7 @@ fn case(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
             }
         }
     }
-    let branch_values = coerce_string_temporal_values(branch_values, &function_context)?;
+    let branch_values = coerce_conditional_branches(branch_values, &function_context)?;
     let when_then_expr = conditions
         .into_iter()
         .zip(branch_values)
@@ -53,7 +55,7 @@ fn if_expr(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     } = input;
     let (when_expr, then_expr, else_expr) = arguments.three()?;
     let (then_expr, else_expr) =
-        coerce_string_temporal_values(vec![then_expr, else_expr], &function_context)?.two()?;
+        coerce_conditional_branches(vec![then_expr, else_expr], &function_context)?.two()?;
     Ok(expr::Expr::Case(expr::Case {
         expr: None,
         when_then_expr: vec![(Box::new(when_expr), Box::new(then_expr))],
@@ -68,6 +70,244 @@ fn coalesce(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     } = input;
     let arguments = coerce_string_temporal_values(arguments, &function_context)?;
     Ok(expr_fn::coalesce(arguments))
+}
+
+/// Unifies the result branches of `CASE`/`IF`, in the order the two steps require.
+///
+/// String and temporal branches must be homogenised FIRST: DataFusion's `type_union_coercion`
+/// answers `Date32` for `(Utf8, Date32)`, the opposite of Spark's non-ANSI `stringPromotion`, so
+/// running the type fold first would type a string/date conditional as `date`.
+fn coerce_conditional_branches(
+    values: Vec<expr::Expr>,
+    function_context: &FunctionContextInput<'_>,
+) -> PlanResult<Vec<expr::Expr>> {
+    let values = coerce_string_temporal_values(values, function_context)?;
+    coerce_branch_values(values, function_context)
+}
+
+/// Casts the result branches of `CASE`/`IF` to their common type.
+///
+/// Spark's `CaseWhenCoercion` and `IfCoercion` unify every `THEN` value **and** the `ELSE` value
+/// before the expression is typed, and leave the branches untouched when the values have no common
+/// type. `Expr::Case::get_type` instead reports the type of the first non-null `THEN` branch, so
+/// without this the resolved plan carries a narrower type than Spark declares.
+fn coerce_branch_values(
+    values: Vec<expr::Expr>,
+    function_context: &FunctionContextInput<'_>,
+) -> PlanResult<Vec<expr::Expr>> {
+    let data_types = values
+        .iter()
+        .map(|value| value.get_type(function_context.schema))
+        .collect::<Result<Vec<_>, _>>()?;
+    let Some(common_type) = common_branch_type(&data_types) else {
+        return Ok(values);
+    };
+    Ok(values
+        .into_iter()
+        .map(|value| value.cast_to(&common_type, function_context.schema))
+        .collect::<Result<Vec<_>, _>>()?)
+}
+
+/// The type Spark gives a conditional whose branches have the types `data_types`.
+///
+/// Folds the branches left to right, as Spark's ANSI `findWiderCommonType` does
+/// (<https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/AnsiTypeCoercion.scala#L150-L156>). The non-ANSI dialect first partitions the string-typed
+/// branches and folds those ahead of the rest, because `findWiderTypeForTwo` is not associative
+/// for `StringType` (<https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala#L175-L186>); that partition is not reproduced here, and is
+/// currently masked by `coerce_string_temporal_values`, which already homogenises the string and
+/// temporal branches before this runs.
+///
+/// Returns `None` when the branches have no common type, so the caller leaves them untouched and
+/// the mismatch is reported later, the way Spark's `.getOrElse(c)` does.
+///
+/// `Decimal256` is decided for the whole slice at once. Spark decimals never exceed 38 digits, so
+/// a wider one describes a value Spark has no type for and the rules below would narrow it; Sail
+/// does accept such decimals (`CAST(1 AS DECIMAL(50, 10))`, which Spark rejects outright), so they
+/// reach here. Applying that exception per PAIR instead would let a fold alternate between two
+/// rule sets and make the answer depend on the order the branches are written in — the very thing
+/// this module exists to remove, and the check reaches inside containers for the same reason: the
+/// rules below recurse into element types, so a `Decimal256` element must switch them off too.
+fn common_branch_type(data_types: &[DataType]) -> Option<DataType> {
+    let (first, rest) = data_types.split_first()?;
+    if data_types.iter().any(contains_decimal256_type) {
+        return rest
+            .iter()
+            .try_fold(first.clone(), |left, right| {
+                type_union_coercion(&left, right)
+            });
+    }
+    rest.iter().try_fold(first.clone(), |left, right| {
+        branch_type_coercion(&left, right)
+    })
+}
+
+/// The wider of two branch types, following Spark where `type_union_coercion` does not.
+///
+/// Two arms of Spark's `findWiderTypeForDecimal` (<https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercionHelper.scala#L186-L198>) disagree
+/// with DataFusion, and both decide the type a conditional reports:
+///
+/// * a float against a decimal widens to `DOUBLE` (<https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercionHelper.scala#L194>), whereas DataFusion turns the float
+///   into `Decimal128(30, 15)` — which then overflows on values Spark represents perfectly well;
+/// * once the widened precision passes 38, Spark drops fractional digits to keep the integral
+///   ones (`DecimalType.boundedPreferIntegralDigits`, <https://github.com/apache/spark/blob/v4.2.0/sql/api/src/main/scala/org/apache/spark/sql/types/DecimalType.scala#L148-L158>), whereas
+///   DataFusion clamps precision and scale independently and so drops the integral digits.
+///
+/// Every other pair is left to `type_union_coercion`, and both rules also reach the element type
+/// of a container: Spark unifies `array`/`map`/`struct` branches by recursing `findWiderTypeForTwo`
+/// into the element (`findTypeForComplex`), whereas `type_union_coercion` recurses into itself and
+/// so would answer the element with DataFusion's rules.
+fn branch_type_coercion(left: &DataType, right: &DataType) -> Option<DataType> {
+    if (left.is_floating() && right.is_decimal()) || (left.is_decimal() && right.is_floating()) {
+        return Some(DataType::Float64);
+    }
+    if let Some(data_type) = wider_decimal_type(left, right) {
+        return Some(data_type);
+    }
+    let coerced = type_union_coercion(left, right)?;
+    Some(coerce_nested_types(left, right, coerced))
+}
+
+/// Replaces the element type DataFusion chose for a container with the one Spark's rules give.
+///
+/// The container shape — field names, `containsNull`, `valueContainsNull`, the fixed-size length —
+/// is left exactly as `type_union_coercion` decided it; only the element data type is substituted.
+/// Matching on the COERCED type rather than on the inputs matters: DataFusion legitimately answers
+/// a different container variant than either branch (`List` for two `FixedSizeList`s of different
+/// lengths, `LargeList` for a `List` against a `LargeList`), and requiring all three to agree would
+/// silently skip the substitution exactly there.
+fn coerce_nested_types(left: &DataType, right: &DataType, coerced: DataType) -> DataType {
+    if let (Some(left_field), Some(right_field), Some(coerced_field)) = (
+        container_element_field(left),
+        container_element_field(right),
+        container_element_field(&coerced),
+    ) {
+        let field = coerce_element_field(left_field, right_field, coerced_field);
+        return with_element_field(&coerced, field);
+    }
+    if let (DataType::Struct(left), DataType::Struct(right), DataType::Struct(coerced_fields)) =
+        (left, right, &coerced)
+    {
+        if let Some(fields) = coerce_struct_fields(left, right, coerced_fields) {
+            return DataType::Struct(fields);
+        }
+    }
+    coerced
+}
+
+/// The single field a container holds its elements in: the element for a list, the key-value
+/// entries struct for a map.
+fn container_element_field(data_type: &DataType) -> Option<&FieldRef> {
+    match data_type {
+        DataType::List(field)
+        | DataType::LargeList(field)
+        | DataType::FixedSizeList(field, _)
+        | DataType::Map(field, _) => Some(field),
+        _ => None,
+    }
+}
+
+fn with_element_field(data_type: &DataType, field: Field) -> DataType {
+    let field = Arc::new(field);
+    match data_type {
+        DataType::List(_) => DataType::List(field),
+        DataType::LargeList(_) => DataType::LargeList(field),
+        DataType::FixedSizeList(_, size) => DataType::FixedSizeList(field, *size),
+        DataType::Map(_, sorted) => DataType::Map(field, *sorted),
+        other => other.clone(),
+    }
+}
+
+fn coerce_element_field(left: &Field, right: &Field, coerced: &Field) -> Field {
+    match branch_type_coercion(left.data_type(), right.data_type()) {
+        Some(data_type) => coerced.clone().with_data_type(data_type),
+        None => coerced.clone(),
+    }
+}
+
+/// `None` when the branches do not name their fields the same way at the same positions.
+///
+/// DataFusion pairs struct fields by NAME and emits them in the left branch's order, so pairing
+/// them positionally here would coerce fields that are not the same column and attach the answer
+/// to a third field's name. Spark rejects such branches outright — `findTypeForComplex` requires
+/// `resolver(field1.name, field2.name)` at every position — so leaving DataFusion's answer alone
+/// is both safe and the closer of the two to Spark.
+///
+/// The comparison is case-insensitive because Spark's resolver is: under the default
+/// `spark.sql.caseSensitive = false`, `struct<A: …>` and `struct<a: …>` are the same field and
+/// Spark merges them. DataFusion's own name matching is case-sensitive, so it reaches those two
+/// through its positional path — which pairs them the way Spark does, and is therefore safe to
+/// build on.
+fn coerce_struct_fields(left: &Fields, right: &Fields, coerced: &Fields) -> Option<Fields> {
+    if left.len() != right.len() || left.len() != coerced.len() {
+        return None;
+    }
+    if left
+        .iter()
+        .zip(right.iter())
+        .zip(coerced.iter())
+        .any(|((left, right), coerced)| {
+            !left.name().eq_ignore_ascii_case(right.name())
+                || !left.name().eq_ignore_ascii_case(coerced.name())
+        })
+    {
+        return None;
+    }
+    Some(Fields::from(
+        left.iter()
+            .zip(right.iter())
+            .zip(coerced.iter())
+            .map(|((left, right), coerced)| coerce_element_field(left, right, coerced))
+            .collect::<Vec<_>>(),
+    ))
+}
+
+/// Spark's `widerDecimalType` (<https://github.com/apache/spark/blob/v4.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/DecimalPrecisionTypeCoercion.scala#L195-L199>), applied only when
+/// the result overflows 38 digits — below that DataFusion already agrees with Spark.
+fn wider_decimal_type(left: &DataType, right: &DataType) -> Option<DataType> {
+    let (left_precision, left_scale) = decimal_type_for(left)?;
+    let (right_precision, right_scale) = decimal_type_for(right)?;
+    let scale = left_scale.max(right_scale);
+    let range = (left_precision - left_scale).max(right_precision - right_scale);
+    let max_precision = i32::from(ARROW_DECIMAL128_MAX_PRECISION);
+    if range + scale <= max_precision {
+        return None;
+    }
+    // Spark keeps the integral digits and gives what is left to the scale, so the result is
+    // `38 - range`. `range` is at least 1 here, so this is in `[0, 37]`.
+    let scale = i8::try_from((max_precision - range).max(0)).unwrap_or(0);
+    Some(DataType::Decimal128(ARROW_DECIMAL128_MAX_PRECISION, scale))
+}
+
+/// The decimal Spark uses for a type: decimals as they are, integrals via `DecimalType.forType`.
+/// Anything else has no decimal form, and neither does `Decimal256` — `common_branch_type` keeps
+/// those out of these rules entirely, and answering `None` holds that line here too.
+fn decimal_type_for(data_type: &DataType) -> Option<(i32, i32)> {
+    match data_type {
+        DataType::Decimal32(precision, scale)
+        | DataType::Decimal64(precision, scale)
+        | DataType::Decimal128(precision, scale) => {
+            Some((i32::from(*precision), i32::from(*scale)))
+        }
+        DataType::Int8 | DataType::UInt8 => Some((3, 0)),
+        DataType::Int16 | DataType::UInt16 => Some((5, 0)),
+        DataType::Int32 | DataType::UInt32 => Some((10, 0)),
+        DataType::Int64 | DataType::UInt64 => Some((20, 0)),
+        _ => None,
+    }
+}
+
+fn contains_decimal256_type(data_type: &DataType) -> bool {
+    match data_type {
+        DataType::Decimal256(_, _) => true,
+        DataType::List(field)
+        | DataType::LargeList(field)
+        | DataType::FixedSizeList(field, _)
+        | DataType::Map(field, _) => contains_decimal256_type(field.data_type()),
+        DataType::Struct(fields) => fields
+            .iter()
+            .any(|field| contains_decimal256_type(field.data_type())),
+        _ => false,
+    }
 }
 
 fn coerce_string_temporal_values(
--- a/crates/sail-plan/src/function/scalar/conditional.rs
+++ b/crates/sail-plan/src/function/scalar/conditional.rs
@@ -151,10 +151,9 @@ fn coerce_nested_types(left: &DataType, right: &DataType, coerced: DataType) -> 
     }
     if let (DataType::Struct(left), DataType::Struct(right), DataType::Struct(coerced_fields)) =
         (left, right, &coerced)
+        && let Some(fields) = coerce_struct_fields(left, right, coerced_fields)
     {
-        if let Some(fields) = coerce_struct_fields(left, right, coerced_fields) {
-            return DataType::Struct(fields);
-        }
+        return DataType::Struct(fields);
     }
     coerced
 }
--- a/crates/sail-plan/src/function/scalar/conditional.rs
+++ b/crates/sail-plan/src/function/scalar/conditional.rs
@@ -72,11 +72,9 @@ fn coalesce(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
     Ok(expr_fn::coalesce(arguments))
 }
 
-/// Unifies the result branches of `CASE`/`IF`, in the order the two steps require.
-///
-/// String and temporal branches must be homogenised FIRST: DataFusion's `type_union_coercion`
-/// answers `Date32` for `(Utf8, Date32)`, the opposite of Spark's non-ANSI `stringPromotion`, so
-/// running the type fold first would type a string/date conditional as `date`.
+/// Unifies the result branches of `CASE`/`IF`. The order matters: `type_union_coercion` answers
+/// `Date32` for `(Utf8, Date32)`, the opposite of Spark's non-ANSI `stringPromotion`, so the
+/// string/temporal pass has to run first.
 fn coerce_conditional_branches(
     values: Vec<expr::Expr>,
     function_context: &FunctionContextInput<'_>,
@@ -85,12 +83,9 @@ fn coerce_conditional_branches(
     coerce_branch_values(values, function_context)
 }
 
-/// Casts the result branches of `CASE`/`IF` to their common type.
-///
-/// Spark's `CaseWhenCoercion` and `IfCoercion` unify every `THEN` value **and** the `ELSE` value
-/// before the expression is typed, and leave the branches untouched when the values have no common
-/// type. `Expr::Case::get_type` instead reports the type of the first non-null `THEN` branch, so
-/// without this the resolved plan carries a narrower type than Spark declares.
+/// Casts the result branches to their common type, the way `CaseWhenCoercion` / `IfCoercion` do
+/// (`TypeCoercionHelper.scala:498-538`). Without this the plan takes `Expr::Case::get_type`, which
+/// reports the first non-null `THEN` branch and ignores the `ELSE`.
 fn coerce_branch_values(
     values: Vec<expr::Expr>,
     function_context: &FunctionContextInput<'_>,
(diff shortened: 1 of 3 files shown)


shehabgamin:
@davidlghellin Here's an initial review from Codex!

---
## Review comments

### 1. ANSI coercion dialect is ignored

Place on [conditional.rs line 97](https://github.com/lakehq/sail/blob/a5993abbabed68d171c463e04bf3a6f4584b9c61/crates/sail-plan/src/function/scalar/conditional.rs#L97).

Exact comment:

> **[P1] Select the common type using the active Spark coercion dialect.**
>
> `common_branch_type` never receives `ansi_mode`, so every conditional follows DataFusion’s union lattice. That differs from Spark ANSI coercion: `if(false, CAST(1 AS INT), '2')` becomes `string` here instead of `bigint`, while `INT` plus `FLOAT` becomes `float` and rounds values such as `2147483647` instead of widening to `double`. Spark implements these rules in `AnsiTypeCoercion.scala:113-147` and `AnsiStringPromotionTypeCoercion.scala:92-107`. Please thread `PlanConfig` through scalar and recursive branch coercion, implement ANSI string promotion and integral/Float32 widening, and enable the currently `@sail-bug` ANSI tests.

Observed behavior:

| Query | PR | Spark |
|---|---|---|
| ANSI `INT` / `STRING` | `string`, `"2"` | `bigint`, `2` |
| ANSI `INT_MAX` / `FLOAT` | `float`, `2147483648` | `double`, `2147483647` |

### 2. DataFusion’s fallback is not Spark’s atomic/datetime lattice

Place on [conditional.rs line 136](https://github.com/lakehq/sail/blob/a5993abbabed68d171c463e04bf3a6f4584b9c61/crates/sail-plan/src/function/scalar/conditional.rs#L136).

Exact comment:

> **[P1] Do not use unrestricted DataFusion union coercion as Spark’s fallback.**
>
> This returns types for pairs Spark rejects and misses Spark datetime widening. On this head, `DATE` plus `INT` becomes `DATE` and the integer is interpreted as days since the epoch, while Spark rejects that pair with ANSI both enabled and disabled. Conversely, DataFusion chooses an unzoned nanosecond timestamp for `TIMESTAMP_LTZ` plus `DATE`; Sail currently fails while translating that schema, whereas Spark returns `TimestampType` and preserves LTZ semantics. `TypeCoercionHelper.findWiderDateTimeType` handles DATE/LTZ/NTZ explicitly and returns no type for unrelated atomic families. Please implement that allowlisted Spark lattice before any DataFusion helper and turn `None` into an analysis error so DataFusion cannot perform a second permissive coercion pass.

Observed behavior:

| Query | PR | Spark |
|---|---|---|
| `DATE` / `INT` | succeeds; `2` becomes `1970-01-03` | `DATA_DIFF_TYPES` |
| `TIMESTAMP_LTZ` / `DATE` | unzoned-nanosecond schema failure | `timestamp` |
| `TIMESTAMP_NTZ` / `DATE` | wrong unit/flavor path | `timestamp_ntz` |
| `TIMESTAMP_LTZ` / `TIMESTAMP_NTZ` | `timestamp` | `timestamp` |

### 3. Nested coercion omits Spark nullability and map-key safety

Place on [conditional.rs line 149](https://github.com/lakehq/sail/blob/a5993abbabed68d171c463e04bf3a6f4584b9c61/crates/sail-plan/src/function/scalar/conditional.rs#L149).

Exact comment:

> **[P1] Recurse with Spark’s complex-type rules instead of patching DataFusion’s chosen container.**
>
> `coerce_string_temporal_values` only sees top-level types, and this applies the Spark-specific overrides after DataFusion has already selected the nested type. With ANSI disabled, `if(false, array('2020-01-01'), array(DATE '2020-01-01'))` is therefore `array<date>` instead of Spark’s `array<string>`. The rebuilt field also preserves DataFusion’s nullability, while Spark ORs the source nullability with `Cast.forceNullable`; and map keys must be rejected whenever either inserted cast can create nulls. The current code accepts ANSI `map<string,...>`/`map<date,...>` keys even though Spark rejects them. Please implement the Array/Map/Struct recursion from `findTypeForComplex`, including non-ANSI string-first folding, `Cast.forceNullable`, and the map-key guard.

Confirmed examples:

| Case | PR | Spark |
|---|---|---|
| ANSI off, `array<string>` / `array<date>` | `array<date>` | `array<string>` |
| ANSI string→date array element | `containsNull=false` | `containsNull=true` |
| ANSI string/date map keys | accepted | `DATA_DIFF_TYPES` |
| ANSI string→date map value | misses forced nullability | `valueContainsNull=true` |

### 4. Struct resolution uses neither Spark’s configured resolver nor positional casting

Place on [conditional.rs line 203](https://github.com/lakehq/sail/blob/a5993abbabed68d171c463e04bf3a6f4584b9c61/crates/sail-plan/src/function/scalar/conditional.rs#L203).

Exact comment:

> **[P1] Use the session resolver and cast accepted structs positionally.**
>
> Spark zips struct fields by position and applies `spark.sql.caseSensitive` to the name at each position. This hard-codes case-insensitive matching, but line 102 subsequently sends the branch through DataFusion’s case-sensitive, by-name struct cast. Consequently, default-case-insensitive `STRUCT<A:int>`/`STRUCT<a:bigint>` is rejected even though Spark returns `struct<A:bigint>`. Conversely, when this helper returns `None`, `coerce_nested_types` still returns DataFusion’s by-name result, so reordered `a,b`/`b,a` fields are accepted although Spark rejects them. Please propagate incompatibility, use `PlanConfig.case_sensitive`, preserve the left field names, and apply Sail’s existing `SparkStructRename` positional adapter before casting.

Observed behavior:

| Struct branches | PR | Spark |
|---|---|---|
| `A` / `a`, case-insensitive | cast failure | `struct<A:bigint>` |
| `A` / `a`, case-sensitive | failure | failure |
| `a,b` / `b,a` | accepted as `{4, 3}` | `DATA_DIFF_TYPES` |

## Spark parity verification

I confirmed every finding against the vendored Spark implementation under `opt/spark`, including:

- [AnsiTypeCoercion.scala](/Users/r/Desktop/lakesail/sail/opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/AnsiTypeCoercion.scala:101)
- [AnsiStringPromotionTypeCoercion.scala](/Users/r/Desktop/lakesail/sail/opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/AnsiStringPromotionTypeCoercion.scala:92)
- [TypeCoercion.scala](/Users/r/Desktop/lakesail/sail/opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercion.scala:76)
- [TypeCoercionHelper.scala](/Users/r/Desktop/lakesail/sail/opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/TypeCoercionHelper.scala:137)
- [Cast.scala](/Users/r/Desktop/lakesail/sail/opt/spark/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala:421)

I also ran the affected queries against the bundled Spark runtime and the exact PR head, not merely against source inspection. The PR’s Python tests are collected and wired correctly; the gaps survive because two are explicitly marked `@sail-bug`, one is encoded as `@sail-only`, and the remaining type combinations are absent.

I excluded `spark.sql.legacy.decimal.retainFractionDigitsOnTruncate` because Sail does not currently expose that setting through `PlanConfig`, placing it outside the requested supported-capability scope.

## Exact proposed diff

This is the combined patch I recommend. It is shown first as requested and was not applied.

```diff
diff --git a/crates/sail-plan/src/resolver/expression/mod.rs b/crates/sail-plan/src/resolver/expression/mod.rs
--- a/crates/sail-plan/src/resolver/expression/mod.rs
+++ b/crates/sail-plan/src/resolver/expression/mod.rs
@@ -11,7 +11,7 @@ use crate::resolver::PlanResolver;
 use crate::resolver::state::PlanResolverState;

 mod attribute;
-mod cast;
+pub(super) mod cast;
 mod function;
 mod grouping;
 mod lambda;
diff --git a/crates/sail-plan/src/resolver/expression/cast.rs b/crates/sail-plan/src/resolver/expression/cast.rs
--- a/crates/sail-plan/src/resolver/expression/cast.rs
+++ b/crates/sail-plan/src/resolver/expression/cast.rs
@@ -220,7 +220,7 @@ impl PlanResolver<'_> {
 /// Returns true if the cast from `from` to `to` involves a Struct
 /// (possibly nested in a List/LargeList/FixedSizeList/Map) whose field names
 /// don't share enough overlap for DataFusion's struct cast validator.
-fn needs_struct_field_rename(from: &DataType, to: &DataType) -> bool {
+pub(crate) fn needs_struct_field_rename(from: &DataType, to: &DataType) -> bool {
     match (from, to) {
         (DataType::Struct(a), DataType::Struct(b)) => {
             a.len() == b.len()
@@ -245,7 +245,7 @@ fn needs_struct_field_rename(from: &DataType, to: &DataType) -> bool {
 /// Build a target type that has the names from `to` but the data types from
 /// `from`. The result is what `SparkStructRename` produces; the subsequent
 /// regular CAST then handles any leaf-type conversion.
-fn build_rename_target_type(from: &DataType, to: &DataType) -> DataType {
+pub(crate) fn build_rename_target_type(from: &DataType, to: &DataType) -> DataType {
     match (from, to) {
         (DataType::Struct(src_fields), DataType::Struct(tgt_fields))
             if src_fields.len() == tgt_fields.len() =>
diff --git a/crates/sail-plan/src/resolver/mod.rs b/crates/sail-plan/src/resolver/mod.rs
--- a/crates/sail-plan/src/resolver/mod.rs
+++ b/crates/sail-plan/src/resolver/mod.rs
@@ -17,6 +17,9 @@ mod schema;
 mod state;
 mod tree;

+pub(crate) use expression::cast::{
+    build_rename_target_type, needs_struct_field_rename,
+};
+
 pub struct PlanResolver<'a> {
     ctx: &'a SessionContext,
     config: Arc<PlanConfig>,
diff --git a/crates/sail-plan/src/function/scalar/conditional.rs b/crates/sail-plan/src/function/scalar/conditional.rs
--- a/crates/sail-plan/src/function/scalar/conditional.rs
+++ b/crates/sail-plan/src/function/scalar/conditional.rs
@@ -1,17 +1,21 @@
 use std::sync::Arc;

-use arrow::datatypes::{DataType, Field, FieldRef, Fields, TimeUnit};
+use arrow::datatypes::{DataType, Field, Fields, IntervalUnit, TimeUnit};
 use datafusion::functions::expr_fn;
 use datafusion_common::ScalarValue;
 use datafusion_expr::type_coercion::binary::type_union_coercion;
 use datafusion_expr::{ExprSchemable, ScalarUDF, cast, expr, lit};
 use sail_common::spec::ARROW_DECIMAL128_MAX_PRECISION;
 use sail_common_datafusion::utils::items::ItemTaker;
 use sail_function::scalar::datetime::spark_date::SparkDate;
 use sail_function::scalar::datetime::spark_timestamp::SparkTimestamp;
+use sail_function::scalar::spark_struct_rename::SparkStructRename;
 use sail_function::scalar::spark_to_string::SparkToUtf8;

-use crate::error::PlanResult;
+use crate::config::PlanConfig;
+use crate::error::{PlanError, PlanResult};
 use crate::function::common::{FunctionContextInput, ScalarFunction, ScalarFunctionInput};
+use crate::resolver::{build_rename_target_type, needs_struct_field_rename};

 fn case(input: ScalarFunctionInput) -> PlanResult<expr::Expr> {
@@ -94,126 +98,300 @@ fn coerce_branch_values(
         .iter()
         .map(|value| value.get_type(function_context.schema))
         .collect::<Result<Vec<_>, _>>()?;
-    let Some(common_type) = common_branch_type(&data_types) else {
-        return Ok(values);
-    };
+    let common_type =
+        common_branch_type(&data_types, function_context.plan_config.as_ref()).ok_or_else(|| {
+            PlanError::analysis(format!(
+                "[DATATYPE_MISMATCH.DATA_DIFF_TYPES] incompatible conditional branch types: \
+                 {data_types:?}"
+            ))
+        })?;
     Ok(values
         .into_iter()
-        .map(|value| value.cast_to(&common_type, function_context.schema))
+        .zip(data_types)
+        .map(|(value, source_type)| {
+            if needs_struct_field_rename(&source_type, &common_type) {
+                let renamed_type = build_rename_target_type(&source_type, &common_type);
+                let renamed =
+                    ScalarUDF::new_from_impl(SparkStructRename::new(renamed_type.clone()))
+                        .call(vec![value]);
+                if renamed_type == common_type {
+                    Ok(renamed)
+                } else {
+                    renamed.cast_to(&common_type, function_context.schema)
+                }
+            } else {
+                value.cast_to(&common_type, function_context.schema)
+            }
+        })
         .collect::<Result<Vec<_>, _>>()?)
 }

 /// The type Spark gives a conditional, folding the branches left to right as `findWiderCommonType`
-/// does. `None` means no common type, and the caller then leaves the branches alone the way Spark's
-/// `.getOrElse(c)` does. Spark's non-ANSI string partition is not reproduced.
+/// does. `None` means Spark has no common branch type and analysis must fail.
 ///
 /// A `Decimal256` anywhere in any branch switches the rules below off for the WHOLE conditional.
 /// Spark has no such type, so they would only narrow it, and deciding per pair instead would let a
 /// fold alternate between two rule sets and depend on branch order.
-fn common_branch_type(data_types: &[DataType]) -> Option<DataType> {
+fn common_branch_type(
+    data_types: &[DataType],
+    plan_config: &PlanConfig,
+) -> Option<DataType> {
     let (first, rest) = data_types.split_first()?;
     if data_types.iter().any(contains_decimal256_type) {
         return rest.iter().try_fold(first.clone(), |left, right| {
             type_union_coercion(&left, right)
         });
     }
+
+    let ordered = spark_ordered_branch_types(data_types, plan_config.ansi_mode);
+    let (first, rest) = ordered.split_first()?;
     rest.iter().try_fold(first.clone(), |left, right| {
-        branch_type_coercion(&left, right)
+        branch_type_coercion(&left, right, plan_config)
     })
 }

-/// The wider of two branch types, following the two arms of Spark's `findWiderTypeForDecimal`
-/// (`TypeCoercionHelper.scala:186-198`) that DataFusion answers differently. Both also reach a
-/// container's element type, since `findTypeForComplex` recurses into it. Everything else is left
-/// to `type_union_coercion`.
-fn branch_type_coercion(left: &DataType, right: &DataType) -> Option<DataType> {
+fn spark_ordered_branch_types(
+    data_types: &[DataType],
+    ansi_mode: bool,
+) -> Vec<DataType> {
+    if ansi_mode {
+        return data_types.to_vec();
+    }
+
+    let mut result = Vec::with_capacity(data_types.len());
+    for data_type in data_types
+        .iter()
+        .filter(|data_type| contains_string_type(data_type))
+    {
+        if !result.contains(data_type) {
+            result.push(data_type.clone());
+        }
+    }
+    result.extend(
+        data_types
+            .iter()
+            .filter(|data_type| !contains_string_type(data_type))
+            .cloned(),
+    );
+    result
+}
+
+fn contains_string_type(data_type: &DataType) -> bool {
+    match data_type {
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => true,
+        DataType::List(field)
+        | DataType::LargeList(field)
+        | DataType::FixedSizeList(field, _) => contains_string_type(field.data_type()),
+        _ => false,
+    }
+}
+
+fn branch_type_coercion(
+    left: &DataType,
+    right: &DataType,
+    plan_config: &PlanConfig,
+) -> Option<DataType> {
+    if left == right {
+        return Some(left.clone());
+    }
+    match (left, right) {
+        (DataType::Null, other) | (other, DataType::Null) => return Some(other.clone()),
+        (DataType::List(left), DataType::List(right)) => {
+            let field = coerce_element_field(left, right, plan_config)?;
+            return Some(DataType::List(Arc::new(field)));
+        }
+        (DataType::LargeList(left), DataType::LargeList(right)) => {
+            let field = coerce_element_field(left, right, plan_config)?;
+            return Some(DataType::LargeList(Arc::new(field)));
+        }
+        (
+            DataType::FixedSizeList(left, left_size),
+            DataType::FixedSizeList(right, right_size),
+        ) if left_size == right_size => {
+            let field = coerce_element_field(left, right, plan_config)?;
+            return Some(DataType::FixedSizeList(Arc::new(field), *left_size));
+        }
+        (DataType::Map(left, sorted), DataType::Map(right, _)) => {
+            return coerce_map_type(left, right, *sorted, plan_config);
+        }
+        (DataType::Struct(left), DataType::Struct(right)) => {
+            return coerce_struct_fields(left, right, plan_config).map(DataType::Struct);
+        }
+        _ if is_complex_type(left) || is_complex_type(right) => return None,
+        _ => {}
+    }
+
+    if is_string_type(left) || is_string_type(right) {
+        return string_branch_type(left, right, plan_config.ansi_mode);
+    }
+    if is_temporal_type(left) || is_temporal_type(right) {
+        if is_temporal_type(left) && is_temporal_type(right) {
+            return wider_temporal_type(left, right, &plan_config.session_timezone);
+        }
+        return None;
+    }
+    if is_time_type(left) || is_time_type(right) {
+        return None;
+    }
+
+    if left.is_numeric() && right.is_numeric() {
+        if plan_config.ansi_mode
+            && ((is_integral_type(left) && matches!(right, DataType::Float32))
+                || (matches!(left, DataType::Float32) && is_integral_type(right)))
+        {
+            return Some(DataType::Float64);
+        }
+        if (left.is_floating() && right.is_decimal())
+            || (left.is_decimal() && right.is_floating())
+        {
+            return Some(DataType::Float64);
+        }
+        if let Some(data_type) = wider_decimal_type(left, right) {
+            return Some(data_type);
+        }
+        return type_union_coercion(left, right);
+    }
+
+    if is_binary_type(left) && is_binary_type(right) {
+        return type_union_coercion(left, right);
+    }
+
+    None
+}
+
+fn string_branch_type(
+    left: &DataType,
+    right: &DataType,
+    ansi_mode: bool,
+) -> Option<DataType> {
+    if is_string_type(left) && is_string_type(right) {
+        return type_union_coercion(left, right);
+    }
+
+    let (string_type, other) = if is_string_type(left) {
+        (left, right)
+    } else {
+        (right, left)
+    };
+
+    if !ansi_mode {
+        return (is_spark_atomic_type(other)
+            && !matches!(other, DataType::Boolean)
+            && !is_binary_type(other))
+        .then(|| string_type.clone());
+    }
+
+    if is_integral_type(other) {
+        return Some(DataType::Int64);
+    }
+    if other.is_floating() || other.is_decimal() {
+        return Some(DataType::Float64);
+    }
+    if is_ansi_interval_type(other) {
+        return None;
+    }
+    is_spark_atomic_type(other).then(|| other.clone())
+}
+
+fn wider_temporal_type(
+    left: &DataType,
+    right: &DataType,
+    session_timezone: &Arc<str>,
+) -> Option<DataType> {
+    if matches!(left, DataType::Timestamp(_, Some(_)))
+        || matches!(right, DataType::Timestamp(_, Some(_)))
+    {
+        return Some(DataType::Timestamp(
+            TimeUnit::Microsecond,
+            Some(Arc::clone(session_timezone)),
+        ));
+    }
+    if matches!(left, DataType::Timestamp(_, None))
+        || matches!(right, DataType::Timestamp(_, None))
+    {
+        return Some(DataType::Timestamp(TimeUnit::Microsecond, None));
+    }
+    if is_date_type(left) && is_date_type(right) {
+        return Some(DataType::Date32);
+    }
+    None
+}
+
+fn coerce_map_type(
+    left: &Field,
+    right: &Field,
+    sorted: bool,
+    plan_config: &PlanConfig,
+) -> Option<DataType> {
+    let (DataType::Struct(left_fields), DataType::Struct(right_fields)) =
+        (left.data_type(), right.data_type())
+    else {
+        return None;
+    };
+    if left_fields.len() != 2 || right_fields.len() != 2 {
+        return None;
+    }
+
+    let left_key = left_fields[0].as_ref();
+    let right_key = right_fields[0].as_ref();
+    let key_type =
+        branch_type_coercion(left_key.data_type(), right_key.data_type(), plan_config)?;
+    if spark_cast_force_nullable(left_key.data_type(), &key_type)
+        || spark_cast_force_nullable(right_key.data_type(), &key_type)
+    {
+        return None;
+    }
+    let key = left_key
+        .clone()
+        .with_data_type(key_type)
+        .with_nullable(false);
+
+    let value = coerce_element_field(
+        left_fields[1].as_ref(),
+        right_fields[1].as_ref(),
+        plan_config,
+    )?;
+    let entries = left
+        .clone()
+        .with_data_type(DataType::Struct(Fields::from(vec![key, value])))
+        .with_nullable(false);
+    Some(DataType::Map(Arc::new(entries), sorted))
+}
+
+fn coerce_element_field(
+    left: &Field,
+    right: &Field,
+    plan_config: &PlanConfig,
+) -> Option<Field> {
+    let data_type =
+        branch_type_coercion(left.data_type(), right.data_type(), plan_config)?;
+    let nullable = left.is_nullable()
+        || right.is_nullable()
+        || spark_cast_force_nullable(left.data_type(), &data_type)
+        || spark_cast_force_nullable(right.data_type(), &data_type);
+    Some(
+        left.clone()
+            .with_data_type(data_type)
+            .with_nullable(nullable),
+    )
+}
+
+fn coerce_struct_fields(
+    left: &Fields,
+    right: &Fields,
+    plan_config: &PlanConfig,
+) -> Option<Fields> {
+    if left.len() != right.len() {
+        return None;
+    }
+    let fields = left
+        .iter()
+        .zip(right.iter())
+        .map(|(left, right)| {
+            if !field_names_match(
+                left.name(),
+                right.name(),
+                plan_config.case_sensitive,
+            ) {
+                return None;
+            }
+            coerce_element_field(left, right, plan_config)
+        })
+        .collect::<Option<Vec<_>>>()?;
+    Some(Fields::from(fields))
+}
+
+fn field_names_match(left: &str, right: &str, case_sensitive: bool) -> bool {
+    if case_sensitive {
+        left == right
+    } else {
+        left.eq_ignore_ascii_case(right)
+    }
+}
+
+fn spark_cast_force_nullable(from: &DataType, to: &DataType) -> bool {
+    if matches!(from, DataType::Null) || from == to {
+        return false;
+    }
+    if is_string_type(from) {
+        return !is_string_type(to) && !is_binary_type(to);
+    }
+    if is_string_type(to) {
+        return false;
+    }
+
+    match (from, to) {
+        (
+            DataType::Timestamp(_, _),
+            DataType::Int8 | DataType::Int16 | DataType::Int32,
+        ) => true,
+        (
+            DataType::Time32(_) | DataType::Time64(_),
+            DataType::Int8 | DataType::Int16,
+        ) => true,
+        (
+            DataType::Float16 | DataType::Float32 | DataType::Float64,
+            DataType::Timestamp(_, _),
+        ) => true,
+        (DataType::Timestamp(_, _), to) if is_date_type(to) => false,
+        (_, to) if is_date_type(to) => true,
+        (from, DataType::Timestamp(_, _)) if is_date_type(from) => false,
+        (from, _) if is_date_type(from) => true,
+        (_, DataType::Interval(IntervalUnit::MonthDayNano)) => true,
+        (_, to) if to.is_decimal() => !decimal_cast_is_null_safe(from, to),
+        (from, to)
+            if (from.is_floating() || from.is_decimal())
+                && is_integral_type(to) =>
+        {
+            true
+        }
+        _ => false,
+    }
+}
+
+fn decimal_cast_is_null_safe(from: &DataType, to: &DataType) -> bool {
+    let Some((target_precision, target_scale)) = decimal_type_for(to) else {
+        return false;
+    };
+    let target_range = target_precision - target_scale;
+    if matches!(from, DataType::Boolean) {
+        return target_range >= 1 && target_scale >= 0;
+    }
+    let Some((source_precision, source_scale)) = decimal_type_for(from) else {
+        return false;
+    };
+    let source_range = source_precision - source_scale;
+    (target_range >= source_range && target_scale >= source_scale)
+        || (from.is_decimal() && target_range > source_range)
+}
+
+fn is_complex_type(data_type: &DataType) -> bool {
+    matches!(
+        data_type,
+        DataType::List(_)
+            | DataType::LargeList(_)
+            | DataType::FixedSizeList(_, _)
+            | DataType::Map(_, _)
+            | DataType::Struct(_)
+    )
+}
+
+fn is_integral_type(data_type: &DataType) -> bool {
+    matches!(
+        data_type,
+        DataType::Int8
+            | DataType::Int16
+            | DataType::Int32
+            | DataType::Int64
+            | DataType::UInt8
+            | DataType::UInt16
+            | DataType::UInt32
+            | DataType::UInt64
+    )
+}
+
+fn is_binary_type(data_type: &DataType) -> bool {
+    matches!(
+        data_type,
+        DataType::Binary | DataType::LargeBinary | DataType::BinaryView
+    )
+}
+
+fn is_time_type(data_type: &DataType) -> bool {
+    matches!(data_type, DataType::Time32(_) | DataType::Time64(_))
+}
+
+fn is_ansi_interval_type(data_type: &DataType) -> bool {
+    matches!(
+        data_type,
+        DataType::Duration(_)
+            | DataType::Interval(IntervalUnit::YearMonth | IntervalUnit::DayTime)
+    )
+}
+
+fn is_spark_atomic_type(data_type: &DataType) -> bool {
+    data_type.is_numeric()
+        || matches!(
+            data_type,
+            DataType::Boolean
+                | DataType::Binary
+                | DataType::LargeBinary
+                | DataType::BinaryView
+                | DataType::Date32
+                | DataType::Date64
+                | DataType::Timestamp(_, _)
+                | DataType::Time32(_)
+                | DataType::Time64(_)
+                | DataType::Duration(_)
+                | DataType::Interval(IntervalUnit::YearMonth | IntervalUnit::DayTime)
+        )
+}
+
+/// Spark's `widerDecimalType`, applied where DataFusion's decimal result differs.
+fn wider_decimal_type(left: &DataType, right: &DataType) -> Option<DataType> {
     if (left.is_floating() && right.is_decimal()) || (left.is_decimal() && right.is_floating()) {
         return Some(DataType::Float64);
     }
-    if let Some(data_type) = wider_decimal_type(left, right) {
-        return Some(data_type);
-    }
-    let coerced = type_union_coercion(left, right)?;
-    Some(coerce_nested_types(left, right, coerced))
-}
-
-/// Substitutes the element type of a container, keeping the shape `type_union_coercion` decided.
-/// Matching on the COERCED type is deliberate: DataFusion can answer a different container variant
-/// than either branch, and requiring all three to agree would skip the substitution exactly there.
-fn coerce_nested_types(left: &DataType, right: &DataType, coerced: DataType) -> DataType {
-    if let (Some(left_field), Some(right_field), Some(coerced_field)) = (
-        container_element_field(left),
-        container_element_field(right),
-        container_element_field(&coerced),
-    ) {
-        let field = coerce_element_field(left_field, right_field, coerced_field);
-        return with_element_field(&coerced, field);
-    }
-    if let (DataType::Struct(left), DataType::Struct(right), DataType::Struct(coerced_fields)) =
-        (left, right, &coerced)
-        && let Some(fields) = coerce_struct_fields(left, right, coerced_fields)
-    {
-        return DataType::Struct(fields);
-    }
-    coerced
-}
-
-/// The field a container holds its elements in: the element for a list, the entries struct for a map.
-fn container_element_field(data_type: &DataType) -> Option<&FieldRef> {
-    match data_type {
-        DataType::List(field)
-        | DataType::LargeList(field)
-        | DataType::FixedSizeList(field, _)
-        | DataType::Map(field, _) => Some(field),
-        _ => None,
-    }
-}
-
-fn with_element_field(data_type: &DataType, field: Field) -> DataType {
-    let field = Arc::new(field);
-    match data_type {
-        DataType::List(_) => DataType::List(field),
-        DataType::LargeList(_) => DataType::LargeList(field),
-        DataType::FixedSizeList(_, size) => DataType::FixedSizeList(field, *size),
-        DataType::Map(_, sorted) => DataType::Map(field, *sorted),
-        other => other.clone(),
-    }
-}
-
-fn coerce_element_field(left: &Field, right: &Field, coerced: &Field) -> Field {
-    match branch_type_coercion(left.data_type(), right.data_type()) {
-        Some(data_type) => coerced.clone().with_data_type(data_type),
-        None => coerced.clone(),
-    }
-}
-
-/// `None` when the fields are not named the same way at the same positions. DataFusion pairs struct
-/// fields by NAME, so pairing them positionally here would coerce fields that are not the same
-/// column; Spark rejects those branches outright, so standing aside is the closer answer. The
-/// comparison ignores case because Spark's resolver does under the default `caseSensitive = false`.
-fn coerce_struct_fields(left: &Fields, right: &Fields, coerced: &Fields) -> Option<Fields> {
-    if left.len() != right.len() || left.len() != coerced.len() {
-        return None;
-    }
-    if left
-        .iter()
-        .zip(right.iter())
-        .zip(coerced.iter())
-        .any(|((left, right), coerced)| {
-            !left.name().eq_ignore_ascii_case(right.name())
-                || !left.name().eq_ignore_ascii_case(coerced.name())
-        })
-    {
-        return None;
-    }
-    Some(Fields::from(
-        left.iter()
-            .zip(right.iter())
-            .zip(coerced.iter())
-            .map(|((left, right), coerced)| coerce_element_field(left, right, coerced))
-            .collect::<Vec<_>>(),
-    ))
-}
-
-/// Spark's `widerDecimalType` (`DecimalPrecisionTypeCoercion.scala:195-199`), applied only past 38
-/// digits — below that DataFusion already agrees. Assumes the non-legacy `boundedPreferIntegralDigits`,
-/// which is the default (`spark.sql.legacy.decimal.retainFractionDigitsOnTruncate = false`).
-fn wider_decimal_type(left: &DataType, right: &DataType) -> Option<DataType> {
     let (left_precision, left_scale) = decimal_type_for(left)?;
     let (right_precision, right_scale) = decimal_type_for(right)?;
     let scale = left_scale.max(right_scale);
diff --git a/python/pysail/tests/spark/function/features/conditional/case_branch_type_coercion.feature b/python/pysail/tests/spark/function/features/conditional/case_branch_type_coercion.feature
--- a/python/pysail/tests/spark/function/features/conditional/case_branch_type_coercion.feature
+++ b/python/pysail/tests/spark/function/features/conditional/case_branch_type_coercion.feature
@@ -235,6 +235,30 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         | NULL              |
         | CAST(NULL AS INT) |

+  Rule: Datetime branches preserve Spark's timestamp flavor
+
+    Scenario Outline: <case> yields <result>
+      When query
+        """
+        SELECT typeof(<expr>) AS t FROM VALUES (false) AS t(c)
+        """
+      Then query result
+        | t        |
+        | <result> |
+
+      Examples:
+        | case               | expr                                                                                           | result        |
+        | CASE LTZ then DATE | CASE WHEN c THEN CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ) ELSE DATE '2020-01-01' END      | timestamp     |
+        | CASE DATE then LTZ | CASE WHEN c THEN DATE '2020-01-01' ELSE CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ) END      | timestamp     |
+        | IF LTZ and DATE    | if(c, CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ), DATE '2020-01-01')                        | timestamp     |
+        | IF DATE and LTZ    | if(c, DATE '2020-01-01', CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ))                        | timestamp     |
+        | IF NTZ and DATE    | if(c, CAST('2020-01-02 03:04:05' AS TIMESTAMP_NTZ), DATE '2020-01-01')                        | timestamp_ntz |
+        | IF LTZ and NTZ     | if(c, CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ), CAST('2020-01-01' AS TIMESTAMP_NTZ))       | timestamp     |
+
   Rule: Nested element types widen along with the branches
@@ -359,6 +383,90 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         | map       | the 38 bound    | map('k', CAST(1.5 AS DECIMAL(20,15)))          | map('k', CAST(2 AS DECIMAL(30,0)))          | map<string,decimal(38,8)> |
         | array     | float to double | array(CAST(1.5 AS DOUBLE))                     | array(CAST(2.5 AS DECIMAL(10,2)))           | array<double>             |

+  Rule: String and temporal coercion recurses into conditional containers
+
+    Scenario Outline: With ANSI disabled, <container> string and date branches widen to string
+      Given config spark.sql.ansi.enabled = false
+      When query
+        """
+        SELECT typeof(if(false, <string_branch>, <date_branch>)) AS t
+        """
+      Then query result
+        | t        |
+        | <result> |
+
+      Examples:
+        | container | string_branch                   | date_branch                          | result           |
+        | array     | array('2020-01-01')             | array(DATE '2020-01-01')             | array<string>    |
+        | struct    | named_struct('f', '2020-01-01') | named_struct('f', DATE '2020-01-01') | struct<f:string> |
+        | map value | map(1, '2020-01-01')            | map(1, DATE '2020-01-01')            | map<int,string>  |
+        | map key   | map('2020-01-01', 1)            | map(DATE '2020-01-01', 1)            | map<string,int>  |
+
+    Scenario: With ANSI disabled, nested strings are folded before earlier branches
+      Given config spark.sql.ansi.enabled = false
+      When query
+        """
+        SELECT
+          typeof(CASE WHEN c = 1 THEN array(TIMESTAMP '2020-01-01 00:00:00')
+                      WHEN c = 2 THEN array(1)
+                      ELSE array('x') END) AS t,
+          CAST(CASE WHEN c = 1 THEN array(TIMESTAMP '2020-01-01 00:00:00')
+                    WHEN c = 2 THEN array(1)
+                    ELSE array('x') END AS STRING) AS v
+        FROM VALUES (2) AS t(c)
+        """
+      Then query result
+        | t             | v   |
+        | array<string> | [1] |
+
+  Rule: Inserted nested casts carry Spark nullability
+
+    Scenario: An ANSI string-to-date array cast makes the element nullable
+      Given config spark.sql.ansi.enabled = true
+      When query
+        """
+        SELECT if(false, array('2020-01-01'), array(DATE '2020-01-01')) AS r
+        """
+      Then query schema
+        """
+        root
+         |-- r: array (nullable = false)
+         |    |-- element: date (containsNull = true)
+        """
+
+    Scenario: An ANSI string-to-date struct cast makes the field nullable
+      Given config spark.sql.ansi.enabled = true
+      When query
+        """
+        SELECT if(false,
+                  named_struct('f', '2020-01-01'),
+                  named_struct('f', DATE '2020-01-01')) AS r
+        """
+      Then query schema
+        """
+        root
+         |-- r: struct (nullable = false)
+         |    |-- f: date (nullable = true)
+        """
+
+    Scenario: An ANSI string-to-date map-value cast makes the value nullable
+      Given config spark.sql.ansi.enabled = true
+      When query
+        """
+        SELECT if(false,
+                  map(1, '2020-01-01'),
+                  map(1, DATE '2020-01-01')) AS r
+        """
+      Then query schema
+        """
+        root
+         |-- r: map (nullable = false)
+         |    |-- key: integer
+         |    |-- value: date (valueContainsNull = true)
+        """
+
+    Scenario: An ANSI cast that can create a null map key is rejected
+      Given config spark.sql.ansi.enabled = true
+      When query
+        """
+        SELECT if(false,
+                  map('2020-01-01', 1),
+                  map(DATE '2020-01-01', 2)) AS r
+        """
+      Then query error .*DATA_DIFF_TYPES.*
+
   Rule: The decimal rules hold with ANSI disabled too
@@ -405,9 +513,35 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         | t                |
         | struct<a:bigint> |

-    # Spark rejects a reordering outright; Sail resolves it by name instead. Pinned so the values
-    # cannot silently land under the wrong field name if the pairing is ever revisited.
-    @sail-bug
+    Scenario: case-only field-name differences resolve under the default resolver
+      Given config spark.sql.caseSensitive = false
+      When query
+        """
+        SELECT
+          typeof(if(c,
+                    named_struct('A', CAST(1 AS INT)),
+                    named_struct('a', CAST(2 AS BIGINT)))) AS t,
+          CAST(if(c,
+                  named_struct('A', CAST(1 AS INT)),
+                  named_struct('a', CAST(2 AS BIGINT))) AS STRING) AS v
+        FROM VALUES (false) AS t(c)
+        """
+      Then query result
+        | t                | v   |
+        | struct<A:bigint> | {2} |
+
+    Scenario: case-only field-name differences are rejected by the case-sensitive resolver
+      Given config spark.sql.caseSensitive = true
+      When query
+        """
+        SELECT if(c,
+                  named_struct('A', CAST(1 AS INT)),
+                  named_struct('a', CAST(2 AS BIGINT))) AS r
+        FROM VALUES (false) AS t(c)
+        """
+      Then query error .*DATA_DIFF_TYPES.*
+
     Scenario: struct branches listing the same fields in a different order are rejected
       When query
         """
@@ -416,18 +550,5 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         FROM VALUES (false) AS t(c)
         """
       Then query error .*DATA_DIFF_TYPES.*
-
-    @sail-only
-    Scenario: a reordered struct still resolves each field by its own name
-      When query
-        """
-        SELECT CAST(if(c, named_struct('a', CAST(1 AS INT), 'b', CAST(2 AS BIGINT)),
-                          named_struct('b', CAST(3 AS BIGINT), 'a', CAST(4 AS INT))) AS STRING) AS v
-        FROM VALUES (false) AS t(c)
-        """
-      Then query result
-        | v      |
-        | {4, 3} |
@@ -468,18 +589,54 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         SELECT if(c, true, 1) AS r FROM VALUES (true) AS t(c)
         """
       Then query error .*(DATA_DIFF_TYPES|Failed to coerce then).*

+    Scenario Outline: <case> is rejected with ANSI set to <ansi>
+      Given config spark.sql.ansi.enabled = <ansi>
+      When query
+        """
+        SELECT <expr> AS r FROM VALUES (false) AS t(c)
+        """
+      Then query error .*(DATA_DIFF_TYPES|Failed to coerce then).*
+
+      Examples:
+        | case            | ansi  | expr                                                       |
+        | CASE date / int | true  | CASE WHEN c THEN DATE '2020-01-01' ELSE CAST(1 AS INT) END |
+        | CASE int / date | false | CASE WHEN c THEN CAST(1 AS INT) ELSE DATE '2020-01-01' END |
+        | IF date / int   | true  | if(c, DATE '2020-01-01', CAST(1 AS INT))                  |
+        | IF date / int   | false | if(c, DATE '2020-01-01', CAST(1 AS INT))                  |
+
+  Rule: ANSI string promotion applies to conditional branches
+
+    Scenario Outline: ANSI promotes <case> to <result>
+      Given config spark.sql.ansi.enabled = true
+      When query
+        """
+        SELECT typeof(<expr>) AS t, CAST(<expr> AS STRING) AS v
+        FROM VALUES (false) AS t(c)
+        """
+      Then query result
+        | t        | v       |
+        | <result> | <value> |
+
+      Examples:
+        | case              | expr                                         | result  | value |
+        | IF int / string   | if(c, CAST(1 AS INT), '2')                   | bigint  | 2     |
+        | CASE int / string | CASE WHEN c THEN CAST(1 AS INT) ELSE '2' END | bigint  | 2     |
+        | IF decimal/string | if(c, CAST(1.25 AS DECIMAL(10,2)), '2.5')    | double  | 2.5   |
+        | IF boolean/string | if(c, true, 'false')                          | boolean | false |
+
   Rule: An integral branch against a float branch widens per dialect

     # ANSI's `findTightestCommonType` returns DoubleType for integral-meets-float
     # (`AnsiTypeCoercion.scala:113-124`); the non-ANSI path (`TypeCoercion.scala:88-92`) answers
-    # plain FLOAT, which is what Sail answers — hence the explicit ANSI pin.
-    #
-    # Still open, and not specific to CASE: `coalesce`, `greatest`, `least`, `array`, `UNION ALL`,
-    # `+` and `*` all answer `float` too, so this belongs in a shared numeric-coercion fix rather
-    # than in the conditional builder.
+    # plain FLOAT. CASE/IF own their branch common type, so this builder must select between them
+    # using the session dialect.
     #
     # The ANSI-off counterpart, which Sail already gets right.
@@ -491,7 +648,6 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
         | t     |
         | float |

-    @sail-bug
     Scenario Outline: <case> widens to double under ANSI
       Given config spark.sql.ansi.enabled = true
@@ -512,7 +668,6 @@ Feature: CASE WHEN and IF type their result from every branch, not just the first
     # combined scenario could not tell "Sail answers float" from "Sail rounds through float32", and
     # would stay green if only one of the two were ever fixed. Read back through DECIMAL(20,0) so
     # the assertion pins the number and not how each engine formats a float.
-    @sail-bug
     Scenario Outline: <case> keeps an integral branch exact instead of rounding it through float32
       Given config spark.sql.ansi.enabled = true
diff --git a/python/pysail/tests/spark/function/test_conditional_arrow_schema.py b/python/pysail/tests/spark/function/test_conditional_arrow_schema.py
--- a/python/pysail/tests/spark/function/test_conditional_arrow_schema.py
+++ b/python/pysail/tests/spark/function/test_conditional_arrow_schema.py
@@ -111,3 +111,24 @@ def test_conditional_arrow_schema_holds_the_widened_value(spark, query, expected):

     assert df.toArrow().column("r").to_pylist() == [expected]
     assert df.collect()[0][0] == expected
+
+
+@pytest.mark.parametrize(
+    ("expression", "expected_timezone"),
+    [
+        pytest.param(
+            "if(false, CAST('2020-01-02 03:04:05' AS TIMESTAMP_LTZ), DATE '2020-01-01')",
+            "UTC",
+            id="ltz_and_date",
+        ),
+        pytest.param(
+            "if(false, CAST('2020-01-02 03:04:05' AS TIMESTAMP_NTZ), DATE '2020-01-01')",
+            None,
+            id="ntz_and_date",
+        ),
+    ],
+)
+def test_conditional_arrow_schema_preserves_timestamp_flavor(
+    spark, expression, expected_timezone
+):
+    arrow = spark.sql(f"SELECT {expression} AS r").toArrow()
+
+    assert arrow.schema.field("r").type.tz == expected_timezone
```

---

PR #2418, opened 2026-08-12
refactor: unify lake row-level operation entry points

lonless9:
## Summary

Introduce a format-neutral row-level planning contract for lake sources.

- add `RowLevelOperation::{Delete, Update, Merge}` to `LakeSource`
- introduce `RowLevelTarget` for the resolved table name, format, location, partitioning, options, and catalog execution context
- represent expanded work as `RowLevelEffect::{WriteRows, TouchFiles, DeleteRows}`
- retain format-specific commit inputs in `RowLevelCommitInfo`
- replace the MERGE-specific logical node with a unified `RowLevelWriteNode`
- route Delta and Iceberg planning through the common lake-source entry point
- preserve the tri-state Iceberg `expected_snapshot_id` contract

This PR establishes the shared entry points and plan representation without adding new DML behavior.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #2417, opened 2026-08-12
refactor: separate data and lake sources

lonless9:
## Summary

Separate generic data-source behavior from lake-format capabilities.

- replace the former `TableFormat` abstraction with a generic `DataSource`
- introduce `LakeSource: DataSource` for lake-specific metadata, DDL, and DML behavior

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #2473, opened 2026-08-21
feat: support scoped Iceberg overwrite

Robin-Everaars:
## Summary

- add dynamic partition overwrite for Iceberg tables
- add predicate overwrite when identity partition values prove that a whole file matches
- preserve unaffected manifests while replacing only the selected live data files
- write conservative per-field bounds and use them for file pruning

This pull request starts from current main instead of reusing or stacking the earlier #2300, #2309 and #2311 branches. It includes the snapshot rewrite pieces that current main still needs for scoped overwrite, then adds both write modes and their coverage. It supersedes #2300, #2309 and #2311.

## Behavior

Dynamic partition overwrite removes live files only from partitions present in the incoming data. An empty input leaves the table unchanged. Predicate overwrite is limited to conditions that identity partition values can prove for a whole file. An empty predicate overwrite still removes matching files.

Both modes keep optimistic snapshot validation, reuse unaffected manifests and recompute overwrite summary totals. They reject tables with active delete files and reject effective Iceberg v3 writes until row lineage is preserved. Requested removal paths must be live in the parent snapshot. Bounds with incompatible types are omitted instead of coerced.

## Validation

- `cargo nextest run -p sail-iceberg -p sail-execution`: 187 passed
- full `cargo test`, including a gold-data update followed by a clean run: 1217 passed and 13 ignored in each run
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- focused scoped-overwrite tests: 7 passed
- scoped feature scenarios: 4 passed with 10 snapshots
- Python formatting checks: clean
- Python suites in CI at this head, including the Iceberg and Delta tests: pass on Spark 3.5.9, 4.0.4, 4.1.3 and 4.2.0


Robin-Everaars:
@lonless9 gentle nudge: if you have found anything already please let me know; I have time tomorrow and afterwards I'm gone for over a week, so I'll try my best to resolve anything before to not have this wait too long.

crates/sail-execution/proto/sail/plan/physical.proto
```diff
@@ -995,6 +995,8 @@ enum IcebergSnapshotUpdateKind {
   ICEBERG_SNAPSHOT_UPDATE_KIND_FAST_APPEND = 1;
   ICEBERG_SNAPSHOT_UPDATE_KIND_FULL_OVERWRITE = 2;
   ICEBERG_SNAPSHOT_UPDATE_KIND_ROW_DELTA = 3;
+  ICEBERG_SNAPSHOT_UPDATE_KIND_COPY_ON_WRITE = 4;
+  reserved 5;
```
lonless9: We can remove this line. 
`reserved 5;`

lonless9 (commented):
Thanks @Robin-Everaars 
- https://github.com/lakehq/sail/pull/2417 
- https://github.com/lakehq/sail/pull/2418 might cause a minor git conflict with this PR, but since they only involve the entry point changes, they should be easy to resolve. Once the conflicts are resolved, we can merge this PR!

linhr:
@Robin-Everaars Thanks for the contribution!

We'd like to merge the PR but it seems that you may have been out. Unfortunately, we cannot resolve the merge conflict from our side, since we cannot push to your fork because it's owned by an organization. The "Maintainers are allowed to edit this pull request." setting only allows maintainers to push to personal forks.

We have heavy development work planned around DML for Iceberg and Delta Lake in the coming week so more merge conflicts may arise. But we understand that you are not available the next week or so, so don't feel pressured to take care of the PR soon. :) We'll leave this PR open, and then when you get back we can evaluate the situation together and see how to bring this to a ready state.

Robin-Everaars:
@linhr I updated the organization fork, so maintainers no longer need to resolve the branch conflict from your side.

The branch is rebased onto the current `main`. The conflicts from #2417 and #2418 retain their unified row-level entry points. I also removed `reserved 5;` as requested.

During the refresh I found and fixed a typed identity-partition issue in predicate overwrite. A new DATE regression reproduced the prior Date32/Int32 mismatch and now passes.

Fresh checks at `2c69b750`:

- Rust and Python format checks passed
- 129 `sail-iceberg` tests passed
- scoped-overwrite protobuf round-trip passed
- 8 scoped-overwrite Spark tests passed
- 4 focused overwrite feature cases passed
- `sail-iceberg` Clippy passed with warnings denied

GitHub now reports the branch as mergeable. Please continue the merge review when convenient.

linhr:
Oh this is great!

I'll let @lonless9 go over the PR quickly tomorrow morning and then we'll merge it and take it from here. Thanks for helping with this!

lonless9 (approved):
Thanks! @Robin-Everaars @linhr

Robin-Everaars:
@lonless9 and @linhr you're very welcome!

PR #2480, opened 2026-08-21
fix: supervise cluster task panics

### LINKED ISSUE #2390: Improve cluster mode execution to return errors early on panic

rohankumardubey:
Closes #2390

Following up on the discussion in #2438, this uses Tokio task supervision instead of `catch_unwind` and `AssertUnwindSafe`.

`TaskMonitor` now runs in a nested Tokio task wrapped with `AbortOnDropHandle`. The supervisor observes its `JoinError` and reports a failed task status when the monitor panics. Normal completion and intentional cancellation remain unchanged.

Synchronous panics during `execute_plan` are intentionally left out of scope.

### Testing

- `cargo test -p sail-execution`
- `cargo clippy -p sail-execution --all-targets -- -D warnings`
- 21 local-cluster execution tests
- Confirmed the original reproducer returns `task panicked: primitive array` instead of hanging


linhr (approved):
This looks very clean! Thanks for the contribution!

PR #2492, opened 2026-08-24
fix: schema query for MS SQL Server

artemklevtsov:
Hi,

Added condition for the MS SQL query: TOP instead LIMIT.

~~
Regards

### DIFF
--- a/python/pysail/spark/datasource/jdbc.py
+++ b/python/pysail/spark/datasource/jdbc.py
@@ -435,12 +435,11 @@ class JdbcDataSource(DataSource):
         dbtable = resolved["dbtable"]
         query = resolved["query"]
 
-        if conn_str.startswith("jdbc:mssql"):
+        if conn_str.startswith("mssql"):
             if query is not None:
-                schema_query = f"TOP 0 SELECT * FROM ({query}) AS _cx_schema_q"  # noqa: S608
+                schema_query = f"SELECT TOP 0 * FROM ({query}) AS _cx_schema_q"  # noqa: S608
             else:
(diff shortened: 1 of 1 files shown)


python/pysail/spark/datasource/jdbc.py
```diff
@@ -435,10 +435,15 @@ def schema(self) -> pa.Schema:
...
-            schema_query = f"SELECT * FROM {dbtable} LIMIT 0"  # noqa: S608
+        if conn_str.startswith("mssql"):
+            if query is not None:
+                schema_query = f"SELECT TOP 0 * FROM ({query}) AS _cx_schema_q"  # noqa: S608
+            else:
+                schema_query = f"SELECT TOP 0 * FROM {dbtable}"  # noqa: S608: S608
+            if query is not None:
+                schema_query = f"SELECT * FROM ({query}) AS _cx_schema_q LIMIT 0"  # noqa: S608
+            else:
+                schema_query = f"SELECT * FROM {dbtable} LIMIT 0"  # noqa: S608
```
linhr, line 447: The logic doesn't seem right. The two `if ... else ...` statements are indented at the same level. It's fine to skip adding tests for now, but could you test the changes manually?

artemklevtsov: Sorry. Typo after fixes first commit.

linhr:
Thanks for the fix, but the latest commit still failed with the same Python lint error as the initial commit.

To avoid more back-and-forth, I'll merge the other PR #2495 which is more complete.

But thanks for reporting the issue in the first place!

PR #2495, opened 2026-08-24
fix: use TOP for MSSQL schema queries

### LINKED ISSUE #2449: Read from SQL Server fails with syntax error
Hi,

Seems sail form incorrect query for the MS SQL Server.

Code to reproduce:

```python
from pysail.spark.datasource.jdbc import JdbcDataSource

server = SparkConnectServer()
server.start()

address, port = server.listening_address
spark = SparkSession.builder \
  .remote(f"sc://{address}:{port}") \
  .getOrCreate()

spark.dataSource.register(JdbcDataSource)

df_ms = spark.read \
  .format("jdbc") \
  .option("url", "jdbc:mssql://localhost:1433/test") \
  .option("query", "SELECT 1 AS n") \
  .option("user", "sa") \
  .option("password", "pass") \
  .load()

df_ms.show()
```

Output:

```python
pyspark.errors.exceptions.connect.SparkRuntimeException: Python error: [jdbc::schema] RuntimeError: Failed to infer schema from JDBC source. Query: 'SELECT * FROM (SELECT 1 AS n) AS _cx_schema_q LIMIT 0'. Error: Token error: 'Incorrect syntax near 'LIMIT'.' on server DBPROD6\DBPROD6 executing  on line 1 (code: 102, state: 1, class: 15)
Traceback:
Traceback (most recent call last):
  File "/data/work/reconciliation/dwh/.venv/lib/python3.13/site-packages/pysail/spark/datasource/jdbc.py", line 447, in schema
    raise RuntimeError(msg) from e
```

`LIMIT` not supported in MS SQL. Need `TOP` instead.

~~
Regards

ErikBPF:
## What

- use `SELECT TOP 0` for SQL Server schema inference
- keep `LIMIT 0` for other JDBC sources
- cover both `query` and `dbtable` options in default unit CI

Fixes #2449 — [Read from SQL Server fails with syntax error](https://github.com/lakehq/sail/issues/2449)

## Testing

- `.venvs/default/bin/pytest -q python/pysail/tests/test_jdbc.py` — 2 passed
- nearby JDBC unit checks — 7 passed
- `hatch fmt --check`

The PostgreSQL testcontainer suite needs Docker and could not run locally.

### DIFF
--- a/python/pysail/spark/datasource/jdbc.py
+++ b/python/pysail/spark/datasource/jdbc.py
@@ -435,10 +435,11 @@ class JdbcDataSource(DataSource):
         dbtable = resolved["dbtable"]
         query = resolved["query"]
 
-        if query is not None:
-            schema_query = f"SELECT * FROM ({query}) AS _cx_schema_q LIMIT 0"  # noqa: S608
+        source = f"({query}) AS _cx_schema_q" if query is not None else dbtable
+        if conn_str.startswith("mssql://"):
+            schema_query = f"SELECT TOP 0 * FROM {source}"  # noqa: S608
         else:
-            schema_query = f"SELECT * FROM {dbtable} LIMIT 0"  # noqa: S608
+            schema_query = f"SELECT * FROM {source} LIMIT 0"  # noqa: S608
 
         try:
             table: pa.Table = cx.read_sql(conn_str, schema_query, return_type="arrow")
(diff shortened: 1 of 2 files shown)


python/pysail/tests/test_jdbc.py
```diff
@@ -0,0 +1,34 @@
+"""Unit tests for the JDBC data source."""
```
linhr: Let's move the test to `python/pysail/tests/spark/datasource/test_jdbc.py` instead of having this separate file.

ErikBPF: Moved it into `python/pysail/tests/spark/datasource/test_jdbc.py` and kept both the `query` and `dbtable` cases there. Thanks.

linhr (approved):
Nice! Thanks!

PR #2501, opened 2026-08-25
fix: derive VALUES column nullability

pomykalakyle:
DataFusion infers every column in a `VALUES` relation as nullable, even when all expressions in a column are non-nullable. This causes Sail schemas to differ from Spark for queries such as integer literals alongside timestamp casts.

Derive each `VALUES` field's nullability from its resolved expressions while retaining DataFusion's existing type inference and plan construction. Remove the known-bug marker from the existing regression scenario and update affected explain snapshots.

### DIFF
--- a/crates/sail-plan/src/resolver/query/values.rs
+++ b/crates/sail-plan/src/resolver/query/values.rs
@@ -28,7 +28,7 @@ impl PlanResolver<'_> {
             Ok::<_, PlanError>(results)
         }
         .await?;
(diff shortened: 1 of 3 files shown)


lonless9:
Thanks for bringing this up! It seems this could be naturally resolved with the upgrade to DataFusion 55. Let's wait and see how things look once this PR https://github.com/lakehq/sail/pull/2383 is merged.

pomykalakyle:
This will be supported by Data Fusion's update. I'm closing the PR, will need to re-work when the update is ready.

PR #2502, opened 2026-08-25
fix: validate CSV headers when schema enforcement is disabled

pomykalakyle:
## Summary

- support the Spark CSV `enforceSchema` read option
- validate each CSV header against the requested schema during file execution
- preserve the option when physical plans are serialized to workers


pomykalakyle:
This will be supported by Data Fusion's update. I'm closing the PR, will need to re-work when the update is ready.

PR #2511, opened 2026-08-26
fix: support Spark Connect CloneSession

ErikBPF:
## Summary

- implement `CloneSession` through the shared session manager/factory lifecycle
- validate source identity, observed server ID, target UUID, duplicates, tombstones, and concurrent target claims
- snapshot SQL config, DataFusion registries, temp views, functions, and current catalog/database into a fresh session
- preserve independent driver, execution, streaming, timeout, telemetry, and release ownership

## Design

The RPC only resolves protocol fields and delegates to `SessionManager::clone_session_context`. The manager actor performs atomic source/target checks and creates the target through `SessionFactory::clone_session`, so clones use normal bookkeeping and cleanup.

The target receives fresh runtime/session resources. Mutable Spark config and Sail catalog state are copied as snapshots. Active executors, operations, streaming queries, and in-flight work are not copied.

Public acceptance coverage is a direct PySpark 4.2 interface test under `spark/session`. Focused Rust tests remain only for identity, concurrency, timeout, and catalog-state seams.

## Prerequisite

#2514 independently fixes the Kubernetes worker launch/shutdown race exposed during CloneSession Minikube cleanup validation. Merge it before the distributed GDR canary; it is intentionally outside this protocol/session PR.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p sail-spark-connect -p sail-catalog --all-targets --no-deps -- -D warnings -A unfulfilled-lint-expectations`
- focused CloneSession RPC tests: 2 passed
- catalog clone-state test: 1 passed
- PySpark 4.2 CloneSession acceptance: 3 passed
- prior full CI: Rust build/tests, lint, docs, and PySpark 4.2 passed
- isolated Gemini Minikube acceptance in both release orders with #2514 applied; distinct source/clone workers returned to zero and Sail remained Ready

## Compatibility boundary

- artifact cloning is deferred because Sail artifact RPC/state is not implemented
- server-side session IDs continue to mirror client session IDs, matching Sail existing integrity model
- production queue admission and the real DataFoundation/GDR canary remain deployment gates
- clone support does not imply later PostgreSQL, Delta, SQL, or DataFrame parity

Closes #2512

PR #2514, opened 2026-08-26
fix: serialize Kubernetes worker shutdown

ErikBPF:
## Summary

- serialize worker pod creation with worker-manager shutdown
- ignore launches after shutdown begins
- add a fake Kubernetes API regression test for the late-pod race

## Why

A pod POST can be in flight while `stop()` deletes the current worker collection. Without synchronization, that POST may complete afterward and leave an orphan worker. CloneSession Minikube validation in #2511 exposed the race, but it affects any concurrent launch and shutdown.

## Validation

- regression test failed against `main`: deletion raced ahead of blocked creation
- `cargo test -p sail-execution stop_waits_for_in_flight_worker_creation_before_deleting_pods --lib`
- `cargo clippy -p sail-execution --all-targets --no-deps -- -D warnings -A unfulfilled-lint-expectations`
- `cargo fmt --all -- --check`

Refs #2512
Prerequisite cleanup fix for #2511.

### DIFF
--- a/crates/sail-execution/src/worker_manager/kubernetes.rs
+++ b/crates/sail-execution/src/worker_manager/kubernetes.rs
@@ -17,7 +17,7 @@ use sail_common::actor::ActorSystem;
 use sail_common::config::ClusterConfigEnv;
 use sail_common::telemetry::ContextPropagationEnv;
 use sail_common::utils::retry::RetryStrategy;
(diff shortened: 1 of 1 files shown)


linhr:
See the decision here: <https://github.com/lakehq/sail/issues/2512#issuecomment-5433792181>

Issue #2519, opened 2026-08-27
Cannot cast Timestamp(Millisecond/Nanosecond/Second, tz) to Spark type when analyzing Parquet schema


amit-odaia:
## Correction (see comment below for context)

The original report in this issue claimed the failure was triggered by `to_date()`/`to_timestamp()`/`current_date()` followed by `.write.parquet(...)`. That repro does **not** reproduce on Sail 0.7.1 — verified directly against a live cluster, with and without `.persist()`, and with many unioned DataFrames. This issue has been corrected below with the actual minimal repro and root cause.

## Environment

- **Sail Version:** 0.7.1
- **Spark Version:** 4.2.0 (via Spark Connect)
- **Client:** pyspark 3.5.7 (Spark Connect)

## Actual Root Cause

The failure is **not** about timezone, and **not** about `to_date`/`to_timestamp`/write. It occurs when Sail merely *analyzes the schema* (`.columns` / `.schema` — no data read, no write) of a Parquet file whose timestamp column was written by a non-Spark tool (e.g. `pyarrow`, or any external export pipeline) at any Arrow time unit other than **microseconds**.

Spark's internal timestamp representation is microsecond-based. Sail's Parquet reader can convert an Arrow timestamp column to Spark's type only when the Arrow unit is already `us`. Millisecond, nanosecond, and second-precision Arrow timestamps all fail the cast — independent of whether a timezone is attached.

## Minimal Reproducible Example

```python
import pyarrow as pa
import pyarrow.parquet as pq
from pyspark.sql import SparkSession

# Write a parquet file the way an external tool (not Spark) would --
# e.g. pyarrow, pandas, or any non-Spark export pipeline.
table = pa.table({
    "id": pa.array(["A", "B"], type=pa.string()),
    "updated_at": pa.array([1700000000000, 1700000001000], type=pa.timestamp("ms", tz="UTC")),
})
pq.write_table(table, "/tmp/repro.parquet")
# upload /tmp/repro.parquet to s3://<bucket>/repro.parquet

spark = SparkSession.builder.remote("sc://<sail-cluster>:50051").getOrCreate()

df = spark.read.parquet("s3a://<bucket>/repro.parquet")
df.columns  # <-- fails here; no .show()/.collect()/.write() needed
```

**Expected:** returns `['id', 'updated_at']`
**Actual:** `UnsupportedOperationException: cast Timestamp(Millisecond, Some("UTC")) to Spark data type`

[...]

PR #2522, opened 2026-08-29
fix: read Parquet files with second and millisecond timestamps

james-willis:
## Problem

Reading a Parquet file whose timestamp columns are millisecond-precision fails:

```
cast Timestamp(Millisecond, None) to Spark data type
```

The error comes from the Arrow-to-Spark type conversion in `crates/sail-spark-connect/src/proto/data_type_arrow.rs`, which maps only `Timestamp(Microsecond, _)` and has an explicit error arm for `Second`, `Millisecond`, and `Nanosecond`.

`timestamp[ms]` is common in the wild — I hit this on a public benchmark dataset whose pickup/dropoff columns are `timestamp[ms]` — so today those files are simply unreadable by path.

### Minimal repro

```python
import pyarrow as pa, pyarrow.parquet as pq
from datetime import datetime

pq.write_table(
    pa.table({"t": pa.array([datetime(2024, 1, 2, 3, 4, 5)], type=pa.timestamp("ms"))}),
    "/tmp/ms.parquet",
)

spark.read.parquet("/tmp/ms.parquet").show()   # fails
```

## Root cause

The cast itself already works. Passing an explicit schema on the same file succeeds and returns the right values:

```python
spark.read.schema("t TIMESTAMP_NTZ").parquet("/tmp/ms.parquet").show()   # works
```

Both branches of `ListingDataSource::create_reader` end up building the same `ListingTableSource`, and DataFusion's schema adapter casts each file to the table schema as it scans — so when the table schema says microseconds, the ms→µs widening happens transparently and correctly. The only gap is that the **inference** branch leaves the unit at whatever the file physically stores, and that millisecond type then survives all the way to the Spark type boundary.

## Fix

`rewrite_utf8view_fields` in `crates/sail-data-source/src/listing/utils.rs` was already a normalization pass over the inferred schema for exactly this class of problem (`// TODO: Spark doesn't support Utf8View`). This generalizes it to `rewrite_unsupported_fields` and adds a `Timestamp(Second | Millisecond, tz)` → `Timestamp(Microsecond, tz)` coercion, preserving the time zone, nullability, and field/schema metadata. The single call site in the inference arm of `listing/source.rs` is renamed to match.

Spark's `TimestampType` is documented as microsecond precision, so this is a lossless widening.

Nanoseconds are deliberately left to error rather than silently truncated.

## Why this matches Spark

Spark does exactly this normalization; its Parquet reader is native, so it does the widening on the way in and never reaches its own strict Arrow boundary. Checked against Spark v4.0.4:

- `sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:297-298`
accepts both units as the same Spark type: `case timestamp: TimestampLogicalTypeAnnotation if timestamp.getUnit == TimeUnit.MICROS || timestamp.getUnit == TimeUnit.MILLIS => ...` yielding `TimestampType` / `TimestampNTZType`.
- `sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/ParquetVectorUpdaterFactory.java:144,156`
selects `LongAsMicrosUpdater`, which widens values via `DateTimeUtils.millisToMicros(...)` (line 768).
- `ParquetSchemaConverter.scala:304-308` rejects `NANOS` as a timestamp (SPARK-40819),
readable only as `LongType` under a legacy flag — so leaving nanoseconds to error here matches Spark.
- `sql/api/src/main/scala/org/apache/spark/sql/util/ArrowUtils.scala` `fromArrowType` is
just as strict as Sail's conversion (microsecond only, everything else throws). Sail's strict boundary is faithful to Spark; Sail is Arrow-native end to end, which is why the millisecond type reaches it at all.

## Verification

- New unit test `test_rewrite_unsupported_fields` in `listing/utils.rs`, covering the
second/millisecond widening, time-zone preservation, nullability and metadata preservation, and that microsecond, nanosecond, and unrelated types are untouched. `cargo test -p sail-data-source --lib` → 67 passed, 0 failed.
- `cargo +nightly fmt -- --check` and
`cargo clippy -p sail-data-source --all-targets --all-features -- -D warnings` are clean.
- Behavior verified end to end against a real `timestamp[ms]` Parquet dataset: the values
returned match `pyarrow` exactly for the same row, confirming this is a genuine cast and not a reinterpret.

## Known scope

- Like the pre-existing `Utf8View` rewrite, this pass only walks **top-level** fields.
Timestamps nested inside struct/list/map fields are still left at their physical unit and will hit the same error. Happy to extend it recursively if you'd prefer that in this PR, but I kept the change narrow.
- This is the read path only. Issue #2519 reports the same error message from the write
path (`createDataFrame` + `to_date` + `df.write.parquet`, no file read involved), which this does not address — same error family, different site.
- Per the development guide, Rust tests are reserved for internal utilities with
easy-to-construct inputs, which `rewrite_unsupported_fields` is (schema in, schema out). If you'd rather have a Python regression test alongside it, `python/pysail/tests/spark/datasource/test_parquet.py` is the natural home — just say the word and I'll add one.

### DIFF
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -7,6 +7,7 @@ use datafusion::datasource::physical_plan::parquet::CachedParquetFileReaderFacto
 use datafusion::datasource::physical_plan::parquet::metadata::{
     DFParquetMetadata, ordering_from_parquet_metadata,
 };
+use datafusion::parquet::arrow::parquet_to_arrow_schema;
 use datafusion_common::config::TableParquetOptions;
 use datafusion_common::parsers::CompressionTypeVariant;
 use datafusion_common::{DataFusionError, Result, plan_err};
@@ -32,6 +33,33 @@ fn fail_for_encryption_factory(options: &TableParquetOptions) -> Result<()> {
     } else {
         Ok(())
     }
+}
+
+/// Reinterprets top-level `Timestamp(Second, _)` fields that are physically stored as `INT64`
+/// back to `Int64`. Parquet has no seconds timestamp logical type: an Arrow writer encodes an
+/// Arrow `timestamp[s]` as an unannotated physical `INT64` and preserves the Arrow type only in
+/// the `ARROW:schema` hint. Spark ignores that hint and exposes the column as a long, so we match
+/// it here rather than widening the hinted seconds to a microsecond timestamp (which would change
+/// both the reported type and the values). Genuine INT96 timestamps are unaffected because they
+/// are handled by the `coerce_int96` option and are not physically `INT64`.
+fn reinterpret_second_timestamps_as_int64(schema: Schema, physical: &Schema) -> Schema {
+    let fields = schema
+        .fields()
+        .iter()
+        .map(|field| {
+            let physical_type = physical
+                .field_with_name(field.name())
+                .ok()
+                .map(|physical_field| physical_field.data_type());
+            match (field.data_type(), physical_type) {
+                (DataType::Timestamp(TimeUnit::Second, _), Some(DataType::Int64)) => {
+                    field.as_ref().clone().with_data_type(DataType::Int64)
+                }
+                _ => field.as_ref().clone(),
+            }
+        })
+        .collect::<Vec<_>>();
+    Schema::new_with_metadata(fields, schema.metadata().clone())
 }
 
 #[async_trait::async_trait]
@@ -71,12 +99,18 @@ impl ReadFormat for ParquetReadFormat {
 
         let mut schemas: Vec<(object_store::path::Path, Schema)> = futures::stream::iter(objects)
             .map(|(store, object)| async {
-                let schema = DFParquetMetadata::new(store.as_ref(), object)
+                let reader = DFParquetMetadata::new(store.as_ref(), object)
                     .with_metadata_size_hint(metadata_size_hint)
                     .with_file_metadata_cache(Some(Arc::clone(&metadata_cache)))
-                    .with_coerce_int96(coerce_int96)
-                    .fetch_schema()
-                    .await?;
+                    .with_coerce_int96(coerce_int96);
+                let schema = reader.fetch_schema().await?;
+                // Compare the Arrow-hinted schema against the metadata-free physical schema so
+                // that Arrow-only `timestamp[s]` hints over physical INT64 are read as long,
+                // matching Spark. The metadata cache above makes this second fetch cheap.
+                let metadata = reader.fetch_metadata().await?;
+                let physical =
+                    parquet_to_arrow_schema(metadata.file_metadata().schema_descr(), None)?;
+                let schema = reinterpret_second_timestamps_as_int64(schema, &physical);
                 Ok::<_, DataFusionError>((object.location.clone(), schema))
             })
             .boxed() // Workaround for https://github.com/rust-lang/rust/issues/64552
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -16,6 +16,7 @@ use object_store::{ObjectMeta, ObjectStore};
 use sail_common_datafusion::schema_evolution::SchemaEvolutionPhysicalExprAdapterFactory;
 
 use crate::listing::source::{ListingFileMeta, ListingFileSample, ListingScanInput, ReadFormat};
+use crate::listing::utils::try_merge_normalized;
 use crate::options::r#gen::ParquetReadOptions;
 
 #[derive(Debug, Clone)]
@@ -90,9 +91,9 @@ impl ReadFormat for ParquetReadFormat {
         let schemas = schemas.into_iter().map(|(_, schema)| schema);
 
         let merged = if options.global.skip_metadata {
-            Schema::try_merge(schemas.map(clear_metadata))
+            try_merge_normalized(schemas.map(clear_metadata))
         } else {
-            Schema::try_merge(schemas)
+            try_merge_normalized(schemas)
         }?;
 
         let merged = if options.global.binary_as_string {
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -122,6 +122,9 @@ impl ReadFormat for ParquetReadFormat {
         // Ensure deterministic ordering for stable schema inference.
         schemas.sort_unstable_by(|(location1, _), (location2, _)| location1.cmp(location2));
 
+        // TODO: Retag inferred microsecond-only adjusted-to-UTC timestamps with Spark's session
+        // timezone instead of retaining the producer's `ARROW:schema` timezone identifier. Spark
+        // renders such a column in the session zone; Sail currently keeps the producer's label.
         let schemas = schemas.into_iter().map(|(_, schema)| schema);
 
         let merged = if options.global.skip_metadata {
--- a/crates/sail-data-source/src/listing/source.rs
+++ b/crates/sail-data-source/src/listing/source.rs
@@ -26,7 +26,7 @@ use url::Url;
 
 use crate::listing::table::{ListingTableSource, ListingTableSourceConfig};
 use crate::listing::utils::{
-    infer_partitions, rewrite_utf8view_fields, sample_listing_files, validate_partitions,
+    infer_partitions, rewrite_unsupported_fields, sample_listing_files, validate_partitions,
 };
 use crate::listing::write::{FileWriteNode, FileWriteOptions};
 use crate::resolve_listing_urls;
@@ -239,7 +239,7 @@ impl<T: FormatFactory> DataSource for ListingDataSource<T> {
                 let schema = read_format
                     .infer_schema(ctx, &sampled_files, compression)
                     .await?;
-                let schema = rewrite_utf8view_fields(schema);
+                let schema = rewrite_unsupported_fields(schema);
 
                 let partition_by = if partition_by.is_empty() {
                     infer_partitions(&sampled_files)?
--- a/crates/sail-data-source/src/listing/utils.rs
+++ b/crates/sail-data-source/src/listing/utils.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use arrow_schema::FieldRef;
-use datafusion::arrow::datatypes::{DataType, Field, Schema};
+use datafusion::arrow::datatypes::{DataType, Field, Schema, TimeUnit};
 use datafusion::datasource::listing::helpers::expr_applicable_for_cols;
 use datafusion::execution::cache::TableScopedPath;
 use datafusion::execution::cache::cache_manager::CachedFileList;
@@ -20,17 +20,29 @@ use object_store::{ObjectMeta, ObjectStore, ObjectStoreExt};
 use crate::listing::source::ListingFileSample;
 use crate::url::PathGlobFilter;
 
-pub fn rewrite_utf8view_fields(schema: Arc<Schema>) -> Arc<Schema> {
-    // TODO: Spark doesn't support Utf8View
+/// Rewrites inferred field types that have no Spark counterpart.
+///
+/// Spark's type system is narrower than Arrow's, and the Arrow-to-Spark conversion rejects
+/// what it cannot represent. Coercing the inferred schema here, rather than letting the
+/// conversion fail later, keeps files readable that DataFusion can already read: the listing
+/// table casts each file to this schema as it scans.
+pub fn rewrite_unsupported_fields(schema: Arc<Schema>) -> Arc<Schema> {
     let new_fields: Vec<Field> = schema
         .fields()
         .iter()
-        .map(|field| {
-            if matches!(field.data_type(), &DataType::Utf8View) {
-                field.as_ref().clone().with_data_type(DataType::Utf8)
-            } else {
-                field.as_ref().clone()
-            }
+        .map(|field| match field.data_type() {
+            // TODO: Spark doesn't support Utf8View
+            DataType::Utf8View => field.as_ref().clone().with_data_type(DataType::Utf8),
+            // Spark timestamps are microseconds, so second and millisecond timestamps are
+            // widened here; the conversion would otherwise reject them even though the
+            // widening is lossless. Nanoseconds are left alone so that they are still
+            // reported rather than silently truncated, which matches Spark: its Parquet
+            // reader accepts `MILLIS` and `MICROS` but not `NANOS` (SPARK-40819).
+            DataType::Timestamp(TimeUnit::Second | TimeUnit::Millisecond, tz) => field
+                .as_ref()
+                .clone()
+                .with_data_type(DataType::Timestamp(TimeUnit::Microsecond, tz.clone())),
+            _ => field.as_ref().clone(),
         })
         .collect();
 
@@ -273,6 +285,8 @@ pub fn can_be_evaluated_for_partition_pruning(
 #[expect(clippy::unwrap_used)]
 #[cfg(test)]
 mod tests {
+    use std::collections::HashMap;
+
     use super::*;
 
     #[test]
@@ -317,5 +331,62 @@ mod tests {
             &file,
             &Path::from("data/_data.json")
         ));
+    }
+
+    #[test]
+    fn test_rewrite_unsupported_fields() {
+        let utc = Some("UTC".into());
+        let metadata = HashMap::from([("k".to_string(), "v".to_string())]);
+        let schema = Arc::new(
+            Schema::new(vec![
+                Field::new("view", DataType::Utf8View, true),
+                Field::new(
+                    "ms",
+                    DataType::Timestamp(TimeUnit::Millisecond, None),
+                    false,
+                )
+                .with_metadata(metadata.clone()),
+                Field::new(
+                    "ms_tz",
+                    DataType::Timestamp(TimeUnit::Millisecond, utc.clone()),
+                    true,
+                ),
+                Field::new("s", DataType::Timestamp(TimeUnit::Second, None), true),
+                Field::new("us", DataType::Timestamp(TimeUnit::Microsecond, None), true),
+                Field::new("ns", DataType::Timestamp(TimeUnit::Nanosecond, None), true),
+                Field::new("other", DataType::Int64, true),
+            ])
+            .with_metadata(metadata.clone()),
+        );
+        let schema = rewrite_unsupported_fields(schema);
+        let field = |name: &str| schema.field_with_name(name).unwrap().clone();
+
+        // Spark has no `Utf8View` type.
+        assert_eq!(field("view").data_type(), &DataType::Utf8);
+
+        // Second and millisecond timestamps widen to microseconds, keeping the time zone.
+        let us = DataType::Timestamp(TimeUnit::Microsecond, None);
+        assert_eq!(field("ms").data_type(), &us);
+        assert_eq!(field("s").data_type(), &us);
+        assert_eq!(
+            field("ms_tz").data_type(),
+            &DataType::Timestamp(TimeUnit::Microsecond, utc)
+        );
+
+        // Microsecond timestamps and unrelated types are left alone.
+        assert_eq!(field("us").data_type(), &us);
+        assert_eq!(field("other").data_type(), &DataType::Int64);
+
+        // Nanoseconds are still reported rather than silently truncated.
+        assert_eq!(
+            field("ns").data_type(),
+            &DataType::Timestamp(TimeUnit::Nanosecond, None)
+        );
+
+        // Rewriting a field preserves its nullability and metadata, and the schema keeps
+        // its own metadata.
+        assert!(!field("ms").is_nullable());
+        assert_eq!(field("ms").metadata(), &metadata);
+        assert_eq!(schema.metadata(), &metadata);
     }
 }
--- a/crates/sail-data-source/src/listing/utils.rs
+++ b/crates/sail-data-source/src/listing/utils.rs
@@ -27,6 +27,25 @@ use crate::url::PathGlobFilter;
 /// conversion fail later, keeps files readable that DataFusion can already read: the listing
 /// table casts each file to this schema as it scans.
 pub fn rewrite_unsupported_fields(schema: Arc<Schema>) -> Arc<Schema> {
+    Arc::new(normalize_unsupported_fields(&schema))
+}
+
+/// Merges per-file inferred schemas, normalizing each one before the merge.
+///
+/// [`Schema::try_merge`] rejects fields whose data types differ, so the normalization has to
+/// happen before the merge rather than after it. A directory holding the same column as
+/// `timestamp[ms]` in one file and `timestamp[us]` in another would otherwise fail to merge,
+/// even though Spark represents both as a single timestamp type, and that failure would occur
+/// before [`rewrite_unsupported_fields`] ever ran.
+pub fn try_merge_normalized(schemas: impl IntoIterator<Item = Schema>) -> Result<Schema> {
+    Ok(Schema::try_merge(
+        schemas
+            .into_iter()
+            .map(|schema| normalize_unsupported_fields(&schema)),
+    )?)
+}
+
+fn normalize_unsupported_fields(schema: &Schema) -> Schema {
     let new_fields: Vec<Field> = schema
         .fields()
         .iter()
@@ -46,10 +65,7 @@ pub fn rewrite_unsupported_fields(schema: Arc<Schema>) -> Arc<Schema> {
         })
         .collect();
 
-    Arc::new(Schema::new_with_metadata(
-        new_fields,
-        schema.metadata().clone(),
-    ))
+    Schema::new_with_metadata(new_fields, schema.metadata().clone())
 }
 
 fn ends_with_ignore_ascii_case(s: &str, suffix: &str) -> bool {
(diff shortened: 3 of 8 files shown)


crates/sail-data-source/src/listing/source.rs
```diff
@@ -239,7 +239,7 @@ impl<T: FormatFactory> DataSource for ListingDataSource<T> {
                 let schema = read_format
                     .infer_schema(ctx, &sampled_files, compression)
                     .await?;
-                let schema = rewrite_utf8view_fields(schema);
+                let schema = rewrite_unsupported_fields(schema);
```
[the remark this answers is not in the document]
james-willis: This is a pre-existing field from teh UTF8View fix but I went ahead and fixed

crates/sail-data-source/src/listing/utils.rs
```diff
@@ -20,24 +20,52 @@ use object_store::{ObjectMeta, ObjectStore, ObjectStoreExt};
...
         .iter()
-        .map(|field| {
-            if matches!(field.data_type(), &DataType::Utf8View) {
-                field.as_ref().clone().with_data_type(DataType::Utf8)
-            } else {
-                field.as_ref().clone()
-            }
+        .map(|field| match field.data_type() {
+            // TODO: Spark doesn't support Utf8View
+            DataType::Utf8View => field.as_ref().clone().with_data_type(DataType::Utf8),
```
shehabgamin: We should probably handle `BinaryView` as well, right?

shehabgamin (commented):
@james-willis I did a manual review and then had Codex do a review as well.

---
# Codex Review

## Actionable findings

### 1. High (P1): Arrow `timestamp[s]` metadata is silently exposed as a timestamp instead of Spark’s `BIGINT`

New-side location: [`crates/sail-data-source/src/listing/utils.rs:60`](https://github.com/lakehq/sail/blob/09290c49b31c70f24cd29efb303348b93cae33fc/crates/sail-data-source/src/listing/utils.rs#L60)

Exact review comment:

> Parquet has no seconds timestamp logical annotation. Arrow-rs writes an Arrow `timestamp[s]` as an unannotated physical `INT64` and stores the Arrow type only in `ARROW:schema`; Spark ignores that key, infers `BIGINT`, and returns the raw epoch-second values. This branch instead widens the Arrow hint to `TIMESTAMP_NTZ`, silently changing both the public type and the values. Please compare the Arrow-hinted schema with the metadata-free physical Parquet schema and reinterpret only top-level `Timestamp(Second, _)` fields backed by physical `INT64`. That preserves actual INT96 fields using the existing `coerce_int96=s` option.

Spark/source evidence:

- Arrow-rs explicitly emits `timestamp[s]` as unannotated `INT64`: [arrow-rs schema conversion:635](/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parquet-59.2.0/src/arrow/schema/mod.rs:635).
- Arrow-rs restores the logical Arrow type from the `ARROW:schema` hint: [arrow-rs schema conversion:49](/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/parquet-59.2.0/src/arrow/schema/mod.rs:49).
- Spark consults only its own schema metadata, otherwise converting the physical Parquet schema: [ParquetFileFormat.scala:564](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala:564).
- Spark maps unannotated `INT64` to `LongType` and recognizes only `MILLIS`/`MICROS` timestamp annotations: [ParquetSchemaConverter.scala:289](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:289).

Reproduction using physical `optional int64 ts`, an `ARROW:schema` hint of `timestamp[s]`, and raw values `[1704164645, -1]`:

```text
Spark 4.1.1:
schema = struct<ts:bigint>
rows   = [Row(ts=-1), Row(ts=1704164645)]

Sail at 09290c49:
schema = struct<ts:timestamp_ntz>
rows   = [1969-12-31 23:59:59, 2024-01-02 03:04:05]
```

Exact code diff:

```diff
diff --git a/crates/sail-data-source/src/formats/parquet/read.rs b/crates/sail-data-source/src/formats/parquet/read.rs
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -5,6 +5,7 @@ use datafusion::catalog::Session;
 use datafusion::datasource::physical_plan::ParquetSource;
 use datafusion::datasource::physical_plan::parquet::CachedParquetFileReaderFactory;
 use datafusion::datasource::physical_plan::parquet::metadata::{
     DFParquetMetadata, ordering_from_parquet_metadata,
 };
+use datafusion::parquet::arrow::parquet_to_arrow_schema;
 use datafusion_common::config::TableParquetOptions;
@@ -33,6 +34,29 @@ fn fail_for_encryption_factory(options: &TableParquetOptions) -> Result<()> {
         Ok(())
     }
 }

+/// Parquet has no seconds timestamp logical type. Arrow writers encode an Arrow
+/// `timestamp[s]` as an unannotated physical `INT64` and preserve the Arrow type only
+/// in `ARROW:schema`; Spark ignores that Arrow-specific hint and exposes the column as long.
+fn reinterpret_second_timestamps_as_int64(schema: Schema, physical: &Schema) -> Schema {
+    let fields = schema
+        .fields()
+        .iter()
+        .map(|field| {
+            let physical_type = physical
+                .field_with_name(field.name())
+                .ok()
+                .map(|field| field.data_type());
+            match (field.data_type(), physical_type) {
+                (DataType::Timestamp(TimeUnit::Second, _), Some(DataType::Int64)) => {
+                    field.as_ref().clone().with_data_type(DataType::Int64)
+                }
+                _ => field.as_ref().clone(),
+            }
+        })
+        .collect::<Vec<_>>();
+    Schema::new_with_metadata(fields, schema.metadata().clone())
+}
+
 #[async_trait::async_trait]
 impl ReadFormat for ParquetReadFormat {
@@ -72,10 +96,13 @@ impl ReadFormat for ParquetReadFormat {
         let mut schemas: Vec<(object_store::path::Path, Schema)> = futures::stream::iter(objects)
             .map(|(store, object)| async {
-                let schema = DFParquetMetadata::new(store.as_ref(), object)
+                let reader = DFParquetMetadata::new(store.as_ref(), object)
                     .with_metadata_size_hint(metadata_size_hint)
                     .with_file_metadata_cache(Some(Arc::clone(&metadata_cache)))
-                    .with_coerce_int96(coerce_int96)
-                    .fetch_schema()
-                    .await?;
+                    .with_coerce_int96(coerce_int96);
+                let schema = reader.fetch_schema().await?;
+                let metadata = reader.fetch_metadata().await?;
+                let file_metadata = metadata.file_metadata();
+                let physical = parquet_to_arrow_schema(file_metadata.schema_descr(), None)?;
+                let schema = reinterpret_second_timestamps_as_int64(schema, &physical);
                 Ok::<_, DataFusionError>((object.location.clone(), schema))
             })
```

Exact test diff:

```diff
diff --git a/python/pysail/tests/spark/datasource/test_parquet.py b/python/pysail/tests/spark/datasource/test_parquet.py
--- a/python/pysail/tests/spark/datasource/test_parquet.py
+++ b/python/pysail/tests/spark/datasource/test_parquet.py
@@ -1,3 +1,4 @@
+import base64
 from collections.abc import Mapping
 from datetime import UTC, datetime

@@ -168,6 +169,25 @@ def test_parquet_explicit_schema_allows_missing_fields(spark, tmp_path):

     assert rows == [Row(id=1, missing=None)]


+def test_parquet_arrow_second_timestamp_is_bigint(spark, tmp_path):
+    path = tmp_path / "arrow_second_timestamp.parquet"
+    encoded_arrow_schema = base64.b64encode(
+        pa.schema([pa.field("ts", pa.timestamp("s"))]).serialize().to_pybytes()
+    ).decode()
+    table = pa.table({"ts": pa.array([1704164645, -1], type=pa.int64())})
+    with pq.ParquetWriter(path, table.schema, store_schema=False) as writer:
+        writer.add_key_value_metadata({"ARROW:schema": encoded_arrow_schema})
+        writer.write_table(table)
+
+    df = spark.read.parquet(str(path))
+
+    assert df.schema.simpleString() == "struct<ts:bigint>"
+    assert df.orderBy("ts").collect() == [
+        Row(ts=-1),
+        Row(ts=1704164645),
+    ]
+
+
 @pytest.mark.parametrize("requested_type", ["TIMESTAMP", "TIMESTAMP_NTZ"])
 def test_parquet_explicit_schema_determines_int96_timestamp_family(spark, tmp_path, requested_type):
```

Why contained: this changes only the PR’s new top-level Parquet-seconds behavior. Comparing with the physical schema avoids changing genuine INT96 timestamps or general schema inference.

### 2. High (P1): Mixed-unit adjusted timestamps still fail when Arrow timezone labels differ

New-side location: [`crates/sail-data-source/src/listing/utils.rs:63`](https://github.com/lakehq/sail/blob/09290c49b31c70f24cd29efb303348b93cae33fc/crates/sail-data-source/src/listing/utils.rs#L63)

Exact review comment:

> The new pre-merge pass widens timestamp units but preserves Arrow’s producer-specific timezone labels. A directory containing `timestamp[ms, UTC]` and `timestamp[us, America/New_York]` therefore reaches `Schema::try_merge` as two unequal Arrow types and remains unreadable. Spark maps both adjusted-to-UTC physical timestamps to one `TimestampType`; Arrow also defines changing one nonempty timezone label to another as metadata-only. Please retag nonempty labels with the Spark session timezone for fields participating in this new millisecond widening, while preserving `None` versus `Some` so `TIMESTAMP_NTZ` remains distinct. Limiting this to fields containing a millisecond input avoids changing the pre-existing microsecond-only path.

Spark/source evidence:

- Spark maps both adjusted `MILLIS` and `MICROS` to the same `TimestampType`: [ParquetSchemaConverter.scala:305](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:305).
- The vectorized reader widens milliseconds while preserving the instant: [ParquetVectorUpdaterFactory.java:148](//sail/opt/spark/sql/core/src/main/java/org/apache/spark/sql/execution/datasources/parquet/ParquetVectorUpdaterFactory.java:148).
- Arrow specifies that every nonempty timezone uses the UTC epoch and nonempty-to-nonempty relabeling is metadata-only: [arrow-schema datatype.rs:135](/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-schema-59.2.0/src/datatype.rs:135).

Reproduction with the Spark session timezone set to `America/Los_Angeles`:

```text
Files:
a_us_new_york.parquet: timestamp[us, America/New_York]
b_ms_utc.parquet:      timestamp[ms, UTC]

Spark 4.1.1:
schema = struct<ts:timestamp>
rows   = [
  2024-01-01 19:04:05.123,
  2024-01-01 19:04:06.123456
]

Sail at 09290c49:
Schema error: Fail to merge schema field 'ts' because the from data_type =
Timestamp(µs, "America/New_York") does not equal Timestamp(µs, "UTC")
```

Exact code diff:

```diff
diff --git a/crates/sail-data-source/src/formats/parquet/read.rs b/crates/sail-data-source/src/formats/parquet/read.rs
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -84,8 +84,16 @@ impl ReadFormat for ParquetReadFormat {
             .buffer_unordered(metadata_fetch_concurrency)
             .try_collect()
             .await?;

+        let session_timezone = ctx
+            .config_options()
+            .execution
+            .time_zone
+            .as_deref()
+            .unwrap_or("UTC");
+        normalize_widened_timestamp_timezones(&mut schemas, session_timezone);
+
         // Ensure deterministic ordering for stable schema inference.
         schemas.sort_unstable_by(|(location1, _), (location2, _)| location1.cmp(location2));

@@ -317,6 +325,46 @@ fn is_binary_type(data_type: &DataType) -> bool {
             | DataType::FixedSizeBinary(_)
     )
 }

+/// For fields participating in millisecond widening, replace producer-specific Arrow
+/// timezone identifiers with the Spark session timezone while preserving `None`.
+fn normalize_widened_timestamp_timezones(
+    schemas: &mut [(object_store::path::Path, Schema)],
+    session_timezone: &str,
+) {
+    let widened_timestamp_fields = schemas
+        .iter()
+        .flat_map(|(_, schema)| schema.fields().iter())
+        .filter_map(|field| match field.data_type() {
+            DataType::Timestamp(TimeUnit::Millisecond, Some(_)) => {
+                Some(field.name().clone())
+            }
+            _ => None,
+        })
+        .collect::<Vec<_>>();
+
+    for (_, schema) in schemas {
+        let metadata = schema.metadata().clone();
+        let fields = schema
+            .fields()
+            .iter()
+            .map(|field| match field.data_type() {
+                DataType::Timestamp(
+                    unit @ (TimeUnit::Millisecond | TimeUnit::Microsecond),
+                    Some(_),
+                ) if widened_timestamp_fields.contains(field.name()) => field
+                    .as_ref()
+                    .clone()
+                    .with_data_type(DataType::Timestamp(
+                        *unit,
+                        Some(session_timezone.into()),
+                    )),
+                _ => field.as_ref().clone(),
+            })
+            .collect::<Vec<_>>();
+        *schema = Schema::new_with_metadata(fields, metadata);
+    }
+}
+
 /// Clears all metadata (Schema level and field level) for a schema.
 fn clear_metadata(schema: Schema) -> Schema {
```

Exact test diff:

```diff
diff --git a/python/pysail/tests/spark/datasource/test_parquet.py b/python/pysail/tests/spark/datasource/test_parquet.py
--- a/python/pysail/tests/spark/datasource/test_parquet.py
+++ b/python/pysail/tests/spark/datasource/test_parquet.py
@@ -15,6 +15,37 @@ def safe_sort_key(row):
         return tuple((v is not None, v) for _, v in sorted(row.items()))
     return tuple((v is not None, v) for v in row)


+@pytest.mark.parametrize("session_timezone", ["America/Los_Angeles"], indirect=True)
+def test_parquet_merges_mixed_timestamp_units_with_different_arrow_timezones(
+    spark, tmp_path, session_timezone
+):
+    _ = session_timezone
+    path = tmp_path / "mixed_timestamp_units_and_timezones"
+    path.mkdir()
+    files = [
+        ("a_us_new_york.parquet", "us", "America/New_York", datetime(2024, 1, 2, 3, 4, 6, 123456, tzinfo=UTC)),
+        ("b_ms_utc.parquet", "ms", "UTC", datetime(2024, 1, 2, 3, 4, 5, 123000, tzinfo=UTC)),
+    ]
+    for name, unit, timezone, value in files:
+        pq.write_table(
+            pa.table({"ts": pa.array([value], type=pa.timestamp(unit, tz=timezone))}),
+            path / name,
+        )
+
+    df = spark.read.parquet(str(path))
+    rows = (
+        df.selectExpr("CAST(ts AS STRING) AS value")
+        .orderBy("value")
+        .collect()
+    )
+
+    assert df.schema.simpleString() == "struct<ts:timestamp>"
+    assert rows == [
+        Row(value="2024-01-01 19:04:05.123"),
+        Row(value="2024-01-01 19:04:06.123456"),
+    ]
+
+
 def test_parquet_read_write_basic(spark, sample_df, tmp_path):
```

Why contained: the retagging applies only when at least one top-level file schema has the newly supported aware millisecond type. Microsecond-only fields and the `TimestampType`/`TimestampNTZType` distinction remain unchanged.

## Deferred parity debt

These receive no severity and no paste-ready review comment.

### Nested timestamp leaves

The PR explicitly limits normalization to top-level fields. Spark recursively converts timestamp leaves through structs, arrays, and maps, as shown in [ParquetSchemaConverter.scala:168](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:168), [line 420](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:420), and [line 460](//sail/opt/spark/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetSchemaConverter.scala:460).

Exact-head Sail failures:

```text
Single nested millisecond file:
UnsupportedOperationException: cast Timestamp(Millisecond, None) to Spark data type

Mixed nested ms/us directory:
Schema error: Fail to merge schema field 'inner' because Timestamp(µs) does not equal Timestamp(ms)
```

### Pre-existing microsecond-only session-timezone inference

For a microsecond-only file carrying the producer label `UTC`, Spark renders the instant in the configured session timezone. Sail retains the producer label. This predates the PR and is therefore deferred rather than folded into Finding 2.

Proposed TODO/FIXME diff:

```diff
diff --git a/crates/sail-data-source/src/formats/parquet/read.rs b/crates/sail-data-source/src/formats/parquet/read.rs
--- a/crates/sail-data-source/src/formats/parquet/read.rs
+++ b/crates/sail-data-source/src/formats/parquet/read.rs
@@ -88,6 +88,8 @@ impl ReadFormat for ParquetReadFormat {
         // Ensure deterministic ordering for stable schema inference.
         schemas.sort_unstable_by(|(location1, _), (location2, _)| location1.cmp(location2));

+        // TODO: Retag inferred microsecond-only adjusted-to-UTC timestamps with Spark's
+        // session timezone instead of retaining producer `ARROW:schema` timezone identifiers.
         let schemas = schemas.into_iter().map(|(_, schema)| schema);

         let merged = if options.global.skip_metadata {
diff --git a/crates/sail-data-source/src/listing/utils.rs b/crates/sail-data-source/src/listing/utils.rs
--- a/crates/sail-data-source/src/listing/utils.rs
+++ b/crates/sail-data-source/src/listing/utils.rs
@@ -48,6 +48,8 @@ pub fn try_merge_normalized(schemas: impl IntoIterator<Item = Schema>) -> Result<
 fn normalize_unsupported_fields(schema: &Schema) -> Schema {
+    // TODO: Apply Spark-compatible type normalization recursively inside structs, lists,
+    // and maps. Nested millisecond timestamps currently remain unsupported by Sail.
     let new_fields: Vec<Field> = schema
         .fields()
         .iter()
```

Proposed strict Sail-only expected-failure tests:

```diff
diff --git a/python/pysail/tests/spark/datasource/test_parquet.py b/python/pysail/tests/spark/datasource/test_parquet.py
--- a/python/pysail/tests/spark/datasource/test_parquet.py
+++ b/python/pysail/tests/spark/datasource/test_parquet.py
@@ -8,6 +8,7 @@ import pytest
 from pandas.testing import assert_frame_equal
 from pyspark.sql import Row

+from pysail.testing.spark.utils.common import is_jvm_spark
 from pysail.testing.spark.utils.files import get_data_directory_size
 from pysail.testing.spark.utils.sql import escape_sql_identifier, escape_sql_string_literal

@@ -191,6 +192,116 @@ def test_parquet_explicit_schema_determines_int96_timestamp_family(spark, tmp_pa
     )

     assert rows == [Row(value="2024-01-02 03:04:05")]


+@pytest.mark.xfail(
+    not is_jvm_spark(),
+    reason="Sail does not normalize nested millisecond Parquet timestamps",
+    strict=True,
+)
+@pytest.mark.parametrize(
+    ("case", "value", "arrow_type", "expected_type"),
+    [
+        pytest.param(
+            "struct",
+            {"inner": datetime(2024, 1, 2, 3, 4, 5, 123000)},
+            pa.struct([pa.field("inner", pa.timestamp("ms"))]),
+            "struct<inner:timestamp_ntz>",
+            id="struct",
+        ),
+        pytest.param(
+            "array",
+            [datetime(2024, 1, 2, 3, 4, 5, 123000)],
+            pa.list_(pa.timestamp("ms")),
+            "array<timestamp_ntz>",
+            id="array",
+        ),
+        pytest.param(
+            "map",
+            [("key", datetime(2024, 1, 2, 3, 4, 5, 123000))],
+            pa.map_(pa.string(), pa.timestamp("ms")),
+            "map<string,timestamp_ntz>",
+            id="map",
+        ),
+    ],
+)
+def test_parquet_reads_nested_millisecond_timestamp(
+    spark, tmp_path, case, value, arrow_type, expected_type
+):
+    path = tmp_path / f"nested_millisecond_{case}.parquet"
+    table = pa.table({"value": pa.array([value], type=arrow_type)})
+    pq.write_table(table, path)
+
+    df = spark.read.parquet(str(path))
+
+    assert df.schema.simpleString() == f"struct<value:{expected_type}>"
+    assert df.count() == 1
+
+
+@pytest.mark.xfail(
+    not is_jvm_spark(),
+    reason="Sail does not normalize nested millisecond timestamps before schema merging",
+    strict=True,
+)
+def test_parquet_merges_nested_millisecond_and_microsecond_timestamps(spark, tmp_path):
+    path = tmp_path / "mixed_nested_timestamps"
+    path.mkdir()
+    ms_value = datetime(1969, 12, 31, 23, 59, 59, 999000)
+    us_value = datetime(2024, 1, 2, 3, 4, 5, 123456)
+
+    for name, unit, value in [
+        ("a_ms.parquet", "ms", ms_value),
+        ("b_us.parquet", "us", us_value),
+    ]:
+        arrow_type = pa.struct([pa.field("inner", pa.timestamp(unit))])
+        table = pa.table(
+            {"value": pa.array([{"inner": value}], type=arrow_type)}
+        )
+        pq.write_table(table, path / name)
+
+    df = spark.read.parquet(str(path))
+    rows = df.select("value.inner").orderBy("inner").collect()
+
+    assert df.schema.simpleString() == (
+        "struct<value:struct<inner:timestamp_ntz>>"
+    )
+    assert rows == [
+        Row(inner=ms_value),
+        Row(inner=us_value),
+    ]
+
+
+@pytest.mark.xfail(
+    not is_jvm_spark(),
+    reason="Sail retains the producer timezone on inferred microsecond Parquet timestamps",
+    strict=True,
+)
+@pytest.mark.parametrize("session_timezone", ["America/Los_Angeles"], indirect=True)
+def test_parquet_inferred_microsecond_timestamp_uses_session_timezone(
+    spark, tmp_path, session_timezone
+):
+    _ = session_timezone
+    path = tmp_path / "microsecond_timestamp_session_timezone.parquet"
+    value = datetime(2024, 1, 2, 3, 4, 5, 123456, tzinfo=UTC)
+    table = pa.table(
+        {
+            "ts": pa.array(
+                [value],
+                type=pa.timestamp("us", tz="UTC"),
+            )
+        }
+    )
+    pq.write_table(table, path)
+
+    row = (
+        spark.read
+        .parquet(str(path))
+        .selectExpr(
+            "CAST(ts AS STRING) AS ts",
+            "hour(ts) AS hour",
+        )
+        .first()
+    )
+
+    assert row == Row(ts="2024-01-01 19:04:05.123456", hour=19)
+
+
 def test_parquet_write_with_bloom_filter(spark, tmpdir):
```

The Python tests are appropriate here because they must construct specific physical Parquet schemas and `ARROW:schema` metadata.

---

crates/sail-data-source/src/listing/utils.rs
james-willis: I think we should actually remove Utf8View. I dont think this solves a problem any more. Will give details in a bit.

james-willis:
I applied Fix 1, really nice catch!

Working on the rest.

crates/sail-data-source/src/listing/utils.rs
james-willis: I can't get a view type to reach the client under any combination of `expand_views_at_output` × `schema_force_view_types` — `df.toArrow()` comes back plain binary/string every time, and BinaryView already arrives non‑view with no rewrite at all.

So I'd hold off on adding `BinaryView` here — and the same evidence says the existing `Utf8View` → `Utf8` branch is redundant and worth removing rather than mirroring.

The downcast also isn't free: it duplicates the buffer bytes on every join output (an interleave copy per value) instead of sharing them, so leaning on the output normalization is the cheaper path.

shehabgamin: @james-willis We def need to keep them, here's a reproducible example that i had Codex generate
```
"""Show why DataFrame.toArrow() does not reveal the received Arrow types.

Tested with the available Sail 0.7.1 native build, PySpark 4.2.0, and PyArrow 24.0.0.
Run each configuration in a fresh process:

    python repro_arrow_views.py false
    python repro_arrow_views.py true

The first run must receive both view types and reject collect(); the second
must receive conventional types and collect successfully. No Sail changes or
mocks are used. This tests client compatibility, not where to place conversion.
"""

import os
import sys
import tempfile
from pathlib import Path

# Sail configuration must be set before importing its native module.
expand_views = sys.argv[1] if len(sys.argv) > 1 else "false"
if expand_views not in ("true", "false"):
    raise SystemExit("Usage: python repro_arrow_views.py [false|true]")
os.environ["SAIL_OPTIMIZER__EXPAND_VIEWS_AT_OUTPUT"] = expand_views

import pyarrow as pa
import pyarrow.parquet as pq
import pysail
import pyspark
from pyspark.errors import PySparkTypeError

print(f"Sail {pysail.__version__}; PySpark {pyspark.__version__}; PyArrow {pa.__version__}")
print(f"expand_views_at_output={expand_views}")

server = SparkConnectServer()
server.start()
spark = None
try:
    with tempfile.TemporaryDirectory(prefix="sail-arrow-views-") as directory:
        path = str(Path(directory) / "views.parquet")
        # The existing inferred-schema rewrite handles only top-level Utf8View.
        # Nesting retains both view types; selecting a child exposes each one
        # independently, so neither failure can hide the other.
        payload = {"text": "a string longer than twelve bytes", "bytes": b"binary data longer than twelve bytes"}
        nested_type = pa.struct([("text", pa.string_view()), ("bytes", pa.binary_view())])
        pq.write_table(pa.table({"nested": pa.array([payload, None], nested_type)}), path)
        df = spark.read.parquet(path)

        for name, view_type, plain_type in (
            ("text", pa.string_view(), pa.string()),
            ("bytes", pa.binary_view(), pa.binary()),
        ):
            column = df.selectExpr(f"nested.{name} AS value")
            # Diagnostic private API: examine the received table before toArrow's cast.
            raw, _ = column._to_table()
            raw_type = raw.schema.field("value").type
            public = column.toArrow()
            public_type = public.schema.field("value").type
            print(f"{name}: received={raw_type}; toArrow()={public_type}")
            assert public_type == plain_type
            assert public.column("value").to_pylist() == [payload[name], None]

            if expand_views == "false":
                assert raw_type == view_type, f"Expected {view_type}, received {raw_type}"
                try:
                    column.collect()
                except PySparkTypeError as error:
                    assert "UNSUPPORTED_DATA_TYPE_FOR_ARROW_CONVERSION" in str(error)
                    assert str(view_type) in str(error)
                    print(f"  collect(): {error}")
                else:
                    raise AssertionError(f"collect() unexpectedly accepted {view_type}")
            else:
                assert raw_type != view_type
                assert [row.value for row in column.collect()] == [payload[name], None]
                print("  collect(): succeeds")
finally:
    try:
        if spark is not None:
            spark.stop()
    finally:
        server.stop()
```

james-willis: ok. I have added binary view plus two test cases:

1. collect, similar to what is above
2. `INSERT INTO` is another case that may trigger the issue

crates/sail-spark-connect/src/service/config_manager.rs
```diff
@@ -29,7 +30,24 @@ pub(crate) fn handle_config_set(
...
+    let sets_session_timezone = {
+        let key = SparkConfigKey::SPARK_SQL_SESSION_TIME_ZONE.to_string();
+        kv.iter().any(|pair| pair.key == key)
+    };
     spark.set_config(kv)?;
+    if sets_session_timezone {
+        // Mirror `spark.sql.session.timeZone` onto the DataFusion session configuration. Parquet
+        // listing-table schema inference reads `execution.time_zone` to reconcile mixed-timezone
+        // timestamps and to render adjusted timestamps in the session zone; Sail otherwise keeps
+        // the session timezone only in `PlanConfig`, leaving `execution.time_zone` unset.
```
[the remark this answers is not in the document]
james-willis: fixed

crates/sail-data-source/src/formats/parquet/read.rs
```diff
@@ -70,12 +99,18 @@ impl ReadFormat for ParquetReadFormat {
...
                     .with_file_metadata_cache(Some(Arc::clone(&metadata_cache)))
-                    .with_coerce_int96(coerce_int96)
-                    .fetch_schema()
-                    .await?;
+                    .with_coerce_int96(coerce_int96);
+                let schema = reader.fetch_schema().await?;
+                // Compare the Arrow-hinted schema against the metadata-free physical schema so
+                // that Arrow-only `timestamp[s]` hints over physical INT64 are read as long,
+                // matching Spark. The metadata cache above makes this second fetch cheap.
+                let metadata = reader.fetch_metadata().await?;
```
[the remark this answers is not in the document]
james-willis: fixed

linhr:
@james-willis Thanks for the contribution! This PR looks good. Would you like to resolve the merge conflict so that we can merge it and include it in the upcoming release (planned this weekend)?

james-willis:
@linhr done!

PR #2235, opened 2026-07-11
perf: title tbd

shehabgamin:
Also address Clickbench misalignment: https://github.com/shehabgamin/ClickBench/tree/sail-bump

Issue #1916, opened 2026-05-16
Memory issue


kannankvsp:
spark.stop() is not releasing memory to os. i need to kill sail server  each time to get the memory back. 

Also it seems it taking more memory than apache spark.

linhr:
Thanks for reaching out. Could you provide a minimum reproducible example?

kannankvsp:
```from __future__ import annotations

import time
from pathlib import Path

from pyspark.sql import DataFrame, SparkSession
from pyspark.sql import functions as F


def timed(label: str, fn):
    start = time.perf_counter()
    result = fn()
    duration = time.perf_counter() - start
    print(f"[TIMER] {label}: {duration:.2f}s")
    return result


def build_spark(app_name: str, shuffle_partitions: int) -> SparkSession:
    spark = (
        SparkSession.builder.remote("sc://localhost:50051")
        .config("spark.sql.files.maxPartitionBytes", str(128 * 1024 * 1024))
        .config("spark.sql.adaptive.enabled", "true")
        .config("spark.sql.adaptive.coalescePartitions.enabled", "true")
        .config("spark.sql.adaptive.advisoryPartitionSizeInBytes", str(128 * 1024 * 1024))
        .getOrCreate()
    )
    return spark


def normalize_columns(df: DataFrame) -> DataFrame:
    # Keep common ecommerce columns when present.
    preferred = [
        "event_time",
        "event_type",
        "product_id",
        "category_id",
        "category_code",
        "brand",
        "price",
        "user_id",
        "user_session",
    ]
    available = [c for c in preferred if c in df.columns]
    if available:
        df = df.select(*available)

    if "event_time" in df.columns:
        # Strip trailing UTC literal before strict timestamp parsing.
        event_time_clean = F.regexp_replace(F.col("event_time"), r"\s+UTC$", "")
        df = df.withColumn("event_ts", F.to_timestamp(event_time_clean, "yyyy-MM-dd HH:mm:ss"))
        df = df.withColumn("event_ts", F.coalesce(F.col("event_ts"), F.to_timestamp("event_time")))
    else:
        df = df.withColumn("event_ts", F.current_timestamp())

    if "price" in df.columns:
        df = df.withColumn("price", F.col("price").cast("double"))
    else:
        df = df.withColumn("price", F.lit(0.0))

    if "event_type" not in df.columns:
        df = df.withColumn("event_type", F.lit("unknown"))
    if "brand" not in df.columns:
[...]

PR #2545, opened 2026-09-03
feat: use mimalloc as the Python wheel's global allocator

james-willis:
> [!NOTE]
> This PR carves the Python-wheel allocator change out of #2235 (@shehabgamin) so it can land on its own, independently of the perf/ClickBench work still in flight there. The approach here — including the mimalloc `v2` pin — follows that PR.

### The problem

Sail sets a mimalloc `#[global_allocator]` in `sail-cli`, but a `#[global_allocator]` only takes effect in the crate that roots the *final artifact*. That static lives in `sail-cli`'s **binary** target (`main.rs`), which isn't linked into the `sail-python` **cdylib**. So the `pysail` wheel silently falls back to the **system allocator (glibc)**.

glibc keeps freed chunks on per-arena free lists and doesn't return them to the OS without an explicit `malloc_trim`. In a long-lived Spark Connect server, that reads as memory that never comes back between queries — matching the report in #1916 (*"spark.stop() is not releasing memory to os… taking more memory than apache spark"*).

On workloads that run several memory-intensive queries back to back in one long-lived server, that retention can accumulate until the process is **OOM-killed**, even though no single query's peak comes close to the limit. The work around is making explicit `malloc_trim` calls between queries which most users wont be savvy enough for.

### SedonaDB vs Sail

[SedonaDB](https://github.com/apache/sedona-db) is the same shape — a Rust engine shipped as a pyo3 cdylib wheel — and it sets its mimalloc `#[global_allocator]` in **both** roots (its CLI *and* `python/sedonadb/src/lib.rs`). Because Sail set it only in the CLI, two otherwise-similar engines behaved very differently on the same long-running workload: SedonaDB's memory plateaus across queries; Sail's climbs. This PR closes that gap by giving the wheel the allocator the CLI already has.

### The change

- Add the mimalloc `#[global_allocator]` to the `sail-python` cdylib root (+ the `mimalloc` feature, enabled via the maturin build).
- Pin mimalloc to **v2**. The crate currently defaults to v3.3.2, which segfaults under multi-threaded aligned allocation (microsoft/mimalloc#1287) and SIGABRTs on macOS when statically linked into a second shared library (microsoft/mimalloc#1301) — both fatal for a threaded cdylib. `local_dynamic_tls` selects the local-dynamic TLS model so the static loads cleanly from the extension module. Once a `libmimalloc-sys > 0.1.49` vendors the 1287 and 1301 fixes, the `v2` pin can be dropped.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8114,6 +8114,7 @@ dependencies = [
  "fastrace",
  "futures",
  "log",
+ "mimalloc",
  "pyo3",
  "sail-catalog",
  "sail-catalog-glue",
--- a/crates/sail-cli/Cargo.toml
+++ b/crates/sail-cli/Cargo.toml
@@ -22,7 +22,10 @@ sail-telemetry = { path = "../sail-telemetry" }
 sail-spark-connect = { path = "../sail-spark-connect" }
 sail-session = { path = "../sail-session" }
 
-mimalloc = { workspace = true, optional = true }
+# TODO: Drop "v2" after:
+#       - libmimalloc-sys vendors the fix for https://github.com/microsoft/mimalloc/issues/1287 and
+#       - exposes MI_TLS_MODEL_LOCAL for macOS coexistence https://github.com/microsoft/mimalloc/issues/1301
+mimalloc = { workspace = true, optional = true, features = ["v2"] }
 clap = { workspace = true }
 tokio = { workspace = true }
 log = { workspace = true }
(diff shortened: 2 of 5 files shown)


shehabgamin:
@james-willis We've been blocked by https://github.com/purpleprotocol/mimalloc_rust/pull/169

At this point, I think we should maybe just fork the repo so that we don't have to downgrade to v2

linhr:
Thanks for raising this issue! I'm going to move this PR to draft as we need to decide what's the best approach here given that the mimalloc binding is not actively maintained.

Issue #2561, opened 2026-09-08
[draft] CoalesceExec over Delta MERGE writer input stalls non-zero output partitions


zemin-piao:
## Summary

`CoalesceExec` placed directly under `DeltaWriterExec` stalls every output partition except partition 0 during Delta `MERGE`. The writer's partition 0 runs to completion; partitions 1..n-1 enter their read loop and never receive a batch nor finish, so the statement never returns.

**This is not reachable on `main` today.** It was found while experimenting with a cap on Delta writer parallelism, and it is the reason that cap was abandoned. Reproducing it requires the small experimental patch below, which makes the Delta write path emit `ExplicitRepartitionExec(UnknownPartitioning(4))` in front of the writer so that `RewriteExplicitRepartition` lowers it to `CoalesceExec`.

Filing it because the stall looks like a scheduling/planner issue in the interaction between `CoalesceExec`'s sequential per-range draining and Delta merge plan shapes, rather than anything specific to the abandoned cap.

## Experimental patch

Applies to `crates/sail-delta-lake/src/physical_plan/planner/utils.rs`, in `prepare_delta_writer_input`:

```rust
    // EXPERIMENT: cap writer parallelism at 4 and let `RewriteExplicitRepartition` resolve it
    // against the optimized input partition count.
    let cap = |plan: Arc<dyn ExecutionPlan>| -> Arc<dyn ExecutionPlan> {
        Arc::new(
            sail_physical_plan::repartition::ExplicitRepartitionExec::new(
                plan,
                Partitioning::UnknownPartitioning(4),
            ),
        )
    };

    if partition_columns.is_empty() && sort_order.is_none() {
        return Ok(cap(input));
    }
    let plan = cap(create_projection(input, partition_columns.to_vec())?);
    Ok(create_sort(plan, partition_columns.to_vec(), sort_order)?)
```

`Partitioning` is already imported in that module; no other change is needed.

## Minimal reproduction

With `SAIL_EXECUTION__DEFAULT_PARALLELISM=4`:

```python
import tempfile
from pathlib import Path

from pysail.testing.spark.session import spark_connect_server, spark_session_factory

with spark_connect_server() as server, spark_session_factory(server.remote) as sessions:
    spark = sessions.create()
    root = Path(tempfile.mkdtemp())
    spark.sql(f"CREATE TABLE tgt (id INT, event_time TIMESTAMP) USING DELTA LOCATION '{root / 'tgt'}'")
[...]

PR #2562, opened 2026-09-08
fix: let Delta writes inherit input parallelism

zemin-piao:
## Summary

Delta writes were sharded by table partition key: `DeltaWriterExec` required a `KeyPartitioned` input distribution on the partition columns, so all rows for a given table partition were routed to one writer task. A write whose scan produced 30 partitions still routed all rows for each date through one writer task, serializing that table partition's output.

Writer parallelism now follows the optimized input plan:

- the hash-by-table-partition distribution requirement is removed from `DeltaWriterExec`;
- `DeltaWriterExec` reports that it does not benefit from input partitioning, so no synthetic repartitioning is requested on its behalf;
- all write paths share one `prepare_delta_writer_input` helper — `build_standard_write_layers`, both `op_write` paths, the COW DELETE path, and both writer inputs in `build_row_level_rewrite_plan` (which covers DELETE, UPDATE and MERGE) — so the grouping guarantee is defined in a single place;
- unpartitioned writes with no user sort order no longer get an identity projection and a sort on a literal.

Correctness is unchanged. `DeltaWriter` keeps at most one open file per task and requires its input to be grouped by partition key; that guarantee comes from the task-local `SortExec` (`preserve_partitioning=true`), not from the distribution requirement. Data file names embed a per-writer UUID, so concurrent writers in the same directory cannot collide.

## Behavior changes

Delta writers no longer request synthetic repartitioning. Writer parallelism follows the optimized input plan and is not capped by `execution.default_parallelism`.

Multiple writer tasks may write files into the same Delta table partition. Files per table partition can reach the number of input tasks touching that partition.

A single-partition source may still be parallelized by upstream operators that benefit from partitioning; the writer itself neither adds nor removes that parallelism.

The clearest illustration is in the regenerated `update.yaml` snapshots: five UPDATE plans previously had `RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=1` directly beneath the writer, i.e. a single input partition fanned out into four writer tasks purely because the writer advertised that it benefited from partitioning. Those nodes are gone.

## Follow-up

Bounding the writer task count at `target_partitions` is deliberately not part of this change. The obvious implementation — `ExplicitRepartitionExec(UnknownPartitioning(target))` resolved late by `RewriteExplicitRepartition` — deadlocks Delta MERGE deterministically; reported with a reproduction patch in #2561.

## Test plan

- [x] `cargo test -p sail-delta-lake --lib`: 288 passed
- [x] `cargo clippy -p sail-delta-lake --all-targets`: clean
- [x] `cargo +nightly fmt --check`: clean
- [x] `pytest python/pysail/tests/spark/delta python/pysail/tests/spark/execution`: 531 passed, 4 skipped, 2 xfailed — run 10 consecutive times with no variation and no snapshot drift
- [x] `operationMetrics` snapshots unchanged, including all file- and byte-count assertions
- [x] Regenerated plan snapshots: `delta/delete.yaml`, `delta/merge.yaml`, `delta/update.yaml`, `execution/lakehouse_commit.yaml` (the last one is the `distributed_delta_commit` scenario, so the change is covered in local-cluster mode)
- [x] New regression test: four input partitions writing into one table partition produce four files
- [x] Throughput sanity check: 5M rows completed in 2.26s from a single-partition source and 1.74s from a four-partition source, producing four files in both cases


lonless9 (approved):
Love it! Thanks!

PR #2572, opened 2026-09-12
feat: explore Lance table format compatibility

jonasdedden:
Lance (https://lance.org) is a columnar format for multimodal and AI data, written in Rust. This registers it as a Sail table format, so `spark.read.format("lance")`, `df.write.format("lance")` and `CREATE TABLE ... USING lance` work without the JVM connector Spark needs.

`crates/sail-lance` implements `TableFormat` for Lance:

* reads with projection, filters, limit and vector search pushed into the Lance scanner, one output partition per fragment group, and time travel by version or tag;
* writes for all four Spark save modes, streamed into a single Lance write transaction.

Filters are pushed down as DataFusion expressions rather than rendered to SQL, so Lance evaluates the same expression tree Sail built and the pushdown is exact.

Three changes outside the crate make this build. Lance needs DataFusion's `sql` feature, which Sail leaves off, and a Cargo build has one feature set per crate:

* the workspace manifest turns the feature on
* `sail-common-datafusion` handles the `DataFusionError::SQL` variant the feature adds
* `sail-plan` builds the Spark wildcard options from `sqlparser` types rather than from strings, which is what DataFusion's stand-ins are when the feature is off

This branch is based on v0.7.1, the release built against the DataFusion and Arrow versions Lance uses.


jonasdedden:
Closed since the approach of a Rust plugin is introducing too many disadvantages into the Sail codebase.

PR #2576, opened 2026-09-12
fix: handle URI-encoded file paths in Delta MERGE and UPDATE

### LINKED ISSUE #2570: MERGE INTO ... WHEN MATCHED THEN UPDATE duplicates the entire file containing a match, not just the matched rows
## Summary

After the fix in #2478 (target-side pruning), `MERGE INTO ... WHEN MATCHED THEN UPDATE` now scans the right partition, but the write side has a separate bug: it duplicates the **entire physical file** containing a matched row, rather than updating matched rows in place. This happens regardless of `ON` clause shape or match key, and can silently affect far more rows than the ones actually being updated.

## Evidence

Delta table partitioned by a timestamp column. Updated a batch of pre-existing rows (single-column change, matched on a unique id column), confirmed present in the target beforehand.

- A compound `ON` (static partition predicate + join, as a pruning-aware upsert would emit) and a **plain join-only `ON`** (no static predicate at all) both fail **identically**: row count grows by exactly the row count of the one physical file the updated rows happened to live in — not by the number of rows updated, and not by the full partition. This rules out the pruning predicate as the cause.
- Matching on a **non-partition** column (one-to-many, matches scattered across many different partition files) fails the same way, at much larger scale: updating a small number of matching rows inflated the table by orders of magnitude more rows than were actually touched, because every file containing any match gets fully duplicated — including unrelated rows that happened to share that file.
- No error or warning in any case.

## What we checked before filing

- #2478 — confirmed fixed target-side pruning; this is a distinct, write-side issue that surfaced after that fix
- #1472 — earlier acknowledged file-touching gap in UPDATE/DELETE; this looks related but is specifically about MERGE's write path duplicating rather than replacing

## Why this looks like a write-side bug specifically

Pruning already works correctly now (per #2478) — the scan reaches the right file. The row count growing by exactly one file's worth, regardless of `ON` clause or match key, points at the commit/rewrite step: it looks like the touched file is being added as a new file rather than replacing the original, so the old file's `remove` action either doesn't happen or doesn't happen for the right file.

## Question for maintainers

Is this in scope for #2419 (the MERGE/UPDATE/DELETE row-operations unification), or a separate fix?

lonless9:
Fixes #2570


KabirBhutti:
Thank you for resolving this issue. Could you please advise which upcoming release this fix is targeted for, or provide an approximate timeline for the next pysail release?

lonless9:
@KabirBhutti Thanks for the report! This fix will be released soon in 0.7.2.

KabirBhutti:
> @KabirBhutti Thanks for the report! This fix will be released soon in 0.7.2.

When will the version 0.7.2 be released?

lonless9:
@KabirBhutti This weekend!

PR #2580, opened 2026-09-13
feat: support duplicated field names in nested Spark struct schemas

### LINKED ISSUE #325: Support duplicated field names in Spark schema

Jorge-Polanco-Roque:
Fixes #325.

## What / why
Spark allows duplicate field names within a struct, but Sail didn't deduplicate them when converting a Spark schema to Arrow. Arrow addresses struct fields by name, so `struct<x: string, x: int>` produced two `x` fields Arrow can't tell apart — which is what breaks `createDataFrame`/`toPandas` on such schemas.

## The fix
Mirrors Spark's `deduplicateFieldNames` (from `ArrowUtils.scala`): when resolving a struct, duplicate names get a deterministic `name_0`, `name_1`, ... suffix in order of appearance. Struct resolution recurses, so this applies at every nesting depth automatically; unique names are untouched.

To keep the *logical* schema the user sees intact (Spark only dedups inside the Arrow IPC, invisibly), each renamed field's original name is stashed in Arrow field metadata (`SAIL::original_field_name`) and restored in the Arrow→Spark conversion. So the physical Arrow schema carries `x_0, x_1` (unique, as Arrow requires) while the client still sees `x, x`.

On the `errorOnDuplicatedFieldNames` mode from the PySpark tests: in Spark Connect the server always converts with that flag off — the error in the `arrow + legacy` `toPandas` case is raised client-side in PySpark, not by the server. So there's nothing server-side to add for it; emitting deduplicated Arrow IPC (which this does) is the whole job.

## Testing
Rust unit tests for the dedup helper, the metadata round-trip, and end-to-end struct resolution via a real `PlanResolver` (fails without the fix, passes with it). The PySpark compat tests named in the issue (`test_createDataFrame_duplicate_field_names`, `test_toPandas_duplicate_field_names`) run against a live Sail server + downloaded Spark, which I couldn't stand up locally — I walked through their v3.5.9 source against this implementation and they line up (collect restores `x,x`; `toPandas` dict mode expects the physical `x_0,x_1`). Would appreciate a CI run to confirm. Build/test for the touched crates, `rustfmt`, and clippy are clean.

---
_Disclosure: I use AI assistance for code review and drafting (English isn't my first language); I verify every change myself._


linhr:
As of now, this contributor opened at least 29 PRs across 27 repositories this month. This is a clear sign of AI-driven massive contribution that does not show human thoughtfulness and engagement. This is a waste of the maintainer's time since the maintainer could just prompt AI directly to get the same result.

> Disclosure: I use AI assistance for code review and drafting (English isn't my first language); I verify every change myself.

I doubt this disclosure reflects the reality faithfully given the observation I mentioned above.

I'm closing this PR. If this comes up again this user will be blocked from the organization.

PR #2583, opened 2026-09-13
feat: forward positional load paths to Python datasources

jonasdedden:
**Positional load-path forwarding** (`adapter.rs`): a single `.load(path)` is now injected as `options["path"]` (explicit option wins case-insensitively), matching PySpark and making positional and option-style reads equivalent. Zero or multiple paths are left untouched such that pathless sources like JDBC see no spurious option. Regression test (`test_python.py`): `test_python_read_load_path_passed` proves end-to-end that `.load("/my/test/path")` reaches a custom datasource as `options["path"]` and that an explicit option keeps precedence, mirroring the pre-existing `.save(path)` test.

```Python
# Positional load paths reach Python datasources (previously: missing-path error)
spark.read.format("vortex").load("/data/file.vortex").show()
```


linhr (approved):
Thanks for the contribution!

PR #2584, opened 2026-09-13
fix: preserve row counts in zero-column Python datasource batches

jonasdedden:
**Zero-column row-count preservation** (`arrow_utils.rs`): Preserve row counts when converting Python rows with an empty schema into an Arrow batch. Zero-column conversion now uses the serialized row count without deserializing Python values. Regression tests cover empty and nonempty input. This change does not implement projection pushdown.
```Python
# COUNT(*) no longer pulls payload columns across the boundary
spark.read.format("my_format").load("/data/x").count()
```


jonasdedden:
Addressed review in commit https://github.com/lakehq/sail/pull/2584/commits/2d4fc9408bc0ea2762c2ff9db8cf7ed8e06477e6

jonasdedden:
Copilot's remark "Push the projection into the Python read/schema before constructing the exec (and add an end-to-end count regression test)." would require some (breaking?) changes in the Python data source API, AFAIS. It's not done in this PR but https://github.com/lakehq/sail/pull/2581 explores this partly.

linhr (approved):
This looks good! Thanks!

PR #2617, opened 2026-09-19
feat: implement vector L2 distance

rohankumardubey:
Add Spark-compatible support for the `vector_l2_distance` function.

- implement L2 distance for `ARRAY<FLOAT>` vectors
- support mixed Arrow `List` and `LargeList` inputs
- match Spark semantics for nulls, empty vectors, dimension mismatches, and extreme values
- register the function with scalar planning and distributed execution codecs
- add BDD, Python, Parquet, Rust, codec round-trip, and gold-data coverage


shehabgamin (approved):
Looks great!

